PR #25, opened 2020-06-14
Added SymbolTable and Reader

zslayton:
*Description of changes:*

Added:
* SymbolTable
* Format-agnostic impl of Reader
* SymbolTableEventHandler trait
* New cursor function `raw_value_bytes()`, which allows safe access to a value's unparsed bytes without advancing the cursor. This is helpful for performance-sensitive searching/filtering and is only available for cursors over in-memory data sources like byte arrays.

Moved some functions from `BinaryIonCursor` into the `Cursor` trait itself:
* `depth()`
* `string_ref_map()`

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -1,0 +1,26 @@
+pub(crate) mod v1_0 {
+    pub const SYSTEM_SYMBOLS: &[&str] = &[
+        "$0",                       // $0
+        "$ion",                     // $1
+        "$ion_1_0",                 // $2
+        "$ion_symbol_table",        // $3
+        "name",                     // $4
+        "version",                  // $5
+        "imports",                  // $6
+        "symbols",                  // $7
+        "max_id",                   // $8
+        "$ion_shared_symbol_table", // $9
+    ];
+
+    pub(crate) mod system_symbol_ids {
+        pub const ION: usize = 1;
+        pub const ION_1_0: usize = 2;
+        pub const ION_SYMBOL_TABLE: usize = 3;
+        pub const NAME: usize = 4;
+        pub const VERSION: usize = 5;
+        pub const IMPORTS: usize = 6;
+        pub const SYMBOLS: usize = 7;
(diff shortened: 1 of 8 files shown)


src/constants.rs
```diff
@@ -0,0 +1,26 @@
...
+        "name",                     // $4
+        "version",                  // $5
+        "imports",                  // $6
+        "symbols",                  // $7
+        "max_id",                   // $8
+        "$ion_shared_symbol_table", // $9
+    ];
+
+    pub(crate) mod system_symbol_ids {
+        pub const ION: usize = 1;
```
therapon, line 16: Are we leaving symbol 0 for later?

src/reader.rs
```diff
@@ -0,0 +1,310 @@
...
+    spooky: PhantomData<D>,
+    symtab_event_handler: Option<Box<dyn SymbolTableEventHandler>>,
+}
+
+impl<D: IonDataSource, C: Cursor<D>> Reader<D, C> {
+    pub fn new(cursor: C) -> Reader<D, C> {
+        Reader {
+            cursor,
+            symbol_table: SymbolTable::new(),
+            spooky: PhantomData,
```
therapon, line 30: ha! :)

src/reader.rs
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    pub fn symbol_table(&self) -> &SymbolTable {
+        &self.symbol_table
+    }
+
+    // The Reader needs to expose many of the same functions as the Cursor, but only some of those
+    // need to be re-defined to allow for system value processing. Any method listed here will be
+    // delegated to self.cursor directly.
+    delegate! {
```
therapon, line 174: cute macro

src/constants.rs
zslayton: Right, there's currently no code in place to handle SID zero. I hope to add `ion-tests`' bad files soon though, and I'm sure `$0` will lead to lots of failures. 😛

PR #30, opened 2020-06-29
Tracking subfield offsets

zslayton:
Adds logic to track the byte offsets for a value's field ID and annotations (if present), header, and
encoded value.

Also adds functions to get byte slices for each subfield's representation to the in-memory Reader
implementation.

### DIFF
--- a/src/binary/cursor.rs
+++ b/src/binary/cursor.rs
@@ -22,6 +22,8 @@ use crate::{
 };
 use std::io;
 
+use std::ops::Range;
+
 /// Information about the value over which the Cursor is currently positioned.
 #[derive(Clone, Debug)]
 struct EncodedValue {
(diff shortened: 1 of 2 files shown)


src/binary/cursor.rs
```diff
@@ -22,18 +22,114 @@ use crate::{
...
+    // Only valid if `value_length` is non-zero.
+    value_offset: usize,
+    // The number of bytes used to encode the value, not including
+    // the header byte or length fields.
+    value_length: usize,
+
+    // The first offset beyond the end of this value, calculated as:
+    //     value_offset + value_length
+    // Stored here to avoid having to recalculate it.
+    value_end: usize,
```
zslayton, line 62: `value_end` (an exclusive range end) replaced the `last_byte` field, which was an inclusive range end. This allowed me to simplify some bounds checking.

src/binary/cursor.rs
```diff
@@ -49,8 +145,12 @@ impl Default for EncodedValue {
             annotations: Vec::new(),
```
therapon, line 145: I am not sure what the balance is here: We could make annotations an `Option<...>` and avoid the new Vec here but that will add extra code around any operation no annotations to check optionality... Food for though mostly

src/binary/cursor.rs
```diff
@@ -22,18 +22,114 @@ use crate::{
...
+        if let Some(length) = self.field_id_length() {
+            let start = self.field_id_offset;
+            let end = start + length;
+            return Some(Range {start, end});
+        }
+        None
+    }
+
+    /// Returns the number of bytes used to encode this value's annotations, if any.
+    /// While annotations envelope the value that they modify, this function does not include
```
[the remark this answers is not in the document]
zslayton: I was trying to avoid saying "annotate" twice in one comment. I like `decorate`. 👍

src/binary/cursor.rs
zslayton: I went with `Vec` because [`Vec::new()` doesn't allocate](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.new) (so it should be about the same expense as making an `Option<Vec<_>>`) and (as you mention) it simplifies the code in places where it's used.

Issue #37, opened 2020-08-03
Add Easier APIs for Using Ion C


almann:
With #34, using the Ion C bindings is pretty low-level.  In particular, one needs to deal with `iERR` and also allocate and bind raw pointers.  It would be nice if we could expose safer versions of the low-level APIs that integrate with `Result`, make working with `ION_TYPE`/`ION_READER`/`ION_WRITER`/etc. easier.

This would not be a replacement for the traits we have in the native Rust implementation, but at least make interacting with IonC in its full capacity easier.

almann:
Resolving in favor of related issues that are more granular.

PR #40, opened 2020-08-04
Adds call wrapper, Result/Error integration.

almann:
* Adds `IonCError` to model `iERR` in a more idiomatic way.
* Adds type alias for `IonCResult`.
* Adds `ionc!` macro that wraps Ion C API calls with `unsafe` and
  conversion to `IonCResult`.
* Adds `as_str()` for `ION_STRING` to get a `str` slice in a convenient
  way.
* Retrofits unit tests and doc tests to use the new APIs.

First step towards #37.

### DIFF
--- a/ion-c-sys/src/lib.rs
+++ b/ion-c-sys/src/lib.rs
@@ -2,100 +2,76 @@
 //!
 //! These bindings are created with `bindgen` and are considerably low-level.
 //!
-//! ## Direct C API Examples
+//! ## Examples
 //! Using `ion-c-sys` directly is a pretty verbose affair, and requires checking the
-//! error code for most calls.
+//! error code for most calls.  This crate provides the [`result`](result/index.html)
+//! module to make it easier to integrate with `std::result::Result` with respect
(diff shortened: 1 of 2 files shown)


ion-c-sys/src/lib.rs
```diff
@@ -2,166 +2,127 @@
...
 //! use ion_c_sys::*;
+//! use ion_c_sys::result::*;
+//! # fn main() -> IonCResult {
 //!
 //! // output buffer
 //! let mut buf: Vec<u8> = vec![0; 128];
 //!
 //! // writer options--emit binary
-//! let mut writer_options: ION_WRITER_OPTIONS = Default::default();
+//! let mut writer_options = ION_WRITER_OPTIONS::default();
```
therapon, line 91: did we add `#[derive(Default)]` for `ION_WRITER_OPTIONS` in another PR?

almann: Yes, I added it in the doc test PR #39 because I used that there.  It is actually a feature of `bindgen`.

Issue #36, opened 2020-08-03
Add Better Testing for Ion C Bindings


almann:
As per #34, we should have better testing for Ion C bindings.  Running some tests against the test vectors at a minimum, and really refactoring that against what we have in `good.rs`.  In addition, we should exercise the bindings in a reasonable way in addition to the high-level test vectors.

PR #41, opened 2020-08-05
Adds ION_TYPE constants and ION_STRING::from.

almann:
* Adds `ion_type!` macro to generate the various `ION_TYPE_*` constants.
  These are not generated by bindgen.  This uses the `paste` proc macro
  crate to generate identifiers.
* Adds `ION_STRING::from` for the common case against a `&mut str`.
  If there is a use case, we could consider adding unsafe versions
  against `&str` because *most* cases are not actually mutating the
  underlying `*u8`.
* Updates unit/doc tests to use these APIs.

More progress towards #37.

### DIFF
--- a/ion-c-sys/src/lib.rs
+++ b/ion-c-sys/src/lib.rs
@@ -183,7 +183,7 @@ impl ION_STRING {
 }
 
 /// Generates easier to use constants for `ION_TYPE`
-/// These exist as CPP macros that don't get translated over from `bindgen`.
+/// These exist as C macros in `ion_types.h` that don't get translated over from `bindgen`.
 ///
 /// Using `ion_types!(NULL)` will generate a constant of the form:
 /// ```
@@ -205,6 +205,7 @@ ion_types!(
     none,
     EOF,
     NULL,
+    BOOL,
     INT,
     FLOAT,
     DECIMAL,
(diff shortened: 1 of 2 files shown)


ion-c-sys/src/lib.rs
```diff
@@ -163,7 +182,42 @@ impl ION_STRING {
...
+            paste! {
+                pub const [<ION_TYPE_ $name:upper>]: *mut ion_type =
+                    [<tid_ $name _INT>] as *mut ion_type;
+            }
+        )*
+    };
+}
+
+ion_types!(
+    none,
```
therapon, line 205: This list is missing BOOL (`tid_bool_INT`). Did the PR deal with BOOL and I missed it?

almann: Good catch (don't have enough tests obviously)--need to get to #36.

ion-c-sys/src/lib.rs
```diff
@@ -163,7 +182,42 @@ impl ION_STRING {
     }
 }
 
-pub mod result;
+/// Generates easier to use constants for `ION_TYPE`
+/// These exist as CPP macros that don't get translated over from `bindgen`.
```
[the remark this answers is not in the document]
almann: Well C Pre-Processor (CPP) is what I was going for here, but I can change it :).

Issue #47, opened 2020-08-07
Add Writer Facade APIs for Ion C


almann:
Much like #46, we should add the similar APIs for the Ion C Writer.

Related to #37.

PR #52, opened 2020-08-07
Adds Facade API on IonCWriterHandle.

almann:
This makes using common Ion C APIs much easier and more idiomatic
to Rust.

* Adds `new_buf_mode` to be an easier entry point to enable binary/text
  modes.
* Adds writer APIs over `ionc!` calls with doc examples.
* Adds support for immutable slices for IonCReaderHandle.
* Removes mutability requirement for input buffer as Ion C does not
  mutate read buffers (modulo bugs).
* Updates main doc example tests to use new APIs.
* Changes `ION_STRING::from_str` to `from_mut_str`
* Changes `ION_STRING::from_bytes` to `from_mut_bytes`
* Adds internal `ION_STRING::from_str` for immutable cases.

Implements #47.

### DIFF
--- a/ion-c-sys/src/reader.rs
+++ b/ion-c-sys/src/reader.rs
@@ -23,11 +23,10 @@ use crate::string::IonCStringRef;
 /// # use std::convert::*;
 /// # use std::ptr;
 /// # fn main() -> IonCResult<()> {
-/// let mut buf = b"hello".to_vec();
-/// let mut reader = IonCReaderHandle::try_from(buf.as_mut_slice())?;
+/// let mut reader = IonCReaderHandle::try_from(b"\xE0\x01\x00\xEA\x85hello".as_ref())?;
 ///
 /// let tid = reader.next()?;
-/// assert_eq!(ION_TYPE_SYMBOL, tid);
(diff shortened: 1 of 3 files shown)


ion-c-sys/src/reader.rs
```diff
@@ -23,11 +23,10 @@ use crate::string::IonCStringRef;
 /// # use std::convert::*;
 /// # use std::ptr;
 /// # fn main() -> IonCResult<()> {
-/// let mut buf = b"hello".to_vec();
-/// let mut reader = IonCReaderHandle::try_from(buf.as_mut_slice())?;
+/// let mut reader = IonCReaderHandle::try_from(b"\xE0\x01\x00\xEA\x85hello".as_ref())?;
```
rgantt, line 26: Forgive the stupid question: this byte sequence shows up all over the place--is it just the [binary version marker](http://amzn.github.io/ion-docs/docs/binary.html#value-streams)?

almann: Yes, this sequence is the BVM: `E0 01 00 EA`

PR #53, opened 2020-08-07
Adds Lifetime Safe Field Writing.

### LINKED ISSUE #51: IonCWriterHandle Lifetime Safe Field APIs
The field writing API has a requirement that the `ION_STRING` passed to them must live until a value is written.  The trivial API surface makes this contract implicit:

```rust
let mut buf = vec![0; 128];
let len = {
    let mut writer = IonCWriterHandle::new_buf_mode(buf.as_mut_slice(), WriterMode::Binary)?;
    writer.start_container(ION_TYPE_STRUCT)?;
    {
        // no problem! 'static life for "name"
        writer.set_field_name("name")?;
        writer.write_string("kumo")?;
    }
    writer.finish_container()?;
    writer.finish()?
}
```

An example where this breaks is here:

```rust
let mut buf = vec![0; 128];
let len = {
    let mut writer = IonCWriterHandle::new_buf_mode(buf.as_mut_slice(), WriterMode::Binary)?;
    writer.start_container(ION_TYPE_STRUCT)?;
    {
        {
            let s = "name".to_string();
            writer.set_field_name(s.as_str())?;
        }
        writer.write_string("kumo")?;
    }
    writer.finish_container()?;
    writer.finish()?
}
```

In practice, users are probably not going to shoot themselves in the foot by separating the value writing calls from the field setting, but it would be nice to take away the foot gun.

almann:
* Adds `IonCFieldWriterContext` which is a partial bind through references of the
  field name and writer handle.
* Changes `set_field_name` to `start_field_name`.
* Adds `IonCValueWriter` trait to capture the value writing API surface.
* Updates doc test with this API.

Resolves #51.

### DIFF
--- a/ion-c-sys/src/writer.rs
+++ b/ion-c-sys/src/writer.rs
@@ -429,7 +429,7 @@ impl Drop for IonCWriterHandle<'_> {
 /// `ion_writer_write_*` function is called for writing a value.
 ///
 /// This context tracks the lifetime of the string slice that a user wants to
-/// write as as a field name along with a mutable borrow of the writer handle
+/// write as a field name along with a mutable borrow of the writer handle
 /// and ensures that calls to `ion_writer_write_field` happen before invoking the
 /// `ion_writer_write_*` call for the value in a lifetime correct way.
 ///
--- a/ion-c-sys/src/writer.rs
+++ b/ion-c-sys/src/writer.rs
@@ -15,6 +15,41 @@ pub enum WriterMode {
     Binary = 1,
 }
 
+/// The API for Ion C value writing.
+pub trait IonCValueWriter {
+    /// Writes a `null` value.
+    fn write_null(&mut self, tid: ION_TYPE) -> IonCResult<()>;
+
+    /// Writes a `bool` value.
+    fn write_bool(&mut self, value: bool) -> IonCResult<()>;
+
+    /// Writes an `int` value.
+    fn write_i64(&mut self, value: i64) -> IonCResult<()>;
+
+    // TODO ion-rust/#50 - support ION_INT (arbitrary large integer) writes
+
+    /// Writes a `float` value.
+    fn write_f64(&mut self, value: f64) -> IonCResult<()>;
+
+    // TODO ion-rust/#42 - support decimal writes
+    // TODO ion-rust/#43 - support timestamp writes
+
+    /// Writes a `symbol` value.
+    fn write_symbol(&mut self, value: &str) -> IonCResult<()>;
+
+    /// Writes a `string` value.
+    fn write_string(&mut self, value: &str) -> IonCResult<()>;
+
+    /// Writes a `clob` value.
+    fn write_clob(&mut self, value: &[u8]) -> IonCResult<()>;
+
+    /// Writes a `blob` value.
+    fn write_blob(&mut self, value: &[u8]) -> IonCResult<()>;
+
+    /// Starts  a container.
+    fn start_container(&mut self, tid: ION_TYPE) -> IonCResult<()>;
+}
+
 /// Wrapper over `hWRITER` to make it easier to writers in IonC correctly.
 ///
 /// Specifically supports the `Drop` trait to make sure `ion_writer_close` is run.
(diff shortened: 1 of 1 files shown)


ion-c-sys/src/writer.rs
```diff
@@ -112,14 +142,27 @@ impl<'a> IonCWriterHandle<'a> {
     /// # Ok(())
     /// # }
     /// ```
-    pub fn set_field_name(&mut self, field: &str) -> IonCResult<()>{
-        // Ion C promises that it won't do mutation for this call!
-        let mut ion_field = ION_STRING::from_str(field);
-        ionc!(ion_writer_write_field_name(self.writer, &mut ion_field))
+    pub fn start_field<'b, 'c>(&'b mut self, field: &'c str) -> IonCFieldWriterContext<'a, 'b, 'c> {
```
marcbowes, line 145: Did you consider `field` or `write_field`?

The previous name made sense. The new one doesn't (to me!). Where is `end_field`, for example. What does 'start' even mean?

ion-c-sys/src/writer.rs
```diff
@@ -140,7 +183,7 @@ impl<'a> IonCWriterHandle<'a> {
     /// # }
     /// ```
     #[inline]
-    pub fn write_null(&mut self, tid: ION_TYPE) -> IonCResult<()> {
+    fn write_null(&mut self, tid: ION_TYPE) -> IonCResult<()> {
```
marcbowes, line 186: Nice!

ion-c-sys/src/writer.rs
```diff
@@ -15,6 +15,41 @@ pub enum WriterMode {
...
+    fn write_string(&mut self, value: &str) -> IonCResult<()>;
+
+    /// Writes a `clob` value.
+    fn write_clob(&mut self, value: &[u8]) -> IonCResult<()>;
+
+    /// Writes a `blob` value.
+    fn write_blob(&mut self, value: &[u8]) -> IonCResult<()>;
+
+    /// Starts  a container.
+    fn start_container(&mut self, tid: ION_TYPE) -> IonCResult<()>;
```
therapon, line 50: Don't we need an `end_container` here, or did I miss something?

ion-c-sys/src/writer.rs
```diff
@@ -383,4 +413,83 @@ impl Drop for IonCWriterHandle<'_> {
             ionc!(ion_writer_close(self.writer)).unwrap()
         }
     }
+}
+
+/// Context for writing fields in with proper reference lifetimes.
```
marcbowes, line 418: I think you stated the problem in the chat group nicely. Can we get some of that context here? Specifically, that it ensures usages where both the field name and value are supplied without lifetime issues.

ion-c-sys/src/writer.rs
almann: I wanted to describe that the it wasn't a writing a field, but a partial application for writing a field.  I am good with `field` if that is more descriptive.

ion-c-sys/src/writer.rs
almann: Good point, let me add some additional color here about what the point of this thing is.

ion-c-sys/src/writer.rs
almann: I didn't add it because it is not relevant to all of the writing contexts, though I think that is potentially confusing.

ion-c-sys/src/writer.rs
marcbowes: Actually, after seeing therapon's question, I agree with the name you picked!

ion-c-sys/src/writer.rs
```diff
@@ -416,6 +422,22 @@ impl Drop for IonCWriterHandle<'_> {
...
 
 /// Context for writing fields in with proper reference lifetimes.
+///
+/// Specifically, writing fields with Ion C with `ion_writer_write_field`
+/// is particularly problematic because the lifetime of the `ION_STRING`
+/// passed into that function, must be valid until a corresponding
+/// `ion_writer_write_*` function is called for writing a value.
+///
+/// This context tracks the lifetime of the string slice that a user wants to
+/// write as as a field name along with a mutable borrow of the writer handle
```
therapon, line 432:
```suggestion
/// write as a field name along with a mutable borrow of the writer handle
```

ion-c-sys/src/writer.rs
almann: Funny enough--I came around to `field` as a useful name because it technically doesn't start anything, it just partially applies the function (so to speak).  Let me know if you prefer I change it back.

PR #55, opened 2020-08-08
Adds copyright notices to ion-c-sys.

almann:
Another maintenance task--add the proper copyright headers.


almann:
> We could add the license header to each file, but if we have a LICENSE in the root of the repository then adding the license header is optional (I verified)
> 
> Just FYI

Good to know--I will tighten this up and push.

PR #58, opened 2020-08-09
[ion-c-sys] Implements Annotation Writing.

### LINKED ISSUE #56: Add Annotation Integration to Writer Ion C Facade
Much like #48, we should add the writer equivalents.

almann:
Changes `IonCFieldWriterContext` to a unified `IonCAnnotationsFieldWriterContext`.
This enables the partial application function of building up annotations/field name
to be done completely with references and writing either or both for a particular
value.

Resolves #56.

### DIFF
--- a/ion-c-sys/src/writer.rs
+++ b/ion-c-sys/src/writer.rs
@@ -486,6 +486,7 @@ pub struct IonCAnnotationsFieldWriterContext<'a, 'b, 'c> {
 }
 
 impl<'a, 'b, 'c> IonCAnnotationsFieldWriterContext<'a, 'b, 'c> {
+    #[inline]
     fn new_field(handle: &'b mut IonCWriterHandle<'a>, field: &'c str) -> Self {
         Self {
             handle,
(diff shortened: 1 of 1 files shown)


ion-c-sys/src/writer.rs
```diff
@@ -428,97 +461,134 @@ impl Drop for IonCWriterHandle<'_> {
...
+
+    /// Sets the field name for the context.
+    pub fn field(&mut self, field: &'c str) -> &mut Self {
+        self.field = Some(field);
+        self
     }
 }
 
-macro_rules! write_field {
+macro_rules! write_annotations_field {
```
therapon, line 518:
```suggestion
macro_rules! write_field_annotations {
```

almann: I refactored this into an inline non-public function called `write_annotations_and_field()`, is there a particular reason why you want the ordering of the names here?  I ask because this context is for either field writing, annotation writing, or both.

PR #68, opened 2020-08-09
[ion-c-sys] Ion binary reader API tests.

### LINKED ISSUE #63: Basic Binary Read Tests for Ion C
Towards #36.  We should have reasonable coverage similar to #62.

almann:
* Adds `bin` field for test cases.
* Adds Ion binary equivalent `u8` array constants for each test.
* Adds a `TestMode` to create a binary/text testing matrix.
* Removes `reader.get_type()` before `next()` assertion because of
  an Ion C bug (amzn/ion-c#186).

Resolves #63. 

### DIFF
--- a/ion-c-sys/tests/value.rs
+++ b/ion-c-sys/tests/value.rs
@@ -57,116 +57,157 @@ use Val::*;
 #[derive(Debug)]
 struct TestCase {
     lit: &'static str,
+    bin: &'static [u8],
     elem: Elem,
+}
+
+enum TestMode {
+    Text,
+    Binary,
 }
 
 #[rstest(c,
     case::null(TestCase {
         lit: "null",
+        bin: &[0xE0, 0x01, 0x00, 0xEA, 0x0F],
         elem: Elem::new(Null(ION_TYPE_NULL)),
     }),
     case::null_bool(TestCase {
         lit: "null.bool",
+        bin: &[0xE0, 0x01, 0x00, 0xEA, 0x1F],
         elem: Elem::new(Null(ION_TYPE_BOOL)),
     }),
     case::null_int(TestCase {
         lit: "null.int",
+        bin: &[0xE0, 0x01, 0x00, 0xEA, 0x2F],
         elem: Elem::new(Null(ION_TYPE_INT)),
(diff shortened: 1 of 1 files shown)


ion-c-sys/tests/value.rs
```diff
@@ -57,116 +57,157 @@ use Val::*;
 #[derive(Debug)]
 struct TestCase {
     lit: &'static str,
+    bin: &'static [u8],
     elem: Elem,
 }
```
rgantt, line 62: It feels like `rstest` is exerting some weird pressure here. The `TestCase` struct has enough information to run a full test case, but you can't actually do anything with it until `rstest` gives you more information in the form of a `TestMode`. That seems backwards from how I would probably set it up--but maybe that's just the easiest way to "go with the grain" in `rstest`. This is where I appreciate using a `@DataProvider` (TestNG) or `@Parameterized` (JUnit) for data-driven tests: so that I can explode the tests according to my own logic, rather than the framework's.

If I were starting from scratch, your `TestCase` seems like a good place to begin, but would probably change `elem` to `expected` and `input` to be a vec containing 1 or more representations that should yield that `expected` value once they're consumed by the reader. Usually that means 1 of each, but are there cases where you could end up with the same `expected` from more than one text or binary input value? Are there `expected` values that can be gotten form a binary input, but not a text one? I think that's where the tests start to get really interesting.

almann: The thinking of the test case design (given the compile-time limitation of rstest)  was to have a 3-way equivalence between a sort of DOM-like representation, text, and binary, and use that as a template to drive reading and a writing test matrix with maybe some flavor of rountrip testing in the mix.

almann: The real problem is that Rust's test framework is purely static--you can't dynamically create tests like various test frameworks can, so you see a cottage industry of `proc_macro` to simulate it at compile time.  You can get some of the way there, but it is weak.

We *could* do a binary crate with our own test driver, but tooling integration would be weak (debugging a single test case, IDE integration, etc.).  I don't know if this route is worth it, but it is an option.

PR #69, opened 2020-08-10
ci: refactors and adds codecoverage

therapon:
*Description of changes:*

* refactors rust.yml to use os matrix and rust stable 
* adds codecoverage as a separate task using tarpaulin
    * I did try [grcov](https://github.com/actions-rs/grcov), but it requires compiler flags that treat certain ion-c warnings as errors. Did not spend too much time debugging.

See  

* https://github.com/therapon/ion-rust-1/runs/966113292?check_suite_focus=true
* https://codecov.io/gh/therapon/ion-rust-1/branch/ts-ci-codecov

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -51,9 +51,9 @@ jobs:
           toolchain: nightly
           override: true
       - name: Rust Cargo Tarpaulin
-        uses: actions-rs/tarpaulin@v0.1
+        eses: actions-rs/tarpaulin@v0.1
         with:
-          args: '-v --run-types AllTargets --run-types Doctests  --out Html --workspace -- --test-threads=1'
+          args: '--release -v --run-types AllTargets --run-types Doctests  --out Html --workspace -- --test-threads=1'
       - name: Upload to codecov.io
         uses: codecov/codecov-action@v1.0.2
         with:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -4,7 +4,7 @@ on: [push]
 
 jobs:
   build:
-    name: Build and Test
+    name: CI Build
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
@@ -48,12 +48,12 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: stable
+          toolchain: nightly
           override: true
       - name: Rust Cargo Tarpaulin
         uses: actions-rs/tarpaulin@v0.1
         with:
-          args: '--release --out Html --workspace'
+          args: '-v --run-types AllTests --run-types Doctests  --out Html --workspace -- --test-threads=1'
       - name: Upload to codecov.io
         uses: codecov/codecov-action@v1.0.2
         with:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -53,7 +53,7 @@ jobs:
       - name: Rust Cargo Tarpaulin
         uses: actions-rs/tarpaulin@v0.1
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -1,46 +1,67 @@
-name: Build
+name: ion-rust CI Build
```
almann, line 1: This name is the name and URL of the badge, so whatever it is will have to be changed in the two README files (and we have to be happy with the text).  Personally, if we're going to change the text `CI Build` is better, or `Ion Rust CI Build` just to be case consistent and keep the badge label reasonable.

image

.github/workflows/rust.yml
```diff
@@ -1,46 +1,67 @@
...
 
 jobs:
-  linux:
-    runs-on: ubuntu-latest
+  build:
+    name: Build and Test
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
```
almann, line 11: :+1:   Nice job doing the needful of eliminating my lack of DRY.

.github/workflows/rust.yml
```diff
@@ -1,46 +1,67 @@
...
-        run: cargo test --verbose --workspace
-
-  windows:
-    runs-on: windows-latest
-    env:
-      # needed for bindgen
-      LIBCLANG_PATH: C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\Llvm\x64\bin
+          profile: minimal
+          toolchain: stable
+          override: true
```
almann, line 27: Is this specifically to choose the tool chain versus what is installed in the base image?

therapon:
> I looked at the coverage report and it doesn't seem accurate. For example, `ion-c-sys::writer` claims zero percent coverage, which is impossible due to the doc tests, are those not getting run?

The lib claims to run all tests. I am trying to get it to work but it seems to either fail on my machine or not take into account doctests. I will dig into it a little more and see if I can manage to obtain results for *all* tests

.github/workflows/rust.yml
therapon: I forgot about the badge. I will update the name.

.github/workflows/rust.yml
therapon: Yes. While trying grcov I needed nightly and this was the GitHub Action that I found which allows you to pick Rust versions as well as (when we define a minimum version) a minimum version feature to build and run.

.github/workflows/rust.yml
almann: We have two badges, FYI for the READMEs in both Crates.  I did this because when we publish `ion-c-sys`, I wanted it to also have a reasonable README.

.github/workflows/rust.yml
```diff
@@ -1,46 +1,67 @@
...
+      - name: Rust Toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
+          override: true
+      - name: Rust Cargo Tarpaulin
+        uses: actions-rs/tarpaulin@v0.1
+        with:
+          args: '--release -v --run-types AllTargets --run-types Doctests  --out Html --workspace -- --test-threads=1'
```
almann, line 56: Is there a bug or something requiring single threaded tests?

therapon: There is something there. It succeed on my (old) personal laptop but fails on GitHub Actions.

therapon:
Closing this PR since I am trying out things that seem to fail only on GitHub and it causes this PR to update. 

Here is the coverage that I get with both tests and doctest measured 
[tarpaulin-report.html.zip](https://github.com/amzn/ion-rust/files/5054469/tarpaulin-report.html.zip)

PR #74, opened 2020-08-11
[ion-c-sys] Adds Big Integer Support.

### LINKED ISSUE #50: Add ion-c-sys support for ION_INT
We are currently limited to `i64` integers in the bindings, we should have some idiomatic integration for `ION_INT` which is Ion C's arbitrary integer type.

almann:
Adds support for converting to/from `num_bigint::BigInt`
and `ION_INT` and reader/writer support.

* Adds `IonIntPtr` as safe pointer to `ION_INT`.
* Adds some module documentation.
* Updates Cargo.toml to have more explicit version
requirements.
* Adds `rstest_reuse` crate for reusing test tables.

Resolves #50.

### DIFF
--- a/ion-c-sys/Cargo.toml
+++ b/ion-c-sys/Cargo.toml
@@ -5,10 +5,12 @@ edition = "2018"
 
 [dependencies]
 paste = "1.0.0"
+num-bigint = "0.3.0"
 
 [build-dependencies]
-cmake = "0.1"
-bindgen = "0.54"
+cmake = "0.1.44"
+bindgen = "0.54.1"
 
 [dev-dependencies]
 rstest = "0.6.4"
+rstest_reuse = "0.1.0"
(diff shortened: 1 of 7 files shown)


ion-c-sys/Cargo.toml
```diff
@@ -5,10 +5,12 @@ edition = "2018"
 
 [dependencies]
 paste = "1.0.0"
+num-bigint = "0.3.0"
 
 [build-dependencies]
-cmake = "0.1"
-bindgen = "0.54"
+cmake = "0.1.44"
+bindgen = "0.54.1"
```
almann, line 12: I believe this is closer to what we'd want based on [the Cargo book](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html).  The `x.y.z` specifier is basically `>= x.y` and `< x.(y + 1)`,

Issue #79, opened 2020-08-13
[ion-c-sys] Errors Converting Large BigDecimal -> ION_DECIMAL


almann:
For cases that are larger than `decQuad` there are issues using `ion_decimal_from_ion_int` to convert the coefficient creating a status error of invalid context leading to generating a `nan` `decNumber`.  Talking with @tgregg, I believe this is a bug in Ion C, and needs to be reproduced there and fixed upstream.

Issue #80, opened 2020-08-13
[ion-c-sys] ION_DECIMAL Assignment from BigDecimal Breaks Ion C Encapsulation


almann:
There appears to be an issue with `ion_decimal_from_int` that requires a specifically allocated `ION_DECIMAL`.  Since this is abstracted away from the public API.

Specifically, the problem appears to be [`ion_decimal.c` L287-L293][1]:
```c
iERR ion_decimal_from_ion_int(ION_DECIMAL *value, decContext *context, ION_INT *p_int) {
    ION_DECIMAL_SWITCH (
        value,
        IONCHECK(ion_int_to_decimal(p_int, quad_value, context)),
        IONCHECK(_ion_int_to_decimal_number(p_int, num_value, context))
    );
}
```

As a result, our bindings do manual initialization and use internal Ion C APIs.  Once this is fixed upstream, we should eliminate this hack.

[1]: https://github.com/amzn/ion-c/blob/d23756567e7b3765b86a58f314e3ea6a83341ca4/ionc/ion_decimal.c#L287-L293

Issue #78, opened 2020-08-13
[ion-c-sys] Convert BigDecimal -> ION_DECIMAL Inexact


almann:
Cases around ~35 digits or so, convert to `ION_DECIMAL` with a loss of precision, which is likely due to `decQuad` limitations, we should probably check the status code and do `decNumber` conversion or be more aggressive with usage of `decNumber`.

PR #81, opened 2020-08-13
[ion-c-sys] Adds support for BigDecimal.

### LINKED ISSUE #42: Add ion-c-sys decNumber/decQuad Integration/Utility
Provide APIs to bridge `decQuad`, `decNumber`, and `ION_DECIMAL` to the more idiomatic types in Rust.

Related to #37.

almann:
Implements `bigdecimal::BigDecimal` conversions to/from
`ION_DECIMAL`.

* Adds `IonDecimalPtr` smart pointer to manage `ION_DECIMAL` instances.
* Adds `ION_DECIMAL` methods:
  - `try_assign_bigdecimal` to convert from `BigDecimal`.
  - `try_to_bigdecimal` to convert to `BigDecimal`.
* Adds `IonCReaderHandle.read_bigdecimal`.
* Adds `IonCWriterHandle.write_bigdecimal`.
* Renames `ION_INT` conversion methods to be consistent.
* Works around construction of `ION_DECIMAL` from `ION_INT` (#80).
  - Adds `bindings.h` header and updates `build.rs` to expose internal
    Ion decimal APIs.
* Adds error cases and test skips for:
  - `ION_INT` to `decNumber` conversion problems (#79).

There is no limit on *reading*, however default options are still 37
digits in Ion C.  Currently, *writing* decimals with over 34 digits
may error out (#79), but should *never* emit incorrect data.

Resolves #42, #78.

### DIFF
--- a/ion-c-sys/Cargo.toml
+++ b/ion-c-sys/Cargo.toml
@@ -21,7 +21,8 @@ edition = "2018"
 
 [dependencies]
 paste = "1.0.0"
-num-bigint = "0.3.0"
+num-bigint = "0.2.6"
+bigdecimal = "0.1.2"
 
 [build-dependencies]
 cmake = "0.1.44"
--- a/ion-c-sys/src/lib.rs
+++ b/ion-c-sys/src/lib.rs
@@ -116,6 +116,7 @@
 #![allow(non_camel_case_types)]
 #![allow(non_snake_case)]
 
+pub mod decimal;
 pub mod int;
 pub mod reader;
 pub mod result;
@@ -124,14 +125,18 @@ pub mod writer;
 
 include!(concat!(env!("OUT_DIR"), "/ionc_bindings.rs"));
 
+use crate::decimal::IonDecimalPtr;
+use crate::int::IonIntPtr;
 use crate::result::*;
 
 use std::cmp::min;
 use std::convert::TryInto;
 use std::marker::PhantomData;
+use std::ptr;
 use std::str::Utf8Error;
 use std::{slice, str};
 
+use bigdecimal::BigDecimal;
 use num_bigint::{BigInt, Sign};
 use paste::paste;
 
@@ -144,7 +149,7 @@ impl ION_INT {
     /// Note that since `BigInt` does not have a ***view*** into its digits,
     /// this method will make an intermediate copy as the big-endian encoded
     /// byte vector that will then be stored into this `ION_INT`
-    pub fn assign_from_bigint(&mut self, src: &BigInt) -> IonCResult<()> {
+    pub fn try_assign_bigint(&mut self, src: &BigInt) -> IonCResult<()> {
         let (sign, mut raw_mag) = src.to_bytes_be();
         let is_neg = match sign {
             Sign::Minus => 1,
@@ -221,7 +226,7 @@ impl ION_INT {
 }
 
 #[cfg(test)]
-mod test {
+mod test_bigint {
     use crate::int::*;
     use crate::result::*;
     use crate::*;
@@ -265,7 +270,7 @@ mod test {
     fn bigint(lit: &str, sign: Sign) {}
 
     #[apply(bigint)]
-    fn assign_from_bigint(lit: &str, sign: Sign) -> IonCResult<()> {
+    fn try_assign_bigint(lit: &str, sign: Sign) -> IonCResult<()> {
         let bval = BigInt::parse_bytes(lit.as_bytes(), 10).unwrap();
         let mut ival = IonIntPtr::try_from_bigint(&bval)?;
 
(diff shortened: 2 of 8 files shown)


ion-c-sys/src/lib.rs
```diff
@@ -306,6 +311,315 @@ mod test {
...
+    ctx.digits = DEC_MAX_DIGITS as i32;
+    ctx.emax = DEC_MAX_EMAX as i32;
+    ctx.emin = DEC_MIN_EMIN as i32;
+    // make sure we don't pad decQuads
+    ctx.clamp = 0;
+
+    ctx
+}
+
+const DEC_QUAD_DIGITS: u64 = 37;
```
therapon, line 331: I should look this up, but hopefully you have the answer in your RAM :) 
I thought the capacity for DIGITS is what we pass at compile time https://github.com/amzn/ion-rust/blob/90efa71bf3af14e4f4bd54f4be06fd0a3bfbcdcf/ion-c-sys/build.rs#L36

Why is it 37? 

And from lines 349 - 350 in this PR I am sure I am missing something that has to do with Ion C :)

almann: IIRC, this is to enable the `decQuad` type which are 128-bit decimals that have 37 digits of precision.  One can still use the `decNumber` type for decimal numbers that are beyond this.

ion-c-sys/src/lib.rs
```diff
@@ -306,6 +311,315 @@ mod test {
...
+            _ => Err(IonCError::from(ion_error_code_IERR_INTERNAL_ERROR)),
+        }
+    }
+
+    /// Converts this `ION_DECIMAL` to a `BigDecimal`.
+    ///
+    /// Special decimal values such as NaN and infinity are not supported for conversion.
+    ///
+    /// This implementation borrows mutably, to avoid a copy of the underlying
+    /// decimal implementation, but does not change the value.
```
zslayton, line 388: Can you expand on this? Why is the `&mut` necessary if the value isn't being changed? Tough to tell at a glance without an IDE, but maybe one of the nested function calls demands it?

ion-c-sys/src/reader.rs
```diff
@@ -347,7 +347,32 @@ impl<'a> IonCReaderHandle<'a> {
...
+    /// ## Usage
+    /// ```
+    /// # use std::convert::*;
+    /// # use bigdecimal::BigDecimal;
+    /// # use ion_c_sys::*;
+    /// # use ion_c_sys::decimal::*;
+    /// # use ion_c_sys::reader::*;
+    /// # use ion_c_sys::result::*;
+    /// # fn main() -> IonCResult<()> {
+    /// let mut reader = IonCReaderHandle::try_from("3.0")?;
```
zslayton, line 361: I don't know if there's a community or `rustfmt`-driven recommendation here, but for the humans among us I think indenting the contents of this function would be helpful. :)

ion-c-sys/src/writer.rs
```diff
@@ -325,7 +327,35 @@ impl IonCValueWriter for IonCWriterHandle<'_> {
         ionc!(ion_writer_write_double(self.writer, value))
     }
 
-    // TODO ion-rust/#42 - support decimal writes
+    /// Writes an `decimal` value.
```
zslayton, line 330: Nit: ~~an~~ a `decimal` value

ion-c-sys/Cargo.toml
```diff
@@ -21,7 +21,8 @@ edition = "2018"
 
 [dependencies]
 paste = "1.0.0"
-num-bigint = "0.3.0"
+num-bigint = "0.2.6"
```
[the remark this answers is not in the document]
almann: Bigdecimal has this dependency, so when I added it, I ended up with a conflict.

ion-c-sys/src/lib.rs
almann: The value is being changed because I am scaling it from/to the exponent to extract the coefficient.  I can add a comment to this effect, but I am doing the operation in place.

ion-c-sys/src/reader.rs
almann: I pulled this style from one of "the books" ;-)--the example docs don't get the `main` scaffolding, so the block isn't there in the code.

If I do indent the code, I believe it will have an extra level of indentation in the rendered docs...

ion-c-sys/src/writer.rs
almann: Fixed.

ion-c-sys/src/lib.rs
almann: Added more documentation about this.

ion-c-sys/src/reader.rs
almann: The style is consistent with:

https://doc.rust-lang.org/rustdoc/documentation-tests.html

I'm not a fan, to be honest, but it will make our docs look a bit uglier to indent.

If you like I can cut an issue to address this globally.

zslayton: > If I do indent the code, I believe it will have an extra level of indentation in the rendered docs...

I didn't realize this, thanks for pointing it out. If this is the prescribed approach I'm fine with keeping it. :thumbsup:

ion-c-sys/src/lib.rs
almann: Ah, here it is in the `CMakeLists.txt` for Ion C:

```cmake
# NOTE: DECNUMDIGITS must be set across all compilation units to at least DECQUAD_Pmax (34), so that the value is
# guaranteed to be consistent between ionc and decNumber. This is required for conversions between decQuad and
# decNumber. This is NOT the limit on decimal precision; ION_DECIMAL can handle arbitrarily large precision.
add_definitions(-DDECNUMDIGITS=34)
```

[1]: https://github.com/amzn/ion-c/blob/1e911eb689a879427aa8842fe2ca7c78546aeed1/CMakeLists.txt#L22-L26

almann: I've updated the `build.rs` doc in [this commit](https://github.com/amzn/ion-rust/pull/81/commits/aeaa47606ea2c3d814a00c82f3958da8000cbc7f).

I've also re-thought what the heck I was doing there and fixed #79 with [this commit](https://github.com/amzn/ion-rust/pull/81/commits/a6ed8e093d7705f67983fafb1c492ce72e39f5db).

(All of the above are in this PR and reviewable separately)

Issue #43, opened 2020-08-05
Add ion-c-sys Timestamp Integration/Utility


almann:
Provide bridge APIs for `ION_TIMESTAMP` into the idiomatic Rust APIs.

Related to #37.

PR #86, opened 2020-08-18
Adds value read tests for decimal/timestamp.

almann:
* Minor cleanup on BigDecimal conversion error message.
* Makes the dereference for the `ION_DECIMAL` explicit to get at fraction
  because it makes CLion confused (and is technically clearer as to what is going on).

Related to #42 and #43.

### DIFF
--- a/ion-c-sys/src/lib.rs
+++ b/ion-c-sys/src/lib.rs
@@ -335,6 +335,9 @@ fn make_context() -> decContext {
     ctx
 }
 
+// TODO amzn/ion-rust#79 - we need to fix decNumber support.
+const DEC_NUMBER_ERROR_MSG: &str = "DecNumber Not Supported";
+
 impl ION_DECIMAL {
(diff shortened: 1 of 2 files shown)


ion-c-sys/src/lib.rs
```diff
@@ -596,9 +599,12 @@ mod test_bigdecimal {
                 assert_eq!(exponent, actual_exponent, "Testing exponents");
             }
             Err(e) => match e.code {
-                ion_error_code_IERR_INVALID_STATE => {
-                    println!("Ignoring amzn/ion-rust#79 for {}", d_lit)
-                }
+                ion_error_code_IERR_INVALID_STATE => match e.additional {
+                    "DecNumber Not Supported" => {
```
zslayton, line 603: Since we're producing and later matching on this `&str` literal, I think we should make it a constant.

almann: Done.

PR #91, opened 2020-08-26
Added basic TextWriter impl

zslayton:
*Description of changes:*

Adds a basic implementation of a user-level `TextWriter`.

This implementation does not:
* Support `clob`
* Provide pretty-printing
* Attempt to produce optimal `symbol` representations. All symbols are wrapped in single quotes.
* Attempt to produce optimal delimiters. All values in a container have a trailing delimiter.

The API is more or less cloned from `ion-java` (in particular, `set_annotations` and `set_field_name`). I fully expect to change it when we've had some time to do design work.


src/result.rs
```diff
@@ -62,3 +62,12 @@ impl From<io::Error> for IonError {
         }
     }
 }
+
+/// Allows [`std::fmt::Error`]s to be converted to an IonError and propagated using the `?` operator.
+impl From<std::fmt::Error> for IonError {
+    fn from(error: std::fmt::Error) -> Self {
+        IonError::IoError {
+            description: format!("Encountered an IO error: {:?}", error),
```
zslayton, line 70: This surfaces `fmt::Error`s (which would happen when encoding scalar values) to the end user as IO issues. This isn't strictly correct, but I don't think it's worth adding another error variant to communicate the distinction. (Our error handling in general could use some more thought, FWIW.)

Cargo.toml
```diff
@@ -25,6 +25,7 @@ members = [
 ]
 
 [dependencies]
+base64 = "~0.12.3"
```
zslayton, line 28: Needed to base64-encode byte arrays as `blob`s.

src/text/encoder.rs
```diff
@@ -0,0 +1,264 @@
+use crate::result::IonResult;
+use crate::IonType;
+use bigdecimal::BigDecimal;
+use chrono::{DateTime, FixedOffset};
+use std::io;
+
+/// Utility type for writing scalar Rust values as Ion text.
+pub struct TextEncoder {}
```
zslayton, line 8: The `TextEncoder` type can encode any scalar value to an `io::Write` sink. `TextWriter` wraps a `TextEncoder`; it provides a buffer and manages container/annotation state.

src/text/encoder.rs
```diff
@@ -0,0 +1,264 @@
...
+        output: &mut W,
+        value: I,
+    ) -> IonResult<()> {
+        let fixed_offset_datetime: DateTime<FixedOffset> = value.into();
+        write!(output, "{}", fixed_offset_datetime.to_rfc3339())?;
+        Ok(())
+    }
+
+    /// Writes a string value to `output` as an Ion symbol.
+    pub fn encode_str_as_symbol<S: AsRef<str>, W: io::Write>(
```
zslayton, line 102: Naming is a bit tricky here. While most Rust types have an obvious mapping to Ion (`bool` -> `boolean`, `i64` -> `integer`, etc), there are multiple Rust text representations and multiple Ion text representations. I've opted to be more explicit in the names, which makes them accurate but a bit unwieldy. Open to suggestions for these.

src/text/encoder.rs
```diff
@@ -0,0 +1,264 @@
...
+
+    /// Writes a byte array slice to `output` as an Ion blob.
+    pub fn encode_blob<W: io::Write>(output: &mut W, value: &[u8]) -> IonResult<()> {
+        // Rust format strings escape curly braces by doubling them. The following string is:
+        // * The opening {{ from a text Ion blob, with each brace doubled to escape it.
+        // * A {} pair used by the format string to indicate where the base64-encoded bytes
+        //   should be inserted.
+        // * The closing }} from a text Ion blob, with each brace doubled to escape it.
+        // TODO: Provide a re-usable encoding buffer instead of allocating a String each time.
+        write!(output, "{{{{{}}}}}", base64::encode(value))?;
```
zslayton, line 132: lol, `{{{{{}}}}}}`.

src/text/encoder.rs
```diff
@@ -0,0 +1,264 @@
...
+    #[test]
+    fn encode_i64() {
+        encoder_test(|output| TextEncoder::encode_i64(output, -99999), "-99999");
+        encoder_test(|output| TextEncoder::encode_i64(output, -42), "-42");
+        encoder_test(|output| TextEncoder::encode_i64(output, -1), "-1");
+        encoder_test(|output| TextEncoder::encode_i64(output, 0), "0");
+        encoder_test(|output| TextEncoder::encode_i64(output, 1), "1");
+        encoder_test(|output| TextEncoder::encode_i64(output, 42), "42");
+        encoder_test(|output| TextEncoder::encode_i64(output, 99999), "99999");
+    }
```
almann, line 179: Not a blocking concern, but you might want to take a look at some of my use of [`rstest`][1] here.

[1]: https://github.com/amzn/ion-rust/blob/c1479ea3f09432f34254d385eba91fb95af236ec/ion-c-sys/src/lib.rs#L466-L614

src/text/writer.rs
```diff
@@ -0,0 +1,352 @@
...
+use crate::IonType;
+use bigdecimal::BigDecimal;
+use chrono::{DateTime, FixedOffset};
+use std::io::{BufWriter, Write};
+
+pub struct TextWriter<W: Write> {
+    output: BufWriter<W>,
+    annotations: Vec<String>,
+    field_name: Option<String>,
+    containers: Vec<IonType>,
```
almann, line 12: Interesting, so this type copies in these concerns?  What is the purpose of the containers `Vec`? Is that the current stack of where we "are" in the writing?

src/text/writer.rs
```diff
@@ -0,0 +1,352 @@
...
+            annotations: vec![],
+            field_name: None,
+            containers: vec![],
+        }
+    }
+
+    /// Returns a reference to the underlying io::Write implementation.
+    pub fn output(&self) -> &W {
+        self.output.get_ref()
+    }
```
almann, line 30: Hmm, is this a useful API to expose?  What is the use case for it (out of curiosity)?

src/text/writer.rs
```diff
@@ -0,0 +1,352 @@
...
+
+    /// Causes any buffered data to be written to the underlying io::Write implementation.
+    pub fn flush(&mut self) -> IonResult<()> {
+        self.output.flush()?;
+        Ok(())
+    }
```
almann, line 36: In other APIs, we have a difference of `flush` vs. `finish`, do you have thoughts about that here?

src/text/writer.rs
```diff
@@ -0,0 +1,352 @@
...
+    /// ignored.
+    pub fn set_field_name(&mut self, name: &str) {
+        self.field_name = Some(name.to_string());
+    }
+
+    /// Sets a list of annotations that will be applied to the next value that is written.
+    pub fn set_annotations(&mut self, annotations: &[&str]) {
+        self.annotations
+            .extend(annotations.iter().map(|s| s.to_string()));
+    }
```
almann, line 49: As you know, it would be nice if we didn't have to do the copy here :wink:

src/text/writer.rs
```diff
@@ -0,0 +1,352 @@
...
+
+    /// Writes the provided byte array slice as an Ion blob.
+    pub fn write_blob(&mut self, value: &[u8]) -> IonResult<()> {
+        self.write_scalar(|output| TextEncoder::encode_blob(output, value))
+    }
+
+    /// Writes the provided byte array slice as an Ion clob.
+    pub fn write_clob(&mut self, _value: &[u8]) -> IonResult<()> {
+        todo!()
```
almann, line 181: Technically you don't really need to `todo!()` here, can't you just delegate downward to `TextEncoder` (which has the `todo!()`)?

Cargo.toml
zslayton: I was following suit with the existing dependencies here, but having re-read [the docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements) just now I do think that `^` is the way to go. I'll change all of the dependencies to `^` in a separate PR if you're alright with that.

> ^ dependencies (the default)?

What do you mean by 'the default'? I would think that no special operators at all (i.e. simply specifying a version like `"0.12.3"`) would be considered the default.

src/text/encoder.rs
zslayton: >  this formatting string is resolved statically right?

Yes.

src/text/encoder.rs
```diff
@@ -0,0 +1,264 @@
...
+    pub fn encode_i64<W: io::Write>(output: &mut W, value: i64) -> IonResult<()> {
+        write!(output, "{}", value)?;
+        Ok(())
+    }
+
+    /// Writes a BigDecimal value to `output` as an Ion decimal.
+    pub fn encode_big_decimal<W: io::Write>(output: &mut W, value: &BigDecimal) -> IonResult<()> {
+        write!(output, "{}", &value)?;
+        Ok(())
+    }
```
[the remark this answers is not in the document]
zslayton: I considered this. The trade-off is that you replace `output: &mut W` everywhere with `&mut self` and then callers need to create an instance of `TextEncoder` anywhere they want to use it. Having functions instead of methods felt like a minor ergonomics win.

I have a `TODO` for creating a reusable `String` for formatting `blob` values; having `TextEncoder` own the `writer` *and* a `String` buffer seems like enough value to tip things in the other direction. I'll go ahead and make that change.

src/text/encoder.rs
```diff
@@ -0,0 +1,264 @@
...
+    pub fn encode_str_as_symbol<S: AsRef<str>, W: io::Write>(
+        output: &mut W,
+        value: S,
+    ) -> IonResult<()> {
+        write!(output, "'{}'", value.as_ref())?;
+        Ok(())
+    }
+
+    /// Writes a string value to `output` as an Ion string.
+    pub fn encode_str_as_string<S: AsRef<str>, W: io::Write>(
```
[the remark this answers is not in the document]
zslayton: This is commonly done to allow users to pass a `&str`, `String`, `&String`, `&mut String`, etc. I almost always write functions that take a `&str` first and then stub my toe on it repeatedly.

Certainly worth weighing its value if it complicates writing a trait, but I find it to be a very helpful default style.

src/text/encoder.rs
```diff
@@ -0,0 +1,264 @@
...
+
+    /// Writes a DateTime value to `output` as an Ion timestamp.
+    pub fn encode_datetime<W: io::Write, I: Into<DateTime<FixedOffset>>>(
+        output: &mut W,
+        value: I,
+    ) -> IonResult<()> {
+        let fixed_offset_datetime: DateTime<FixedOffset> = value.into();
+        write!(output, "{}", fixed_offset_datetime.to_rfc3339())?;
+        Ok(())
+    }
```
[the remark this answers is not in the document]
zslayton: > pushing that concern trivially to the caller?

"Trivial" is a matter of taste in this case. Compare:
```
let fixed_offset_now: DateTime<FixedOffset> = Utc::now().into();
TextEncoder::encode_datetime(output, fixed_offset_now);
```
to:
```
TextEncoder::encode_datetime(output, Utc::now());
```

While the `chrono` crate provides [`Local`](https://docs.rs/chrono/0.4.15/chrono/offset/struct.Local.html#method.now) and [`Utc`](https://docs.rs/chrono/0.4.15/chrono/offset/struct.Utc.html#method.now) flavors of `now()`, it does not provide a way to get the current time as a `DateTime<FixedOffset>`, so I would expect this sort of code to be common.

Similar to overriding methods in Java, accepting a generic parameter is just a nice API convenience.

src/text/writer.rs
zslayton: > Is that the current stack of where we "are" in the writing?

Yes. It's being used to track which value delimiters to insert between child values (`,`/`\n`/` `) and whether a field name is needed.

>  so this type copies in these concerns? 

Sorry, unsure what you mean here. Could you rephrase?

src/text/writer.rs
zslayton: Yep. I could eagerly write out the `field_name` and `annotations` as soon as the user provides them, but I'd need to rewrite things to track which function calls would be legal in which states. I also considered keeping a small pool of reusable annotations `String`s to avoid the cost of new allocations. Both seemed like more effort than they were worth for an initial pass at the less-performance-sensitive text writer. I'll open an issue to track it.

Cargo.toml
almann: I mean that with no sigil, it is a `^` dependency.

>The string "0.1.12" is a semver version requirement. Since this string does not have any operators in it, it is interpreted the same way as if we had specified "^0.1.12", which is called a caret requirement.

src/text/encoder.rs
almann: But why do we need both?  `TextEncoder` if I squint is really just some parts of `TextWriter` factored in a way to make it useful in other contexts.  What would that context be other than `TextWriter`?  In other words what is the separation buying us other than splitting some of the encoding concerns?

I think that intent should be captured in the doc comment or somewhere, because it did confuse me.

src/text/encoder.rs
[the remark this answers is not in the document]
almann: > Having functions instead of methods felt like a minor ergonomics win.

What is the ergonomic win? The call sites seem roughly the same to me (except the writer is a parameter vs. bound to the method receiver) so it seems like a wash, but I could be missing something.

src/text/writer.rs
zslayton: It's an API pattern I've encountered; things that own a `io::Write` tend to provide a reference to it and a way to retrieve it when the wrapper is no longer needed. See `BufWriter`'s [`get_ref`](https://doc.rust-lang.org/std/io/struct.BufWriter.html#method.get_ref) and [`into_inner`](https://doc.rust-lang.org/std/io/struct.BufWriter.html#method.into_inner) methods for an example.

It provides a nice way to do read-only queries against the underlying value (like getting [the peer IP address](https://doc.rust-lang.org/std/net/struct.TcpStream.html#method.peer_addr) from a `TcpStream` struct) and then ultimately do something else with it when you're done (like reading a TCP response back after writing the request).

I haven't added an `into_inner` method yet, but I think it's a good idea.

src/text/writer.rs
zslayton: The strongest use case I'm aware of for having both is LST appends. Since the text writer doesn't maintain an LST itself and we don't yet have a `Writer` trait to code to, I'm for punting on this.

src/text/writer.rs
zslayton: See issue #92.

src/text/writer.rs
zslayton: Yep, good call.

src/text/encoder.rs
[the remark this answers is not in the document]
almann: Why not have the call site be:

```rust
TextEncoder::encode_datetime(output, &Utc::now().into());
```

The problem I have with this API is not the generics per se (although I do think it adds cognitive overhead for such a trivial operation), it is more the *consumption* where it need not have those semantics.

If we really wanted it to be generic, `AsRef<DateTime<FixedOffset>>` makes more sense to me, but even still that is overkill and doesn't work for the use case you have in mind.

src/text/writer.rs
almann: Sorry, "this type stores copies of the working fields and annotations" is what I meant.

src/text/encoder.rs
```diff
@@ -0,0 +1,264 @@
...
+
+    /// Writes a string value to `output` as an Ion symbol.
+    pub fn encode_str_as_symbol<S: AsRef<str>, W: io::Write>(
+        output: &mut W,
+        value: S,
+    ) -> IonResult<()> {
+        write!(output, "'{}'", value.as_ref())?;
+        Ok(())
+    }
```
almann, line 108: Note that this API has an interesting implication of consuming `value` (might not be what the user wants if they pass a `String`), now looking down at the writer, you already only take in `&str` so it is a trivial copy, but then I have to ask, why are we being generic here and not just make it `&str`?

To illustrate my point:

```rust
use std::convert::*;

fn print_str<S: AsRef<str>>(s: S) {
    println!("{}", s.as_ref());
}

fn main() {
    let s = String::from("blah");
    print_str(s);
    
    // Cannot do this! s is moved into `print_str`.
    //print_str(s);
}
```

src/text/writer.rs
almann: The use case around `into_write` or some such makes sense to me (assuming your at a place that it makes sense to return the underlying `Write`), breaking encapsulation to get at the underlying `Write` by reference seems a little weak (kind of makes sense for a generic API like `BufWriter`, but less clear about what I can do in any reasonable way around an Ion writer.

It's not a big deal, but I always like to have less API unless we have good reasons to expose these underlying bits.

src/text/encoder.rs
almann: I am okay with `AsRef<str>` here after talking to you offline--I think that it is "free" so that works well.

Cargo.toml
zslayton: Huh! I didn't know that, thanks.

Issue #17, opened 2020-06-06
Fractional seconds support for binary Timestamps


zslayton:
The `BinaryCursor` currently skips the fractional seconds support when reading a `Timestamp` into a `DateTime<FixedOffset>`.

zslayton:
Closing this out as the `BinaryCursor` now reads fractional seconds:

https://github.com/amzn/ion-rust/blob/0e3d46ee0b115c496a5192a0b2bb124c8717a9d7/src/binary/cursor.rs#L581-L593

It always returns the Timestamp as a `DateTime` and so converts the fractional seconds to nanoseconds. #274 tracks lifting this limitation.

PR #116, opened 2021-01-11
Added system-level binary writer.

zslayton:
*Description of changes:*

This PR adds a system-level binary writer to the crate.

Notes:
* Rather than not supporting `timestamp` fractional seconds at all, the reader and writer now support nanosecond precision (and *only* nanosecond precision). This is a stepping stone to full support for timestamps.

*Out of scope for this PR:*
* A user-level binary writer, with symbol table management and LST appends.
* A proper `SymbolToken`-centric API.
* A way to specify precision when writing a given `DateTime` as a timestamp. (#17 will be updated to indicate that the reader and writer now support nanoseconds precision and nothing else.)

### DIFF
--- a/src/binary/cursor.rs
+++ b/src/binary/cursor.rs
@@ -548,7 +548,11 @@ impl<R: IonDataSource> Cursor for BinaryIonCursor<R> {
             subsecond_coefficient = self.read_int(coefficient_size_in_bytes)?.value();
         }
 
-        // FIXME: This always assumes nanosecond precision.
+        // This turns the encoded decimal fractional seconds into a number of nanoseconds to set on
+        // the DateTime value we're building.
+        // TODO: This assumes that the provided DateTime had nanosecond precision.
+        //      Offer an read_* API that surfaces information about the encoded timestamp's
+        //      precision.
         const NANOSECONDS_PER_SECOND: f64 = 1_000_000_000f64;
         let fractional_seconds = subsecond_coefficient as f64 * 10f64.powf(subsecond_exponent as f64);
         let nanoseconds = (fractional_seconds * NANOSECONDS_PER_SECOND).round() as u32;
--- a/src/binary/cursor.rs
+++ b/src/binary/cursor.rs
@@ -558,7 +558,6 @@ impl<R: IonDataSource> Cursor for BinaryIonCursor<R> {
             .with_nanosecond(nanoseconds);
 
         if naive_datetime.is_none() {
-            println!("Depth: {}, Index at depth: {}", self.cursor.depth, self.cursor.index_at_depth);
             return decoding_error(
                 format!(
                     "{}: year={}, month={}, day={}, hour={}, minute={}, second={}, exp={}, coeff={}, nanos={}",
--- a/src/binary/cursor.rs
+++ b/src/binary/cursor.rs
@@ -333,6 +333,10 @@ impl<R: IonDataSource> Cursor for BinaryIonCursor<R> {
         self.cursor.value.header.ion_type
     }
 
+    fn is_null(&self) -> bool {
+        self.cursor.value.is_null
+    }
+
     fn annotation_ids(&self) -> &[SymbolId] {
         let num_annotations = self.cursor.value.number_of_annotations as usize;
         if num_annotations == 0 {
@@ -492,6 +496,8 @@ impl<R: IonDataSource> Cursor for BinaryIonCursor<R> {
     fn read_datetime(&mut self) -> IonResult<Option<DateTime<FixedOffset>>> {
         read_safety_checks!(self, IonType::Timestamp);
 
+        let datetime_start_offset = self.cursor.bytes_read;
+
         let offset_minutes = self.read_var_int()?.value();
         let year = self.read_var_uint()?.value();
 
@@ -500,6 +506,9 @@ impl<R: IonDataSource> Cursor for BinaryIonCursor<R> {
         let mut hour = 0;
         let mut minute = 0;
         let mut second = 0;
+
+        let mut subsecond_exponent = 0;
+        let mut subsecond_coefficient = 0;
 
         loop {
             if self.finished_reading_value() {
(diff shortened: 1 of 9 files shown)


src/binary/cursor.rs
```diff
@@ -333,6 +333,10 @@ impl<R: IonDataSource> Cursor for BinaryIonCursor<R> {
         self.cursor.value.header.ion_type
     }
 
+    fn is_null(&self) -> bool {
+        self.cursor.value.is_null
+    }
+
```
zslayton, line 339: In this file: 
* Exposing an `is_null` method for use in the user-level `Reader`
* Adding support for reading timestamps with fractional seconds. For interop with the `chrono` crate, fractional seconds are turned into `nanoseconds` on the resulting `DateTime` value.

src/binary/int.rs
```diff
@@ -55,6 +56,26 @@ impl Int {
         })
     }
```
zslayton, line 58: In this file:
* Adding write support for [the low-level encoding primitive `Int`](https://amzn.github.io/ion-docs/docs/binary.html#uint-and-int-fields), not to be confused with the Ion `integer` type.

src/binary/uint.rs
```diff
@@ -47,7 +47,7 @@ impl UInt {
     }
```
zslayton, line 47: In this file:
* A previous commit had added write support for [the low-level primitive `UInt`](https://amzn.github.io/ion-docs/docs/binary.html#uint-and-int-fields). However, the method name didn't follow the naming convention used elsewhere, which refers to the Rust data type being accepted as a value to write out. `UInt::write_uint` is thus being renamed `UInt::write_u64`.

src/binary/var_int.rs
```diff
@@ -1,6 +1,7 @@
 use crate::data_source::IonDataSource;
 use crate::result::{decoding_error, IonResult};
 use std::mem;
```
zslayton, line 3: In this file:
* Adding write support for [the low-level encoding primitive `VarInt`](https://amzn.github.io/ion-docs/docs/binary.html#varuint-and-varint-fields).

src/binary/var_uint.rs
```diff
@@ -1,6 +1,7 @@
 use crate::data_source::IonDataSource;
```
zslayton, line 1: In this file:
* Adding write support for [the low-level encoding primitive `VarUInt`](https://amzn.github.io/ion-docs/docs/binary.html#varuint-and-varint-fields).

zslayton (commented):
Added some file-level comments to explain the smaller diffs outside of `src/binary/writer.rs`.

src/binary/writer.rs
```diff
@@ -0,0 +1,1324 @@
...
+
+    // Visits all of the IoRanges belonging to this EncodingLevel and notes their total length.
+    // This length will be written out as a length prefix for the container.
+    fn calculate_final_size(&self, io_ranges: &mut [Range<usize>]) -> usize {
+        io_ranges[self.td_io_range_index..]
+            .iter()
+            .map(|r| r.len())
+            .sum()
+    }
+}
```
therapon, line 105: I am missing something here, or rather, I am not sure I get the whole picture. The input `io_ranges` and the stored `td_io_range.index` are `usize` based ranges. How do we know when iterating over `io_range` that we are not going over values in the range that are *not* for our container but some other container? I'll keep reading to see if I can figure out the answer, but I am assuming when this function is called the tail end of `io_ranges` starting at `td_io_range_index` contains the information for this container which is also the last "processed" container.

I am assuming an explicit contract here that this method is called at the end of a container which would guarantee that the tail range in `io_ranges` contains the relevant information and nothing extra.

src/binary/writer.rs
```diff
@@ -0,0 +1,1324 @@
...
+    io_ranges: Vec<IoRange>,
+    // Stack for tracking step_in/step_out
+    levels: Vec<EncodingLevel>,
+    // An io::Write implementation to be used as a sink for encoded data.
+    out: W,
+    // The field ID of the current value. If the writer is not in a struct, this will be None.
+    field_id: Option<SymbolId>,
+    // A Vec into which all EncodingLevels can store their annotations. Sharing this Vec avoids
+    // allocating new space for each container the writer steps into.
+    annotations_all_levels: Vec<SymbolId>,
```
therapon, line 127:
```suggestion
    annotations_acc: Vec<SymbolId>,  // accumulator of all annotations seen thus far
```

src/binary/writer.rs
```diff
@@ -0,0 +1,1324 @@
...
+            for annotation_id in annotations {
+                VarUInt::write_u64(&mut writer.buffer, *annotation_id as u64)?;
+            }
+            Ok(())
+        })?);
+        let annotation_sequence_encoded_length = annotations_seq_io_range.len();
+
+        // Encode the length of the annotations sequence as a VarUInt.
+        let _ = mem::replace(annotations_seq_length_io_range, self.encode_to_buffer(|writer| {
+            let _num_bytes = VarUInt::write_u64(
```
therapon, line 289: do we need `_num_bytes`. It looks that is it not used? Replace with `_`?

src/binary/writer.rs
zslayton: >  I am assuming when this function is called the tail end of io_ranges starting at td_io_range_index contains the information for this container which is also the last "processed" container.

This is correct. 

When the writer steps into a container, it pushes a number of placeholder `IoRange`s that will need to be populated later, when more length information is available. Afterwards, as the writer continues writing new nested values inside the container (potentially including other containers), all of their `IoRange`s will be pushed *after* the placeholders we already added. When `step_out()` is called, we look at the `EncodingLevel` that's popped off the stack. By definition, all of the `IoRange`s that follow its type descriptor `IoRange` (at offset `self.td_io_range_index`) belong to nested values.

> I am assuming an explicit contract here that this method is called at the end of a container which would guarantee that the tail range in io_ranges contains the relevant information and nothing extra.

Yep! `calculate_final_size()` is only called during `step_out()`. At the moment a streaming writer is stepping out of a container, there cannot be any values after the container yet.

therapon (approved):

src/binary/writer.rs
zslayton: I'm wary of using `_` by itself too often, because as a reader visiting new code it usually means I have to fire up an IDE to get hints about what's being discarded. I still do it sometimes, including just above this code snippet; [the first time I wrote an explanation for why we could ignore it](https://github.com/amzn/ion-rust/blob/f246c0a6189ee8065a537299800eca24c204a7c8/src/binary/writer.rs#L275-L277), but didn't add the same disclaimer elsewhere.

Prefixing the variable name with `_` is enough to satisfy the compiler/style guidelines as it flags it as intentionally unused. I happened to know what the value I was ignoring was, so I used a meaningful variable name instead of adding a comment.

```
// Ignore the number of bytes that were written.
let _ = VarUInt::write_u64(...);
// vs. 
let _num_bytes = VarUInt::write_u64(...);
```

I'm unsure if there's a formal style guideline around this; if so, I'm happy to defer to it.

PR #117, opened 2021-01-20
Added convenience methods to the public API

zslayton:
*Description of changes:*

This PR builds on pending PR #116, which adds a system-level binary writer. The diff has been generated from the `bin-writer` branch for easier review. Once `bin-writer` is merged, this PR will be rebased against `master`.

This PR tweaks a number of APIs:

* The binary reader now exposes functions that return the offset, length, and range for each value's field name, annotations, header, and body.
* Renamed `SymbolTableEventHandler` trait to `SystemEventHandler so it can be used for reacting to IVMs as well as symbol table definitions/appends.
* The text and binary writers now offer `output()` and `output_mut()` functions that allow the user to get a reference to the underlying io::Write sink.
* `blob_ref_map` and `clob_ref_map` are now part of the `Reader` trait.

Many of these changes are being leveraged in the new `ion-cli` command `inspect`, which displays a given binary Ion stream side-by-side with its equivalent text Ion. A PR introducing the `inspect` command will be opened in `ion-cli` after this new functionality is merged.

### DIFF
--- a/src/system_event_handler.rs
+++ b/src/system_event_handler.rs
@@ -1,0 +1,15 @@
+use crate::SymbolTable;
+
+/// Functions that will be called when the reader handles system-level events that would otherwise
+/// not be surfaced to the user level.
+pub trait SystemEventHandler {
+    // TODO: It would be better to create structs to hold the arguments to each of these
+    //       callbacks so information can be added over time without it being a breaking change.
+
+    /// Invoked when the cursor encounters an Ion Version Marker.
+    fn on_ivm(&mut self, _ion_version: (u8, u8)) {}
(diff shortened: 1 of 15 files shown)


src/system_event_handler.rs
```diff
@@ -0,0 +1,15 @@
+use crate::SymbolTable;
+
+/// Functions that will be called when the reader handles system-level events that would otherwise
+/// not be surfaced to the user level.
+pub trait SystemEventHandler {
+    // TODO: It would be better to create structs to hold the arguments to each of these
+    //       callbacks so information can be added over time without it being a breaking change.
```
therapon, line 7: +1

therapon (approved):

PR #120, opened 2021-01-28
Field/annotation lengths are cleared as needed

zslayton:
* When the cursor reads a value that has no annotations, the internal field that records the number of bytes used to encode the current value's annotations is set to 0.
* When the cursor reads a value that is not a struct field, the internal field that records the number of bytes used to encode the current value's field ID is set to 0.
* Added a unit test to verify this behavior.

*Description of changes:*

This will fix amzn/ion-cli#9 when it's able to update its `ion-rust` dependency.


jobarr-amzn (approved):
As discussed out of band I'd still like to see named sub arrays rather than byte offsets in the test, but that's aesthetic to a degree and optional. 

E.g. 
```rust
assert_eq!(cursor.raw_bytes(), Some(&foo_bytes));
assert_eq!(cursor.raw_field_id_bytes(), None);
assert_eq!(cursor.raw_annotations_bytes(), None);
assert_eq!(cursor.raw_header_bytes(), Some(&foo_header));
assert_eq!(cursor.raw_value_bytes(), Some(&foo_value));
```
 or however the Rust syntax works out.

PR #121, opened 2021-02-12
Initial impl of Decimal, Timestamp, and text scalar parsing

zslayton:
This PR lays the groundwork for a fully functional text Ion reader.

Unfortunately, this PR is quite large. It was initially going to only be text scalar parsing, but in order to fully test each type's parser I ended up needing to write proper `Decimal` and `Timestamp` types.

**Changes:**
* Added a `Decimal` type that:
   * supports arbitrary precision
   * avoids allocating whenever possible
   * defines equality in accordance with the Ion data model (including support for negative zero)
* Added a `Timestamp` type that:
  * supports variable precision, including fractional seconds of arbitrary precision
  * defines equality in accordance with the Ion data model (including offset and fractional second precision)
  * provides a builder API that restricts users to calling a set of methods that will lead to a valid timestamp
* Adds a text parsing module for each of the following scalar types:
  * `null`
  * `bool`
  * `int`
  * `float`
  * `decimal`
  * `timestamp`
  * `string`
  * `symbol`
  * `blob`
* Created a placeholder `TextReader` that can read a single top-level Ion value of one of the above types.

Some of the APIs are incomplete. For example, the `Timestamp` type can be instantiated and compared to other `Timestamp`s for equality but does not yet offer accessors for the individual time unit fields. The functionality present is largely what was needed for parsing and unit testing. Future PRs will shore this up.

**Out of scope:**
* Clobs
* S-expressions
* Annotations
* Field names
* Full text reader functionality like `step_in` and `step_out`
* Zero-allocation text parsing. A variety of constructs require temporary buffers, including escape characters, multi-part long strings, etc. With some refactoring these can be modified to leverage reusable buffers. I have tried to mark each of these with a `TODO`.
* Integers of arbitrary size (currently limited to `i64`)
* Changing the binary reader/writer to work with the new `Decimal` and `Timestamp` types instead of `BigDecimal` and `DateTime<FixedOffset>`.
* A `SymbolToken` type + APIs.


src/text/parsers/float.rs
```diff
@@ -0,0 +1,133 @@
...
+use nom::combinator::{map_res, opt, recognize, map};
+use nom::bytes::streaming::{tag};
+use nom::character::streaming::one_of;
+use crate::text::parsers::numeric_support::{digits_before_dot, floating_point_number, exponent_digits};
+use std::num::ParseFloatError;
+use std::str::FromStr;
+
+/// Matches the text representation of a float value and returns the resulting [f64]
+/// as a [TextStreamItem::Float].
+pub(crate) fn parse_float(input: &str) -> IResult<&str, TextStreamItem> {
```
therapon, line 15: Mostly a curiosity comment: there are similarities between float parsing and double parsing, could we, in the future, refactor a bit or does `nom` make that difficult?

src/text/parsers/numeric_support.rs
```diff
@@ -0,0 +1,97 @@
...
+pub(crate) fn digits_before_dot(input: &str) -> IResult<&str, &str> {
+    alt((
+        tag("0"),
+        recognize(
+            pair(leading_digit, trailing_digits)
+        )
+    ))(input)
+}
+
+/// Recognizes the first digit of a multi-digit base-10 integer. (i.e. Any digit but zero.)
```
therapon, line 33:
```suggestion
/// Recognizes the first non-0 digit of a multi-digit base-10 integer. (i.e. Any digit but zero.)
```

src/text/parsers/timestamp.rs
```diff
@@ -0,0 +1,374 @@
...
+    let number_of_digits = fractional.len();
+    // If the precision is less than or equal to nanoseconds...
+    if number_of_digits <= 9 {
+        // Convert the number to nanoseconds and make a note of its original precision.
+        let power = 9 - number_of_digits;
+        let nanoseconds = trim_zeros_expect_u32(fractional, "fractional seconds") * 10u32.pow(power as u32);
+        setter = setter.with_nanoseconds_and_precision(nanoseconds, number_of_digits as u32);
+    } else {
+        // Otherwise, the number's precision is great enough that we'll need to construct a Decimal
+        // to store it with loss of fidelity.
```
therapon, line 139:
```suggestion
        // to store it without loss of fidelity.
```

src/types/coefficient.rs
```diff
@@ -0,0 +1,133 @@
...
+            fn from(value: $t) -> Coefficient {
+                let sign = if value < <$t>::zero() { Sign::Negative } else { Sign::Positive };
+                Coefficient::new(sign, value)
+            }
+        }
+    )*)
+}
+impl_coefficient_from_signed_int_types!(i8, i16, i32, i64, i128, isize);
+
+// `BigInt` can't represent -0, so this is technically a lossy operation.
```
therapon, line 83: Ah that old chestnut again. Is the plan to "wrap" the type and track the `-0` case?

therapon (approved):
Minor comments and I admit towards the end I got a little cross-eyed :) 

This is great for a start. We can iterate and create smaller PRs after this making it easier to go over :) 

:shipit:

src/text/parsers/blob.rs
```diff
@@ -0,0 +1,87 @@
...
+use crate::text::TextStreamItem;
+
+/// Matches the text representation of a blob value, decodes it, and returns the resulting bytes
+/// as a [TextStreamItem::Blob].
+pub(crate) fn parse_blob(input: &str) -> IResult<&str, TextStreamItem> {
+    map_res::<_, _, _, _, DecodeError, _, _>(
+        delimited(
+            pair(tag("{{"), opt(whitespace)),
+            recognize_base64_data,
+            pair(opt(whitespace), tag("}}")),
```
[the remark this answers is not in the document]
zslayton: It's not required, but Rust style allows (and even encourages) it because it makes it easy to add and remove items from the list. Any line can be cut/pasted into a different position without needing to rejigger the commas. After this PR is merged, I'll do a crate-wide pass with `rustfmt` to clean up anything it disagrees with.

src/text/parsers/decimal.rs
```diff
@@ -0,0 +1,211 @@
...
+        parse_equals("-0d5 ", Decimal::negative_zero_with_exponent(5));
+        parse_equals("-0d-5 ", Decimal::negative_zero_with_exponent(-5));
+        parse_equals("305d1 ", Decimal::new(305, 1));
+        parse_equals("305d-1 ", Decimal::new( 305, -1));
+        parse_equals("111_111d222 ", Decimal::new(111_111, 222));
+        parse_equals("111_111d-222 ", Decimal::new(111_111, -222));
+        parse_equals("111_111d222_222 ", Decimal::new(111_111, 222_222));
+        parse_equals("-279d0 ", Decimal::new(-279, 0));
+        parse_equals("-999_9d9_9 ", Decimal::new(-9_999, 99));
+        parse_equals("-999_9d-9_9 ", Decimal::new(-9_999, -99));
```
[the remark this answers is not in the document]
zslayton: Yes, Ion's text encoding requires that many of the scalar types end in a stop character. Here's an excerpt from [the spec section on Real Numbers](https://amzn.github.io/ion-docs/docs/spec.html#real-numbers):
> In the text notation, real values must be followed by one of the fifteen numeric stop-characters: `{}[](),\"\'\ \t\n\r\v\f`.

(It's tough to see in the font above but one of those characters is a space.)

Additionally, `nom` will reject any input that might be incomplete data. For example, the text `314` might be a complete integer or it might be the first half of a `314159`. Adding a trailing space removes this ambiguity.

src/text/parsers/float.rs
zslayton: Very probably; I would need to carefully review which functions are used where. There's probably some room for consolidation.

src/types/coefficient.rs
zslayton: I've replaced the `From<Decimal> for BigDecimal` and `From<Coefficient> for BigInt` implementations with `TryFrom` implementations that fail if the value being converted is a negative zero (and so would cause data loss).

If we need one, we can add a `From<Decimal> for (Sign, BigDecimal)` or other option for guaranteed conversion.

src/types/decimal.rs
```diff
@@ -0,0 +1,258 @@
...
+        // From the BigInt docs: "Note that a positive exponent indicates a negative power of 10."
+        let exponent = -negative_exponent;
+
+        Decimal::new(Coefficient::new(sign, magnitude), exponent)
+    }
+}
+
+/// Make a BigDecimal from a Decimal. BigDecimal doesn't have a distinct -0, so this is technically
+/// a lossy operation.
+impl From<Decimal> for BigDecimal {
```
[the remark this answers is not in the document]
zslayton: Fixed.

src/text/parsers/numeric_support.rs
zslayton: I'd prefer to leave this as-is. In my opinion the rephrasing suggests that this will look at several digits and match the first non-zero one rather than matching the first digit, which must be non-zero.

src/text/parsers/blob.rs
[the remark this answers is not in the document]
therapon: Maybe add a TODO (or Issue) to run rustfmt as part of the CI or commit.

src/text/parsers/decimal.rs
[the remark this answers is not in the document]
therapon: Thanks. I forgot the Ion spec.

therapon (approved):

Issue #124, opened 2021-02-23
Implement text clob writing


zslayton:
The `TextWriter` has text clob writing [stubbed out with a `todo!()` implementation](https://github.com/amzn/ion-rust/blob/master/src/text/writer.rs#L262).

Here's the [relevant portion of the spec](https://amzn.github.io/ion-docs/docs/spec.html#clob).

zslayton:
@desaikd addressed this in #128.

PR #128, opened 2021-03-06
added changes for writing clob value

desaikd:
*Issue #124*

*Description of changes:*
This PR works on adding support for writing `clob` using text writer.

*Test:*
added test for writing `clob`.

### DIFF
--- a/src/text/writer.rs
+++ b/src/text/writer.rs
@@ -9,6 +9,37 @@ pub struct TextWriter<W: Write> {
     annotations: Vec<String>,
     field_name: Option<String>,
     containers: Vec<IonType>,
+    STRING_ESCAPE_CODES: Vec<String>,
+}
+
+pub const ZERO_PADDING: [&str; 8] = ["", "0", "00", "000", "0000", "00000", "000000", "0000000"];
+
+/**
+ * String escape codes, for Ion Clob.
+ */
+fn string_escape_code_init() -> Vec<String> {
+    let mut string_escape_codes = vec!["".to_string(); 256];
+    string_escape_codes[0x00] = "\\0".to_string();
+    string_escape_codes[0x07] = "\\a".to_string();
+    string_escape_codes[0x08] = "\\b".to_string();
+    string_escape_codes['\t' as usize] = "\\t".to_string();
+    string_escape_codes['\n' as usize] = "\\n".to_string();
+    string_escape_codes[0x0B] = "\\v".to_string();
+    string_escape_codes[0x0C] = "\\f".to_string();
+    string_escape_codes['\r' as usize] = "\\r".to_string();
+    string_escape_codes['\\' as usize] = "\\\\".to_string();
+    string_escape_codes['\"' as usize] = "\\\"".to_string();
+    for i in 1..0x20 {
+        if string_escape_codes[i] == "" {
+            let s = format!("{:x}", i);
+            string_escape_codes[i] = "\\x".to_owned() + ZERO_PADDING[2 - s.chars().count()] + &s;
+        }
+    }
+    for i in 0x7F..0x100 {
+        let s = format!("{:x}", i);
+        string_escape_codes[i] = "\\x".to_owned() + &s;
+    }
+    return string_escape_codes;
 }
 
 impl<W: Write> TextWriter<W> {
@@ -20,6 +51,7 @@ impl<W: Write> TextWriter<W> {
             annotations: vec![],
             field_name: None,
             containers: vec![],
+            STRING_ESCAPE_CODES: string_escape_code_init(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/text/writer.rs
```diff
@@ -259,7 +291,21 @@ impl<W: Write> TextWriter<W> {
 
     /// Writes the provided byte array slice as an Ion clob.
     pub fn write_clob(&mut self, _value: &[u8]) -> IonResult<()> {
```
zslayton, line 293: Now that this function is defined, the `_value` variable is being used. The `_` name prefix is a hint to the compiler that we're ignoring this value intentionally; it's no longer being ignored, so you can remove the `_`.

src/text/writer.rs
```diff
@@ -259,7 +291,21 @@ impl<W: Write> TextWriter<W> {
...
-        todo!()
+        // clob_value to be written based on defined STRING_ESCAPE_CODES.
+        let mut clob_value = String::from("");
+        for i in 0.._value.len() {
+            let c = (_value[i] & 0xff) as char;
+            let escaped_byte = &self.string_escape_codes[c as usize];
```
zslayton, line 298: Java doesn't have an unsigned byte type, so when developers go to convert a `byte` to an `int` or `char`, they use `& 0xff` to promote the signed byte to a 32-bit integer that can represent values >=128 as positive. For example:

```java
class Main {  
  public static void main(String args[]) {
    byte b = (byte) 0x81;
    int i = b & 0xff;
    System.out.println(b);
    System.out.println(i);
  } 
}
/* 
* == Output ==
* -127
* 129
*/
```

In contrast, Rust does have an unsigned byte type: `u8`.  Here, `value` is known to be a slice of unsigned bytes (`&[u8]`), so calling `& 0xff` on each value in the slice will have no effect.

src/text/writer.rs
```diff
@@ -259,7 +291,21 @@ impl<W: Write> TextWriter<W> {
...
+            if escaped_byte != "" {
+                clob_value = format!("{}{}", clob_value, escaped_byte);
+            } else {
+                clob_value = format!("{}{}", clob_value, &c);
```
zslayton, line 302: This is clever! Unfortunately, it requires `clob_value` to be reallocated in each loop iteration. Here are some ways you can append more text to the end of the existing string without making a new one:
* [`String::push`](https://doc.rust-lang.org/std/string/struct.String.html#method.push) will add the contents a given `char` to the end of an existing `String`.
* [`String::push_str`](https://doc.rust-lang.org/std/string/struct.String.html#method.push_str) will add the contents of a `&str` to the end of an existing `String`.
* [`write!`](https://doc.rust-lang.org/std/macro.write.html) also appends to the end of a given `String` (or other output), but allows you to use the same formatting flags as `format!` and `println!`.

src/text/writer.rs
```diff
@@ -364,9 +410,11 @@ mod tests {
     }
 
     #[test]
-    #[ignore]
     fn write_clob() {
-        todo!();
+        writer_test(
+            |w| w.write_clob(&[b'a', b'"', b'\'', b'\n']),
```
zslayton, line 415: For an easier way to get a byte array representing text, I suggest making a `&str` and then calling [`as_bytes()`](https://doc.rust-lang.org/std/primitive.str.html#method.as_bytes) on it:
```rust
"Algeria".as_bytes()
```
You can also use `r""` syntax to create a [raw string literal](https://doc.rust-lang.org/reference/tokens.html#raw-string-literals), allowing you to write characters you would otherwise have to escape.

src/text/writer.rs
```diff
@@ -364,9 +410,11 @@ mod tests {
...
+            "{{\"a\\\"'\\n\"}}\n",
```
zslayton, line 416: I know the other unit tests in this module are pretty sparse, but would you mind adding a few more `writer_test`s to this one? Try an ascii string, an ascii string with special escape characters, some emoji, etc.

src/text/writer.rs
```diff
@@ -290,16 +286,16 @@ impl<W: Write> TextWriter<W> {
     }
 
     /// Writes the provided byte array slice as an Ion clob.
-    pub fn write_clob(&mut self, _value: &[u8]) -> IonResult<()> {
+    pub fn write_clob(&mut self, value: &[u8]) -> IonResult<()> {
         // clob_value to be written based on defined STRING_ESCAPE_CODES.
-        let mut clob_value = String::from("");
-        for i in 0.._value.len() {
-            let c = (_value[i] & 0xff) as char;
+        let mut clob_value = String::new();
```
zslayton, line 291: We can save ourselves a few reallocations by making an educated guess at how long `clob_value` will be and using [`String::with_capacity`](https://doc.rust-lang.org/std/string/struct.String.html#method.with_capacity) to set aside that much memory.

In the best case, each byte from `value` will be displayable ASCII and so will add one byte to `clob_value`. In the worst case, each byte from `value` will be non-ASCII and will add 4 bytes (`\xHH`) to `clob_value`. (Escaped ASCII values like `\n` add two bytes and are neither the best nor worst case.)

In any case, we also have to add enough space for the `{{` and `}}`, 4 more bytes.

To be very conservative, we can start by accommodating the worst case:
```suggestion
        const NUM_DELIMITER_BYTES: usize = 4; // {{}}
        const NUM_HEX_BYTES_PER_BYTE: usize = 4; // \xHH
        // Set aside enough memory to hold a clob containing all hex-encoded bytes
        let mut clob_value = String::with_capacity((value.len() * NUM_HEX_BYTES_PER_BYTE) + NUM_DELIMITER_BYTES);
```

We could also go with a middle-ground approach that multiplied `value.len()` by 1, 2, or 3 instead, occasionally requiring reallocations but wasting less memory in the common case.

zslayton (approved):
Nice work!

src/text/writer.rs
```diff
@@ -288,7 +288,13 @@ impl<W: Write> TextWriter<W> {
     /// Writes the provided byte array slice as an Ion clob.
     pub fn write_clob(&mut self, value: &[u8]) -> IonResult<()> {
         // clob_value to be written based on defined STRING_ESCAPE_CODES.
-        let mut clob_value = String::new();
+        const NUM_DELIMITER_BYTES: usize = 4; // {{}}
+        const NUM_HEX_BYTES_PER_BYTE: usize = 4; // \xH
```
zslayton, line 292: nit: the 4-byte format comment should have two Hs: `\xHH`

zslayton (approved):

Issue #125, opened 2021-02-23
Implement text clob parsing


zslayton:
This can largely piggy back on the parser for `string`, which supports both short and long-form strings. The main differences are:

* Long-form `string`s allow comments to appear between `'''`-delimited segments while `clob`s [do not permit comments anywhere in the body](https://amzn.github.io/ion-docs/docs/spec.html#clob):

> In the text format, clob values use similar syntax to blob, but the data between braces must be one string. The string may only contain legal 7-bit ASCII characters, using the same escaping syntax as string and symbol values. This guarantees that the value can be transmitted unscathed while remaining generally readable (at least for western language text). Like blobs, clobs disallow comments everywhere within the value.

* Clobs only allow displayable ASCII characters and `\n` to be included in their encoding as-is. All other characters [must be written as unicode escape sequences](https://amzn.github.io/ion-docs/docs/stringclob.html#ion-clob):

> The rules for the quoted strings within a clob follow the similarly to the string type, with the following exceptions. Unicode newline characters in long strings and all verbatim ASCII characters are interpreted as their ASCII octet values. Non-printable ASCII and non-ASCII Unicode code points are not allowed un-escaped in the string bodies. Furthermore, the following table describes the clob string escape sequences that have direct octet replacement for both all strings.

desaikd:
This has been fixed with #130

PR #130, opened 2021-03-18
added changes for text clob parsing

desaikd:
*Issue #125*

*Description of changes:*
This PR works on creating text `clob` parser. 

*Changes:*
- Created `parse_clob` similar to `string` parser with an extra parsing step to verify `{{}}` for `clob` value.
- Instead of returning `string` in the helper parsing methods, changed to use `Vec<u8>` for correctly parsing extended ASCII values. 
- Changed the order of arguments for `assert_eq!` to be same as `parse_equals` method for consistency.
```
assert_eq!(actual, expected)

actual -> value parsed with ion-rust text parsers 
expected -> expected value to be used for equivalence
```

*Test:*
Added unit tests for `clob` type check and parsing tests.

### DIFF
--- a/src/text/parsers/clob.rs
+++ b/src/text/parsers/clob.rs
@@ -1,0 +1,152 @@
+use crate::text::parsers::text_support::{escaped_char, escaped_newline, StringFragment};
+use crate::text::parsers::whitespace;
+use crate::text::TextStreamItem;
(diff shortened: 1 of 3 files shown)


zslayton (approved):
🚢!

Issue #114, opened 2021-01-07
Ion Element/Value API


rhinomode:
Hello, this is a request for enhancing/extending the functionality provided to Rust Ion users.

The [IonJava ](https://github.com/amzn/ion-java) and [IonElement](https://github.com/amzn/ion-element-kotlin) libraries provide useful abstractions over Ion readers and writers (which themselves are very useful).

As a user of the `ion-rs` crates, I find myself wanting a API similar to those found in the Java libraries (above) or the [`serde-rs/json`](https://github.com/serde-rs/json) crate. 

The reason is simple: to make sensible code, I must build structures (`enum` even) around the `ion-rs/*` functions and constants. Then I must also consider all of the other low-level concerns of (lazily) materializing, (avoiding) clones of effectively immutable data to satisfy the borrow checker, etc. etc.

I suspect can and should be "solved once" as in the Java prior art.

Thanks,
Chuck

rhinomode:
There's some[ interesting test code](https://github.com/amzn/ion-rust/blob/master/ion-c-sys/tests/value.rs) from Almann along these lines in the `ion-c-sys` crate - obviously not meant for consumption outside the test harness.

almann:
#132 also makes further progress on this.

zslayton:
The `Element` API has been heavily revised (see: #479, #485, #486, #487); these changes will be included in the next release.

PR #131, opened 2021-03-23
Adds a preliminary `Element` DOM-like API.

almann:
Provides an _borrowed_/_owned_ implementations of the traits in the `value` module.  The borrowed implementation is tied to a lifetime for string/binary data (but potentially more in the future).  The owned implementation is completely self contained.

* Defines `SymbolToken` and `ImportSource` traits that is the basis for full symbol/field name/annotation support in `ion-rs`, but also is required for the `Element` trait.
* APIs that return `Iterator` types currently use a `Box` to avoid the need for generic associated types (GAT).
* Captures the variance of integers in Ion with the `IntAccess` trait and the `AnyInt` enum.
* Currently implements the `Struct` trait via `Vec` of tuples, which will have to be refactored for O(N) access but is suitable for the iterator APIS.

Starts to address #114, but is not nearly complete.  This commit is to provide the initial structure of this API for collaborative development, and to get early review.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -7,6 +7,7 @@ pub mod cursor;
 pub mod data_source;
 pub mod text;
 pub mod types;
+pub mod value;
 
 pub mod constants;
 mod reader;
--- a/src/value/borrowed.rs
+++ b/src/value/borrowed.rs
@@ -1,0 +1,229 @@
+// Copyright Amazon.com, Inc. or its affiliates.
+
+//! Provides borrowed implementations of [`SymbolToken`], [`Element`] and its dependents.
+//!
+//! Specifically, all implementations are tied to some particular lifetime, generally linked
+//! to a parser implementation of some sort or some context from which the borrow can occur.
+//! For simple values, the values are inlined (see [`BorrowedValue`]), but for things that are
+//! backed by octets or string data, `&[u8]` and `&str` are used.
+
+use super::{Element, ImportSource, Sequence, Struct, SymbolToken};
+use crate::types::SymbolId;
+use crate::value::AnyInt;
+use crate::IonType;
+
+/// A borrowed implementation of [`ImportSource`].
+#[derive(Debug, Copy, Clone)]
+pub struct BorrowedImportSource<'val> {
+    table: &'val str,
+    sid: SymbolId,
+}
+
+impl<'val> BorrowedImportSource<'val> {
+    pub fn new(table: &'val str, sid: SymbolId) -> Self {
+        Self { table, sid }
+    }
+}
+
+impl<'val> ImportSource for BorrowedImportSource<'val> {
+    fn table(&self) -> &str {
+        self.table
+    }
+
+    fn sid(&self) -> usize {
+        self.sid
+    }
+}
+
+/// A borrowed implementation of [`SymbolToken`].
+#[derive(Debug, Copy, Clone)]
+pub struct BorrowedSymbolToken<'val> {
+    text: Option<&'val str>,
+    local_sid: Option<SymbolId>,
+    source: Option<BorrowedImportSource<'val>>,
+}
+
+impl<'val> BorrowedSymbolToken<'val> {
+    pub fn new(
+        text: Option<&'val str>,
+        local_sid: Option<SymbolId>,
+        source: Option<BorrowedImportSource<'val>>,
+    ) -> Self {
+        Self {
+            text,
+            local_sid,
+            source,
+        }
+    }
+}
+
+impl<'val> From<&'val str> for BorrowedSymbolToken<'val> {
+    fn from(text: &'val str) -> Self {
+        Self::new(Some(text), None, None)
+    }
+}
+
+impl<'val> SymbolToken for BorrowedSymbolToken<'val> {
+    type ImportSource = BorrowedImportSource<'val>;
+
+    fn text(&self) -> Option<&str> {
+        self.text
+    }
+
+    fn local_sid(&self) -> Option<usize> {
+        self.local_sid
+    }
+
+    fn source(&self) -> Option<&Self::ImportSource> {
+        self.source.as_ref()
+    }
+}
+
+/// A borrowed implementation of [`Sequence`].
+#[derive(Debug, Clone)]
+pub struct BorrowedSequence<'val> {
+    children: Vec<BorrowedElement<'val>>,
+}
+
+impl<'val> BorrowedSequence<'val> {
+    pub fn new(children: Vec<BorrowedElement<'val>>) -> Self {
+        Self { children }
+    }
+}
+
+impl<'val> Sequence for BorrowedSequence<'val> {
+    type Element = BorrowedElement<'val>;
+
+    fn iter<'a>(&'a self) -> Box<dyn Iterator<Item = &'a Self::Element> + 'a> {
+        Box::new(self.children.iter())
+    }
+}
+
+/// A borrowed implementation of [`Struct`]
+#[derive(Debug, Clone)]
+pub struct BorrowedStruct<'val> {
+    // TODO model actual map indexing for the struct (maybe as a variant type)
+    //      otherwise struct field lookup will be O(N)
+    fields: Vec<(BorrowedSymbolToken<'val>, BorrowedElement<'val>)>,
+}
+
+impl<'val> BorrowedStruct<'val> {
+    pub fn new(fields: Vec<(BorrowedSymbolToken<'val>, BorrowedElement<'val>)>) -> Self {
+        Self { fields }
+    }
+}
+
+impl<'val> Struct for BorrowedStruct<'val> {
+    type FieldName = BorrowedSymbolToken<'val>;
+    type Element = BorrowedElement<'val>;
+
+    fn iter<'a>(
+        &'a self,
+    ) -> Box<dyn Iterator<Item = (&'a Self::FieldName, &'a Self::Element)> + 'a> {
+        // convert &(k, v) -> (&k, &v)
+        Box::new(self.fields.iter().map(|kv| match &kv {
+            (k, v) => (k, v),
+        }))
+    }
+}
+
+// TODO replace the references with `Cow` and bridge to the owned APIs for mutability
+
+/// Variants for all borrowed version _values_ within an [`Element`].
+#[derive(Debug, Clone)]
+pub enum BorrowedValue<'val> {
+    Null(IonType),
+    Integer(AnyInt),
+    String(&'val str),
+    Symbol(BorrowedSymbolToken<'val>),
+    SExpression(BorrowedSequence<'val>),
+    List(BorrowedSequence<'val>),
+    Struct(BorrowedStruct<'val>),
+    // TODO fill this in with the rest...
+}
+
+/// A borrowed implementation of [`Element`]
+#[derive(Debug, Clone)]
+pub struct BorrowedElement<'val> {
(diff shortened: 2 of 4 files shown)


src/value/borrowed.rs
```diff
@@ -0,0 +1,229 @@
...
+            (k, v) => (k, v),
+        }))
+    }
+}
+
+// TODO replace the references with `Cow` and bridge to the owned APIs for mutability
+
+/// Variants for all borrowed version _values_ within an [`Element`].
+#[derive(Debug, Clone)]
+pub enum BorrowedValue<'val> {
```
rgantt, line 134: Do you imagine having `ToOwned`/`Borrow`/`AsRef` _etc_ for these `enum`s?

(I'm curious about how you envision moving between `BorrowedT`/`T` underlying data, or if that's even necessary; maybe all interaction needs to happen behind the trait and you get what you get)

src/value/borrowed.rs
```diff
@@ -0,0 +1,229 @@
...
+        match &self.value {
+            Null(t) => *t,
+            Integer(_) => IonType::Integer,
+            String(_) => IonType::String,
+            Symbol(_) => IonType::Symbol,
+            SExpression(_) => IonType::SExpression,
+            List(_) => IonType::List,
+            Struct(_) => IonType::Struct,
+        }
+    }
```
rgantt, line 181: Does this effectively strip out type information from null values?

(I chose to encode each of the underlying values as an `Option<T>` which worked pretty well, didn't lose the type info, and made some later matching really easy)

src/value/borrowed.rs
```diff
@@ -0,0 +1,229 @@
...
+
+    fn as_str(&self) -> Option<&str> {
+        match &self.value {
+            BorrowedValue::String(text) => Some(*text),
+            BorrowedValue::Symbol(sym) => sym.text(),
+            _ => None,
+        }
+    }
```
rgantt, line 207: I'm not sure if this is a DOM-like API pattern (if so, FFTI), but I noticed it in the Kotlin IonElement, too: is there a reason to have these `as_*()` methods in the trait signature?

Especially in Rust, it seems like `impl From` would be helpful here, too. Then callers can specify `let f: Whatever = foo.into()` without having to call `let f = foo.as_whatever()`--and `Into<Whatever>` if they want to parameterize their own types. In other words, it seems like it would let `rustc` do more of the work.

rgantt (commented):
I think the approach of pulling ownership up a level and abstracting it behind some traits turned out pretty nice here (and it's something that my IonAny utility is missing, as we talked about "in person" a couple weeks ago).

But it does strike me as a lot of extra work solely to avoid having to commit to a lifetime, a la `Cow<'a, T>` (since both the `BorrowedT`/`T` types and `impls` are largely duplicates).

src/value/borrowed.rs
almann: No, this explicitly models it as the `Null(IonType)` variant so one does not need to have `Option<V>` for all the other variants.

src/value/borrowed.rs
almann: Yes, this PR doesn't have all the `From`/`Into` conversion operators nor does it deal with `ToOwned` traits either.  Those traits could be defined outside of the trait as potentially blanket implementations of the above traits.

src/value/borrowed.rs
almann: Yes, I want those things, but didn't factor them in this PR since I wanted to make sure it could be even abstracted with this factoring.

almann:
> I think the approach of pulling ownership up a level and abstracting it behind some traits turned out pretty nice here (and it's something that my IonAny utility is missing, as we talked about "in person" a couple weeks ago).
> 
> But it does strike me as a lot of extra work solely to avoid having to commit to a lifetime, a la `Cow<'a, T>` (since both the `BorrowedT`/`T` types and `impls` are largely duplicates).

What would the lifetime of the owned implementation be if using things like a `Cow<'a, T>`?  I mean technically, it can't be `'static` but we could bypass the borrow checker with a `transmute` or some such...

src/value/borrowed.rs
rgantt: Ah, I see it now. I just read this wrong and didn't pay close enough attention to the variant definition. Nice.

rgantt (approved):
> What would the lifetime of the owned implementation be if using things like a `Cow<'a, T>`? I mean technically, it can't be `'static` but we could bypass the borrow checker with a `transmute` or some such...

Yea, I don't think using `Cow<'a, T>` is feasible. And black magic probably isn't the right answer. It probably just doesn't make sense to abstract across ownership at the `struct` level _in case where a value will never be non-owned_ (which is where you'd never use a `Cow<...>` anyway). 

The traits you have here are well-considered, even if, at first glance, it seems like there's duplication. Good stuff!

almann:
> > What would the lifetime of the owned implementation be if using things like a `Cow<'a, T>`? I mean technically, it can't be `'static` but we could bypass the borrow checker with a `transmute` or some such...
> 
> Yea, I don't think using `Cow<'a, T>` is feasible. And black magic probably isn't the right answer. It probably just doesn't make sense to abstract across ownership at the `struct` level _in case where a value will never be non-owned_ (which is where you'd never use a `Cow<...>` anyway).
> 
> The traits you have here are well-considered, even if, at first glance, it seems like there's duplication. Good stuff!

Thanks @rgantt  for your feedback!  I don't like the copy-pasta either--I think as we develop this we should consider how to manage the parallel implementations with one set of code--I suppose one option is to solve this with a macro, but I would suppose that might leave us with two problems :laughing:.

src/value/mod.rs
```diff
@@ -0,0 +1,341 @@
...
+    /// if the value is any `null`.
+    fn as_sequence(&self) -> Option<&Self::Sequence>;
+
+    /// Returns a reference to the [`Struct`] of this element.
+    ///
+    /// This will return `None` in the case that the type is not `struct` or the value is
+    /// any `null`.
+    fn as_struct(&self) -> Option<&Self::Struct>;
+
+    // TODO add all the accessors to the trait
```
zslayton, line 284: Thinking ahead for this: it would be nice to have `Element` also implement `IntAccess` so instead of:
```rust
let int: Option<i64> = element.as_any_int().as_i64();
```
callers could use:
```rust
let int: Option<i64> = element.as_i64();
```
as long as `element` is an `int`. (`Element` could still provide an `as_any_int` method, this would just be an ergonomic shortcut for a common case.)

src/value/owned.rs
```diff
@@ -0,0 +1,228 @@
...
+//! ownership of data to do so.
+
+use super::{AnyInt, Element, ImportSource, Sequence, Struct, SymbolToken};
+use crate::types::SymbolId;
+use crate::IonType;
+
+/// An owned implementation of  [`ImportSource`].
+#[derive(Debug, Clone)]
+pub struct OwnedImportSource {
+    table: String,
```
zslayton, line 15: Given that any symbol table import is going to end up creating lots of references to the same table name (one per `SymbolToken`), I think that the `table` field is probably going to be an `Rc<String>`. That doesn't need to hold up this PR though, since `table` isn't part of the trait.

src/value/owned.rs
```diff
@@ -0,0 +1,228 @@
...
+        self.sid
+    }
+}
+
+/// An owned implementation of [`SymbolToken`].
+#[derive(Debug, Clone)]
+pub struct OwnedSymbolToken {
+    text: Option<String>,
+    local_sid: Option<SymbolId>,
+    source: Option<OwnedImportSource>,
```
zslayton, line 43: Is it ok for an implementer of `SymbolToken` to always return `None` for the source? (e.g. If I want to make a `SkinnySymbolToken` type that doesn't have an extra 9 bytes+alignment, can I decide to not track the import source?)

src/value/owned.rs
```diff
@@ -0,0 +1,228 @@
...
+        }
+    }
+}
+
+impl<T: Into<String>> From<T> for OwnedSymbolToken {
+    /// Constructs an owned token that has only text.
+    fn from(text: T) -> Self {
+        Self::new(Some(text.into()), None, None)
+    }
+}
```
zslayton, line 65: We should provide a `From<SymbolId>` implementation too.

This raises an interesting question: when do we expect a `SymbolToken`'s text to be resolved? The `text()` method doesn't take a `SymbolTable`, so are we assuming it was resolved eagerly at `SymbolToken` creation time?

I can imagine a `LazySymbolToken` implementation that holds an `Rc<dyn SymbolTable>` pointer that's only read if `text()` is called.

src/lib.rs
```diff
@@ -7,6 +7,7 @@ pub mod cursor;
 pub mod data_source;
 pub mod text;
 pub mod types;
+pub mod value;
```
zslayton, line 10: Should this be `element`?

src/value/mod.rs
```diff
@@ -0,0 +1,341 @@
...
+//! of an underlying element as a `Vec<String>`.  Note that filters out any annotation
+//! that may not have text (so data could be dropped):
+//!
+//! ```
+//! use ion_rs::value::{Element, SymbolToken};
+//! use ion_rs::value::borrowed::*;
+//! use ion_rs::value::owned::*;
+//!
+//! fn extract_annotations<T: Element>(elem: &T) -> Vec<String> {
+//!     elem.annotations().flat_map(
```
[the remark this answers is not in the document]
almann: My goal here was to do some kind of text extraction which I will suppose to be a common case.  I could either add a comment to this effect or return `Vec<Option<String>>` or `Option<Vec<String>>`.  Do you have any preference here?

src/value/owned.rs
almann: Added a TODO for this `OwnedSymbolTable::text`.

src/value/mod.rs
zslayton: I missed that, thanks for pointing it out.

src/value/owned.rs
almann: Yes, it is _okay_, but only if the token `text` is not `None`.  So for applications that know the text of all symbols at compile time you could statically use such an implementation or hide that detail with a `dyn SymbolToken`.

src/value/mod.rs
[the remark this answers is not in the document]
zslayton: I like `Vec<Option<String>>` for this.

src/lib.rs
almann: I considered that--though we currently have a bunch of other _value_ like things lurking such as the `SymbolToken` etc.

src/value/mod.rs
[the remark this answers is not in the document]
almann: Updated.

src/value/owned.rs
almann: I don't really think the from `SymbolId` without either text or source is particularly useful though, because if that is true, you technically have a local symbol that has undefined text and that is equivalent to `$0`.

zslayton: This use case is roughly equivalent to a lazily resolved symbol token. I should be able to create a `SymbolToken` from `3usize` (i.e. a `SymbolId`) and give it to a writer without first providing or resolving the text. That doesn't mean there _isn't_ text for that symbol, but I haven't gone to the effort to look it up or otherwise provide it. (One could argue that we shouldn't facilitate this by providing an implementation of `From<SymbolId>` as a distinct concern. My point is that having a "raw" `SymbolToken` whose text is as of yet unknown is useful.)

PR #140, opened 2021-03-30
Added methods for Sequence and Struct (#132)

desaikd:
*Issue #114 *

*Description of changes:*
This issues fills in the API methods for `Sequence` and `Struct` on the Value API.

*Changes:*
- Added `Sequence` methods: `get`, `len`, `is_empty`
- Added `Struct` methods: `get`, `get_all`
- `Struct` uses a `HashMap` with `(key, values) -> (String, Vec<(SymbolToken, Element)>)` as well as  a `Vec<(Symboltoken, Element)>` to store fields with no text `SymbolToken`


almann (approved):
This is just a rebase from #132 which is approved. :shipit:

Issue #134, opened 2021-03-26
Changed `OwnedSymbolToken`/`OwnedStruct` from using `String` to `Rc<str>`


almann:
`OwnedSymbolToken` uses `String` for its text, which is likely to be shared with other instances of itself, the symbol table, and with `OwnedStruct`.  We should replace the internal implementation with `Rc<String>` or the like and also make sure the constructors are friendly to this (and appropriately generic).

This is related to #114 and should be implemented once #132 is pushed.

almann:
Originally, I thought it should be `Rc<String>`, but on review of #141, it is clear that it should be `Rc<str>`.

almann:
This has been implemented in #141.

PR #141, opened 2021-03-30
Changed OwnedSymbolToken, OwnedImportSource, OwnedStruct from using String to Rc<String>

desaikd:
*Issue #134*

*Description of changes:*
This PR works on using `Rc<String>` instead of `String` for `OwnedSymbolToken`, `OwnedImportSource`, `OwnedStruct`.

### DIFF
--- a/src/value/owned.rs
+++ b/src/value/owned.rs
@@ -121,8 +121,6 @@ impl Sequence for OwnedSequence {
 /// An owned implementation of [`Struct`]
 #[derive(Debug, Clone)]
 pub struct OwnedStruct {
-    // TODO model actual map indexing for the struct (maybe as a variant type)
-    //      otherwise struct field lookup will be O(N)
     text_fields: HashMap<Rc<str>, Vec<(OwnedSymbolToken, OwnedElement)>>,
     no_text_fields: Vec<(OwnedSymbolToken, OwnedElement)>,
 }
--- a/src/value/owned.rs
+++ b/src/value/owned.rs
@@ -9,19 +9,20 @@ use super::{AnyInt, Element, ImportSource, Sequence, Struct, SymbolToken};
 use crate::types::SymbolId;
 use crate::IonType;
 use std::collections::HashMap;
+use std::rc::Rc;
 
 /// An owned implementation of  [`ImportSource`].
 #[derive(Debug, Clone)]
 pub struct OwnedImportSource {
     // TODO consider Rc<String> here as we flesh out sharing/symbol tables.
-    table: String,
+    table: Rc<String>,
     sid: SymbolId,
 }
 
 impl OwnedImportSource {
     pub fn new<T: Into<String>>(table: T, sid: SymbolId) -> Self {
         Self {
-            table: table.into(),
+            table: Rc::new(table.into()),
             sid,
         }
     }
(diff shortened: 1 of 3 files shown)


src/value/owned.rs
```diff
@@ -41,14 +42,14 @@ impl ImportSource for OwnedImportSource {
 #[derive(Debug, Clone)]
 pub struct OwnedSymbolToken {
     // TODO consider Rc<String> for a common case of many symbols sharing the same text.
```
almann, line 44: Remove this defunct comment.

src/value/owned.rs
```diff
@@ -9,19 +9,20 @@ use super::{AnyInt, Element, ImportSource, Sequence, Struct, SymbolToken};
 use crate::types::SymbolId;
 use crate::IonType;
 use std::collections::HashMap;
+use std::rc::Rc;
 
 /// An owned implementation of  [`ImportSource`].
 #[derive(Debug, Clone)]
 pub struct OwnedImportSource {
     // TODO consider Rc<String> here as we flesh out sharing/symbol tables.
```
almann, line 17: Remove this defunct comment.

src/value/owned.rs
```diff
@@ -9,19 +9,20 @@ use super::{AnyInt, Element, ImportSource, Sequence, Struct, SymbolToken};
...
-    table: String,
+    table: Rc<String>,
     sid: SymbolId,
 }
 
 impl OwnedImportSource {
     pub fn new<T: Into<String>>(table: T, sid: SymbolId) -> Self {
         Self {
-            table: table.into(),
+            table: Rc::new(table.into()),
```
almann, line 25: I think we need to change the generic here to take `Into<Rc<String>>` or something otherwise it will be impossible to share from an name that is itself an `Rc<String>`.  That said `Into<Rc<String>>` may not be as ergonomic with respect to `&str`, so your mileage may vary...

src/value/owned.rs
```diff
@@ -63,15 +64,18 @@ impl OwnedSymbolToken {
...
 impl SymbolToken for OwnedSymbolToken {
     type ImportSource = OwnedImportSource;
 
     fn text(&self) -> Option<&str> {
-        self.text.as_ref().map(String::as_str)
+        match self.text.as_ref() {
+            None => None,
+            Some(value) => Some(value.as_str()),
+        }
     }
```
almann, line 79: Personally, I have a preference for using `map` over a `match` for this because you don't have to do the un-wrapping and wrapping.  You could write it as `.map(|s| s.as_str())` for example to avoid the `match` boilerplate.

almann (changes requested):
Looks pretty good--comments below.

Something to consider that I didn't think about when writing #134, is that we probably want `Rc<str>` and **not** `Rc<String>`.

With `Rc<str>`, we can accept `Into<Rc<str>>` and my recollection is that you can then accept `Rc<str>`/`&str`/`String` as parameters.

src/value/owned.rs
```diff
@@ -128,14 +123,14 @@ impl Sequence for OwnedSequence {
 pub struct OwnedStruct {
     // TODO model actual map indexing for the struct (maybe as a variant type)
     //      otherwise struct field lookup will be O(N)
```
almann, line 125: *Minor*: Let's eliminate this TODO.

almann (approved):
LGTM :ship:

src/value/owned.rs
almann: I did this directly in the PR, so no action to take unless you object.

Issue #139, opened 2021-03-26
Fill in the Value Variants for the `Element` API


almann:
#131 added the basic structure of the API and its owned/borrrowed implementations.  The rest of the Ion type space (the various scalars including `clob`/`blob`) should be filled in.

Related to #114.

PR #143, opened 2021-03-31
added remaining scalar value types for Element API

desaikd:
*Issue #139*

*Description of changes:*
This PR works on filling in rest of the scalar value types for Element API.

*Changes:*
Added following scalar types with this PR:
- Float
- Decimal
- Timestamp
- Boolean
- Blob
- Clob

### DIFF
--- a/src/value/borrowed.rs
+++ b/src/value/borrowed.rs
@@ -8,12 +8,12 @@
 //! backed by octets or string data, `&[u8]` and `&str` are used.
 
 use super::{Element, ImportSource, Sequence, Struct, SymbolToken};
+use crate::types::decimal::Decimal;
+use crate::types::timestamp::Timestamp;
 use crate::types::SymbolId;
 use crate::value::AnyInt;
 use crate::IonType;
 use std::collections::HashMap;
-use crate::types::decimal::Decimal;
-use crate::types::timestamp::Timestamp;
 
 /// A borrowed implementation of [`ImportSource`].
 #[derive(Debug, Copy, Clone)]
--- a/src/value/borrowed.rs
+++ b/src/value/borrowed.rs
@@ -188,12 +188,12 @@ impl<'val> Struct for BorrowedStruct<'val> {
 pub enum BorrowedValue<'val> {
     Null(IonType),
     Integer(AnyInt),
-    Float(&'val f64),
+    Float(f64),
     Decimal(Decimal),
     Timestamp(Timestamp),
     String(&'val str),
     Symbol(BorrowedSymbolToken<'val>),
-    Boolean(&'val bool),
+    Boolean(bool),
     Blob(&'val [u8]),
     Clob(&'val [u8]),
     SExpression(BorrowedSequence<'val>),
@@ -266,7 +266,7 @@ impl<'val> Element for BorrowedElement<'val> {
 
     fn as_float(&self) -> Option<f64> {
         match &self.value {
-            BorrowedValue::Float(&f) => Some(f),
+            BorrowedValue::Float(f) => Some(*f),
             _ => None,
         }
     }
@@ -302,7 +302,7 @@ impl<'val> Element for BorrowedElement<'val> {
 
     fn as_bool(&self) -> Option<bool> {
         match &self.value {
-            BorrowedValue::Boolean(&b) => Some(b),
+            BorrowedValue::Boolean(b) => Some(*b),
             _ => None,
         }
     }
--- a/src/value/borrowed.rs
+++ b/src/value/borrowed.rs
@@ -12,6 +12,8 @@ use crate::types::SymbolId;
 use crate::value::AnyInt;
 use crate::IonType;
 use std::collections::HashMap;
+use crate::types::decimal::Decimal;
+use crate::types::timestamp::Timestamp;
 
 /// A borrowed implementation of [`ImportSource`].
 #[derive(Debug, Copy, Clone)]
@@ -186,8 +188,14 @@ impl<'val> Struct for BorrowedStruct<'val> {
 pub enum BorrowedValue<'val> {
     Null(IonType),
     Integer(AnyInt),
+    Float(f64),
+    Decimal(Decimal),
+    Timestamp(Timestamp),
     String(&'val str),
     Symbol(BorrowedSymbolToken<'val>),
+    Boolean(&'val bool),
+    Blob(Vec<u8>),
+    Clob(Vec<u8>),
     SExpression(BorrowedSequence<'val>),
     List(BorrowedSequence<'val>),
     Struct(BorrowedStruct<'val>),
@@ -224,8 +232,14 @@ impl<'val> Element for BorrowedElement<'val> {
         match &self.value {
             Null(t) => *t,
             Integer(_) => IonType::Integer,
+            Float(_) => IonType::Float,
+            Decimal(_) => IonType::Decimal,
+            Timestamp(_) => IonType::Timestamp,
             String(_) => IonType::String,
             Symbol(_) => IonType::Symbol,
+            Boolean(_) => IonType::Boolean,
+            Blob(_) => IonType::Blob,
+            Clob(_) => IonType::Clob,
             SExpression(_) => IonType::SExpression,
             List(_) => IonType::List,
             Struct(_) => IonType::Struct,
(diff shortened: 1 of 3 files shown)


src/value/borrowed.rs
```diff
@@ -186,8 +188,14 @@ impl<'val> Struct for BorrowedStruct<'val> {
...
     Null(IonType),
     Integer(AnyInt),
+    Float(f64),
+    Decimal(Decimal),
+    Timestamp(Timestamp),
     String(&'val str),
     Symbol(BorrowedSymbolToken<'val>),
+    Boolean(&'val bool),
+    Blob(Vec<u8>),
+    Clob(Vec<u8>),
```
almann, line 198: These fields should be the borrowed slice: `&'val [u8]`

src/value/owned.rs
```diff
@@ -6,6 +6,9 @@
 //! ownership of data to do so.
 
 use super::{AnyInt, Element, ImportSource, Sequence, Struct, SymbolToken};
+use crate::result::IonError::IllegalOperation;
```
almann, line 9: There are some warnings here around "unused import", are those warnings valid?

src/value/mod.rs
```diff
@@ -270,6 +287,24 @@ pub trait Element {
...
+    ///
+    /// This will return `None` in the case that the type is not `blob` or the value is
+    /// any `null`.
+    fn as_blob(&self) -> Option<&Vec<u8>>;
+
+    /// Returns a reference to the underlying bytes of this element.
+    ///
+    /// This will return `None` in the case that the type is not `clob` or the value is
+    /// any `null`.
+    fn as_clob(&self) -> Option<&Vec<u8>>;
```
almann, line 306: This should return `Option<&[u8]>`.

src/value/mod.rs
```diff
@@ -270,6 +287,24 @@ pub trait Element {
     /// any `null`.
     fn as_sym(&self) -> Option<&Self::SymbolToken>;
 
+    /// Returns a reference to the boolean value of this element.
+    ///
+    ///  This will return `None` in the case that the type is not `bool` or the value is
+    ///   any `null`.
+    fn as_bool(&self) -> Option<&bool>;
```
almann, line 294: This should return `Option<bool>` as `bool` is a cheap value.

src/value/mod.rs
```diff
@@ -258,6 +260,21 @@ pub trait Element {
     /// This will return `None` if the type is not `int` or the value is any `null`.
     fn as_any_int(&self) -> Option<&AnyInt>;
 
+    /// Returns a reference to the underlying float value for this element.
+    ///
+    /// This will return `None` if the type is not `float` or the value is any `null`.
+    fn as_float(&self) -> Option<&f64>;
```
almann, line 266: This should just return `Option<f64>` as `f64` is a cheap value.  The reason why `AnyInt` gets borrowed is due to it not supporting `Copy` trait (because of the `BigInt` variant).

src/value/owned.rs
```diff
@@ -186,8 +189,14 @@ impl Struct for OwnedStruct {
...
     Null(IonType),
     Integer(AnyInt),
+    Float(f64),
+    Decimal(Decimal),
+    Timestamp(Timestamp),
     String(String),
     Symbol(OwnedSymbolToken),
+    Boolean(bool),
+    Blob(Vec<u8>),
+    Clob(Vec<u8>),
```
almann, line 199: These **are** correct for the owned variants (just noting this in contrast to the trait API and the borrowed implementation).

almann (changes requested):
The main bit of feedback is that the APIs for LOBs should return `&[u8]` and the borrowed versions should be implemented in terms of this.

We've been playing a bit fast and loose with doc/unit testing (myself included) as part of just getting the scaffolding of the APIs together.  Since the scope of change is pretty small here, may I ask that you fill in some reasonable unit tests around the value variants?

src/value/owned.rs
desaikd: yes. I will remove it.

src/value/borrowed.rs
```diff
@@ -188,14 +188,14 @@ impl<'val> Struct for BorrowedStruct<'val> {
 pub enum BorrowedValue<'val> {
     Null(IonType),
     Integer(AnyInt),
-    Float(f64),
+    Float(&'val f64),
```
almann, line 191: I don't know if I was particularly clear about it, but it it is a small value, we shouldn't borrow it.

src/value/borrowed.rs
```diff
@@ -188,14 +188,14 @@ impl<'val> Struct for BorrowedStruct<'val> {
...
     Decimal(Decimal),
     Timestamp(Timestamp),
     String(&'val str),
     Symbol(BorrowedSymbolToken<'val>),
     Boolean(&'val bool),
```
almann, line 196: I might have missed this earlier--but similar to `f64`, let's not borrow this.

almann (commented):
Some concerns around borrowing `f64` and `bool` in the API and borrowed implementation.

src/value/owned.rs
```diff
@@ -335,3 +335,118 @@ impl Element for OwnedElement {
         }
     }
 }
+
+#[cfg(test)]
+mod value_tests {
+    use crate::result::IonResult;
+    use crate::types::decimal::Decimal;
+    use crate::types::timestamp::Timestamp;
+    use crate::value::owned::{OwnedElement, OwnedSequence, OwnedSymbolToken, OwnedValue};
```
almann, line 344: `OwnedSequence` appears to be not used.

src/value/owned.rs
```diff
@@ -335,3 +335,118 @@ impl Element for OwnedElement {
...
+    use crate::value::{AnyInt, Element, IntAccess, SymbolToken};
+
+    fn extract_text<T: SymbolToken>(tok: &T) -> String {
+        tok.text().unwrap().into()
+    }
```
almann, line 349: Curious why you need to convert `&str` to `String`?

src/value/owned.rs
```diff
@@ -335,3 +335,118 @@ impl Element for OwnedElement {
...
+
+    #[test]
+    fn test_as_float() -> IonResult<()> {
+        let owned_elem = create_element_from_value(OwnedValue::Float(305e1));
+        let expected_elem = Some(305e1);
+
+        assert_eq!(&expected_elem, &owned_elem.as_float());
+        Ok(())
+    }
```
almann, line 375: Might be worth considering (or adding an issue/TODO) to do the negative testing around the other `as` methods.  Another thing to consider (or add an issue/TODO)  is how to do test fixture style parameterization via something like `rstest`.

Specifically, as I look at each of these cases, they are really just testing the wrapping functionality and the ability to access the value through the `as` methods.  Creating a fixture for a similar equivalence class might be to just parameterize on `OwnedValue` and using a table or something to test that `as` methods are `None` except for the kind of value being wrapped.  You'll get the negative coverage this way and avoid writing a lot of boilerplate.

almann (approved):
I have some minor comments below (and suggested TODO issues we should to follow up on, maybe even as a next change), but otherwise :ship:

src/value/owned.rs
desaikd: No specific reason. I think I tried to replicate from the doc tests of `SymbolToken`. But I think using `&str ` would save an extra step of converting to `String` .

src/value/owned.rs
desaikd: Created an issue for this: #146

src/value/owned.rs
```diff
@@ -279,3 +335,118 @@ impl Element for OwnedElement {
...
+
+    #[test]
+    fn test_as_clob() -> IonResult<()> {
+        let owned_elem =
+            create_element_from_value(OwnedValue::Clob(String::from("{{\"hello\"}}").into_bytes()));
+        let expected_elem = Some("{{\"hello\"}}".as_bytes());
+
+        assert_eq!(&expected_elem, &owned_elem.as_clob());
+        Ok(())
+    }
```
zslayton, line 451: These examples are valid but a bit misleading. In each case, the `String` being turned into lob bytes is in Ion text syntax, but this doesn't have any impact on the value produced.

`"{{aGVsbG8h}}".as_bytes()` is a `&[u8]` representing the literal ascii sequence starting with `{`, `{`, `a`, etc. It's not being base64 decoded or anything.

That's a legal byte sequence to make a *lob (in that any byte sequence is legal!), but making the input string resemble Ion text made me stop and stare at this test a bit longer than necessary. Any reason not to use the string "hello" for both?

src/value/mod.rs
```diff
@@ -270,6 +287,24 @@ pub trait Element {
     /// any `null`.
     fn as_sym(&self) -> Option<&Self::SymbolToken>;
 
+    /// Returns a reference to the boolean value of this element.
+    ///
+    ///  This will return `None` in the case that the type is not `bool` or the value is
+    ///   any `null`.
```
zslayton, line 293: Nit: extra space before `any`.

src/value/mod.rs
```diff
@@ -258,6 +260,21 @@ pub trait Element {
     /// This will return `None` if the type is not `int` or the value is any `null`.
     fn as_any_int(&self) -> Option<&AnyInt>;
 
+    /// Returns a reference to the underlying float value for this element.
```
zslayton, line 263: Nit: the doc strings for `as_float` and `as_boolean` indicate that those methods will return `a reference to the underlying <<value>>`, but it's actually a copy.

src/value/owned.rs
desaikd: This has been changed in https://github.com/amzn/ion-rust/pull/147.

PR #147, opened 2021-04-02
Added unit tests with fixture for `Element` API

desaikd:
*Issue #146*

*Description of changes:*
This PR works on using fixtures for `Element` API unit tests.

*Changes:*
- created a fixture for creating `owned_element` to be used for unit tests.
- added `rstest` to `Cargo.toml`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #146, opened 2021-04-01
Use rstest fixtures for unit tests in `Element` API


desaikd:
Currently the unit tests added to the `Element` API has a lot of boilerplate code. Use [fixtures](https://docs.rs/rstest/0.7.0/rstest/attr.rstest.html#injecting-fixtures) from rstests to remove the boilerplate code for the unit tests. Also, add unit tests for borrow `Element` API as well.

PR #148, opened 2021-04-05
Refactors `value::owned` unit tests to be more parameterized.

almann:
Increases coverage for that module from 54% to 72% as measured with grcov.

Creates a dispatch table style approach where all the `asXXX` methods
are negatively asserted and each test case "overrides" the negative
assertion cases with its own.

Also did some API cleanup to make things more consistent and easier to test:
* Renames `as_float` to `as_f64`.
* Consolidates `as_blob` and `as_clob` to `as_bytes`.
* `PartialEq`/`Eq` for `AnyInt` to facilitate testing.

Addresses some of #146 and is based on work that @desaikd started with #147.

### DIFF
--- a/src/value/borrowed.rs
+++ b/src/value/borrowed.rs
@@ -282,7 +282,7 @@ impl<'val> Element for BorrowedElement<'val> {
         }
     }
 
-    fn as_float(&self) -> Option<f64> {
+    fn as_f64(&self) -> Option<f64> {
         match &self.value {
             BorrowedValue::Float(f) => Some(*f),
             _ => None,
@@ -325,16 +325,9 @@ impl<'val> Element for BorrowedElement<'val> {
         }
     }
 
-    fn as_blob(&self) -> Option<&[u8]> {
+    fn as_bytes(&self) -> Option<&[u8]> {
         match &self.value {
-            BorrowedValue::Blob(bytes) => Some(*bytes),
-            _ => None,
-        }
-    }
-
-    fn as_clob(&self) -> Option<&[u8]> {
-        match &self.value {
-            BorrowedValue::Clob(bytes) => Some(*bytes),
+            BorrowedValue::Blob(bytes) | BorrowedValue::Clob(bytes) => Some(*bytes),
             _ => None,
         }
     }
--- a/src/value/mod.rs
+++ b/src/value/mod.rs
@@ -263,7 +263,7 @@ pub trait Element {
     /// Returns a reference to the underlying float value for this element.
     ///
     /// This will return `None` if the type is not `float` or the value is any `null`.
-    fn as_float(&self) -> Option<f64>;
+    fn as_f64(&self) -> Option<f64>;
 
     /// Returns a reference to the underlying [`Decimal`] for this element.
     ///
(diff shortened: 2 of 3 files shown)


src/value/borrowed.rs
```diff
@@ -282,7 +282,7 @@ impl<'val> Element for BorrowedElement<'val> {
         }
     }
 
-    fn as_float(&self) -> Option<f64> {
+    fn as_f64(&self) -> Option<f64> {
         match &self.value {
```
almann, line 286: Changed this to return what the Rust type is.

src/value/borrowed.rs
```diff
@@ -325,16 +325,9 @@ impl<'val> Element for BorrowedElement<'val> {
...
-            BorrowedValue::Blob(bytes) => Some(*bytes),
-            _ => None,
-        }
-    }
-
-    fn as_clob(&self) -> Option<&[u8]> {
-        match &self.value {
-            BorrowedValue::Clob(bytes) => Some(*bytes),
+            BorrowedValue::Blob(bytes) | BorrowedValue::Clob(bytes) => Some(*bytes),
```
almann, line 330: Same here--changes to both `clob`/`blob` accessing data as `&[u8]` through a single API.

src/value/mod.rs
```diff
@@ -207,6 +207,24 @@ impl IntAccess for AnyInt {
...
+            },
+            BigInt(my_bi) => match other {
+                I64(other_i64) => my_bi.to_i64() == Some(*other_i64),
+                BigInt(other_bi) => my_bi == other_bi,
+            },
+        }
+    }
+}
+
+impl Eq for AnyInt {}
```
almann, line 226: Handy to make the tests easier to write.

src/value/owned.rs
```diff
@@ -357,110 +350,169 @@ impl Element for OwnedElement {
 
 #[cfg(test)]
 mod value_tests {
-    use crate::result::IonResult;
```
almann, line 360: We don't leverage the `?` operator in these tests, so I didn't bother keeping this.

src/value/owned.rs
```diff
@@ -357,110 +350,169 @@ impl Element for OwnedElement {
...
-    #[rstest]
-    fn test_as_bool(
-        #[with(OwnedValue::Boolean(false))] owned_element: OwnedElement,
-    ) -> IonResult<()> {
-        assert_eq!(&Some(false), &owned_element.as_bool());
-        Ok(())
-    }
+    impl IntoIterator for ElemOp {
+        type Item = ElemOp;
+        type IntoIter = <Once<ElemOp> as IntoIterator>::IntoIter;
```
almann, line 387: The common case of using this is just replacing one operation--this allows us to use the single value as a singleton iterator in the fixture.

src/value/owned.rs
```diff
@@ -357,110 +350,169 @@ impl Element for OwnedElement {
...
+            }),
+            (AsF64, &|e| assert_eq!(None, e.as_f64())),
+            (AsDecimal, &|e| assert_eq!(None, e.as_decimal())),
+            (AsTimestamp, &|e| assert_eq!(None, e.as_timestamp())),
+            (AsStr, &|e| assert_eq!(None, e.as_str())),
+            (AsSym, &|e| assert_eq!(true, e.as_sym().is_none())),
+            (AsBytes, &|e| assert_eq!(None, e.as_bytes())),
+            (AsSequence, &|e| assert_eq!(true, e.as_sequence().is_none())),
+            (AsStruct, &|e| assert_eq!(true, e.as_struct().is_none())),
+        ];
```
almann, line 499: One way to think about this is a kind of "vtable" for all the operators and the negative assertion, we then "override" with the input to the fixture.

desaikd (approved):
LGTM. I think there should be a TODO for adding tests for `Struct` and `Sequence` as well.

src/value/owned.rs
```diff
@@ -357,110 +350,170 @@ impl Element for OwnedElement {
...
-    fn test_as_decimal(
-        #[with(OwnedValue::Decimal(Decimal::new(8, 3)))] owned_element: OwnedElement,
-    ) -> IonResult<()> {
-        let decimal_value2 = Decimal::new(80, 2);
-        let expected_elem = Some(&decimal_value2);
-
-        assert_eq!(&expected_elem, &owned_element.as_decimal());
-        Ok(())
+    /// Makes a timestamp from an RFC-3339 string and panics if it can't
+    fn mk_ts<T: AsRef<str>>(text: T) -> Timestamp {
```
zslayton, line 363: IMO this function name is short enough to be cryptic.

Issue #149, opened 2021-04-05
Implement `PartialEq`/`Eq` for `SymbolToken`, `Sequrence`, `Struct `


desaikd:
Implement `PartiEq`/ `Eq` for `SymbolToken`, `Sequrence`, `Struct` for easier unit testing. For `Struct` and `Sequence` it would mean we have to add `PartialEq` / `Eq` for `OwnedElement`.

almann:
Related to #114.

desaikd:
This issue is fixed with #152

PR #152, opened 2021-04-06
added PartialEq/ Eq for value::owned and added  unit tests for container types

desaikd:
*Issue #149 , #147*

*Description of changes:*
This PR works on creating `PartialEq` and `Eq` for `value::Owned` types and adds unit tests for container types.

*Changes:*
- Added `PartialEq` and `Eq` for `OwnedSymbolToken`, `OwnedImportSource`, `OwnedStruct`, `OwnedSequence`, `OwnedElement`. 
- Added unit tests for container types.

### DIFF
--- a/src/value/owned.rs
+++ b/src/value/owned.rs
@@ -57,7 +57,7 @@ pub struct OwnedSymbolToken {
 }
 
 impl OwnedSymbolToken {
-    pub fn new(
+    fn new(
         text: Option<Rc<str>>,
         local_sid: Option<SymbolId>,
         source: Option<OwnedImportSource>,
@@ -68,6 +68,38 @@ impl OwnedSymbolToken {
             source,
         }
     }
+
+    /// Decorates the [`OwnedSymbolToken`] with text.
+    pub fn with_text<T: Into<Rc<str>>>(mut self, text: T) -> Self {
+        self.text = Some(text.into());
+        self
+    }
+
+    /// Decorates the [`OwnedSymbolToken`] with a local ID.
+    pub fn with_local_sid(mut self, local_sid: SymbolId) -> Self {
+        self.local_sid = Some(local_sid);
+        self
+    }
+
+    /// Decorates the [`OwnedSymbolToken`] with an [`OwnedImportSource`].
+    pub fn with_source<T: Into<Rc<str>>>(mut self, table: T, sid: SymbolId) -> Self {
+        self.source = Some(OwnedImportSource::new(table, sid));
+        self
+    }
+}
+
+/// Constructs an [`OwnedSymbolToken`] with unknown text and a local ID.
+/// A common case for binary parsing (though technically relevant in text).
+#[inline]
+pub fn local_sid_token(local_sid: SymbolId) -> OwnedSymbolToken {
+    OwnedSymbolToken::new(None, Some(local_sid), None)
+}
+
+/// Constructs an [`OwnedSymbolToken`] with just text.
+/// A common case for text and synthesizing tokens.
+#[inline]
+pub fn text_token<T: Into<Rc<str>>>(text: T) -> OwnedSymbolToken {
+    OwnedSymbolToken::new(Some(text.into()), None, None)
 }
 
 impl PartialEq for OwnedSymbolToken {
@@ -85,7 +117,7 @@ impl Eq for OwnedSymbolToken {}
 impl<T: Into<Rc<str>>> From<T> for OwnedSymbolToken {
     /// Constructs an owned token that has only text.
     fn from(text: T) -> Self {
-        Self::new(Some(text.into()), None, None)
+        text_token(text)
     }
 }
 
@@ -573,13 +605,55 @@ mod value_tests {
             &|e: &OwnedElement| {
                 assert_eq!(Some(&vec![("greetings", OwnedElement::from(OwnedValue::String("hello".into())))].into_iter().collect()), e.as_struct());
             }
-        )
+        ),
+        // TODO consider factoring this out of the value tests to make it more contained
+        // TODO consider adding non-equivs for this (really symbol token tests are probably better for this)
+        eq_annotations => [
+            // trivially empty is equivalent to another empty
+            vec![vec![], vec![]],
+            // tokens with text
+            vec![
+                vec![text_token("hello"), text_token("world")],
+                // containing local sids only
+                vec![
+                    text_token("hello").with_local_sid(20),
+                    local_sid_token(21).with_text("world"),
+                ],
+                // mix of local sid, but all with sources
+                vec![
+                    text_token("hello").with_source("hello_table", 2),
+                    text_token("world").with_local_sid(59).with_source("world_table", 200)
+                ]
+            ],
+            // tokens without text
+            vec![
+                vec![
+                    // local sid only with no text are all $0 equivalent
+                    local_sid_token(21),
+                    // source import table is the comparator for unknown cases
+                    local_sid_token(22).with_source("hello_table", 2)
+                ],
+                vec![
+                    local_sid_token(0),
+                    local_sid_token(400).with_source("hello_table", 2),
+                ],
+                vec![
+                    local_sid_token(200),
+                    text_token("hello").with_source("hello_table", 2),
+                ],
+                vec![
+                    local_sid_token(99),
+                    text_token("hello").with_local_sid(500).with_source("hello_table", 2),
+                ],
+            ],
+        ]
     )]
     fn owned_element_accessors<O: IntoIterator<Item = ElemOp>>(
         val: OwnedValue,
         ion_type: IonType,
         valid_ops_iter: O,
         op_assert: &ElemAssertFunc,
+        eq_annotations: Vec<Vec<OwnedSymbolToken>>,
     ) {
         // table of negative assertions for each operation
         let neg_table: Vec<(ElemOp, &ElemAssertFunc)> = vec![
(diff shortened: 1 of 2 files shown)


src/value/owned.rs
```diff
@@ -62,6 +74,24 @@ impl OwnedSymbolToken {
...
+            }
+            false
+        } else {
+            if self.source == other.source {
+                return true;
+            }
+            false
+        }
+    }
+}
```
almann, line 91: I might make this a bit more concise without the extra returns and explicit Boolean values

```rust
        if other.text != None && self.text != None {
            // both has text
            other.text == self.text
        } else {
            // one of the tokens has undefined text, the only way they can be equal if they have the same source
            // None implies $0
            self.source == other.source
        }
```

Note that we probably should document the `None` implies `$0` bit in our APIs (if you agree), and we really need tests for the different cases described in:

https://amzn.github.io/ion-docs/docs/symbols.html#data-model

src/value/owned.rs
```diff
@@ -234,6 +293,26 @@ impl OwnedElement {
...
+            IonType::Float => self.as_f64() == other.as_f64(),
+            IonType::Timestamp => self.as_timestamp() == other.as_timestamp(),
+            IonType::Clob | IonType::Blob => self.as_bytes() == other.as_bytes(),
+            IonType::String => self.as_str() == other.as_str(),
+            IonType::Symbol => self.as_sym() == other.as_sym(),
+            IonType::Struct => self.as_struct() == other.as_struct(),
+            IonType::SExpression | IonType::List => self.as_sequence() == other.as_sequence(),
+        }
+    }
+}
```
almann, line 312: Is it possible to lean on a `PartialEq` defined for `OwnedValue`?  Also don't you need to compare the annotations?  This needs its own explicit tests too.

almann (changes requested):
I would really like to see test cases for the non-trivial `PartialEq` methods, they are potentially error prone.

src/value/owned.rs
desaikd: Oh yes! I will change it to be more concise. And yes, I think we should document None implies $0. I will try to add tests for above mentioned cases.

src/value/owned.rs
```diff
@@ -53,6 +49,7 @@ impl ImportSource for OwnedImportSource {
 }
 
 /// An owned implementation of [`SymbolToken`].
+/// Here text as `None` represents SID $0
```
almann, line 52: This comment belongs on the trait definition (since it should be uniformly implemented that way).

almann (approved):
:rocket: LGTM.  Do you mind adding a GH issue to explicitly test symbol token on its own?

almann (changes requested):
I am not sure I grok the test factoring for annotation testing.  I would either factor that out as its own test or maybe explain to me how you're thinking about it.

src/value/owned.rs
```diff
@@ -600,6 +590,16 @@ mod value_tests {
...
         ];
+
+        // tests for element with annotations
+        let annotations = vec!["foo", "bar", "baz"];
+        let owned_elem = OwnedElement::new(
+            annotations.iter().map(|s| (*s).into()).collect(),
+            val.clone(),
+        );
+        let expected: Vec<String> = annotations.iter().map(|s| (*s).into()).collect();
+        assert_eq!(expected, annotation_strings(&owned_elem));
```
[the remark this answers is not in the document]
desaikd: I think I was trying to add 3 annotation to  each value in test cases and test those 3 annotations as you mentioned. But I think it would make more sense to have separate unit test for this. I will try to see if value lists can be used here.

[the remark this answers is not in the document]
almann: Here is a small, contrived example of mixing parameter cases and value lists.

```rust
#[cfg(test)]
mod test {
    use rstest::*;

    #[rstest]
    #[case::range_10_to_100(10, 100)]
    #[case::range_5_to_45(5, 45)]
    fn foo1(#[case] lval: i64, #[case] hval: i64, #[values(20, 30, 40)] other: i64) {
        assert!(lval < other);
        assert!(hval > other);
    }

    #[rstest(
        lval, hval,
        case::range_10_to_100(10, 100),
        case::range_5_to_45(5, 45),
        other => [20, 30, 40]
    )]
    fn foo2(lval: i64, hval: i64, other: i64) {
        assert!(lval < other);
        assert!(hval > other);
    }
}
```

The `foo2` example uses the macros I used, but apparently that is the [*old compact syntax*][1].  Apparently `rstest` 0.7 added this (my `ion-c-sys` tests used 0.6).  We don't need to port over now, but something for us to consider going forward.

[1]: https://docs.rs/rstest/0.7.0/rstest/attr.rstest.html#old-compact-syntax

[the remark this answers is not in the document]
almann: I do think having equivalence and non-equivalence classes of annotations as a value list could be used in the value test to test `Eq` more comprehensively, FWIW.

almann:
This reminds me that we should also have these tests factored for the `borrowed` module as well.  Ideally, though not required, it would be nice to make the tests factored well for both (and possibly consider using [`rstest_reuse`][1] as needed to avoid copy pasting the parameters).

[1]: https://crates.io/crates/rstest_reuse

src/value/owned.rs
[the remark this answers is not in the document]
almann: See #158 (based on this PR) which makes some changes that illustrate what I was trying to articulate here.

src/value/owned.rs
```diff
@@ -472,7 +546,30 @@ mod value_tests {
...
+            &|e: &OwnedElement| {
+                assert_eq!(Some(&OwnedSequence::new(vec![OwnedElement::from(OwnedValue::Boolean(true)), OwnedElement::from(OwnedValue::Boolean(false))])), e.as_sequence());
+            }
+        ),
+        case::sexp(
+            OwnedValue::SExpression(OwnedSequence::new(vec![OwnedElement::from(OwnedValue::Boolean(true)), OwnedElement::from(OwnedValue::Boolean(false))])),
+            IonType::SExpression,
+            AsSequence,
+            &|e: &OwnedElement| {
+                assert_eq!(Some(&OwnedSequence::new(vec![OwnedElement::from(OwnedValue::Boolean(true)), OwnedElement::from(OwnedValue::Boolean(false))])), e.as_sequence());
```
almann, line 562: IIRC, we have an `FromIterator` implementation that should make this a little more compact

src/value/owned.rs
```diff
@@ -472,7 +546,30 @@ mod value_tests {
...
+            }
+        ),
+        case::struct_(
+            OwnedValue::Struct(OwnedStruct::from_iter(vec![("greetings", OwnedElement::from(OwnedValue::String("hello".into())))].into_iter())),
+            IonType::Struct,
+            AsStruct,
+            &|e: &OwnedElement| {
+                assert_eq!(Some(&OwnedStruct::from_iter(vec![("greetings", OwnedElement::from(OwnedValue::String("hello".into())))].into_iter())), e.as_struct());
```
almann, line 570: Likewise, see above comment.

almann (commented):
Minor factoring comments below.

desaikd:
*Changes with https://github.com/amzn/ion-rust/pull/152/commits/9e61da3022578df48ccf67e0dac49be331a391e7*
- Added `FromIterator` for Sequence 
- Modified tests for using `FromIterator` for `Struct` and `Sequence`

src/value/owned.rs
```diff
@@ -233,12 +244,7 @@ impl Struct for OwnedStruct {
...
         for (key, value) in &self.text_fields {
-            let other_values: Vec<&OwnedElement> = other.get_all(key).collect();
-            if other_values.is_empty() {
-                return false;
-            }
-            let this_values: Vec<&OwnedElement> = value.iter().map(|(_s, v)| v).collect();
-            if other_values != this_values {
+            if other.get_all(key).ne(value.iter().map(|(_s, v)| v)) {
                 return false;
             }
```
almann: As I look at this I think some separate struct tests are in order, especially because I think we need to take `no_text_fields` into account.

Specifically, I wouldn't try to roll all of this testing into the value tests, I think they are complicated to have their own sets.

almann: Also, what happens if the vector order is different for a given key, this technically doesn't work, does it?

In other words `{a: 1, a: 2}` should equal `{a: 2, a: 1}`, does that work?

src/value/owned.rs
```diff
@@ -547,27 +553,27 @@ mod value_tests {
...
         ),
         case::list(
-            OwnedValue::List(OwnedSequence::new(vec![OwnedElement::from(OwnedValue::Boolean(true)), OwnedElement::from(OwnedValue::Boolean(false))])),
+            OwnedValue::List(vec![OwnedValue::Boolean(true).into(), OwnedValue::Boolean(false).into()].into_iter().collect()),
```
almann, line 556: Not sure if it is worth it with only two elements, but `into_iter().map(|v| v.into()).collect()` might be be a bit more concise.

almann (changes requested):
Can you verify that the semantics on `OwnedStruct`'s `PartialEq` works for unknown text symbols and duplicate key values out of order?

I don't think the tests are sufficient to verify this as written, but let me know if I missed something.

src/value/owned.rs
almann: I misread the code above--it should work (though tests needed), but it is an O(N**2) comparison, If we want to make this O(N lg N) or O(N), we need an intermediate `BTreeSet` or `HashSet`.

almann: See #160.

almann (approved):
Looks good contingent on fast follow up on #161.

PR #156, opened 2021-04-08
Cleanup as per post-merge feedback from @zslayton on #132 and #148.

almann:
Makes `Sequence::get` implementations to avoid extra wrapping
by using `get` instead of the indexing operator.  Also renames a test
method `mk_ts` to `make_timestamp` to be less obtuse.


zslayton (approved):
👍

PR #158, opened 2021-04-08
Refactors `OwnedElement` tests to do some equivalence testing.

almann:
Adds new top-level constructor functions for `OwnedSymbolToken` that
makes it more ergonomic.  Makes the low-level `OwnedSymbolToken::new`
private in preference of the new functions.

Adds annotation tests on the elements in various equivalences.

TODO refactor this to `borrowed`, add non-equivalence, and refactor
to smaller units.

This is based on #152.

### DIFF
--- a/src/value/mod.rs
+++ b/src/value/mod.rs
@@ -57,12 +57,8 @@
 //!
 //! let owned_elem = OwnedElement::new(
 //!     vec![
-//!         OwnedSymbolToken::new(
-//!             None,
-//!             Some(300),
-//!             Some(OwnedImportSource::new("foo", 12))
-//!         ),
-//!         "hello".into()
+//!         local_sid_token(300).with_source("foo", 12),
+//!         text_token("hello")
 //!     ],
 //!     OwnedValue::String("world".into())
 //! );
--- a/src/value/owned.rs
+++ b/src/value/owned.rs
@@ -57,7 +57,7 @@ pub struct OwnedSymbolToken {
 }
 
 impl OwnedSymbolToken {
-    pub fn new(
+    fn new(
         text: Option<Rc<str>>,
         local_sid: Option<SymbolId>,
         source: Option<OwnedImportSource>,
@@ -68,6 +68,38 @@ impl OwnedSymbolToken {
             source,
         }
     }
+
+    /// Decorates the [`OwnedSymbolToken`] with text.
+    pub fn with_text<T: Into<Rc<str>>>(mut self, text: T) -> Self {
+        self.text = Some(text.into());
+        self
+    }
+
+    /// Decorates the [`OwnedSymbolToken`] with a local ID.
+    pub fn with_local_sid(mut self, local_sid: SymbolId) -> Self {
+        self.local_sid = Some(local_sid);
+        self
+    }
+
+    /// Decorates the [`OwnedSymbolToken`] with an [`OwnedImportSource`].
+    pub fn with_source<T: Into<Rc<str>>>(mut self, table: T, sid: SymbolId) -> Self {
+        self.source = Some(OwnedImportSource::new(table, sid));
+        self
+    }
+}
+
+/// Constructs an [`OwnedSymbolToken`] with unknown text and a local ID.
+/// A common case for binary parsing (though technically relevant in text).
+#[inline]
+pub fn local_sid_token(local_sid: SymbolId) -> OwnedSymbolToken {
+    OwnedSymbolToken::new(None, Some(local_sid), None)
+}
+
+/// Constructs an [`OwnedSymbolToken`] with just text.
+/// A common case for text and synthesizing tokens.
+#[inline]
+pub fn text_token<T: Into<Rc<str>>>(text: T) -> OwnedSymbolToken {
+    OwnedSymbolToken::new(Some(text.into()), None, None)
 }
 
 impl PartialEq for OwnedSymbolToken {
@@ -85,7 +117,7 @@ impl Eq for OwnedSymbolToken {}
 impl<T: Into<Rc<str>>> From<T> for OwnedSymbolToken {
     /// Constructs an owned token that has only text.
     fn from(text: T) -> Self {
-        Self::new(Some(text.into()), None, None)
+        text_token(text)
     }
 }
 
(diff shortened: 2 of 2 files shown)


src/value/mod.rs
```diff
@@ -57,12 +57,8 @@
...
 //!     vec![
-//!         OwnedSymbolToken::new(
-//!             None,
-//!             Some(300),
-//!             Some(OwnedImportSource::new("foo", 12))
-//!         ),
-//!         "hello".into()
+//!         local_sid_tok(300).with_source("foo", 12),
+//!         text_tok("hello")
 //!     ],
```
almann, line 62: Hopefully, this new API looks a whole lot better to folks.

src/value/owned.rs
```diff
@@ -68,6 +68,38 @@ impl OwnedSymbolToken {
...
+        self.local_sid = Some(local_sid);
+        self
+    }
+
+    /// Decorates the [`OwnedSymbolToken`] with an [`OwnedImportSource`].
+    pub fn with_source<T: Into<Rc<str>>>(mut self, table: T, sid: SymbolId) -> Self {
+        self.source = Some(OwnedImportSource::new(table, sid));
+        self
+    }
+}
```
almann, line 89: Are these names idiomatic?  @zslayton might have an opinion.

src/value/owned.rs
```diff
@@ -68,6 +68,38 @@ impl OwnedSymbolToken {
...
+pub fn local_sid_tok(local_sid: SymbolId) -> OwnedSymbolToken {
+    OwnedSymbolToken::new(None, Some(local_sid), None)
+}
+
+/// Constructs an [`OwnedSymbolToken`] with just text.
+/// A common case for text and synthesizing tokens.
+#[inline]
+pub fn text_tok<T: Into<Rc<str>>>(text: T) -> OwnedSymbolToken {
+    OwnedSymbolToken::new(Some(text.into()), None, None)
 }
```
almann, line 103: Notice that we cannot construct a token without a `local_sid` or `text`.  Also note that the mutators only add something, never remove.  This is intentional, though it makes `$0` the odd-case that cannot be constructed, but we can add a function to produce that case.

src/value/owned.rs
```diff
@@ -569,13 +601,55 @@ mod value_tests {
...
+                vec![
+                    local_sid_tok(200),
+                    text_tok("hello").with_source("hello_table", 2),
+                ],
+                vec![
+                    local_sid_tok(99),
+                    text_tok("hello").with_local_sid(500).with_source("hello_table", 2),
+                ],
+            ],
+        ]
```
almann, line 645: Really, the symbol token tests should cover this, but it isn't bad to have it here for now.

src/value/owned.rs
```diff
@@ -590,7 +664,7 @@ mod value_tests {
             (AsDecimal, &|e| assert_eq!(None, e.as_decimal())),
             (AsTimestamp, &|e| assert_eq!(None, e.as_timestamp())),
             (AsStr, &|e| assert_eq!(None, e.as_str())),
-            (AsSym, &|e| assert_eq!(true, e.as_sym().is_none())),
+            (AsSym, &|e| assert_eq!(None, e.as_sym())),
```
almann, line 667: `Eq` works now :smiley:

almann (commented):
Some additional context for reviewers below.

src/value/owned.rs
almann: See #159

desaikd (approved):
Looks good! The `OwnedSymbolToken` changes makes it much easier to be used in tests.

src/value/owned.rs
zslayton: These names are idiomatic, but I'd really prefer to spell out the word `token` in `local_sid_tok` and `text_tok` below.

src/value/owned.rs
zslayton: > `$0`... cannot be constructed ... but we can add a function to produce that case.

👍

src/value/owned.rs
```diff
@@ -569,13 +601,55 @@ mod value_tests {
             &|e: &OwnedElement| {
                 assert_eq!(Some(&OwnedStruct::from_iter(vec![("greetings", OwnedElement::from(OwnedValue::String("hello".into())))].into_iter())), e.as_struct());
             }
-        )
+        ),
+        // TODO consider factoring this out of the value tests to make it more contained
+        // TODO consider adding non-equivs for this (really symbol token tests are probably better for this)
+        eq_annotations => [
```
zslayton, line 607: I'm stumbling on this code block, likely because I'm new to `rstest`. This is called `eq_annotations`, but with the exception of the leading empty vec pair example all of the token vectors below contain non-equivalent tokens.

Could you walk me through this a bit?

zslayton (commented):
I like the builder API!

src/value/owned.rs
almann: This `rs_test`-ism makes a test case for every test case with each top-level `Vec`.

Each `Vec` within that is a list of annotations to apply and check element equality for.

```rust
        eq_annotations => [
            // CASE 1
            vec![vec![], vec![]],

            // CASE 2
            vec![
                vec![text_token("hello"), text_token("world")],
                // containing local sids only
                vec![
                    text_token("hello").with_local_sid(20),
                    local_sid_token(21).with_text("world"),
                ],
                // mix of local sid, but all with sources
                vec![
                    text_token("hello").with_source("hello_table", 2),
                    text_token("world").with_local_sid(59).with_source("world_table", 200)
                ]
            ],

            // CASE 3
            vec![
                vec![
                    // local sid only with no text are all $0 equivalent
                    local_sid_token(21),
                    // source import table is the comparator for unknown cases
                    local_sid_token(22).with_source("hello_table", 2)
                ],
                vec![
                    local_sid_token(0),
                    local_sid_token(400).with_source("hello_table", 2),
                ],
                vec![
                    local_sid_token(200),
                    text_token("hello").with_source("hello_table", 2),
                ],
                vec![
                    local_sid_token(99),
                    text_token("hello").with_local_sid(500).with_source("hello_table", 2),
                ],
            ],
        ]
```

Issue #161, opened 2021-04-08
`OwnedStruct` `PartialEq` Does Not Use `no_text_fields`


desaikd:
*Related to #152*
It seems we need to add separate tests for `no_text_fields` in `OwnedStruct`. It might also lead to a change in `PartialEq` for `OwnedStruct`.

almann:
I changed the title, because the implementation doesn't use `no_text_fields` and therefore has a bug.  The testing is important, but the fact that it is a bug is more so.

PR #163, opened 2021-04-09
 Changed `OwnedStruct``PartialEq` to use `no_text_fields` for equivalence

desaikd:
*Issue #161*

*Description of changes:*
This PR works on adding check for `no_text_fields` in `OwnedStruct` `PartialEq`.

*Changes:*
- added `no_text_fields` check inside `PartialEq` for `OwnedStruct` .

*Tests:*
Added unit tests for `no_text_fields`.

### DIFF
--- a/src/value/owned.rs
+++ b/src/value/owned.rs
@@ -271,12 +271,22 @@ impl Struct for OwnedStruct {
 
 impl PartialEq for OwnedStruct {
     fn eq(&self, other: &Self) -> bool {
-        for (key, value) in &self.text_fields {
-            if other.get_all(key).ne(value.iter().map(|(_s, v)| v)) {
-                return false;
-            }
-        }
-        true
+        // check if both the text_fields have same (field_name,value) pairs
+        self.text_fields.iter().all(|(key, value)| {
+            value.iter().all(|(_my_s, my_v)| {
+                other
+                    .get_all(key)
+                    .find(|other_v| my_v == *other_v)
+                    .is_some()
+            })
+        }) && self.no_text_fields.iter().all(|(my_k, my_v)| {
+            // check if both the no_text_fields have same values
+            other
+                .no_text_fields
+                .iter()
+                .find(|(other_k, other_v)| my_k == other_k && my_v == other_v)
+                .is_some()
+        })
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/value/owned.rs
```diff
@@ -606,6 +609,22 @@ mod value_tests {
...
+            }
+        ),
+        case::struct_with_no_text(
+            OwnedValue::Struct(vec![(local_sid_token(21), OwnedValue::String("hello".into()))].into_iter().collect()),
+            IonType::Struct,
+            AsStruct,
+            &|e: &OwnedElement| {
+                assert_eq!(Some(&vec![(local_sid_token(21), OwnedValue::String("hello".into()))].into_iter().collect()), e.as_struct());
+            }
+        ),
```
almann, line 627: I think this test is too trivial, consider the following:

```rust
        case::struct_with_no_text(
            OwnedValue::Struct(
                vec![
                    (local_sid_token(21), OwnedValue::String("hello".into())),
                    (local_sid_token(22), OwnedValue::String("world".into())),
                ].into_iter().collect()),
            IonType::Struct,
            AsStruct,
            &|e: &OwnedElement| {
                assert_eq!(
                    Some(&vec![
                        (local_sid_token(22), OwnedValue::String("world".into())),
                        (local_sid_token(21), OwnedValue::String("hello".into())),
                    ].into_iter().collect()),
                    e.as_struct()
                );
            }
        ),
```

I am pretty sure this would fail, but if I order the tokens the same it would succeed:

```rust
        case::struct_with_no_text(
            OwnedValue::Struct(
                vec![
                    (local_sid_token(21), OwnedValue::String("hello".into())),
                    (local_sid_token(22), OwnedValue::String("world".into())),
                ].into_iter().collect()),
            IonType::Struct,
            AsStruct,
            &|e: &OwnedElement| {
                assert_eq!(
                    Some(&vec![
                        (local_sid_token(21), OwnedValue::String("hello".into())),
                        (local_sid_token(22), OwnedValue::String("world".into())),
                    ].into_iter().collect()),
                    e.as_struct()
                );
            }
        ),
```

What I am doing here is testing something akin to `{$20: "hello", $21: "world"} == {$21: "world", $20: "hello"}`.

src/value/owned.rs
```diff
@@ -276,6 +276,9 @@ impl PartialEq for OwnedStruct {
                 return false;
             }
         }
+        if other.no_text_fields.ne(&self.no_text_fields) {
+            return false;
+        }
```
almann, line 281: I don't think this is right--you either need to generate the nested loop similar to what you have above O(N**2), or materialize one side into a `BTreeSet`/`HashSet` and do the probe on that index at a cost of O(N lg N)/O(N).

```rust
self.no_text_fields.iter().all(|(my_k, my_v)| {
    other
        .no_text_fields
        .iter()
        .find(|(other_k, other_v)| my_k == other_k && my_v == other_v)
        .is_some()
})
```

src/value/owned.rs
```diff
@@ -606,6 +609,22 @@ mod value_tests {
...
+        case::struct_with_no_text(
+            OwnedValue::Struct(vec![(local_sid_token(21).with_source("hello_table", 2), OwnedValue::String("hello".into()))].into_iter().collect()),
+            IonType::Struct,
+            AsStruct,
+            &|e: &OwnedElement| {
+                assert_eq!(Some(&vec![(local_sid_token(21).with_source("hello_table", 2), OwnedValue::String("hello".into()))].into_iter().collect()), e.as_struct());
+            }
+        ),
+        case::struct_with_no_text(
+            OwnedValue::Struct(vec![(local_sid_token(21), OwnedValue::String("hello".into()))].into_iter().collect()),
```
almann, line 621: This test case has the same name as the one previous to it.

src/value/owned.rs
almann: I also am suspicious of this modified case for named fields:

```rust
        case::struct_(
            OwnedValue::Struct(
                vec![
                    ("greetings", OwnedElement::from(OwnedValue::String("hello".into()))),
                    ("greetings", OwnedElement::from(OwnedValue::String("world".into()))),
                ].into_iter().collect()
            ),
            IonType::Struct,
            AsStruct,
            &|e: &OwnedElement| {
                assert_eq!(
                    Some(
                        &vec![
                            ("greetings", OwnedElement::from(OwnedValue::String("world".into()))),
                            ("greetings", OwnedElement::from(OwnedValue::String("hello".into()))),
                        ].into_iter().collect()
                    ),
                    e.as_struct()
                );
            }
        ),
```

When I pulled this code down this also did not work.  I think you will need to do something similar for the fields with text as my suggested fix for the `no_text_fields`.

almann (changes requested):
I really think some separate tests around `OwnedStruct` is warranted to capture the interesting equivalence classes to test, but if you do want to keep using the value tests, see my non-trivial duplicate field tests below.

I also think `PartialEq` for fields with text and no text with duplicate attributes in *different* orders does not work.  I don't trust my eyes, so I pulled the PR code and actually embedded the test cases in my review feedback to verify that they break.

src/value/owned.rs
```diff
@@ -272,13 +272,20 @@ impl Struct for OwnedStruct {
...
+            });
         }
+        self.no_text_fields.iter().all(|(my_k, my_v)| {
+            other
+                .no_text_fields
+                .iter()
+                .find(|(other_k, other_v)| my_k == other_k && my_v == other_v)
+                .is_some()
+        });
         true
```
almann, line 289: I think this is written with side effects being ignored, so I *think* this always returns true.  In other words, I don't think you're doing anything with the predicate and I would like to see some negative cases being tested (e.g. `{a:1} != {a:2}`.  The end statement `true` is a tell.

Consider rewriting as a single expression instead of with the `for` loop, because I think that is getting in your way, but also add some struct cases that deal with non-equivalence.

For example, something like:
```rust
self.text_fields.all(|(key, values)| {
  values.iter().all(...)
}) && self.no_text_fields.iter().all(...)
```

Would eliminate the need for imperative branching (though I do think it hurts readability for readers not used to functional style, so YMMV).

almann (changes requested):
I don't think `eq` is quite right, I have some suggestions below.  I really would add some negative equivalence, parameterized tests for the interesting cases of `OwnedStruct`.

desaikd:
*Changes with https://github.com/amzn/ion-rust/pull/163/commits/b8c0782ca63bf238c9f765e8308efbd9e85ddfe0, https://github.com/amzn/ion-rust/pull/163/commits/21350e63b96dd7c895129c8712135d15c77cf647:*

- Added documentation note for `PartialEq` for `SymbolToken`
- Changed `PartialEq` for `Struct` as per suggestion to consider both the `text_fields` and `no_text_fields`
- Added more tests for `Struct` equivalence and non-equivalence.

almann (changes requested):
I think we should make sure **not** to violate `PartialEq`, but instead note in the trait documentation how `PartialEq` violates the Ion data model.

src/value/owned.rs
```diff
@@ -102,6 +102,19 @@ pub fn text_token<T: Into<Rc<str>>>(text: T) -> OwnedSymbolToken {
...
+///     c  (nil, 200, ("my_table", 1))
+///     b := ("foobar", 200, ("my_table", 1))
+///     a := ("foobar", nil, nil)
+///
+///     a == b, b == a // true
+///     b == c, c == b // true
+///     a == c // For SymbolToken this should return false (i.e. a != c)
+///    ```
+///
+/// [1] - https://doc.rust-lang.org/std/cmp/trait.PartialEq.html
```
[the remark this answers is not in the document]
desaikd: Oh okay. I think I misunderstood this. Let me modify the doc and add it to `SymbolToken` trait

almann (approved):
Looks good--I do think we're stuffing the value/element unit tests with a lot, but we can refactor that later.

almann:
I also pushed c2c79d2 which refactors your documentation  a bit.  Here is the rendered version.
image

PR #164, opened 2021-04-09
Fixes `no_text_fields` `PartialEq` implementation.

almann:
Also makes struct tests less trivial.


almann:
Based on #163--this has failing tests for duplicate text fields with different order and the fix for `no_text_fields` and tests the verify it.

PR #166, opened 2021-04-09
Configures workflow to not run on PRs within repository.

### LINKED ISSUE #162: Duplicate CI Workflow Runs for `push` and `pull_request`
When creating a PR within the repo, we get a duplicate set of runs:

image

We should look at deduplicating this:

This seems to be related:

https://github.community/t/duplicate-trigger-on-the-same-commit-push-and-pull-request/18473

almann:
Right now we build on `push` or `pull_request`.  This is fine for
PRs from forks, but not so great on internal PRs as we get two runs,
this makes us ignore the `pull_request` run for the latter case.

Resolves #162.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -6,6 +6,9 @@ jobs:
   build:
     name: Build and Test
     runs-on: ${{ matrix.os }}
+    # We want to run on external PRs, but not on internal ones as push automatically builds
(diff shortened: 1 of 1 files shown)


almann:
Here the *internal* PR skips the redundant `pull_request` trigger:

image

#168 is a slightly different commit with the same effective change (to avoid colliding the workflows via the same actual commit):

image

So I would say this is working!

zslayton (approved):

PR #176, opened 2021-04-12
Changes rustup profile to `default` for `nightly`.

almann:
Coverage relies on `nightly`, and the `default` profile seems to be a bit more stable.

I noticed that 2021-04-11 nightly started breaking for `clang_sys`.  It seems that the version one installs by default for nightly was a few weeks behind.  I am trying setting this profile in the hopes that we get a more "stable" nightly for our coverage CI.  If we run into more problems, we might have to start pinning the nightly version to keep things more reliable.

Hopefully someday soon we won't have to rely on unstable `-Zprofile` for code coverage.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Rust Toolchain
         uses: actions-rs/toolchain@v1
         with:
-          profile: minimal
+          profile: default
           toolchain: nightly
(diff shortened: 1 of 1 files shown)


almann:
It is not clear to me why this change would trigger a reduction of coverage other than differences between the current nightly and the version this change runs with the `default` profile.

zslayton (approved):

Issue #153, opened 2021-04-06
Add tests for `SymbolToken` in Element API 


desaikd:
Add unit tests for `SymbolToken` with various test cases, specially covering the unknown text case with respect to the [Ion data model](https://amzn.github.io/ion-docs/docs/symbols.html#data-model).

Related to #114

PR #181, opened 2021-04-12
added more unit tests for `OwnedSymbolToken`

desaikd:
*Issue #153*

*Description of changes:*

This PR works on creating more unit tests for `OwnedSymbolToken`.

### DIFF
--- a/src/value/owned.rs
+++ b/src/value/owned.rs
@@ -558,7 +558,7 @@ mod value_tests {
             AsStr,
             &|e: &OwnedElement| assert_eq!(Some("hello"), e.as_str())
         ),
-        case::sym_text(
+        case::sym_with_text(
             OwnedValue::Symbol("hello".into()),
             IonType::Symbol,
             vec![AsStr, AsSym],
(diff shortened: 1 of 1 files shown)


src/value/owned.rs
```diff
@@ -576,6 +576,51 @@ mod value_tests {
...
+        // SymbolTokens with same texts are always equivalent
+        case::sym_with_text_local_sid_source_as_sym(
+            OwnedValue::Symbol(text_token("foo")),
+            IonType::Symbol,
+            vec![AsStr, AsSym],
+            &|e: &OwnedElement| {
+                assert_eq!(Some(&text_token("foo").with_source("greetings", 1)), e.as_sym());
+                assert_eq!(Some(&text_token("foo").with_source("greetings", 1).with_local_sid(10)), e.as_sym());
+            }
+        ),
```
almann, line 623: The problem with saying this is an `OwnedSymbolToken` set of tests is that it is buried in `OwnedElement` testing.

I would instead push this out into its own tests e.g.:

```rust
/// Each case is a set of tokens that are the same, and a set of tokens that are not ever equal to the first.
/// This should test symmetry/transitivity/commutativity
fn owned_symbol_token_eq(
  #[case] equivalent: Vec<OwnedSymbolToken>,
  #[case] non_equivalent: Vec<OwnedSymbolToken>) {
  ...
}
```

almann (changes requested):
See below.

src/value/owned.rs
```diff
@@ -506,6 +506,74 @@ mod value_tests {
...
+    #[rstest(
+        // SymbolTokens with same text are equivalent
+        case::sym_text(
+            vec![
+                text_token("foo"),
+                text_token("foo").with_local_sid(10),
+                text_token("foo").with_local_sid(10).with_source("greetings", 2)
+            ],
+            vec![
+                text_token("foo"),
```
almann, line 520: This doesn't look right, how come `text_token("foo")` in the second vector is not equivalent to the first set of `"foo"` tokens?

src/value/owned.rs
```diff
@@ -506,6 +506,74 @@ mod value_tests {
...
+        for eq_this_token in &equivalent {
+            for eq_other_token in &equivalent {
+                assert_eq!(eq_this_token, eq_other_token);
+            }
+        }
+
+        // check if non_equivalent vector contains a set of tokens that are not ever equal to the first.
+        for non_eq_token in &non_equivalent[1..] {
+            assert_ne!(&non_equivalent[0], non_eq_token);
+        }
```
almann, line 574: I misunderstood what the second parameter meant.  It might be clearer to model this set as never equivalent to a member of the first set otherwise this vector is disjoint and should be in its own test cases (which is fine too).

src/value/owned.rs
```diff
@@ -506,6 +506,68 @@ mod value_tests {
...
+                local_sid_token(10).with_source("hello_table", 2)
+            ]
+        ),
+        case::sym_local_sid(
+            // local sids with no text are equivalent to each other and to SID $0
+            vec![
+                local_sid_token(200),
+                local_sid_token(100),
+                text_token("foo"),
+                text_token("foo").with_local_sid(200)
```
almann, line 531: Shouldn't `text_token("foo")` **not** be equivalent (and in the second vector)?

src/value/owned.rs
```diff
@@ -506,6 +506,68 @@ mod value_tests {
...
+            vec![
+                text_token("foo"),
+                text_token("foo").with_local_sid(10),
+                text_token("foo").with_local_sid(10).with_source("greetings", 2)
+            ],
+            vec![
+                text_token("bar"),
+                local_sid_token(10).with_source("greetings", 1),
+                local_sid_token(10).with_source("hello_table", 2)
+            ]
```
almann, line 523: I would put tokens without text here too, because all of them should be not equivalent. (e.g. local symbol with no text and local symbol with shared import and no text (with same source as above).

src/value/owned.rs
```diff
@@ -506,6 +506,68 @@ mod value_tests {
...
+            vec![
+                local_sid_token(200).with_source("greetings", 1),
+                local_sid_token(100).with_source("greetings", 1),
+                local_sid_token(100).with_source("greetings", 1).with_text("foo")
+            ],
+            vec![
+                local_sid_token(200).with_source("greetings", 2),
+                local_sid_token(200).with_source("hello_table", 1),
+                local_sid_token(200)
+            ]
```
almann, line 548: I would also add a token with text `"greetings"` and nothing else.

almann (commented):
Looking pretty good!  I suggest a couple of extra cases below on non-equivalence and we can call this good.

src/value/owned.rs
desaikd: I have put it in the equivalent vector because all the token there contains `text = "foo"` hence all are equivalent.

src/value/owned.rs
desaikd: I thought putting local symbol with no text i.e. `local_sid_token(200)` would be equal to `text_token("foo")` because for both `ImportSource` would be `None`. For local_symbol with a source I have already added `local_sid_token(10).with_source("greetings", 1)` for it. Does this make sense or am I missing something?

src/value/owned.rs
desaikd: I will add this change.

src/value/owned.rs
almann: It would be good to have a `local_sid_token` with no source or text for completeness

src/value/owned.rs
```diff
@@ -107,7 +107,26 @@ impl PartialEq for OwnedSymbolToken {
...
+                    // if one token has text and another doesn't have text and
+                    // for both tokens source is unknown(i.e. None)
+                    false
+                }
+            } else {
+                self.source == other.source
+            }
         }
     }
 }
```
almann, line 132: I think this collapses to:

```rust
if other.text != None || self.text != None {
  // if either side has text, we only compare text
  other.text == self.text
} else {
  // no text--so the sources must be the same (all local symbols with no source are the same)
  other.source == self.source
}

almann (commented):
Can you verify or deny that my read of the logic is more simply modeled as below?

almann (approved):
:rocket: LGTM.

PR #182, opened 2021-04-13
Pins Code Coverage to `nightly-2021-03-23`.

almann:
Nightly builds can be really unstable--and has shown to break our
pipeline.  This pins the version to make it explicit to which
version of the compiler we use to generate code coverage.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -22,7 +22,8 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: default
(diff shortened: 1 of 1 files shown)


almann:
https://github.com/almann/ion-rust/runs/2335309770 - build succeeds.

zslayton (approved):

almann:
Chatting with @zslayton -- we should probably update the pinned version with every real release.  It is not the end of the world if we don't, but that seems reasonable (and sadly manual).

Issue #157, opened 2021-04-08
Refactor `owned` Tests to `borrowed` Module 


almann:
This reminds me that we should also have these tests factored for the `borrowed` module as well.  Ideally, though not required, it would be nice to make the tests factored well for both (and possibly consider using [`rstest_reuse`][1] as needed to avoid copy pasting the parameters).

[1]: https://crates.io/crates/rstest_reuse

_Originally posted by @almann in https://github.com/amzn/ion-rust/issues/152#issuecomment-816002186_

PR #183, opened 2021-04-14
Added PartialEq and tests for borrowed module

desaikd:
*Issue #157*

*Description of changes:*
This PR works on refactoring owned tests to add it for borrowed module.

*Changes:*
- Added PartialEq/Eq for `BorrowedSymbolToken`, `BorrowedImportSource`, `BorrowedStruct` and `BorrowedSequence`
- Added `FromIterator` for `BorrowedSequence`
- Added unit tests for borrowed module same as owned tests.

### DIFF
--- a/src/value/borrowed.rs
+++ b/src/value/borrowed.rs
@@ -149,6 +149,17 @@ impl<'val> BorrowedSequence<'val> {
     }
 }
 
+impl<'val> FromIterator<BorrowedElement<'val>> for BorrowedSequence<'val> {
+    /// Returns an owned sequence from the given iterator of elements.
+    fn from_iter<I: IntoIterator<Item = BorrowedElement<'val>>>(iter: I) -> Self {
+        let mut children: Vec<BorrowedElement> = Vec::new();
+        for elem in iter {
+            children.push(elem);
+        }
+        Self { children }
+    }
+}
+
 impl<'val> Sequence for BorrowedSequence<'val> {
     type Element = BorrowedElement<'val>;
 
--- a/src/value/borrowed.rs
+++ b/src/value/borrowed.rs
@@ -56,7 +56,7 @@ pub struct BorrowedSymbolToken<'val> {
 }
 
 impl<'val> BorrowedSymbolToken<'val> {
-    pub fn new(
+    fn new(
         text: Option<&'val str>,
         local_sid: Option<SymbolId>,
         source: Option<BorrowedImportSource<'val>>,
--- a/src/value/borrowed.rs
+++ b/src/value/borrowed.rs
@@ -29,6 +29,14 @@ impl<'val> BorrowedImportSource<'val> {
     }
 }
 
+impl<'val> PartialEq for BorrowedImportSource<'val> {
+    fn eq(&self, other: &Self) -> bool {
+        self.table == other.table && self.sid == other.sid
+    }
+}
+
+impl<'val> Eq for BorrowedImportSource<'val> {}
+
 impl<'val> ImportSource for BorrowedImportSource<'val> {
     fn table(&self) -> &str {
         self.table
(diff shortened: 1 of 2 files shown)


almann (commented):
It seems you are doing this--but I wanted to emphasize that we should make the constructors symmetric (insofar as they make sense) between the two versions.

The real rationale for the former, is that I think we should make the test drivers generic `<E: Element, ...>` and be able to operate on `Borrowed...` and `Owned...` variants.  We can then collapse the tests more easily.

An interesting question is how do we avoid copy/paste for the tests themselves.  I might suggest a generic function that produces some `<E: Element>` or the like and use them for each form.  Ideally, we'd automate the expansion via macro, but I think doing this manually for now is fine as it reduces the copy/paste of the test cases.

That said, if this is too onerous, lets allow copy/paste and cut a follow up issue to make this generic and no duplicative.

src/value/borrowed.rs
```diff
@@ -59,8 +67,54 @@ impl<'val> BorrowedSymbolToken<'val> {
...
+        self.text = Some(text);
+        self
+    }
+
+    /// Decorates the [`BorrowedSymbolToken`] with a local ID.
+    pub fn with_local_sid(mut self, local_sid: SymbolId) -> Self {
+        self.local_sid = Some(local_sid);
+        self
+    }
+
```
almann, line 82: Codecov indicates that this was not tested--is it correct?  Might be worth re-jiggering some test cases to hit this (probably true for `OwnedSymbolToken` tests as well).

src/value/borrowed.rs
```diff
@@ -59,8 +67,54 @@ impl<'val> BorrowedSymbolToken<'val> {
...
+    BorrowedSymbolToken::new(None, Some(local_sid), None)
+}
+
+/// Constructs an [`BorrowedSymbolToken`] with just text.
+/// A common case for text and synthesizing tokens.
+#[inline]
+pub fn borrowed_text_token(text: &str) -> BorrowedSymbolToken {
+    BorrowedSymbolToken::new(Some(text), None, None)
+}
+
```
almann, line 103: I would not call this `borrowed_` I would just call this `text_token` and re-alias in import.  Also codecov is saying these were not tested which surprises me.

src/value/borrowed.rs
```diff
@@ -195,10 +268,33 @@ impl<'val> Struct for BorrowedStruct<'val> {
...
 
+impl<'val> PartialEq for BorrowedStruct<'val> {
+    fn eq(&self, other: &Self) -> bool {
+        // check if both the text_fields have same (field_name,value) pairs
+        self.text_fields.iter().all(|(key, value)| {
+            value.iter().all(|(_my_s, my_v)| {
+                other
+                    .get_all(key)
+                    .find(|other_v| my_v == *other_v)
+                    .is_some()
```
almann, line 279: Codecov is complaining that L275-278 and L280 is not tested, please confirm (could be a misattribution, but worth checking).

almann (commented):
Can you verify that codecov is accurate?  Also, let's not call things `borrowed_` in the `borrowed` submodule.

src/value/borrowed.rs
desaikd: I do not think these lines that codecov shows as not tested are correct. The tests seems to cover these lines.

src/value/mod.rs
```diff
@@ -57,8 +57,8 @@
 //!
 //! let owned_elem = OwnedElement::new(
 //!     vec![
-//!         local_sid_token(300).with_source("foo", 12),
-//!         text_token("hello")
+//!         owned_local_sid_token(300).with_source("foo", 12),
+//!         owned_text_token("hello")
 //!     ],
 //!     OwnedValue::String("world".into())
```
almann, line 63: Should the borrowed symbol token below be rewritten to use these constructor functions (IIRC `OwnedSymbolToken` doesn't have a `new` anymore, we should probably do the same to `BorrowedSymbolToken`).  Those should probably be aliased to be consistent with this.

almann (approved):
Looks good--minor issue on the doc test below.

PR #193, opened 2021-04-18
[ion-c-sys] Short circuits empty case `IonCReader.read_bytes()`.

almann:
amzn/ion-c#233 triggers an assertion failure when reading an empty LOB,
so we avoid this case explicitly.

### DIFF
--- a/ion-c-sys/src/reader.rs
+++ b/ion-c-sys/src/reader.rs
@@ -323,11 +323,13 @@ pub trait IonCReader {
     /// # use ion_c_sys::reader::*;
     /// # use ion_c_sys::result::*;
     /// # fn main() -> IonCResult<()> {
-    /// let mut reader = IonCReaderHandle::try_from("{{\"hello\"}} {{d29ybGQ=}}")?;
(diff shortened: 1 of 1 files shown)


ion-c-sys/src/reader.rs
```diff
@@ -510,6 +512,10 @@ impl<'a> IonCReader for IonCReaderHandle<'a> {
     fn read_bytes(&mut self) -> IonCResult<Vec<u8>> {
         let mut len = 0;
         ionc!(ion_reader_get_lob_size(self.reader, &mut len))?;
+        if len == 0 {
+            // short-circuit the empty case
```
zslayton, line 516: I think the 'why' would be good to add here, which I gather is that `ion_reader_read_lob_bytes` will return an error for a zero-length lob.

zslayton (approved):
One comment suggestion below.

ion-c-sys/src/reader.rs
almann: I believe that to be a point in time though--that is a bug in Ion C, I can add a note about that though.

Issue #138, opened 2021-03-26
Provide an API to parse Ion into `Iterator` of `Element`


almann:
We should have an API that plumbs to either `ion-c-sys` or the native parser to iterate as `Element` instances.  We should try our best to make such an API easy to replace the implementation with especially if we source the `ion-c-sys` parser.

For example, the most basic version of this might be something like:

```rust
fn element_iter_from_buf<'a>(buf: &'a [u8]) -> impl Iterator<Item = BorrowedElement<'a>> {
  ...
}
```

Of course if the above is to wordy, some clever `From<AsRef<[u8]>>` style API could be defined.

More stream oriented APIs and file based APIs would probably be appropriate as well.

PR #196, opened 2021-04-19
Loader

almann:
`Loader` provides a method to iterate over a `&[u8]` as `Element`.

The loader trait is currently concrete and based on `OwnedElement`.
In the future we should attempt to make this more generic and/or
be based on `BorrowedElement`.

Implements #138.

### DIFF
--- a/src/value/loader.rs
+++ b/src/value/loader.rs
@@ -1,13 +1,12 @@
 // Copyright Amazon.com, Inc. or its affiliates.
 
 use crate::result::IonResult;
-use crate::value::owned::{OwnedElement, OwnedStruct, OwnedSymbolToken, OwnedValue};
+use crate::value::owned::{OwnedElement, OwnedSequence, OwnedStruct, OwnedSymbolToken, OwnedValue};
 use crate::value::AnyInt;
 use crate::IonType;
 use ion_c_sys::reader::{IonCReader, IonCReaderHandle};
 use ion_c_sys::ION_TYPE;
 use std::convert::{TryFrom, TryInto};
-use std::iter::FromIterator;
 
 use super::owned::text_token;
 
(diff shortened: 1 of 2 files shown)


src/value/loader.rs
```diff
@@ -0,0 +1,462 @@
...
+///
+/// [gat]: https://rust-lang.github.io/rfcs/1598-generic_associated_types.html
+/// [et]:https://rust-lang.github.io/rfcs/2071-impl-trait-existential-types.html
+pub trait Loader {
+    /// Parses Ion over a given slice of data and yields each top-level value as
+    /// an [`Element`] instance.
+    ///
+    /// The [`Iterator`] will generally return `Ok([Element])` but on a failure of
+    /// parsing it will return an `Err([IonError])` and then a `None` to signal no more
+    /// elements.
```
zslayton, line 31: This makes sense conversationally, but I suggest preserving the `Option` wrapper when discussing the return values.

```suggestion
    /// The [`Iterator`] will generally return `Some(Ok([Element]))` but on a failure of
    /// parsing it will return a `Some(Err([IonError]))` and then a `None` to signal no more
    /// elements.
```

src/value/loader.rs
```diff
@@ -0,0 +1,462 @@
...
+use ion_c_sys::ION_TYPE;
+use std::convert::{TryFrom, TryInto};
+
+use super::owned::text_token;
+
+/// TODO add/refactor trait/implementation for borrowing over some context
+///      we could make it generic with generic associated types or just have a lifetime
+///      scoped implementation
+
+/// Loads Ion data into [`Element`] instances.
```
zslayton, line 17: The docs around this trait are written in terms of the generics we plan to have eventually. For the time being, should we just refer to `OwnedElement` everywhere to align with the current implementation?

almann: I'd like to move this to `BorrowedElement` sooner than later to make sure this is doable with this API.  Also, the real docs for these APIs are in the traits and not the implementations, so not clear that would help anyone.  That said if you think this would be clearer, I could do that, I really do want folks to think of it in terms of `Element`, in other words, the lack of *existential types* in more useful places is really the reason why `OwnedElement` pops out.

zslayton (approved):

Issue #184, opened 2021-04-15
Make the `Element` construction generic


desaikd:
Currently the `Element` construction is separate for both `OwnedElement` and `BorrowedElement` with their own `new` methods. We need to have a generic construction based on `annotations` and `Value`. This will be useful when working with creating generic test case code and generic test cases for `Owned` and `Borrowed` . 

This implementation is trickier in terms of working with a generic `Value` implementation, which currently is an `enum` `BorrowedValue`/`OwnedValue`. It might be worth having a `new` in terms of `Self::Value` and `Self:SymbolToken` trait for this generic implementation.

almann:
Related to #157 and #183.

almann:
@desaikd and I talked about this at length last week, and came up with the following scheme for how to construct `Element` and `SymbolToken` generically.

Where ever possible, we should implement `From<X>` for a `...SymbolToken` and `...Element`.  We can further add this to the traits, but at a minimum it makes generic code such as `<E: Element + From<X> + ...>` possible.  To avoid boilerplate, for the common types we should add those as constraints on the `Element` and `SymbolToken` traits as base requirements for any such implementation.

```rust
trait SymbolToken
where
  Self: From<&str> + From<String> {
  ...
}
```

`clob` and `blob` are interesting cases as it is ambiguous how which one is meant for `&[u8]` or `Vec<u8>` or equivalent.  For those, we can add a builder type that is convertible from `&[u8]` and `Vec<u8>`:

```rust
trait LobBuilder
where
  Self: From<&[u8]> + From<Vec<u8>> {
  type Element: Element;
  ...
  fn into_clob(self) -> Self::Element;
  fn into_blob(self) -> Self::Element;
}
```

This makes the conversion explicit.  Likewise in `Sequence`, we should add `into_list` and `into_sexp` that are similar to the `LobBuilder` above except that `Sequence` is useful in its own right.

Some examples of the above factoring in a generic context:

```rust
fn make_stuff<E: Element>() {
  let clob_elem = E::LobBuilder::from(b"hello").into_clob();
  let blob_elem = E::LobBuilder::from(b"world).into_blob();
  ...
}
```

There is an additional trait bounds on `Element` that would require nested types to have the same element type as `Self`:

```rust
trait Element ... {
  type LobBuilder: LobBuilder<Element=Self> ...;
}
```

See [this playground link][1] for a really trivial version for this

Doing this I think will go a long way towards #136 (but may not satisfy it).

[...]

PR #199, opened 2021-04-20
adding changes for making `Element` construction generic

desaikd:
*Issue #184*

*Description of changes:*
This PR works on making Element construction generic.

*Changes:*
- Added `From<X>` for `Owned/BorrowedElement` 
- Added a LobBuilder trait which specifies construction of `clob` and `blob` using `into_clob` and `into_blob` respectively.
- Also added `into_list` and `into_sexp` in `Sequence` trait similar to `LobBuilder`.

*Test:*
Modified tests to use `From<X>` of `Owned/BorrowedElement`.

### DIFF
--- a/src/value/borrowed.rs
+++ b/src/value/borrowed.rs
@@ -139,28 +139,28 @@ impl<'val> SymbolToken for BorrowedSymbolToken<'val> {
 }
 
 /// An borrowed implementation of [`Builder`].
-impl<'val> Builder for BorrowedValue<'val> {
+impl<'val> Builder for BorrowedElement<'val> {
     type Element = BorrowedElement<'val>;
     type Sequence = BorrowedSequence<'val>;
 
-    fn new_null(e_type: IonType) -> Self {
-        BorrowedValue::Null(e_type)
+    fn new_null(e_type: IonType) -> Self::Element {
+        BorrowedValue::Null(e_type).into()
     }
 
-    fn new_clob(bytes: &'static [u8]) -> Self {
-        BorrowedValue::Clob(bytes)
+    fn new_clob(bytes: &'static [u8]) -> Self::Element {
+        BorrowedValue::Clob(bytes).into()
     }
 
-    fn new_blob(bytes: &'static [u8]) -> Self {
-        BorrowedValue::Blob(bytes)
+    fn new_blob(bytes: &'static [u8]) -> Self::Element {
+        BorrowedValue::Blob(bytes).into()
     }
 
-    fn new_list(seq: Self::Sequence) -> Self {
-        BorrowedValue::List(seq)
+    fn new_list(seq: Self::Sequence) -> Self::Element {
+        BorrowedValue::List(seq).into()
     }
 
-    fn new_sexp(seq: Self::Sequence) -> Self {
-        BorrowedValue::SExpression(seq)
+    fn new_sexp(seq: Self::Sequence) -> Self::Element {
+        BorrowedValue::SExpression(seq).into()
     }
 }
 
@@ -431,7 +431,7 @@ impl<'val> Element for BorrowedElement<'val> {
     type SymbolToken = BorrowedSymbolToken<'val>;
     type Sequence = BorrowedSequence<'val>;
     type Struct = BorrowedStruct<'val>;
-    type Builder = BorrowedValue<'val>;
+    type Builder = BorrowedElement<'val>;
 
     fn ion_type(&self) -> IonType {
         use BorrowedValue::*;
@@ -751,7 +751,7 @@ mod borrowed_value_tests {
             }
         ),
         case::blob(
-            BorrowedValue::new_blob(b"world").into(),
+            BorrowedElement::new_blob(b"world"),
             IonType::Blob,
             AsBytes,
             &|e: &BorrowedElement| {
@@ -759,7 +759,7 @@ mod borrowed_value_tests {
             }
         ),
         case::clob(
-            BorrowedValue::new_clob(b"goodbye").into(),
+            BorrowedElement::new_clob(b"goodbye"),
             IonType::Clob,
             AsBytes,
             &|e: &BorrowedElement| {
@@ -767,7 +767,7 @@ mod borrowed_value_tests {
             }
         ),
         case::list(
-            BorrowedValue::new_list(vec![true.into(), false.into()].into_iter().collect()).into(),
+            BorrowedElement::new_list(vec![true.into(), false.into()].into_iter().collect()),
             IonType::List,
             AsSequence,
             &|e: &BorrowedElement| {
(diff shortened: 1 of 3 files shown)


src/value/borrowed.rs
```diff
@@ -304,8 +347,8 @@ pub enum BorrowedValue<'val> {
     String(&'val str),
     Symbol(BorrowedSymbolToken<'val>),
     Boolean(bool),
-    Blob(&'val [u8]),
-    Clob(&'val [u8]),
+    Blob(BorrowedLobBuilder<'val>),
+    Clob(BorrowedLobBuilder<'val>),
```
almann, line 351: I don't think we need to store this in the value.

almann (changes requested):
I do think we don't want to store the `LobBuilder` in the values, I don't see any reason to do so.

src/value/owned.rs
```diff
@@ -340,10 +385,65 @@ impl From<OwnedValue> for OwnedElement {
     }
 }
 
+impl From<IonType> for OwnedElement {
+    fn from(ion_type: IonType) -> Self {
+        OwnedValue::Null(ion_type).into()
+    }
+}
```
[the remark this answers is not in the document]
desaikd: I don't think I understand this clearly. Did you mean something like this:
```
pub trait Builder {
    type Element: Element + ?Sized;

    /// build an element from Builder
    fn into_elem(self) -> Self::Element;
}
```

where this would then have to use some `enum` for it to return `null`/`clob`/`blob` `Element`. Can you please explain with an example just to be clear on this.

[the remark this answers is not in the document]
almann: Here is a more concrete example definition:

```rust
trait Builder {
    type Element: Element + ?Sized;
    type Sequence: Sequence<Element = Self::Element> + ?Sized;

    fn with_null(e_type: Type) -> Self;
    fn with_clob(seq: Self::Sequence) -> Self;
    fn with_blob(seq: Self::Sequence) -> Self;

    fn into(self) -> Self::Element;
}

trait Element ... {
    type Builder: Builder<Element = Self> + ?Sized;
    ...
}
```

Note that we could flatten this into `Element` (maybe make these methods `new_...` instead of `with...`):

```rust
trait Element ... {
    ...
    new_null(ion_type: IonType) -> Self;
    new_clob(seq: Self::Sequence) -> Self;
    new_blob(seq: Self::Sequence) -> Self;
}
```

Maybe the latter is preferable--I'd love to get your take and @zslayton to weigh in on that.

almann (commented):
Looks good--we just got to close out on what we want to do for the `null` case.

src/value/mod.rs
```diff
@@ -262,7 +262,7 @@ pub trait Element {
     type SymbolToken: SymbolToken + ?Sized;
     type Sequence: Sequence<Element = Self> + ?Sized;
     type Struct: Struct + ?Sized;
```
almann, line 264: *Minor*: Sine you made a similar change in another part of the code, we probably want `Struct<SymbolToken=Self::SymbolToken, Element=Self>`

src/value/mod.rs
```diff
@@ -491,13 +485,22 @@ pub trait Struct {
...
 
-    /// build `clob` from Lob
-    fn into_clob(self) -> Self::Element;
+    /// Build a `list` from Sequence using Builder
+    fn new_list(seq: Self::Sequence) -> Self;
 
-    /// build `blob` from Lob
-    fn into_blob(self) -> Self::Element;
+    /// Build a `sexp` from Sequence using Builder
+    fn new_sexp(seq: Self::Sequence) -> Self;
```
almann: The more I think about this, the more I wonder if this should just all return `Self::Element`, I think this made sense as `Self` when this was `LobBuilder`, but I am not so sure now, what do you think?  It also means we don't need an `Into<Element>` trait implementation (which adds a cycle anyhow).

If we keep with this factoring, we will need some kind of `into_element()` for this to work generically to go from `Builder` to `Element`.

If we do go straight to `Self::Element` you can implement it for `...Element` and avoid returning an `...Value` as the builder intermediate.

almann (commented):
Would like to get your thoughts on the below, but other than that I am just about ready to :ship:

src/value/mod.rs
desaikd: Yes. I think we should go with `Self::Element` as the return type here rather than returning intermediate `....Value`.

almann (approved):
LGTM :rocket:

src/value/borrowed.rs
```diff
@@ -137,6 +138,32 @@ impl<'val> SymbolToken for BorrowedSymbolToken<'val> {
     }
 }
 
+/// An borrowed implementation of [`Builder`].
```
zslayton, line 141: Nit:
```suggestion
/// A borrowed implementation of [`Builder`].
```
or maybe:
```suggestion
/// A borrowing implementation of [`Builder`].
```

src/value/borrowed.rs
```diff
@@ -137,6 +138,32 @@ impl<'val> SymbolToken for BorrowedSymbolToken<'val> {
...
+
+    fn new_null(e_type: IonType) -> Self::Element {
+        BorrowedValue::Null(e_type).into()
+    }
+
+    fn new_clob(bytes: &'static [u8]) -> Self::Element {
+        BorrowedValue::Clob(bytes).into()
+    }
+
+    fn new_blob(bytes: &'static [u8]) -> Self::Element {
```
zslayton, line 154: I was a little surprised to see the `new_*lob` functions requiring a `'static` lifetime, but I see that's caused by `Self::Element` not having a lifetime of its own. My kingdom for a GATs! 😞

src/value/borrowed.rs
```diff
@@ -573,22 +662,22 @@ mod borrowed_value_tests {
...
-            BorrowedValue::Null(IonType::Null),
+            IonType::Null.into(),
             IonType::Null,
             IsNull,
             &|e: &BorrowedElement| assert_eq!(true, e.is_null())
         ),
         // TODO more null testing (probably its own fixture)
         case::bool(
-            BorrowedValue::Boolean(true),
+            true.into(),
```
zslayton, line 674: Nice, the new trait implementations make these short and sweet.

src/value/mod.rs
```diff
@@ -483,3 +484,23 @@ pub trait Struct {
...
+    fn new_clob(bytes: &'static [u8]) -> Self::Element;
+
+    /// Build a `blob` using Builder
+    fn new_blob(bytes: &'static [u8]) -> Self::Element;
+
+    /// Build a `list` from Sequence using Builder
+    fn new_list(seq: Self::Sequence) -> Self::Element;
+
+    /// Build a `sexp` from Sequence using Builder
+    fn new_sexp(seq: Self::Sequence) -> Self::Element;
```
zslayton, line 505: Should these instead take `S where S: Into<Self::Sequence>`? (If so, I'm happy with tracking this in another issue.)

zslayton (approved):
Looks good! I've added a couple of notes and for-your-considerations below.

src/value/mod.rs
desaikd: We are punting on this according to an offline discussion. We already have `Sequence` as the concrete associated type here.

Issue #204, opened 2021-04-22
Fill in `Builder` Trait


almann:
#199 will add a `Builder` trait that is associated with an `Element` to enable generic construction.  With that, we can fill in all of the types to be explicit.  We also have generic support via the `From` trait from obvious scalar values, but this would make that construction a bit more explicit if the user desired and make the `Builder` API more consistent.

We should also make sure the `From` traits are on the `Element` trait as well to make sure the following works:

```rust
fn do_something<E: Element>() {
  // From<i64> should work
  let e1 = 64.into();
  
  // Also the builder should work
  let e2 = E::Builder::new_i64(64);
}
```

Issue #135, opened 2021-03-26
Support Mutability in `Element` APIs


almann:
Once we flesh out `Element` for read cases, we should carefully add mutation APIs.  Specifically, the following probably makes sense at a minimum:

* Mutating the `Struct`/`Sequence` with `insert`/`push`/`pop`/`remove` type operations.
* Mutating the `Element` annotations.

With respect to mutating values themselves, this is less clear if necessary, we may choose to punt on that an cut a separate issue to track.

This probably implies that the borrowed implementation use a `Cow<'val, T>` for string/byte slice references to allow them to be changed or inserted by way of owned objects.

This is related to #114.

PR #206, opened 2021-04-22
Filled `Builder` trait with `new_...` methods for all types

desaikd:
*Issue #204*

*Description of changes:*
This PR works on filling in all `new_...` methods for `Builder` trait created by #199.

*Changes:*
- Added `new_...` method for all types in Builder trait
- Modified tests to use `new_...` method construct element.

*TODO:*
- Add `From<X>` for all types in `Element` trait as defined below:
```rust
trait Element 
where
     Self:From<i64> + From<f64> + … {
```
- Add generic test to verify above mentioned element constructions `e1` and `e2`.

### DIFF
--- a/src/value/borrowed.rs
+++ b/src/value/borrowed.rs
@@ -141,10 +141,44 @@ impl<'val> SymbolToken for BorrowedSymbolToken<'val> {
 /// A borrowed implementation of [`Builder`].
 impl<'val> Builder for BorrowedElement<'val> {
     type Element = BorrowedElement<'val>;
+    type SymbolToken = BorrowedSymbolToken<'val>;
     type Sequence = BorrowedSequence<'val>;
+    type Struct = BorrowedStruct<'val>;
 
     fn new_null(e_type: IonType) -> Self::Element {
         BorrowedValue::Null(e_type).into()
+    }
+
+    fn new_bool(bool: bool) -> Self::Element {
+        BorrowedValue::Boolean(bool).into()
+    }
+
+    fn new_string(str: &'static str) -> Self::Element {
+        BorrowedValue::String(str).into()
+    }
+
+    fn new_symbol(sym: Self::SymbolToken) -> Self::Element {
+        BorrowedValue::Symbol(sym).into()
+    }
+
+    fn new_i64(int: i64) -> Self::Element {
+        BorrowedValue::Integer(AnyInt::I64(int)).into()
+    }
+
+    fn new_big_int(big_int: BigInt) -> Self::Element {
+        BorrowedValue::Integer(AnyInt::BigInt(big_int)).into()
+    }
+
+    fn new_decimal(decimal: Decimal) -> Self::Element {
+        BorrowedValue::Decimal(decimal).into()
+    }
+
+    fn new_timestamp(timestamp: Timestamp) -> Self::Element {
+        BorrowedValue::Timestamp(timestamp).into()
+    }
+
+    fn new_f64(float: f64) -> Self::Element {
+        BorrowedValue::Float(float).into()
     }
 
     fn new_clob(bytes: &'static [u8]) -> Self::Element {
@@ -161,6 +195,10 @@ impl<'val> Builder for BorrowedElement<'val> {
 
     fn new_sexp(seq: Self::Sequence) -> Self::Element {
         BorrowedValue::SExpression(seq).into()
+    }
+
+    fn new_struct(struct_: Self::Struct) -> Self::Element {
+        BorrowedValue::Struct(struct_).into()
     }
 }
 
@@ -664,20 +702,20 @@ mod borrowed_value_tests {
     #[rstest(
         elem, ion_type, valid_ops_iter, op_assert,
         case::null(
-            IonType::Null.into(),
+            BorrowedElement::new_null(IonType::Null),
             IonType::Null,
             IsNull,
             &|e: &BorrowedElement| assert_eq!(true, e.is_null())
         ),
         // TODO more null testing (probably its own fixture)
         case::bool(
-            true.into(),
+            BorrowedElement::new_bool(true),
             IonType::Boolean,
             AsBool,
             &|e: &BorrowedElement| assert_eq!(Some(true), e.as_bool())
         ),
         case::i64(
-            100.into(),
+            BorrowedElement::new_i64(100),
             IonType::Integer,
             AsAnyInt,
             &|e: &BorrowedElement| {
@@ -687,7 +725,7 @@ mod borrowed_value_tests {
             }
         ),
         case::big_int(
-            BigInt::from(100).into(),
+            BorrowedElement::new_big_int(BigInt::from(100)),
             IonType::Integer,
             AsAnyInt,
             &|e: &BorrowedElement| {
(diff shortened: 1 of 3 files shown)


src/value/borrowed.rs
```diff
@@ -141,12 +141,46 @@ impl<'val> SymbolToken for BorrowedSymbolToken<'val> {
...
         BorrowedValue::Null(e_type).into()
     }
 
+    fn new_bool(bool: bool) -> Self::Element {
+        BorrowedValue::Boolean(bool).into()
+    }
+
+    fn new_string(str: &'static str) -> Self::Element {
+        BorrowedValue::String(str).into()
+    }
```
almann, line 158: I am not sure about this--are you able to tie the life to `'val`? If not (e.g. this requires GAT), then I would eliminate it and only generically provide the `String` based constructor (similarly for `clob` and `blob`).

src/value/mod.rs
```diff
@@ -487,11 +487,37 @@ pub trait Struct {
...
+    fn new_decimal(decimal: Decimal) -> Self::Element;
+
+    /// Build a `timestamp` using Builder
+    fn new_timestamp(timestamp: Timestamp) -> Self::Element;
+
+    /// Build a `f64` using Builder
+    fn new_f64(float: f64) -> Self::Element;
+
     /// Build a `clob` using Builder
     fn new_clob(bytes: &'static [u8]) -> Self::Element;
```
almann, line 523: I somehow missed this in an earlier commit--this should not be `'static` for sure, if we cannot have an appropriate lifetime due to GAT, then let's just use the owned type (e.g. `Vec<u8>`)

src/value/mod.rs
```diff
@@ -503,4 +529,7 @@ pub trait Builder {
 
     /// Build a `sexp` from Sequence using Builder
     fn new_sexp(seq: Self::Sequence) -> Self::Element;
+
+    /// Build a `struct` from Struct using Builder
+    fn new_struct(struct_: Self::Struct) -> Self::Element;
```
almann, line 534: I would change these to just spell it out (see also the implementations).

```suggestion
    fn new_struct(structure: Self::Struct) -> Self::Element;
```

almann (changes requested):
I think the builder APIs that take `&'static [u8]` and `&'static str` should be changed to their owned counterparts if you cannot tag the lifetimes correctly (ugh, GAT).

I do think this change implies changing the`Borrowed...` types to use [`Cow<'val, X>`][1].  #135 alludes to this, but it is fine if we just do it now.

[1]: https://doc.rust-lang.org/std/borrow/enum.Cow.html

desaikd:
Changes with https://github.com/amzn/ion-rust/pull/206/commits/83ac1b2f6b0c9a1ade51ec81392a92a26290c437:

- Added `From<X>` for scalar types in `Element`.
- Added generic scalar tests for `Element`.

TODO:

- Add generic tests for container types and `SymbolToken`.
- Add `From<X>` for container types and `SymbolToken`

src/value/mod.rs
```diff
@@ -258,10 +259,13 @@ impl Eq for AnyInt {}
...
-    type SymbolToken: SymbolToken + ?Sized;
-    type Sequence: Sequence<Element = Self> + ?Sized;
-    type Struct: Struct<FieldName = Self::SymbolToken, Element = Self> + ?Sized;
+pub trait Element
+where
+    Self: From<i64> + From<bool> + From<Decimal> + From<Timestamp> + From<f64> + From<BigInt>,
+{
+    type SymbolToken: SymbolToken + ?Sized + Debug + PartialEq;
+    type Sequence: Sequence<Element = Self> + ?Sized + Debug + PartialEq;
+    type Struct: Struct<FieldName = Self::SymbolToken, Element = Self> + ?Sized + Debug + PartialEq;
```
almann, line 268: Since you added `Debug` to all of the downstream traits, should `Element` have `Self: Debug`?

almann (commented):
Looks good overall--I like how the generic tests worked out, should we eliminate the non-generic tests now (I would run coverage before and after removal to be sure that we still are getting everything).

src/value/owned.rs
```diff
@@ -666,20 +719,20 @@ mod value_tests {
     #[rstest(
         elem, ion_type, valid_ops_iter, op_assert,
         case::null(
-            IonType::Null.into(),
+            OwnedElement::new_null(IonType::Null),
             IonType::Null,
             IsNull,
             &|e: &OwnedElement| assert_eq!(true, e.is_null())
         ),
         // TODO more null testing (probably its own fixture)
```
almann, line 727: Do these tests matter now that you have generic ones?

src/value/borrowed.rs
```diff
@@ -664,20 +718,20 @@ mod borrowed_value_tests {
...
-            IonType::Null.into(),
+            BorrowedElement::new_null(IonType::Null),
             IonType::Null,
             IsNull,
             &|e: &BorrowedElement| assert_eq!(true, e.is_null())
         ),
         // TODO more null testing (probably its own fixture)
         case::bool(
-            true.into(),
+            BorrowedElement::new_bool(true),
```
almann, line 728: Do these tests matter now that you have generic ones?

almann (commented):
I think this is pretty close.  Do we really need the owned/borrowed tests now that you've made everything generic?

src/value/owned.rs
desaikd: I meant to do this in separate PR as it would require to add more tests in generic test suite.

src/value/mod.rs
```diff
@@ -258,10 +259,13 @@ impl Eq for AnyInt {}
 
 /// Represents a either a borrowed or owned Ion datum.  There are/will be specific APIs for
 /// _borrowed_ and _owned_ implementations, but this trait unifies operations on either.
-pub trait Element {
-    type SymbolToken: SymbolToken + ?Sized;
-    type Sequence: Sequence<Element = Self> + ?Sized;
-    type Struct: Struct<FieldName = Self::SymbolToken, Element = Self> + ?Sized;
+pub trait Element
+where
+    Self: From<i64> + From<bool> + From<Decimal> + From<Timestamp> + From<f64> + From<BigInt>,
```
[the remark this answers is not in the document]
desaikd: It does seem like a reasonable bound to me as we are not taking `From<AnyInt>` as the bound here.  Did you mean to have something else in mind for `BigInt`?

src/value/borrowed.rs
```diff
@@ -70,20 +69,17 @@ impl<'val> BorrowedSymbolToken<'val> {
...
-    /// Decorates the [`BorrowedSymbolToken`] with a local ID.
-    pub fn with_local_sid(mut self, local_sid: SymbolId) -> Self {
+    fn with_local_sid_borrrowed(mut self, local_sid: SymbolId) -> Self {
         self.local_sid = Some(local_sid);
         self
     }
 
-    /// Decorates the [`BorrowedSymbolToken`] with an [`BorrowedImportSource`].
-    pub fn with_source(mut self, table: &'val str, sid: SymbolId) -> Self {
+    fn with_source_borrowed(mut self, table: &'static str, sid: SymbolId) -> Self {
```
almann, line 82: Why do we have `..._borrowed` and the forms below that don't have the suffix?

src/value/borrowed.rs
```diff
@@ -137,6 +133,18 @@ impl<'val> SymbolToken for BorrowedSymbolToken<'val> {
...
+        self.with_text_borrowed(text)
+    }
+
+    fn with_local_sid(self, local_sid: SymbolId) -> Self {
+        self.with_local_sid_borrrowed(local_sid)
+    }
+
+    fn with_source(self, table: &'static str, sid: SymbolId) -> Self {
+        self.with_source_borrowed(table, sid)
+    }
```
almann, line 147: Let's make sure we have an issue to deal with this `'static` lifetime problem the APIs will be very marginal with this...

src/value/owned.rs
```diff
@@ -71,20 +71,17 @@ impl OwnedSymbolToken {
...
-    pub fn with_local_sid(mut self, local_sid: SymbolId) -> Self {
+    fn with_local_sid_owned(mut self, local_sid: SymbolId) -> Self {
         self.local_sid = Some(local_sid);
         self
     }
 
-    /// Decorates the [`OwnedSymbolToken`] with an [`OwnedImportSource`].
-    pub fn with_source<T: Into<Rc<str>>>(mut self, table: T, sid: SymbolId) -> Self {
+    fn with_source_owned(mut self, table: &'static str, sid: SymbolId) -> Self {
         self.source = Some(OwnedImportSource::new(table, sid));
```
almann, line 85: Same question on this API, why do we need `..._owned` variants when they look the same as the non-suffix ones below?

almann (commented):
I have some questions around the API factoring below.

src/value/borrowed.rs
desaikd: changed this to construct a new `SymbolToken ` for all the `with _...` construction methods.

almann (approved):
As long as we have an issue around our `'static` lifetimes and an issue (or its been removed) for the redundant tests, :rocket:.

PR #210, opened 2021-04-23
Adds doc test example for `Loader` API.

### LINKED ISSUE #208: Add Doc Test Examples for `Loader` API
After publishing `ion-rs` 0.4.0, I noticed that we should have some examples of how to use the APIs in the main docs.  We should also clean up the API documentation for it.

almann:
* Fixes lifetimes on `Loader`.
* Cleans up `loader` module documentation a bit.
* Fixes references in `magnitude` and `timestamp` modules.

Fixes #208.

### DIFF
--- a/src/types/magnitude.rs
+++ b/src/types/magnitude.rs
@@ -3,7 +3,9 @@ use std::cmp::Ordering;
 use bigdecimal::ToPrimitive;
 use num_bigint::{BigUint, ToBigUint};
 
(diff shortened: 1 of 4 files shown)


desaikd (approved):
LGTM! Although is there a specific reason to have an extra lifetime added in the `iterate_over` method?

almann:
> LGTM! Although is there a specific reason to have an extra lifetime added in the `iterate_over` method?

The reason is the borrow of the loader has nothing to do with the slice your iterating over and the example code wouldn't work without a needless bind for loader.

PR #215, opened 2021-04-25
Scaffold a subcrate for ion-hash

marcbowes:
Not sure we want this to live here forever, but this should let us
iterate in lock step with the in-development ion-rs crate.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/ion-hash/src/lib.rs
+++ b/ion-hash/src/lib.rs
@@ -1,0 +1,6 @@
+// Copyright Amazon.com, Inc. or its affiliates.
+
+//! Implement Ion Hash as per the [spec](https://amzn.github.io/ion-hash/docs/spec.html)
+//!
+//! ## Examples
+//! Coming soon.
--- a/ion-hash/src/lib.rs
+++ b/ion-hash/src/lib.rs
@@ -1,6 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates.
 
-//! Implement Ion Hash as per the [spec](https://amzn.github.io/ion-hash/docs/spec.html)
+//! Implements the [Ion Hash specification](https://amzn.github.io/ion-hash/docs/spec.html)
 //!
 //! ## Examples
 //! ```rust
@@ -16,35 +16,36 @@
 //! # }
 //! ```
 
-use ion_rs::{
-    value::{owned::OwnedElement, Element},
-    IonType,
-};
+use ion_rs::{value::Element, IonType};
 
 use sha2::{digest::Output, Digest, Sha256};
 
-pub fn sha256(elem: &OwnedElement) -> Vec<u8> {
+/// Utility to hash an [`Element`] using SHA-256 as the hash function.
+pub fn sha256<E: Element>(elem: &E) -> Vec<u8> {
     let mut hasher = IonHasher::new(Sha256::new());
-    let result = hasher.visit_element(elem);
+    let result = hasher.hash_element(elem);
     Vec::from(result.as_slice())
 }
 
 /// Bytes markers as per the spec.
-mod markers {
+struct Markers;
+impl Markers {
     /// single byte begin marker
-    pub(crate) const B: u8 = 0x0B;
+    const B: u8 = 0x0B;
     /// single byte end marker
-    pub(crate) const E: u8 = 0x0E;
+    const E: u8 = 0x0E;
     /// single byte escape
-    pub(crate) const ESC: u8 = 0x0C;
+    const ESC: u8 = 0x0C;
     /// type qualifier octet consisting of a four-bit type code T followed by a four-bit qualifier Q
     /// (this varies per type)
-    pub(crate) const TQ_STRING: u8 = 0x80;
+    const TQ_STRING: u8 = 0x80;
 }
 
 // TODO: In the fullness of time, we will have a IonHashReader and a
 // IonHashWriter. This will allow reading/writing a value *while* computing a
-// hash. For now, we provide a standalone hasher using the OwnedElement API.
+// hash. For now, we provide a standalone hasher using the Element API.
+/// Provides Ion hash over arbitrary [`Element`] instances with a given
+/// [`Digest`] algorithm.
 pub struct IonHasher<D>
 where
     D: Digest,
@@ -56,11 +57,13 @@ impl<D> IonHasher<D>
 where
     D: Digest,
 {
-    fn new(hasher: D) -> IonHasher<D> {
-        IonHasher { hasher }
+    fn new(hasher: D) -> Self {
+        Self { hasher }
     }
 
-    fn visit_element(&mut self, elem: &OwnedElement) -> Output<D> {
+    /// Computes the Ion hash over an [`Element`] recursively using this
+    /// hasher's [`Digest`]
+    fn hash_element<E: Element + ?Sized>(&mut self, elem: &E) -> Output<D> {
         let serialized_bytes = match elem.ion_type() {
             IonType::Null => todo!(),
             IonType::Boolean => todo!(),
@@ -77,15 +80,17 @@ where
             IonType::Struct => todo!(),
         };
 
+        // TODO: Annotations
+
         self.hasher.update(serialized_bytes);
         self.hasher.finalize_reset()
     }
 
     fn visit_string(&mut self, value: &str) -> Vec<u8> {
         let representation = value.as_bytes();
-        let mut serialized_bytes = vec![markers::B, markers::TQ_STRING];
+        let mut serialized_bytes = vec![Markers::B, Markers::TQ_STRING];
         serialized_bytes.extend(ion_hash_escape(representation));
-        serialized_bytes.push(markers::E);
+        serialized_bytes.push(Markers::E);
         serialized_bytes
     }
 }
@@ -94,7 +99,7 @@ where
 fn ion_hash_escape(representation: &[u8]) -> Vec<u8> {
     let mut out = Vec::with_capacity(representation.len());
     for byte in representation {
-        if let markers::B | markers::E | markers::ESC = *byte {
+        if let Markers::B | Markers::E | Markers::ESC = *byte {
             out.push(0x0C);
         }
         out.push(*byte);
--- a/ion-hash/src/lib.rs
+++ b/ion-hash/src/lib.rs
@@ -3,4 +3,118 @@
 //! Implement Ion Hash as per the [spec](https://amzn.github.io/ion-hash/docs/spec.html)
 //!
 //! ## Examples
-//! Coming soon.
+//! ```rust
+//! use ion_rs::value::loader::{self, Loader};
+//! use ion_rs::result::IonResult;
+//!
+//! # fn main() -> IonResult<()> {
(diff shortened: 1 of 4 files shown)


marcbowes:
Do you want to call this `ion-hash-rust` or `ion-hash-rs` or just `ion-hash`? I chose the latter.

ion-hash/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+impl<D> IonHasher<D>
+where
+    D: Digest,
+{
+    fn new(hasher: D) -> IonHasher<D> {
+        IonHasher { hasher }
+    }
+
+    fn visit_element(&mut self, elem: &OwnedElement) -> Output<D> {
+        let serialized_bytes = match elem.ion_type() {
```
almann, line 64: I think this should be made generic and I think this avoids the dependency on `OwnedElement` (which is good).

```rust
fn hash_element<E: Element + ?Sized>(&mut self, elem: &E) -> Output<D>{ ... }
```

Any thoughts about making `D` a parameter and allowing `IonHasher` to be used in a stateless way?

E.g.

```rust
/// Constructs a [`Digest`] and performs a self-contained hash for a given [`Element`].
fn hash_element<D: Digest, E: Element + ?Sized>(&self, elem: &E) -> Output<D> {
    let mut digest = D::new();
    self.update_hash_element(&mut digest, elem);
    digest.finalize()
}

/// Hashes an element and updates an existing [`Digest`].
fn update_hash_element<D: Digest, E: Element + ?Sized>(&self, digest: &mut D, elem: &E) { ... }
```

This would take the type parameter away from the `IonHasher` and push the parameterization on the invocation, but it would also decouple the usage of the digest (especially around finalization, if the user wants to do other things before `finalize`).  That said, `finalize_reset` might be more optimal for subsequent hashing, so that might be an argument to keep it internalized (and that doesn't mean we couldn't add the update method later).

ion-hash/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+            IonType::String => self.visit_string(elem.as_str().unwrap()),
+            IonType::Clob => todo!(),
+            IonType::Blob => todo!(),
+            IonType::List => todo!(),
+            IonType::SExpression => todo!(),
+            IonType::Struct => todo!(),
+        };
+
+        self.hasher.update(serialized_bytes);
+        self.hasher.finalize_reset()
```
almann, line 81: I think there needs to be a TODO for annotations.

ion-hash/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+    pub(crate) const ESC: u8 = 0x0C;
+    /// type qualifier octet consisting of a four-bit type code T followed by a four-bit qualifier Q
+    /// (this varies per type)
+    pub(crate) const TQ_STRING: u8 = 0x80;
+}
+
+// TODO: In the fullness of time, we will have a IonHashReader and a
+// IonHashWriter. This will allow reading/writing a value *while* computing a
+// hash. For now, we provide a standalone hasher using the OwnedElement API.
+pub struct IonHasher<D>
```
almann, line 48: Adds doc comments for `IonHasher`.

```suggestion
// TODO: In the fullness of time, we will have a IonHashReader and a
// IonHashWriter. This will allow reading/writing a value *while* computing a
// hash. For now, we provide a standalone hasher using the Element API.

/// Provides Ion hash over arbitrary [`Element`] instances with a given [`Digest`]
/// algorithm.
pub struct IonHasher<D>
```

ion-hash/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+    hasher: D,
+}
+
+impl<D> IonHasher<D>
+where
+    D: Digest,
+{
+    fn new(hasher: D) -> IonHasher<D> {
+        IonHasher { hasher }
+    }
```
almann, line 61: Minor style change:

```suggestion
    fn new(hasher: D) -> Self {
        Self { hasher }
    }
```

ion-hash/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+
+    fn visit_element(&mut self, elem: &OwnedElement) -> Output<D> {
```
almann, line 63: Minor doc comment suggestion:

```suggestion
    /// Computes the Ion hash over an [`Element`] recursively using this hasher's [`Digest`]
    fn visit_element(&mut self, elem: &OwnedElement) -> Output<D> {
```

ion-hash/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// single byte begin marker
+    pub(crate) const B: u8 = 0x0B;
+    /// single byte end marker
+    pub(crate) const E: u8 = 0x0E;
+    /// single byte escape
+    pub(crate) const ESC: u8 = 0x0C;
+    /// type qualifier octet consisting of a four-bit type code T followed by a four-bit qualifier Q
+    /// (this varies per type)
+    pub(crate) const TQ_STRING: u8 = 0x80;
+}
```
almann, line 43: Thoughts on associated constants for this--it would eliminate the `pub(crate)` requirement (at least for how it is being used now):

```rust
/// Byte markers as per spec.
struct Marker;

impl Marker {
    /// single byte begin marker
    const B: u8 = 0x0B;
    /// single byte end marker
    const E: u8 = 0x0E;
    /// single byte escape
    const ESC: u8 = 0x0C;
    /// type qualifier octet consisting of a four-bit type code T followed by a four-bit qualifier Q
    /// (this varies per type)
    const TQ_STRING: u8 = 0x80;
}
```

ion-hash/src/lib.rs
```diff
@@ -0,0 +1,120 @@
+// Copyright Amazon.com, Inc. or its affiliates.
+
+//! Implement Ion Hash as per the [spec](https://amzn.github.io/ion-hash/docs/spec.html)
```
almann, line 3: Minor doc comment edit:

```suggestion
//! Implements the [Ion Hash specification](https://amzn.github.io/ion-hash/docs/spec.html)
```

ion-hash/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+//! ```
+
+use ion_rs::{
+    value::{owned::OwnedElement, Element},
+    IonType,
+};
+
+use sha2::{digest::Output, Digest, Sha256};
+
+pub fn sha256(elem: &OwnedElement) -> Vec<u8> {
```
almann, line 26: Minor doc comment and generics edit (assuming you agree with my suggestions on `IonHasher`):

```suggestion
/// Utility to hash an [`Element`] using SHA-256 as the hash function.
pub fn sha256<E: Element>(elem: &E) -> Vec<u8> {
```

almann (changes requested):
Overall, I am good with adding this as a workspace sub-crate, especially since APIs are unstable.  This will make it easier to develop this code base.  When this gets too big we can always move it to its own repository.

That said, I don't want to make this decision unilaterally, so I would like @zslayton  to review.

As for the code itself, I have some minor edit proposals below, I do think this should be made generic with respect to `Element` and eliminate the `OwnedElement` direct dependency.

almann:
> Do you want to call this `ion-hash-rust` or `ion-hash-rs` or just `ion-hash`? I chose the latter.

I prefer `ion-hash`, the only reason why we have `ion-rs` is because of crates.io name clash.

ion-hash/src/lib.rs
marcbowes: > I think this should be made generic and I think this avoids the dependency on OwnedElement (which is good).

Done!

> Any thoughts about making D a parameter and allowing IonHasher to be used in a stateless way?

I'm expecting some churn here in our API; as we use it we'll learn. For now I'm happy to leave as is or take your suggestion - pretty ambivalent.

ion-hash/src/lib.rs
marcbowes: Added one, but I'm expecting the test suite to guide me here.

ion-hash/src/lib.rs
marcbowes: Good idea.

ion-hash/src/lib.rs
almann: Let's not change it--I was more curious then anything.

almann (commented):
Minor point below otherwise this looks pretty good to me.  I am not explicitly **approving** the PR until we get some broader eyes around the sub-crate question.

ion-hash/src/lib.rs
```diff
@@ -0,0 +1,125 @@
...
+    fn hash_element<E: Element + ?Sized>(&mut self, elem: &E) -> Output<D> {
+        let serialized_bytes = match elem.ion_type() {
+            IonType::Null => todo!(),
+            IonType::Boolean => todo!(),
+            IonType::Integer => todo!(),
+            IonType::Float => todo!(),
+            IonType::Decimal => todo!(),
+            IonType::Timestamp => todo!(),
+            IonType::Symbol => todo!(),
+            IonType::String => self.visit_string(elem.as_str().unwrap()),
```
almann, line 75: I missed this earlier, but shouldn't this API return `Result<T, E>` or some such to avoid the unwrap?  At a bare minimum, a TODO screaming at us to fix this would be good because it is a bad idea for a library to panic this way.

marcbowes: Fixed in 00d23d4

ion-hash/src/lib.rs
```diff
@@ -16,15 +16,30 @@
...
 use ion_rs::{value::Element, IonType};
 
 use sha2::{digest::Output, Digest, Sha256};
 
+// A `try`-like macro to workaround the Element API ergonomics. This API
+// requires checking the type and then calling the appropriate getter function
+// (which returns a None if you got it wrong). This macro turns the `None` into
+// an `IonError`.
+// TODO: Remove this once the Element API is reworked.
+macro_rules! t {
```
almann, line 29: Nit, I really don't like this name, how about `try_to` or some such (though @zslayton doesn't like this name either).  That said, we can refactor this when I merge in my `Dumper` API which steals this definition anyhow.

almann (approved):
:ship:

ion-hash/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+            IonType::Struct => todo!(),
+        };
+
+        // TODO: Annotations
+
+        self.hasher.update(serialized_bytes);
+        Ok(self.hasher.finalize_reset())
+    }
+
+    fn visit_string(&mut self, value: &str) -> Vec<u8> {
```
zslayton, line 104: It seems like the `Self` could have a reusable `Vec<u8>` that's updated by each visitor and reset between calls to `hash_element`.

zslayton (approved):

ion-hash/src/lib.rs
marcbowes: Agreed! I plan on reworking this as we get into other types.

ion-hash/src/lib.rs
marcbowes: Ack

PR #217, opened 2021-04-25
Adds `ion-tests` test file integration for `Loader`.

almann:
* Adds `test-generator` create dev dependency to get test per file.
* Adds a test suite with `good`/`bad` files with skip lists.
  - `equivs` and `non-equivs` are also supported with
    `embedded_document` parsing.
* Adds `Loader::load_all` convenience method.
* Adds doc test example in `loader` module docs and fixes the iterator
  example.

Additional notes:
* Added #216 as a result of non-equivs testing and it definitely exposes
  a weakness in our unit tests and bug in our `PartialEq` for `...Struct`.
* Added #218 to track a particular set ion-c integration issues around integer
  sizes and binary reader.
* Added #219 to track adding support for unknown text symbol tokens in `Loader`.
* Added #220 to track adding `IonEq`.
* Added amzn/ion-c#235 as these tests run into problems with negative
  binary integer syntax (but not hex or decimal), indicating a potential
  issue there.

Resolves #214.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #221, opened 2021-04-25
Adds `Loader::load_one` convenience method.

### LINKED ISSUE #211: Add a `load_one` method to `Loader`
When implementing #208, I noticed that a common case of loading a single value should have a convenience method in the `Loader` trait.  Adding a trait method with an implementation using `iterate_over` would be handy.  If there is a single `Element`, return it, otherwise if there are more than one element return an error.

This probably implies adding a new error type to `IonError` as the existing set don't quite cover this case well.

almann:
* Adds doc test examples for this API.
* Adds support in the loader integration tests for `load_one`.

Resolves #211.

Based on #217.

Here is an example of the rendered docs:
image

### DIFF
--- a/src/value/loader.rs
+++ b/src/value/loader.rs
@@ -49,7 +49,7 @@ pub trait Loader {
     }
 
     /// Parses Ion over a given slice into a single [`Element`](super::Element) instance.
-    /// Returning [`IonError`](crate::result::IonError) if any error occurs during the parse
+    /// Returns [`IonError`](crate::result::IonError) if any error occurs during the parse
     /// or there is more than one top-level [`Element`](super::Element) in the data.
     #[inline]
     fn load_one(&self, data: &[u8]) -> IonResult<OwnedElement> {
--- a/src/value/loader.rs
+++ b/src/value/loader.rs
@@ -48,7 +48,7 @@ pub trait Loader {
         self.iterate_over(data)?.collect()
     }
 
-    /// Parses given Ion over a given slice into a single [`Element`](super::Element) instance.
+    /// Parses Ion over a given slice into a single [`Element`](super::Element) instance.
     /// Returning [`IonError`](crate::result::IonError) if any error occurs during the parse
     /// or there is more than one top-level [`Element`](super::Element) in the data.
     #[inline]
--- a/src/value/loader.rs
+++ b/src/value/loader.rs
@@ -3,7 +3,7 @@
 //! Provides utility to load Ion data into [`Element`](super::Element) from different sources such
 //! as slices or files.
 
-use crate::result::IonResult;
+use crate::result::{decoding_error, IonResult};
 use crate::value::owned::{OwnedElement, OwnedSequence, OwnedStruct, OwnedSymbolToken, OwnedValue};
 use crate::value::AnyInt;
 use crate::IonType;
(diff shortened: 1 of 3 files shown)


tests/loader_test_vectors.rs
```diff
@@ -61,20 +67,64 @@ fn concat<'a>(left: &[&'a str], right: &[&'a str]) -> Vec<&'a str> {
...
+                        }
+                    },
+                }
+            }
+        }
+        Err(_) => assert!(
+            single_result.is_err(),
+            "Expected error from load_one: {:?}",
+            single_result
+        ),
```
almann, line 120: Ideally, we do stronger assertions here--but it is a bit complicated to make sure the errors line up consistently.

tests/loader_test_vectors.rs
```diff
@@ -61,20 +67,64 @@ fn concat<'a>(left: &[&'a str], right: &[&'a str]) -> Vec<&'a str> {
...
+                    Err(e) => panic!("Expected element {:?}, got {:?}", elems, e),
+                }
+            } else {
+                match single_result {
+                    Ok(elem) => panic!(
+                        "Did not expect element for duplicates: {:?}, {:?}",
+                        elems, elem
+                    ),
+                    Err(e) => match e {
+                        IonError::DecodingError { description: _ } => (),
```
almann, line 108: This case we know the error type.

tests/loader_test_vectors.rs
```diff
@@ -61,20 +67,64 @@ fn concat<'a>(left: &[&'a str], right: &[&'a str]) -> Vec<&'a str> {
...
+
 fn load_file<L: Loader>(loader: &L, file_name: &str) -> IonResult<Vec<OwnedElement>> {
     // TODO have a better API that doesn't require buffering into memory everything...
     let data = read(file_name)?;
-    loader.load_all(&data)
+    let result = loader.load_all(&data);
+
+    // do some simple single value loading tests
+    let single_result = loader.load_one(&data);
+    match &result {
```
almann, line 89: I think I could've written it with a match on a pair of `(&result, &single_result)` or some such, but I don't think that would've been clearer in my opinion.

src/value/loader.rs
```diff
@@ -43,9 +43,32 @@ pub trait Loader {
 
     /// Parses given Ion over a given slice into an [`Vec`] returning an
     /// [`IonError`](crate::result::IonError) if any error occurs during the parse.
+    #[inline]
     fn load_all(&self, data: &[u8]) -> IonResult<Vec<OwnedElement>> {
         self.iterate_over(data)?.collect()
     }
+
+    /// Parses given Ion over a given slice into a single [`Element`](super::Element) instance.
```
zslayton, line 51:
```suggestion
    /// Parses Ion over a given slice into a single [`Element`](super::Element) instance.
```

src/value/loader.rs
```diff
@@ -43,9 +43,32 @@ pub trait Loader {
...
+    /// Returning [`IonError`](crate::result::IonError) if any error occurs during the parse
```
zslayton, line 52:
```suggestion
    /// Returns [`IonError`](crate::result::IonError) if any error occurs during the parse
```

src/value/loader.rs
```diff
@@ -43,9 +43,32 @@ pub trait Loader {
...
+        let mut iter = self.iterate_over(data)?;
+        match iter.next() {
+            Some(Ok(elem)) => {
+                // make sure there is nothing else
+                match iter.next() {
+                    None => Ok(elem),
+                    Some(Ok(_)) => {
+                        decoding_error("Expected a single element, but there was more than one")
+                    }
+                    Some(other) => other,
```
zslayton, line 65: This branch produces a potentially confusing `IonError`--it won't mention that a value was successfully parsed before invalid bytes were encountered. We should look into adding a source error field to the `IonError` enum variants so we can layer them when necessary.

zslayton (approved):

src/value/loader.rs
almann: Added #222

almann:
Closing this PR in favor of #223 which targets `main` now that #217 is merged.

PR #223, opened 2021-04-26
Adds `Loader::load_one` convenience method.

### LINKED ISSUE #211: Add a `load_one` method to `Loader`

almann:
* Adds doc test examples for this API.
* Adds support in the loader integration tests for `load_one`.

Resolves #211.

### DIFF
--- a/src/value/loader.rs
+++ b/src/value/loader.rs
@@ -3,7 +3,7 @@
 //! Provides utility to load Ion data into [`Element`](super::Element) from different sources such
 //! as slices or files.
 
-use crate::result::IonResult;
+use crate::result::{decoding_error, IonResult};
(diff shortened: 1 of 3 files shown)


almann:
Note that this is a rebase for #221 that targets `main`.  I have addressed (to my knowledge) all of the feedback in that PR, so this should be a clean merge.

zslayton (approved):

PR #227, opened 2021-04-28
Ion hash test

marcbowes:
This commit starts to wire up the ion-hash-test submodule with a runner test file. Work in progress!


ion-hash/src/lib.rs
```diff
@@ -63,31 +83,31 @@ impl Markers {
...
 {
-    fn new(hasher: D) -> Self {
+    pub fn new(hasher: D) -> Self {
         Self { hasher }
     }
 
     /// Computes the Ion hash over an [`Element`] recursively using this
     /// hasher's [`Digest`]
-    fn hash_element<E: Element + ?Sized>(&mut self, elem: &E) -> IonResult<Output<D>> {
+    pub fn hash_element<E: Element + ?Sized>(mut self, elem: &E) -> IonResult<Vec<u8>> {
```
marcbowes, line 101: The change here (from `sha2::Digest::Output` to `Vec<u8>`) is temporary. I just don't really care about the abstraction right now. Make it work.. make it fast.

ion-hash/tests/ion_hash_tests.rs
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    fn finalize(self) -> Vec<u8> {
+        let inner = self.inner.lock().unwrap();
+        inner.digest.clone()
+    }
+}
+
+#[test]
+fn the_first_test() -> IonResult<()> {
```
marcbowes, line 46: The intention is to delete this once the main file loads.

ion-hash/tests/ion_hash_tests.rs
```diff
@@ -0,0 +1,126 @@
...
+        .expect("`expect` should have a field called `identity`")
+        .as_sequence()
+        .expect("`identity` should be a sexp");
+
+    let digest = TestDigest::default();
+    let hasher = IonHasher::new(digest.clone());
+    let _result = hasher.hash_element(ion)?;
+
+    let (recorded_updates, recorded_digest) = digest.inspect();
+    let mut recorded_updates = recorded_updates.into_iter();
```
marcbowes, line 81: I know this is incomplete test coverage (I don't check that all updates/digest were checked against).

ion-hash/tests/ion_hash_tests.rs
```diff
@@ -0,0 +1,126 @@
...
+            inner.updates.push(*b);
+            inner.digest.push(*b);
+        }
+    }
+
+    fn finalize(self) -> Vec<u8> {
+        let inner = self.inner.lock().unwrap();
+        inner.digest.clone()
+    }
+}
```
almann, line 43: I really need some help with explaining why these types are here--my read below makes me think you are doing some kind of mocking or internal aggregation, but I am not sure.  Either way I didn't read any concurrent code below so is that just there in case you do make it multi-threaded or is it just `RefCell`?

ion-hash/tests/ion_hash_tests.rs
```diff
@@ -0,0 +1,126 @@
...
+        .get("identity")
+        .expect("`expect` should have a field called `identity`")
+        .as_sequence()
+        .expect("`identity` should be a sexp");
+
+    let digest = TestDigest::default();
+    let hasher = IonHasher::new(digest.clone());
+    let _result = hasher.hash_element(ion)?;
+
+    let (recorded_updates, recorded_digest) = digest.inspect();
```
almann, line 80: Why not use the result instead of making this `inspect` machinery?  Again, I think a bit of comment as to why would help the reader here...

ion-hash/tests/ion_hash_tests.rs
```diff
@@ -0,0 +1,126 @@
...
+            "update" => {
+                let first = recorded_updates
+                    .nth(0)
+                    .expect("no more calls to `update` were made");
+
+                assert_eq!(&bytes[..], &[first]);
+            }
+            "digest" => {
+                assert_eq!(bytes, recorded_digest);
+            }
```
almann, line 108: Are you asserting a particular call pattern or some such?

almann (changes requested):
I really don't quite grok the motivation behind the test changes--I think some comments will really elucidate that for me at a minimum.

ion-hash/src/lib.rs
```diff
@@ -51,8 +69,10 @@ impl Markers {
     const E: u8 = 0x0E;
     /// single byte escape
     const ESC: u8 = 0x0C;
-    /// type qualifier octet consisting of a four-bit type code T followed by a four-bit qualifier Q
-    /// (this varies per type)
+
+    // type qualifier octet consisting of a four-bit type code T followed by a four-bit qualifier Q
+    // (this varies per type)
+    const TQ_NULL: u8 = 0x0F;
```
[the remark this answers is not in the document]
marcbowes: Because the comment applies to TQs in general as opposed to `TQ_STRING`.

ion-hash/src/lib.rs
```diff
@@ -19,7 +19,25 @@
 use ion_rs::result::{illegal_operation, IonResult};
 use ion_rs::{value::Element, IonType};
 
-use sha2::{digest::Output, Digest, Sha256};
+use sha2::{Digest, Sha256};
+
+pub trait IonHashDigest {
+    fn update(&mut self, bytes: impl AsRef<[u8]>);
+    fn finalize(self) -> Vec<u8>;
+}
```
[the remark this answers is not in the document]
marcbowes: Ack, will do!

ion-hash/tests/ion_hash_tests.rs
```diff
@@ -0,0 +1,126 @@
...
+use ion_rs::value::*;
+use std::{
+    fs::read,
+    sync::{Arc, Mutex},
+};
+
+#[derive(Default, Clone)]
+struct TestDigest {
+    inner: Arc<Mutex<TestDigestInner>>,
+}
```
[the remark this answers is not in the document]
marcbowes: No, not really. We could get away with `Arc<RefCell>` since concurrency is not required. Or, we could find another way to do the assertions..

ion-hash/tests/ion_hash_tests.rs
marcbowes: The latter :)

ion-hash/tests/ion_hash_tests.rs
marcbowes: I probably just don't understand ion-hash-test properly. I read it and how the Python one uses it, and kind of just went along with precedent.

ion-hash/tests/ion_hash_tests.rs
marcbowes: Yeah.

ion-hash/src/lib.rs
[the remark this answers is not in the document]
marcbowes: Not really (although that's how it works as of this PR - hence taking self by value).

I'm more just trying to assemble the bits at this point. My thought is that we'll figure out the API progressively.

ion-hash/src/lib.rs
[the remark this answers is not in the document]
marcbowes: New revision has a TODO that says we should use the `digest` crate.

ion-hash/tests/ion_hash_tests.rs
```diff
@@ -1,71 +1,96 @@
...
-    sync::{Arc, Mutex},
-};
+use std::fs::read;
 
 #[derive(Default, Clone)]
 struct TestDigest {
-    inner: Arc<Mutex<TestDigestInner>>,
+    updates: GenericArray<u8, U256>,
+    position: usize,
 }
```
almann: Is this the identity digest?  If so, documenting this would be nice.

Also, couldn't you just use a `Vec` for this?

ion-hash/tests/ion_hash_tests.rs
```diff
@@ -1,71 +1,96 @@
...
-}
 
-impl IonHashDigest for TestDigest {
     fn update(&mut self, bytes: impl AsRef<[u8]>) {
-        let mut inner = self.inner.lock().unwrap();
         for b in bytes.as_ref() {
-            inner.updates.push(*b);
-            inner.digest.push(*b);
+            self.updates[self.position] = *b;
+            self.position += 1;
```
almann, line 30: This really just looks like `Vec::push` to me...

ion-hash/tests/ion_hash_tests.rs
```diff
@@ -1,71 +1,96 @@
...
+fn test_file(file_name: &str) -> IonResult<()> {
+    let data = read(file_name)?;
+    let elems = loader().load_all(&data)?;
     test_all(elems)
 }
 
 fn test_all<E: Element>(elems: Vec<E>) -> IonResult<()> {
-    for case in elems {
+    for (i, case) in elems.iter().enumerate() {
         let case = case.as_struct().expect("test cases are structs");
```
almann: Lack of table driven testing for Rust's testing framework makes me a sad :panda_face:.

ion-hash/tests/ion_hash_tests.rs
```diff
@@ -75,10 +100,16 @@ fn test_case<E: Element>(ion: &E, strukt: &E) -> IonResult<()> {
...
-    let mut recorded_updates = recorded_updates.into_iter();
+    let annotations: Vec<_> = ion
+        .annotations()
+        .map(|it| it.text().unwrap().to_string())
+        .collect();
+    let test_case_name = match &annotations[..] {
+        [] => format!("unannotated-test-case #{}", case_number), // FIXME: Use Dumper to print the case name
+        [single] => single.clone(),
+        _ => unimplemented!(),
+    };
```
almann, line 111: Is this literally to get the test case name?  I would comment this intent, I stared at it for a bit wondering what it was doing.

ion-hash/tests/ion_hash_tests.rs
```diff
@@ -97,30 +128,22 @@ fn test_case<E: Element>(ion: &E, strukt: &E) -> IonResult<()> {
...
             }
             "digest" => {
-                assert_eq!(bytes, recorded_digest);
+                for (i, byte) in bytes.iter().enumerate() {
+                    assert_eq!(
+                        *byte, result[i],
+                        "case: {}; byte {} failed to match",
+                        test_case_name, i
+                    );
+                }
```
almann, line 142: Are you comparing each byte, byte for byte to make it easier to determine about where the problem is?

almann (commented):
I think this is pretty close--a few questions and we can ship it.

ion-hash/Cargo.toml
```diff
@@ -22,7 +22,5 @@ edition = "2018"
 [dependencies]
 ion-rs = { path = "../", version = "0.5" }
 ion-c-sys = { path = "../ion-c-sys", version = "0.4" }
+digest = "0.9.0"
 sha2 = { version = "0.9.3" }
```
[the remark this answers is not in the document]
marcbowes: I'm preparing to make it an optional dependency so that consumers don't need digest implementations they're not using. I had it as optional, but I had some editor issues, .. For now, I'll just have the version verbatim.

ion-hash/tests/ion_hash_tests.rs
marcbowes: Yes. Will add a comment.

The `Digest` trait uses fixed size arrays. So, we have to play by that if we want to use that trait (which I think we should). While the test digest could use a Vec internally, it would still need to turn it into a generic array as part of the final result. And, that conversion has to match exactly (the vec slice len has to be exactly the len of the associated output size). So you land up with annoying code at some point. I started with Vec (see previous commits) and moved to this because I felt it was simpler.

ion-hash/tests/ion_hash_tests.rs
marcbowes: It is. See the other comment. We either work with a generic array internally and just return it; or we work with a vec internally and then convert it over to a generic array later. I think the above is actually simpler.

ion-hash/tests/ion_hash_tests.rs
marcbowes: yup!

ion-hash/tests/ion_hash_tests.rs
marcbowes: Yes it is. I'll extract a method.

ion-hash/tests/ion_hash_tests.rs
marcbowes: Hold my beer..

So. The ion-hash-tests aren't actually doing any digesting (because that's not the point). Usually, a digest implementation has a fixed size, so the digest trait lets you pick the output size. But, we don't have a fixed size because our final digest is just the u8s we fed to the digester. Which means that test-1 and test-2 have different lens. To play with the generic array trait, I picked a N that is bigger than the largest test case. But.. that means that we have trailing 0s in our array. Or, put another way, `result` is exactly equal to `bytes` plus 0s to the end of the array.

I'll add a comment.

ion-hash/tests/ion_hash_tests.rs
marcbowes: I could use a Vec internally only *and test against it* if we're prepared to do the interior mutability trick I had in a previous version. But.. I actually quite like what I have here. Let me know what you think.

almann (approved):
:rocket: Thanks, the comments on the tests really help with understanding intent!

Issue #232, opened 2021-04-28
Remove `Owned/Borrowed` tests to use generic tests instead


desaikd:
Now with #206 we have generic tests to cover the separate `Owned/Borrowed` tests. We can add more tests in the generic test suite and remove the previous tests for `Owned/Borrowed` entirely.

PR #233, opened 2021-04-29
Add generic tests and remove borrowed/owned tests

desaikd:
*Issue #232*

*Description of changes:*
This PR works on creating more tests for generic test suite and remove the owned/borrowed tests entirely.

*Changes:*
- Added generic tests for all Ion types
- Added `with_annotations` for adding annotations to `Element`.
- remove owned/borrowed tests


almann (commented):
Looks pretty good--my real only concern is making the `with_annotations` generic.

src/value/mod.rs
```diff
@@ -878,64 +879,106 @@ mod generic_value_tests {
         }
     }
 
-    fn bool_case<E: Element>() -> Case<E> {
+    fn bool_case<E: Element>() -> Case<E>
+    where
+        E: Debug + PartialEq,
```
almann, line 884: Should we just add this super-trait on `Element` and friends?

almann (approved):
LGTM, I assume the coverage looks reasonable?

src/value/mod.rs
desaikd: Yes. I will add it.

PR #234, opened 2021-05-01
Include current reader position in errors

marcbowes:
Introduce Position, populate in IonCReader

This commit introduces a `Position` type that represents the bytes,
lines and offset into data. Essentially, this mirrors the ion-c
`ion_reader_get_position` API. The `Reader` trait can produce one of
these via a new method, `Reader::pos()`.

A new proc macro `#[position_error]` automatically calls this API if any
of the `Reader` methods fail. The resulting `Position` is stored in
`IonCError::position`.

The motivation behind this commit is quite simple: while working on
ion-hash, the ion_hash_tests.ion sample failed to load. All I got was a
test failure with a `Result` that said "null": not super helpful! No
backtraces, nada. Now, the error includes the exact place where the
Reader produced the Err variant from!

### DIFF
--- a/ion-c-sys-macros/src/lib.rs
+++ b/ion-c-sys-macros/src/lib.rs
@@ -1,0 +1,89 @@
+use proc_macro::TokenStream;
+use proc_macro2::Span;
+use quote::quote;
+use syn::{parse_quote, FnArg, Ident, PatType, ReturnType};
+
+/// This macro is intended to support the `IonCReaderHandle` type. When any
+/// function returns an `IonCError`, the error should be populated with the
+/// current reader position.
+///
+/// Without this macro, errors could be decorated like so:
+///
+/// ```no_test
+/// fn foo() -> IonCResult<()> {
+///     fn1().map_err(|err| include_current_position(self.reader, err))?
+///     // ...
+///     Ok(())
+/// }
+/// ```
+///
+/// However, functions can fail in many places, so decorating each of these with
+/// `map_err` is tedious and error prone:
+///
+/// ```no_test
+/// fn foo() -> IonCResult<()> {
+///   fn1()?.map_err(..);
+///   fn2()?.map_err(..);
+/// }
+/// ```
+///
+/// This macro rewrites the function such that any error gets the position
+/// added. It does this by moving the existing implementation into a closure and
+/// then calling `map_err` on the result.
+///
+/// ***Note***: to make the variety of possible functions work (which include mutable
+/// borrows, lifetime elision), the closure uses the `move` keyword. There are
+/// some gymnastics (adapted from ***[dtolnay/no-panic][no-panic]***) to make that work out.
+///
+/// [no-panic]: https://github.com/dtolnay/no-panic
+#[proc_macro_attribute]
+pub fn position_error(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let mut function = syn::parse_macro_input!(item as syn::ItemFn);
+
+    // this is heavily adapted for dtolnay/panic - as this code evolves to deal
+    // with rustc/ion-c-sys changes refer to to the above for any changes
+    // here...
+    let mut move_self = None;
+    let mut arg_pat = Vec::new();
+    let mut arg_val = Vec::new();
+    for (i, input) in function.sig.inputs.iter_mut().enumerate() {
+        let numbered = Ident::new(&format!("__arg{}", i), Span::call_site());
+        match input {
+            FnArg::Typed(PatType { pat, .. }) => {
+                arg_pat.push(quote!(#pat));
+                arg_val.push(quote!(#numbered));
+                *pat = parse_quote!(mut #numbered);
+            }
+            FnArg::Receiver(_) => {
+                move_self = Some(quote! {
+                    if false {
+                        loop {}
+                        #[allow(unreachable_code)]
+                        {
+                            let __self = self;
+                        }
+                    }
+                });
+            }
+        }
+    }
+
+    let ret = match &function.sig.output {
+        ReturnType::Default => quote!(-> ()),
+        output @ ReturnType::Type(..) => quote!(#output),
+    };
+    let stmts = function.block.stmts;
+    function.block = Box::new(parse_quote!({
+        let reader = self.reader;
+        let __result = (move || #ret {
+            #move_self
+            #(
+                let #arg_pat = #arg_val;
+            )*
+            #(#stmts)*
+        })();
+        __result.map_err(|err| include_current_position(&reader, err))
+    }));
+
+    TokenStream::from(quote!(#function))
+}
--- a/ion-c-sys/Cargo.toml
+++ b/ion-c-sys/Cargo.toml
@@ -20,6 +20,7 @@ version = "0.4.10"
 edition = "2018"
 
 [dependencies]
+ion-c-sys-macros = { version = "0.1", path = "../ion-c-sys-macros" }
 paste = "1.0"
 num-bigint = "0.3"
 bigdecimal = "0.2"
--- a/ion-c-sys/src/reader.rs
+++ b/ion-c-sys/src/reader.rs
@@ -2,6 +2,7 @@
 
 //! Provides higher-level APIs for Ion C's `hREADER`.
 
+use ion_c_sys_macros::position_error;
 use std::convert::{TryFrom, TryInto};
 use std::marker::PhantomData;
 use std::ops::{Deref, DerefMut};
@@ -334,6 +335,9 @@ pub trait IonCReader {
     /// # }
     /// ```
     fn read_bytes(&mut self) -> IonCResult<Vec<u8>>;
+
+    /// Returns the current position. TODO: blah blah.
+    fn pos(&self) -> IonCResult<Position>;
 }
 
 /// Wrapper over `hREADER` to make it easier to use readers in IonC correctly.
@@ -367,8 +371,8 @@ impl<'a> IonCReaderHandle<'a> {
         })
     }
 }
-
 impl<'a> IonCReader for IonCReaderHandle<'a> {
+    #[position_error]
     #[inline]
     fn next(&mut self) -> IonCResult<ION_TYPE> {
         let mut tid = ptr::null_mut();
@@ -377,6 +381,7 @@ impl<'a> IonCReader for IonCReaderHandle<'a> {
         Ok(tid)
     }
 
+    #[position_error]
     #[inline]
     fn get_type(&self) -> IonCResult<ION_TYPE> {
         let mut tid = ptr::null_mut();
@@ -385,16 +390,19 @@ impl<'a> IonCReader for IonCReaderHandle<'a> {
         Ok(tid)
     }
 
+    #[position_error]
     #[inline]
     fn step_in(&mut self) -> IonCResult<()> {
         ionc!(ion_reader_step_in(self.reader))
     }
 
+    #[position_error]
     #[inline]
     fn step_out(&mut self) -> IonCResult<()> {
         ionc!(ion_reader_step_out(self.reader))
     }
 
+    #[position_error]
     #[inline]
     fn depth(&self) -> IonCResult<i32> {
         let mut depth = 0;
@@ -403,6 +411,7 @@ impl<'a> IonCReader for IonCReaderHandle<'a> {
         Ok(depth)
     }
 
+    #[position_error]
     #[inline]
     fn is_null(&self) -> IonCResult<bool> {
         let mut is_null = 0;
@@ -411,6 +420,7 @@ impl<'a> IonCReader for IonCReaderHandle<'a> {
         Ok(is_null != 0)
     }
 
+    #[position_error]
     #[inline]
     fn is_in_struct(&self) -> IonCResult<bool> {
         let mut is_in_struct = 0;
@@ -419,6 +429,7 @@ impl<'a> IonCReader for IonCReaderHandle<'a> {
         Ok(is_in_struct != 0)
     }
 
+    #[position_error]
     #[inline]
     fn get_field_name(&mut self) -> IonCResult<StrSliceRef> {
         let mut field = ION_STRING::default();
@@ -450,6 +461,7 @@ impl<'a> IonCReader for IonCReaderHandle<'a> {
         Ok(StrSlicesRef::new(self, annotations))
     }
 
+    #[position_error]
     #[inline]
     fn read_bool(&mut self) -> IonCResult<bool> {
         let mut value = 0;
@@ -458,6 +470,7 @@ impl<'a> IonCReader for IonCReaderHandle<'a> {
         Ok(value != 0)
     }
 
+    #[position_error]
     #[inline]
     fn read_i64(&mut self) -> IonCResult<i64> {
         let mut value = 0;
@@ -466,6 +479,7 @@ impl<'a> IonCReader for IonCReaderHandle<'a> {
         Ok(value)
     }
 
+    #[position_error]
     #[inline]
     fn read_bigint(&mut self) -> IonCResult<BigInt> {
         let mut value = ION_INT::default();
@@ -474,6 +488,7 @@ impl<'a> IonCReader for IonCReaderHandle<'a> {
         Ok(value.try_to_bigint()?)
     }
 
+    #[position_error]
     #[inline]
     fn read_f64(&mut self) -> IonCResult<f64> {
         let mut value = 0.0;
@@ -482,6 +497,7 @@ impl<'a> IonCReader for IonCReaderHandle<'a> {
         Ok(value)
     }
 
+    #[position_error]
     #[inline]
     fn read_bigdecimal(&mut self) -> IonCResult<BigDecimal> {
         let mut value = ION_DECIMAL::default();
@@ -490,6 +506,7 @@ impl<'a> IonCReader for IonCReaderHandle<'a> {
         Ok(value.try_to_bigdecimal()?)
     }
 
+    #[position_error]
     #[inline]
     fn read_datetime(&mut self) -> IonCResult<IonDateTime> {
         let mut value = ION_TIMESTAMP::default();
@@ -498,6 +515,7 @@ impl<'a> IonCReader for IonCReaderHandle<'a> {
         Ok(value.try_to_iondt()?)
     }
 
+    #[position_error]
     #[inline]
     fn read_string(&mut self) -> IonCResult<StrSliceRef> {
         let mut value = ION_STRING::default();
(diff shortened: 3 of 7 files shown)


marcbowes:
The current state of this PR is "gather feedback". It's nowhere near ready to merge.

I was trying to debug why the ion-hash-test.ion file couldn't load. Our error infrastructure doesn't include anything to make this easy. Backtraces don't work even. So I wanted to first fix the debug-ability issue before fixing the actual issue.

The core idea is that IonCError will carry file pos information. There is a new field called `pos` that is optionally hydrated by the reader (writers will never set it). The type is currently the C triple verbatim (bytes, line number, offset).

**Q1:** do we want to carry positional information in case of error? (my opinion: yes)
**Q2:** is a field in `IonCError` the right place to carry the information? (my opinion: yes)
**Q3:** should this be enabled by default? (my opinion: yes, and it cannot be disabled)
**Q4:** should we create a type that at least names the fields? (my opinion: yes - and this is not done)

To carry the information, I added a new method to the reader trait that returns the position (using the underlying ionc call).

**Q5:** should this method be part of the reader trait? (my opinion: yes)

With this in place, the question then became: how to use the new reader API to update the `IonCError` generated by the reader. Initially I tried to change all error-creators to do so with a macro. Here's an example of how it looked:

    #[inline]
    fn read_string(&mut self) -> IonCResult<StrSliceRef> {
        let mut value = ION_STRING::default();
        with_pos!(self, ionc!(ion_reader_read_string(self.reader, &mut value)))?;
        // make a str slice that is tied to our lifetime
        let str_ref = value.as_str(PhantomData::<&'a u8>::default())?;
        Ok(StrSliceRef::new(self, str_ref))
    }

Note `with_pos!(self, ...`, which added the result of `self.pos()` to the `IonCError`. The problem is that I had to add this everywhere (no big deal, but the compiler didn't help me..) and I missed one! In particular, see the `value.as_str(..)?` above. That's the *actual bug I was hunting*. So, instead, I did manual delegation:

```
impl<'a> IonCReader for IonCReaderHandle<'a> {
    #[inline]
    fn next(&mut self) -> IonCResult<ION_TYPE> {
        let reader = self.reader;
        self._next()
            .map_err(|err| include_current_position(&reader, err))
    }
```

this is a real pain to maintain as it has to be done for the entire trait surface area, **so I'm looking for suggestions on how to make it less painful**. However, on the plus side, it completely covers all ways errors can be created.

marcbowes:
And, here is it working:

```
     Running [...]/ion-rust/target/debug/deps/ion_hash_tests-43dcc1c1cca61933

running 1 test
test ion_hash_tests ... FAILED

failures:

---- ion_hash_tests stdout ----
Error: IonCError { source: IonCError { code: 8, message: "IERR_NULL_VALUE", additional: "iERR Result", pos: Some((26874, 1, 7)) } }
thread 'ion_hash_tests' panicked at 'assertion failed: `(left == right)`
```

I updated my ion_hash_tests.rs to test the entire suite and it failed. Note the `pos` above:

```
// symbols
{ ion:$0,            expect:{ identity:(update::(0x0b) update::(0x71) update::(0x0e)
                                        digest::(0x0b 0x71 0x0e)) } }
```

(That fails when `loader` calls `read_string()` for the empty symbol.)

almann:
> **Q1:** do we want to carry positional information in case of error? (my opinion: yes)

I agree.

> **Q2:** is a field in `IonCError` the right place to carry the information? (my opinion: yes)

Yes, and we probably want a field that helps capture other context too.  E.g. `ionc!` should be able to `stringify!` the expression itself plus `file!` and `line!` to make it a bit clearer where the error originated--this should actually be the easiest thing to add.

> **Q3:** should this be enabled by default? (my opinion: yes, and it cannot be disabled)

I agree.

> **Q4:** should we create a type that at least names the fields? (my opinion: yes - and this is not done)

Yes, this should be structured in the error somehow--whether nested or not, I don't have a strong opinion.

almann:
> **Q5:** should this method be part of the reader trait? (my opinion: yes)

I agree, this seems reasonable.

almann:
> this is a real pain to maintain as it has to be done for the entire trait surface area, **so I'm looking for suggestions on how to make it less painful**. However, on the plus side, it completely covers all ways errors can be created.

I really think the only place we can do this is with some kind of delegation/decoration macro over the entire reader surface which probably needs some kind of procedural macro (I'm sure someone has done something like this before).  We will invariably miss something, and this code is not too shy around the use of `?` for various places that `IonCError` can pop up including translation from other more standard errors.

Even if we had to manually use a attribute macro on every method, that would be acceptable:

```rust
#[position_error]
fn read_string(...) -> IonCResult<...>
```

I think extending `ionc!` with `stringify!` and `line!` and `file!` is probably the easiest first step.  But something like the above is probably what we want for the code you're writing manually.

Alternatively, we manually decorate every method.  The above macro is just shorthand for something vaguely like:

```rust
fn actual_read_string(...) -> IonCResult<...> {...}

fn read_string(...) -> IonCResult<...> {
  match self.actual_read_string(...) {
    Ok(v) => Ok(v),
    Err(e) => Err(include_current_position(&self.reader, e))
  }
}
```

marcbowes:
```
#[position_error]
```
That's a good suggestion. I'll look into it when I get a chance.

Agree with the `ionc!` suggestion, but while that's easier to implement it also has less utility for me right now.

almann:
> Agree with the `ionc!` suggestion, but while that's easier to implement it also has less utility for me right now.

Added #242 for this.

ion-c-sys-macros/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+///
+/// This macro rewrites the function such that any error gets the position
+/// added. It does this by moving the existing implementation into a closure and
+/// then calling `map_err` on the result.
+///
+/// To make the variety of possible functions work (which include mutable
+/// borrows, lifetime elision), the closure uses the `move` keyword. There are
+/// some gymnastics (adapted from dtolnay/no-panic) to make that work out.
+#[proc_macro_attribute]
+pub fn position_error(_attr: TokenStream, item: TokenStream) -> TokenStream {
```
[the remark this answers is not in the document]
marcbowes: To unit test this macro in isolation, we have to use proc_macro2, which would require some changes. It's doable, but..

I believe we have excellent coverage. Firstly, the compiler exercises this code when we type `cargo test` in the workspace. It builds ion-c-sys, applies the transforms and if it doesn't compile you'll hear about it. That provides really great coverage because we have to fully implement the reader trait. Next up, we run the normal tests (including things like the loader tests) that exercise the inner implementation still works. Finally, I added a specific integration test in reader.rs that shows position errors show up.

ion-c-sys-macros/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+                        loop {}
+                        #[allow(unreachable_code)]
+                        {
+                            let __self = self;
+                        }
+                    }
+                });
+            }
+        }
+    }
```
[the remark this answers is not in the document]
marcbowes: The best way to explain this is: the code is munged in arcane ways that dtolnay invented. If you remove one of the arcane steps, things fail to compile. For example, if we remove the `#move_self` in the final codegen, then the methods that return self references like this one:

```
fn get_field_name(&mut self) -> IonCResult<StrSliceRef>;
```

will fail to compile:

```
error[E0495]: cannot infer an appropriate lifetime for lifetime parameter `'a` due to conflicting requirements
   --> ion-c-sys/src/reader.rs:526:12
    |
526 |         Ok(StrSliceRef::new(self, str_ref))
    |            ^^^^^^^^^^^^^^^^
    |
note: first, the lifetime cannot outlive the lifetime `'_` as defined on the body at 518:5...
   --> ion-c-sys/src/reader.rs:518:5
    |
518 |     #[position_error]
    |     ^^^^^^^^^^^^^^^^^
note: ...so that closure can access `self`
   --> ion-c-sys/src/reader.rs:526:29
    |
526 |         Ok(StrSliceRef::new(self, str_ref))
    |                             ^^^^
note: but, the lifetime must be valid for the anonymous lifetime defined on the method body at 520:20...
   --> ion-c-sys/src/reader.rs:520:20
    |
520 |     fn read_string(&mut self) -> IonCResult<StrSliceRef> {
    |                    ^^^^^^^^^
note: ...so that the expression is assignable
   --> ion-c-sys/src/reader.rs:518:5
    |
518 |     #[position_error]
    |     ^^^^^^^^^^^^^^^^^
    = note: expected `Result<string::StrSliceRef<'_>, _>`
               found `Result<string::StrSliceRef<'_>, _>`
    = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

I don't really know enough about the arcane magic to give it a reasonable comment, hence the sort of handwave one I added already:

```
/// This macro rewrites the function such that any error gets the position
/// added. It does this by moving the existing implementation into a closure and
/// then calling `map_err` on the result.
///
/// To make the variety of possible functions work (which include mutable
/// borrows, lifetime elision), the closure uses the `move` keyword. There are
/// some gymnastics (adapted from dtolnay/no-panic) to make that work out.
```

ion-c-sys-macros/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+    let stmts = function.block.stmts;
+    function.block = Box::new(parse_quote!({
+        let reader = self.reader;
+        let __result = (move || #ret {
+            #move_self
+            #(
+                let #arg_pat = #arg_val;
+            )*
+            #(#stmts)*
+        })();
```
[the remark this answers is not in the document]
marcbowes: I don't know how to answer this better than the previous comment I made. Removing any of these incantations causes compile failures due to the various self/mut-self and elision requirements of our functions. Fully explaining weird move closure stuff in this macro is beyond my knowledge.

ion-c-sys/src/result.rs
```diff
@@ -10,12 +10,15 @@ use std::fmt;
...
+///
+/// `position` is populated when errors come from readers. See [`Position`] for
+/// more information.
 #[derive(Copy, Clone, Debug, PartialEq)]
 pub struct IonCError {
     pub code: i32,
     pub message: &'static str,
     pub additional: &'static str,
-    pub pos: Option<(i64, i32, i32)>,
+    pub position: Option<reader::Position>,
```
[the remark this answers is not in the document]
marcbowes: Ok!

src/value/loader.rs
```diff
@@ -489,7 +489,30 @@ mod loader_tests {
...
     ) -> IonResult<()> {
         let actual: Vec<_> = loader().iterate_over(input)?.collect();
-        assert_eq!(expected, actual);
+        assert_eq!(strip_positions(expected), strip_positions(actual));
         Ok(())
     }
+
+    // FIXME: For discussion:
+    // 1. Is stripping positions the right way forward?
+    // 2. Where should this code live?
```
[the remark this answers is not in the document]
marcbowes: I'll push back a little on this, but defer to you. I think that if we require bytes/offsets in our tests, this is what's going to happen:

1. Developer makes up a random position, runs tests, they fail.
2. Developer copies the position into the test code, things pass.

This workflow is just silly (more work, no real verification things work), but it's the most likely one IMO.

ion-c-sys-macros/src/lib.rs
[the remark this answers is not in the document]
marcbowes: https://github.com/dtolnay/no-panic/commit/90d81328e38d28f63865d92bd19d95b5247686e4

>  Make compatible with receiver lifetime elision

A magician never reveals his secrets..

src/value/loader.rs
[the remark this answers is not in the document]
almann: There is no random position though in _these_ cases.  So it should be explicit if we want to ignore it.  We've caught off by one line/position counting bugs before.   So in the round-trip tests, I would assert the error more weakly, but in these low-level unit tests I would at least try to put a reasonably position.

ion-c-sys-macros/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+/// }
+/// ```
+///
+/// This macro rewrites the function such that any error gets the position
+/// added. It does this by moving the existing implementation into a closure and
+/// then calling `map_err` on the result.
+///
+/// To make the variety of possible functions work (which include mutable
+/// borrows, lifetime elision), the closure uses the `move` keyword. There are
+/// some gymnastics (adapted from dtolnay/no-panic) to make that work out.
```
almann, line 36:
```suggestion
/// ***Note***: to make the variety of possible functions work (which include mutable
/// borrows, lifetime elision), the closure uses the `move` keyword. There are
/// some gymnastics (adapted from ***[dtolnay/no-panic][no-panic]***) to make that work out.
///
/// [no-panic]: https://github.com/dtolnay/no-panic 
```

ion-c-sys-macros/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+#[proc_macro_attribute]
+pub fn position_error(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let mut function = syn::parse_macro_input!(item as syn::ItemFn);
+
+    let mut move_self = None;
+    let mut arg_pat = Vec::new();
+    let mut arg_val = Vec::new();
```
almann, line 43:
```suggestion
    /// XXX this is heavily adapted for dtolnay/panic - as this code evolves to deal with rustc/ion-c-sys changes
    ///  refer to to the above for any changes here...
    let mut move_self = None;
    let mut arg_pat = Vec::new();
    let mut arg_val = Vec::new();
```

almann (commented):
Added some comment suggestions that could help the future person doing something here that _there be dragons_.  Basically, I want the tone of the comments for code we don't understand completely to be blunt about that.

src/value/loader.rs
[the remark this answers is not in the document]
marcbowes: I will remove the strip and add the positions.

marcbowes:
Squashed commit has arrived!

ion-c-sys/src/reader.rs
```diff
@@ -367,8 +371,8 @@ impl<'a> IonCReaderHandle<'a> {
         })
     }
 }
-
 impl<'a> IonCReader for IonCReaderHandle<'a> {
+    #[position_error]
     #[inline]
```
almann, line 376: Nice!

ion-c-sys/src/reader.rs
```diff
@@ -533,6 +552,44 @@ impl<'a> IonCReader for IonCReaderHandle<'a> {
...
+    let mut line = -1;
+    let mut offset = -1;
+    ionc!(ion_reader_get_position(
+        *reader,
+        &mut bytes,
+        &mut line,
+        &mut offset
+    ))?;
+    Ok((bytes, line, offset).into())
+}
```
almann, line 580: Should this just be a method with the reader or does that complicate things?

ion-c-sys/src/reader.rs
```diff
@@ -533,6 +552,44 @@ impl<'a> IonCReader for IonCReaderHandle<'a> {
...
+    if let Some(_) = err.position {
+        return err;
+    }
+
+    let pos = match reader_current_pos(reader) {
+        Ok(pos) => pos.into(),
+        Err(_) => return err, // the original, not the one from failing to read the pos
+    };
+
+    err.with_position(pos)
```
almann, line 592: Ditto.

ion-c-sys/src/result.rs
```diff
@@ -10,11 +10,66 @@ use std::fmt;
...
+/// containing Ion data that is being parsed using an [`IonCReader`].
+///
+/// `bytes` will always be hydrated while `lines` and `offset` will only be
+/// populated for text readers.
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub struct Position {
+    pub bytes: i64,
+    pub line: Option<i32>,
+    pub offset: Option<i32>,
+}
```
almann, line 34: I am probably going to sound nitpicking here but when I see all these options here, that really makes me wonder if `Position` is really an `enum`.  Maybe something like:

```rust
/// Represents a position in a stream/file by line and column.
pub struct LineAndColumn(usize, usize);

/// Possible position information from Ion C.
pub enum Position {
  Unknown,
  /// Number of bytes from the start of the stream/file
  Offset(usize),
  OffsetLineColumn(usize, LineAndColumn),
}
```

FWIW, I went through this evolution of problem on partiql/partiql-lang-rust#8

This has the benefit of collapsing all your `Option` modalities into one thing.  Though this factoring may make expressing logic around this worse, so feel free to push back.

ion-c-sys/src/result.rs
```diff
@@ -10,11 +10,66 @@ use std::fmt;
...
+            None
+        } else {
+            Some(ionc_values.2)
+        };
+        Self {
+            bytes,
+            line: lines,
+            offset,
+        }
+    }
```
almann, line 72: I see what is going on here--though I think it better to just put into a named method.  This seems a little clever to have a call-site like `(...).into()` do this conversion via negatives.

src/value/reader.rs
```diff
@@ -442,7 +442,9 @@ mod reader_tests {
             $10
         "#,
         vec![
-            Err(IonCError::from(ion_error_code_IERR_NULL_VALUE).into())
+            Err(IonCError::from(ion_error_code_IERR_NULL_VALUE)
+                .with_position(Position::new(124, 4, 8).into())
+                .into())
```
almann, line 447: :+1:

almann (commented):
Overall this is shaping up really nicely, some questions/comments below.

ion-c-sys/src/reader.rs
marcbowes: It compiles things because of the `move` closure generated by the macro. Want a comment?

ion-c-sys/src/result.rs
marcbowes: I'll make this change.

ion-c-sys/src/result.rs
marcbowes: I'll remove this method and use the constructors directly.

ion-c-sys/src/reader.rs
almann: Might be good to comment just so someone doesn't stub their toe on this in the future.

almann (approved):
:rocket:

ion-c-sys/Cargo.toml
```diff
@@ -20,6 +20,7 @@ version = "0.4.10"
 edition = "2018"
 
 [dependencies]
+ion-c-sys-macros = { version = "0.1.0", path = "../ion-c-sys-macros" }
```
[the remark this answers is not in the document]
marcbowes: Was copying what you had earlier (but maybe you just changed it..)

Issue #216, opened 2021-04-25
`PartialEq` for `...Struct` Symmetry Bug


almann:
When looking into #214, I discovered a bug where according to our implementation of `PartialEq` for `OwnedStruct`, the following happens:

```plain
{a: 1} == {a: 1, b: 2}
{a: 1, b: 2} != {a: 1}
```

The problem is that when a struct that is the subset of another is compared on the *left* hand side, the algorithm does not check to make sure all elements in the *right* hand side are in the *left*.

While I can conceive of a test that can be added `..._element_accessors` test function, I believe very strongly that we do the proper equivalence class testing and non-equivalence class testing **separate** of those tests to consider this bug fixed.

almann:
When fixing this, we should un-comment these skip lists:

https://github.com/amzn/ion-rust/blob/e2fd5245b8c5ec7e676ee6043864224cfd685da0/tests/loader_test_vectors.rs#L57-L58

PR #239, opened 2021-05-03
changed PartialEq for struct for symmetry bug

desaikd:
*Issue #216 *

*Description of changes:*
This PR works on solving the issue with symmetry in `Struct` of `Value` API.

*Changes:*
- Added length check in the `PartialEq` of `...Struct` for symmetry.

*Tests:*
Added unittests for different length structs in `Value` API.

### DIFF
--- a/src/value/borrowed.rs
+++ b/src/value/borrowed.rs
@@ -375,7 +375,11 @@ impl<'val> PartialEq for BorrowedStruct<'val> {
     fn eq(&self, other: &Self) -> bool {
         // check if both text_fields and no_text_fields have same length
         self.text_fields.len() == other.text_fields.len() && self.no_text_fields.len() == other.no_text_fields.len()
-        // check if text_fields and no_text_fields are equal
+        // check if text_fields and no_text_fields are equal 
+        // we need to test equality in both directions for both text_fields and no_text_fields
+        // A good example for this is annotated vs not annotated values in struct
+        //  { a:4, a:4 } vs. { a:4, a:a::4 } // returns true
+        //  { a:4, a:a::4 } vs. { a:4, a:4 } // returns false 
         && self.eq_text_fields(other) && other.eq_text_fields(self)
             && self.eq_no_text_fields(other) && other.eq_no_text_fields(self)
     }
--- a/src/value/borrowed.rs
+++ b/src/value/borrowed.rs
@@ -273,6 +273,31 @@ pub struct BorrowedStruct<'val> {
     no_text_fields: Vec<(BorrowedSymbolToken<'val>, BorrowedElement<'val>)>,
 }
 
+impl<'val> BorrowedStruct<'val> {
+    fn eq_text_fields(&self, other: &Self) -> bool {
+        // check if both the text_fields have same (field_name,value) pairs
+        self.text_fields.iter().all(|(key, value)| {
+            value.iter().all(|(_my_s, my_v)| {
+                other
+                    .get_all(key)
+                    .find(|other_v| my_v == *other_v)
+                    .is_some()
+            }) && value.len() == other.get_all(key).count()
+        })
+    }
+
+    fn eq_no_text_fields(&self, other: &Self) -> bool {
+        // check if both the no_text_fields are same values
+        self.no_text_fields.iter().all(|(my_k, my_v)| {
+            other
+                .no_text_fields
+                .iter()
+                .find(|(other_k, other_v)| my_k == other_k && my_v == other_v)
+                .is_some()
+        })
+    }
+}
+
 impl<'val, K, V> FromIterator<(K, V)> for BorrowedStruct<'val>
 where
     K: Into<BorrowedSymbolToken<'val>>,
@@ -348,23 +373,11 @@ impl<'val> Struct for BorrowedStruct<'val> {
 
 impl<'val> PartialEq for BorrowedStruct<'val> {
     fn eq(&self, other: &Self) -> bool {
-        // check if both the text_fields have same (field_name,value) pairs
-        self.text_fields.iter().all(|(key, value)| {
-            value.iter().all(|(_my_s, my_v)| {
-                other
-                    .get_all(key)
-                    .find(|other_v| my_v == *other_v)
-                    .is_some()
-            })
-        }) && self.no_text_fields.iter().all(|(my_k, my_v)| {
-            // check if both the no_text_fields have same values
-            other
-                .no_text_fields
-                .iter()
-                .find(|(other_k, other_v)| my_k == other_k && my_v == other_v)
-                .is_some()
-        }) && self.text_fields.len() == other.text_fields.len()
-            && self.no_text_fields.len() == other.no_text_fields.len()
+        // check if both test_fields and no_text_fields have same length
+        self.text_fields.len() == other.text_fields.len() && self.no_text_fields.len() == other.no_text_fields.len()
+        // check if text_fields and no_text_fields are equal
+        && self.eq_text_fields(other) && other.eq_text_fields(self)
+            && self.eq_no_text_fields(other) && other.eq_no_text_fields(self)
     }
 }
 
--- a/src/value/borrowed.rs
+++ b/src/value/borrowed.rs
@@ -373,7 +373,7 @@ impl<'val> Struct for BorrowedStruct<'val> {
 
 impl<'val> PartialEq for BorrowedStruct<'val> {
     fn eq(&self, other: &Self) -> bool {
-        // check if both test_fields and no_text_fields have same length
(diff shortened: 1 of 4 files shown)


src/value/borrowed.rs
```diff
@@ -363,7 +363,8 @@ impl<'val> PartialEq for BorrowedStruct<'val> {
                 .iter()
                 .find(|(other_k, other_v)| my_k == other_k && my_v == other_v)
                 .is_some()
-        })
+        }) && self.text_fields.len() == other.text_fields.len()
+            && self.no_text_fields.len() == other.no_text_fields.len()
```
almann, line 367: I might front load these conditionals (fail fast if the lengths don't line up).

src/value/owned.rs
```diff
@@ -367,7 +367,8 @@ impl PartialEq for OwnedStruct {
                 .iter()
                 .find(|(other_k, other_v)| my_k == other_k && my_v == other_v)
                 .is_some()
-        })
+        }) && self.text_fields.len() == other.text_fields.len()
+            && self.no_text_fields.len() == other.no_text_fields.len()
```
almann, line 371: Same as the other one

almann (changes requested):
As I mentioned in #216 

> While I can conceive of a test that can be added `..._element_accessors` test function, I believe very strongly that we do the proper equivalence class testing and non-equivalence class testing **separate** of those tests to consider this bug fixed.

I strongly believe we should _stop_ overloading these tests.  Doing the equivalence class of tests for equality/non-equality for the `Struct` trait on its own tests would be a lot better and give us higher confidence that there are no bugs and this piece of code has had at least a couple of bugs in it so far due to inadequate testing.

src/value/mod.rs
```diff
@@ -838,6 +838,395 @@ mod generic_value_tests {
...
+                    vec![(
+                        E::SymbolToken::local_sid_token(21),
+                        E::Builder::new_string("hello"),
+                    )]
+                    .into_iter(),
+                ),
+                // SymbolToken with local SID and no text are equivalent to each other and to SID $0
+                E::Builder::new_struct(
+                    vec![(
+                        E::SymbolToken::local_sid_token(22),
```
zslayton, line 862: Highlighting this as part of our offline discussion: this is a case where we're constructing an instance of a type that implements `SymbolToken` and our expectation is that "we didn't provide text, therefore there is no text."

src/value/mod.rs
```diff
@@ -838,6 +838,395 @@ mod generic_value_tests {
         }
     }
 
+    struct CaseStruct<E: Element> {
```
zslayton, line 841: This could use a comment to explain its purpose.

src/value/borrowed.rs
```diff
@@ -348,22 +373,11 @@ impl<'val> Struct for BorrowedStruct<'val> {
...
-        }) && self.no_text_fields.iter().all(|(my_k, my_v)| {
-            // check if both the no_text_fields have same values
-            other
-                .no_text_fields
-                .iter()
-                .find(|(other_k, other_v)| my_k == other_k && my_v == other_v)
-                .is_some()
-        })
+        // check if both test_fields and no_text_fields have same length
+        self.text_fields.len() == other.text_fields.len() && self.no_text_fields.len() == other.no_text_fields.len()
```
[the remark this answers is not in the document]
almann: Our current definition of `SymbolToken` defines that its text must be resolvable via `SymbolToken::text()` if possible (note that this does not imply that an implementation need to be eager/lazy).  So based on your statement above, I would expect (1) to be true.

We do have a definition in `SymbolToken` that this should be consistent with:

https://github.com/amzn/ion-rust/blob/4d0c67b3273bd62603ddc2a3c75d711348b0e4b8/src/value/mod.rs#L223-L259

A more friendly rendering is here:

https://docs.rs/ion-rs/0.5.0/ion_rs/value/trait.SymbolToken.html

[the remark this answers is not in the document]
almann: >The two structs being compared having the same encodings for each field, which would be too strict.

We cannot use `PartialEq` or `Eq` to compare across two different implementations, but I am not sure to what you mean with _same encodings_ for the field.  Do you mind clarifying this statement?

src/value/mod.rs
almann: :+1:

src/value/mod.rs
```diff
@@ -838,6 +838,395 @@ mod generic_value_tests {
...
+        eq_annotations: Vec<E>,
+        ne_annotations: Vec<E>,
```
almann, line 843: Commenting what this is for would be helpful.

src/value/mod.rs
```diff
@@ -1583,28 +1563,6 @@ mod generic_value_tests {
...
-    #[case::borrowed_struct_with_different_length_multiple_fields(struct_with_different_length_multiple_fields_case::<BorrowedElement>())]
-    #[case::owned_struct_with_multiple_fields_not_equal(struct_with_multiple_fields_not_equal_case::<OwnedElement>())]
-    #[case::borrowed_struct_with_multiple_fields_not_equal(struct_with_multiple_fields_not_equal_case::<BorrowedElement>())]
-    #[case::owned_struct_with_text_and_duplicates(struct_with_text_and_duplicates_case::<OwnedElement>())]
-    #[case::borrowed_struct_with_text_and_duplicates(struct_with_text_and_duplicates_case::<BorrowedElement>())]
-    #[case::owned_struct_with_unordered_text_and_duplicates(struct_with_unordered_text_and_duplicates_case::<OwnedElement>())]
-    #[case::borrowed_struct_with_unordered_text_and_duplicates(struct_with_unordered_text_and_duplicates_case::<BorrowedElement>())]
-    #[case::owned_struct_with_unordered_no_text_and_duplicates(struct_with_unordered_no_text_and_duplicates_case::<OwnedElement>())]
-    #[case::borrowed_struct_with_unordered_no_text_and_duplicates(struct_with_unordered_no_text_and_duplicates_case::<BorrowedElement>())]
```
almann: This elimination makes me happy!

src/value/mod.rs
```diff
@@ -838,6 +838,395 @@ mod generic_value_tests {
...
+                    ]
+                    .into_iter(),
+                ),
+            ],
+        }
+    }
+
+    fn struct_with_duplicate_fieldnames_case<E: Element>() -> CaseStruct<E>
+    where
+        E::Builder: Builder<SymbolToken = E::SymbolToken>,
```
almann, line 1114: What about a variant(s) of these that have sources and/or no text either alone or mixed with text?

almann (commented):
Thank you for moving these tests out.  I would like to see more cases around mixing text/no text field names with/without sources to really make sure the equality coverage is there...

src/value/borrowed.rs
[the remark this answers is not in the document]
zslayton: Per our offline discussion: the current `SymbolToken`'s trait had some intentional ambiguity and we're going to revise some of the language.

>>    The two structs being compared having the same encodings for each field, which would be too strict.
>
> We cannot use PartialEq or Eq to compare across two different implementations, but I am not sure to what you mean with same encodings for the field. Do you mind clarifying this statement?

We established offline that `BorrowedSymbolToken` is required to be fully resolved (i.e. `text()` and `local_sid()` will be populated if at all possible), which renders this point moot.

Because `SymbolToken` did not have a hard requirement that its implementations provide resolved data, it was feasible that `BorrowedSymbolToken` could be used for both resolved and 'raw' use cases. If a `BorrowedStruct` was maintaining a collection of raw `BorrowedSymbolToken`s, then the line of code in question would be requiring that the two structs use the same field encodings. (`{foo: true}` and `{$12: true}` would not be considered equal even if `$12=foo`.)

src/value/mod.rs
```diff
@@ -838,8 +838,11 @@ mod generic_value_tests {
         }
     }
 
+    // A struct that defines input case for `struct_accessors` method
```
almann, line 841: Minor, especially since this will not get docs generated, but I like the consistency:

```suggestion
    /// A struct that defines input case for `struct_accessors` method
```

src/value/mod.rs
```diff
@@ -838,8 +838,11 @@ mod generic_value_tests {
...
     struct CaseStruct<E: Element> {
+        // set of struct elements that are the same
         eq_annotations: Vec<E>,
+        // set of struct elements that are never equal to `eq_annotations`
```
almann:
```suggestion
        /// set of struct elements that are the same
        eq_annotations: Vec<E>,
        /// set of struct elements that are never equal to `eq_annotations`
```

almann (approved):
:ship:

Issue #243, opened 2021-05-05
Rename `Dumper`/`Loader` to `ElementWriter`/`ElementReader`


almann:
As per discussion with @zslayton offline around #237, it was pointed out to me that `Dumper` can be potentially offensive.  We came up with the names `ElementWriter` and `ElementReader` to replace the existing names.

Furthermore when referring to _fixed buffers_, we should just use the term _slice_.  Also, we should rename the integration tests for this as well.

PR #254, opened 2021-05-08
Renames `Loader` and `Dumper to `ElementReader` and `ElementWriter`.

### LINKED ISSUE #251: Reconsider `try_...` for constructing `ElementWriter`
We should consider a name for this like `Format::open_writer_for_slice(...)` or the like.

---
I don't think I've seen many instances of `try_NOUN` in constructor names. (Though I think most constructors are infallible, which may be why.)

When this is renamed per #243, how would you feel about something like: 
```rust
fn element_writer_for(O: output) -> IonResult<ElementWriter<Output=O>> where O: io::Write
```
?

(As mentioned in another comment, I'm unsure why constructing an `ElementWriter` impl without doing any serialization operations could fail.)

_Originally posted by @zslayton in https://github.com/amzn/ion-rust/pull/237#discussion_r627511406_

almann:
* Renames the `value::loader` module to `value::reader`.
* Renames the `value::dumper` module to `value::writer`.
* Renames various references to dump/dumper and load/loader.
* Renames `loader_test_vectors.rs` to `element_test_vectors`.
* Minor API cleanup to match the new names.

Fixes #251 and #243.

### DIFF
--- a/ion-hash/src/lib.rs
+++ b/ion-hash/src/lib.rs
@@ -4,11 +4,11 @@
 //!
 //! ## Examples
 //! ```rust
-//! use ion_rs::value::loader::{self, Loader};
+//! use ion_rs::value::reader::{self, ElementReader};
 //! use ion_rs::result::IonResult;
 //!
 //! # fn main() -> IonResult<()> {
-//!   let loader = loader::loader();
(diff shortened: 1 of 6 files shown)


marcbowes (approved):
Thanks!

Issue #258, opened 2021-05-10
Enhance `SymbolToken` to support notion of Unknown Context.


almann:
This is a digest of a conversation that I had with @zslayton and @desaikd.  Specifically there was a mental model misalignment between the three of us, and I think we believe there is an important concept missing in our current conception of these APIs.

The [`SymbolToken`][1] API has the following method that returns the import source of the token:

https://github.com/amzn/ion-rust/blob/a84579536e673643e596fd147691d8d119d72140/src/value/mod.rs#L278-L279

This API returns `None` if no such source exists or is _known_.  The problem is the distinction between a _knowable state_ and an _unknowable state_.  For example, given a symbol token from some raw reader, before that token is resolved, the source is in an _unknowable state_, this means that there _may_ be some state that _could_ provide context for the `text` or `source` of the token, but it hasn't yet been resolved either through the local symbol table and/or a catalog.  Another context is a local symbol with unknown text locally defined or with some import that is not resolvable by the application.  This has a `source` that is _known_ not defined or _known_ to be defined with some source that still renders the `text` _unknown_.

I think the above has a few implications that we should investigate for modeling this in our APIs while preserving the ergonomics of the common case.

## Distinguishing `None`

Today we use `None` to mean two different things depending on context.  We should still preserve this short-hand for applications that don't care about the distinction, but introduce the following concepts:

* **`Unknown`** - which is a _kind_ of `None` that is _potentially knowable_, but is lacking the context (that an application may never satisfy) for it.  An example of `Unknown` is the `text` for a symbol token that came from a shared symbol table that we do not have in our application context.  Another example is a `source` for a symbol token that has not yet been resolved against some symbol table context.
* **`Nothing`** - which is a _kind_ of `None` that is known to be _explicitly_ specified as such and no amount of context can change this.  An example of `Nothing` is any symbol whose `text` is explicitly `null` in the symbol table definition.

```rust
/// Similar to [`Option`] but differentiates `None`.
pub enum Specified<T> {
    /// Indicates that the current processing context does not know if the value exists or not, but it *could*
    /// exist under the correct conditions.
    Unknown,

    /// Indicates that the value is *explicitly* not specified.
    Nothing,

    /// Indicates a value
    Some(T),
}
```

The API consideration of `source` and also the following `text` could be updated to use this revised `enum`.  Consider `text`:

[...]

PR #259, opened 2021-05-11
Adds some clarification around `SymbolToken` API docs.

almann:
Specifically documents more clearly what implementations are allowed
to do for `text` and `source`.  Also adds more examples around
implementations that may have certain behaviors.

Related to #258.


desaikd (approved):
LGTM! Do we want to add some doc example tests to demonstrate this?

src/value/mod.rs
```diff
@@ -269,13 +287,21 @@ pub trait SymbolToken: Debug + PartialEq {
...
 
     /// The ID of the token, which may be `None` if no ID is associated with the token
     /// (e.g. Ion text symbols).
     fn local_sid(&self) -> Option<SymbolId>;
 
     /// The source of this token, which may be `None` if the symbol is locally defined.
+    ///
+    /// An implementation is allowed to return `None` if the source is explicitly known to be
+    /// *undefined* (e.g. a locally defined symbol) or the context that would allow the source
+    /// to be defined  has not been applied or is not available.
```
zslayton, line 304:
```suggestion
    /// to be defined has not been applied or is not available.
```

zslayton (approved):

almann:
> Do we want to add some doc example tests to demonstrate this?

I am not sure I can--it alludes to implementations that may or may not exist.  I mean I could draft one, but I don't think that would be useful.

PR #264, opened 2021-05-12
Refactor ion-hash implementation

marcbowes:
*Description of changes:*

This PR has two commits that make ion-hash (which is still pretty early days) a bit nicer.

The first commit refactors `IonTypeCode` to play nicer with the `std::convert` traits and makes the type public. This allows ion-hash to convert from IonType -> IonTypeCode without reinventing the wheel.

The next commit moves away from hardcoded typequalifiers (such as `TQ_STRING = 0x80`) and onto a more logical setup. I think that the new implementation looks a lot more like the spec, which is a good thing for maintainability. I tried to use `const fn` to have easy to read code that also performs like a magic lookup table.

Beyond that, there is no new ion-hash functionality. Things should work as before.

While I was at it, I fixed up the tests to use `ElementWriter` to print their names instead of relying on case numbering.


ion-hash/src/lib.rs
```diff
@@ -141,3 +127,22 @@ fn ion_hash_escape(representation: &[u8]) -> Vec<u8> {
...
+
+fn _binary_repr<E: Element + ?Sized>(elem: &E) -> IonResult<Vec<u8>> {
+    // TODO: Fix this once the writer API supports non-fixed-size buffers.
+    let mut buf = vec![0u8; 4096];
+    let mut writer = Format::Binary.element_writer_for_slice(&mut buf)?;
+    writer.write(elem)?;
+    let len = writer.finish()?.len();
+    buf.truncate(len);
+    Ok(buf)
+}
```
almann, line 148: Do you want the full Ion binary encoding including the IVM + symbol table?  I don't think that is right.

ion-hash/src/type_qualifier.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+impl TypeQualifier {
+    /// Computes a [`TypeQualifier`] from an [`Element`] according to the rules
+    /// laid out in the spec. In many cases, the `T` is determined by the Ion
+    /// binary type code.
+    ///
+    /// This code implements the TQ computation logically, but leans on `const
+    /// fn` to have the same runtime performance as if we wrote a jump table out
+    /// by hand.
```
almann, line 33: I don't think this is a jump table, but just a compile time generated lookup table, right?  I just want to make sure I am not misunderstanding the `const fn` above.

ion-hash/src/type_qualifier.rs
```diff
@@ -0,0 +1,91 @@
...
+            IonType::SExpression => SExpression,
+            IonType::Struct => Struct,*/
+            _ => todo!(),
+        }
+    }
+
+    /// Convenient transform to feed to a `Digest`.
+    pub(crate) fn into_bytes(self) -> [u8; 1] {
+        [self.0]
+    }
```
almann, line 80: Would it make more sense to make this return a slice to itself?

```rust
fn as_bytes(&self) -> &[u8] {
    std::slice::from_ref(self)
}
```

src/binary/type_code.rs
```diff
@@ -88,3 +96,29 @@ impl IonTypeCode {
         Ok(ion_type_code)
     }
 }
+
+/// Turns an [`IonTypeCode`] into a byte. This does not use `From` so that it
+/// can be a `const fn`.
+impl IonTypeCode {
+    pub const fn to_u8(self) -> u8 {
```
almann, line 103:
```suggestion
impl IonTypeCode {
    /// Constant function to convert an [`IonTypeCode`] into a `u8`.
    pub const fn to_u8(self) -> u8 {
```

src/binary/type_code.rs
```diff
@@ -32,11 +34,13 @@ pub(crate) enum IonTypeCode {
     Reserved,         // 15
 }
 
-impl IonTypeCode {
+impl TryFrom<IonTypeCode> for IonType {
+    type Error = IonError;
+
```
almann, line 39: :+1:

ion-hash/src/lib.rs
marcbowes: Nope. The other implementations factor out the binary writer and the "raw" writer. 

https://github.com/amzn/ion-hash-java/blob/b855f0836a527b05cca2895f1099c0028283c0a9/src/com/amazon/ionhash/HasherImpl.java#L353
https://github.com/amzn/ion-java/blob/0eb62ba92633f03e9f3175bb77f1b3214fcc1cdd/src/com/amazon/ion/impl/bin/_PrivateIon_HashTrampoline.java#L35
https://github.com/amzn/ion-java/blob/0eb62ba92633f03e9f3175bb77f1b3214fcc1cdd/src/com/amazon/ion/impl/bin/IonRawBinaryWriter.java#L1085

ion-hash/src/type_qualifier.rs
marcbowes: yeah, let me update that

ion-hash/src/type_qualifier.rs
marcbowes: probably - https://docs.rs/digest/0.9.0/digest/trait.Digest.html#tymethod.update

almann: Yes, then I think my `as_bytes` avoids an extra copy to get you to a slice. (not that the copy is expensive since it is copying a byte, but I think the API makes more sense than returning a size 1 array).

ion-hash/src/lib.rs
almann: Yeah, I would feel more comfortable just ripping this dead code out for now...  The crux is you want the raw encoding functions, and we should figure that problem out.

marcbowes: yeah! For now, I'm just going to duplicate them by hand since it's not very complicated.

Latest commit adds ints and removes this code.

ion-hash/src/type_qualifier.rs
marcbowes: Fixed in latest commit.

ion-hash/src/representation.rs
```diff
@@ -0,0 +1,37 @@
+use ion_rs::value::AnyInt;
+
+// FIXME: Some types (e.g. BigInt) can't spit out a representation as a
+// reference, and thus need to allocate. Change this trait so that not all types
+// have to pay the cost of allocation.
```
marcbowes, line 5: TODO: I will document this trait and fix the FIXME before merging.

src/binary/type_code.rs
```diff
@@ -88,3 +96,28 @@ impl IonTypeCode {
         Ok(ion_type_code)
     }
 }
+
+impl IonTypeCode {
+    /// Constant function to convert an [`IonTypeCode`] into a `u8`.
+    pub const fn to_u8(self) -> u8 {
```
zslayton, line 102: Just curious: why write a `to_u8` function instead of providing an `impl From<IonTypeCode> for u8`?

I thought it might be because you couldn't impl a trait for a type outside the current crate, but [that seems to have changed (at least for the `From` trait?) in Rust 1.41](https://doc.rust-lang.org/std/convert/trait.Into.html#implementing-into-for-conversions-to-external-types-in-old-versions-of-rust). (See working playground example [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=93b7f1923030d832cd65c33c6a237f0a)).

`const` could be buying us a minor performance benefit; is that the motivation?

ion-hash/src/type_qualifier.rs
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+
+    /// Convenient transform to feed to a `Digest`.
+    pub(crate) fn as_bytes(&self) -> &[u8] {
+        slice::from_ref(&self.0)
+    }
+}
+
+fn is_anyint_positive(value: Option<&AnyInt>) -> bool {
```
zslayton, line 91: This seems generally useful enough that we should consider adding it as a method on `AnyInt`.

ion-hash/src/representation.rs
marcbowes: Latest commit fixes.

ion-hash/tests/ion_hash_tests.rs
```diff
@@ -89,20 +89,20 @@ fn test_file(file_name: &str) -> IonResult<()> {
...
             .expect("test cases have an `expect` value");
-        test_case(i, ion, expect)?;
+        test_case(ion, expect)?;
     }
 
     Ok(())
 }
 
-fn test_case<E: Element>(case_number: usize, ion: &E, strukt: &E) -> IonResult<()> {
+fn test_case<E: Element>(ion: &E, strukt: &E) -> IonResult<()> {
```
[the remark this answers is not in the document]
marcbowes: I can! I named it after the structure in the tests:

```

// ints
{ ion: 0, expect:{ identity:(update::(0x0b) update::(0x20)                update::(0x0e) digest::(0x0b      0x20 0x0e)) } }
{ ion:-0, expect:{ identity:(update::(0x0b) update::(0x20)                update::(0x0e) digest::(0x0b      0x20 0x0e)) } }
//^^^^^
```

what would you suggest?

src/binary/type_code.rs
marcbowes: Yeah, it's all about const. The previous commit message said something like "this is not From because of const" and Almann suggested I rewrite it to the above.

I haven't benchmarked `const`. I just did it like this because it was the weekend and I was having fun!

ion-hash/src/type_qualifier.rs
marcbowes: Good idea. Are you ok with a follow up PR?

zslayton: Yep, that works!

almann (approved):
:ship:

ion-hash/tests/ion_hash_tests.rs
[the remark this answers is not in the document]
zslayton: How about `ion`->`actual` and `strukt`->`expected` or similar?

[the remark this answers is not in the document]
marcbowes: Thanks! Will make those changes in one of my next commits.

PR #265, opened 2021-05-14
TestDigest now uses interior mutability

marcbowes:
The fixed-length output requirements of the Digest trait make
integrating with the ion-hash-test suite a little annoying! Nothing
wrong with the trait; rather it's a consequence of the design of the
test suite..

The tests define correct as "the underlying digest was fed a series of
updates that look like this" - kind of like a mock/expect setup. Because
different test cases have different sequences, this design clashes with
the fixed length expectations of the Digest trait.

To work around this, we completely ignore the fixed-length result in our
tests and return assert on a result vec that lives outside of the Digest
trait implementation.

One wrinkle is that IonHasher wants to own the Digest so that it can be
sure nobody else is writing to it. To work around that, we use interior
mutability to allow our mock to be cloned and our tests keep a concrete
reference to it.

As a direct consequence of this change, I can enable the intLength512
test which was longer that our previous 256 bit length limit. Also, the
nonsense around trailing zeros is gone!

-------

*Issue #, if available:*

*Description of changes:*


ion-hash/tests/ion_hash_tests.ion
```diff
@@ -20,6 +20,50 @@
...
 { ion:-6, expect:{ identity:(update::(0x0b) update::(0x30) update::(0x06) update::(0x0e) digest::(0x0b 0x30 0x06 0x0e)) } }
+intLength512::{
+  ion:-0xfe95f4cff19a8ee2edbbee30c7c0acbb83bfc4c0a58e8b94bb6250aeeaf3db8f41b0acdbb94b990c518d96c5ee3c02e276e06e07570a2b6e5dea9fe4fac8475a84efca8a8432d6d463bf0ceb470b4ad9b3b0c80730492e5ee660bca86932d933c471f178140c5256affe4ef5c0404d74b4b7776e77178b3281e1c5b65ad8866bcbaa6225c4e1c5b9624b19dcc6001afc3535a3769c8e937b7e3f9073ab0053cc0ffeb34124d5d570749d0181f4d4dedced7d28f038247befa18ce02a3d1293da637bb1ab6598bb6617a6a5ce0512c390236dbca283adf0291e6903fbd6819d4c5a8216c69e2083da5b3fefb0928b208345a39207c8461e38f793036146107559adf2f40612d25f14d45d7e2780b45e2cf9b5790d91aaaf327af3323e20242c2632a64725844f1d9e218aab0d56ee99ae486034d7b3fbfc4dce8c9cc2a793ce93affe81dee7158dad7f0623ce692c8ed0975dbeef9a717a0b63f90af4febc96785a6ff4e06b090a65d33c98932df39f7c5b807956a19897e0c3463046df2eb4df624c7c43bef48fab381a857b9f5b6c1bdbd6b3270c107cd3bc1c41fe04e1ddac69f14119de961af773285544f819f3951542f704b501ff0364bf54d14a86e19bec39394c85a6b256c6233da801a44f5db98ccdd8d9bb6788c014216dd57cb64573333ceed5b5c72a4ee296e75b3e32ed69083675a6a8f6b8ac85deaed88ad0a7,
```
almann, line 24: Are these in the original files as well?  or is this something you added?

almann (approved):
The limitations of the `Digest` trait makes sense for interior mutability, thanks for spelling it out.  I do wonder if it is worth just creating a large fixed size (high water mark) to avoid the complexity.  IIRC, you had done that on one of your original factorings.

ion-hash/tests/ion_hash_tests.ion
marcbowes: Yes, th ey are.

marcbowes:
> The limitations of the `Digest` trait makes sense for interior mutability, thanks for spelling it out. I do wonder if it is worth just creating a large fixed size (high water mark) to avoid the complexity. IIRC, you had done that on one of your original factorings.

That's what I did based on your previous feedback (and what I'm undoing here!). I think it was actually more complex (having to use generic array, etc.) and harder to understand. The code as written in this PR uses standard rust idioms (interior mutability and shallow cloning) and matches "spy objects" as found in other languages.

PR #266, opened 2021-05-14
Refactor out `raw::i64_magnitude`

marcbowes:
This commit refactors the code that produces the "magnitude" subfield
out of `BinarySystemWriter`, allowing the code to be re-used in
ion-hash.

The implementation returns a I64Magnitude type rather than a slice. This
avoids an allocation (Vec, for example).

ion-hash now avoids i64 -> BigInt as a result of this change.

-------

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/ion-hash/src/representation.rs
+++ b/ion-hash/src/representation.rs
@@ -7,10 +7,10 @@
 //! and not speed.
 
 use ion_rs::{
+    binary,
     value::{AnyInt, Element},
     IonType,
 };
-use num_bigint::BigInt;
 
 // TODO: Finish ion-rust's binary writer and factor it such that the binary
 // representations can be written by the "raw" writer (ref. the Java
@@ -39,16 +39,16 @@ pub trait Representation {
 
 impl Representation for Option<&AnyInt> {
     fn repr(&self) -> BinaryRepr {
-        fn bigint_repr(b: &BigInt) -> Vec<u8> {
-            b.magnitude().to_bytes_be()
-        }
-
         match self {
             Some(AnyInt::I64(v)) => match v {
                 0 => None,
-                _ => Some(bigint_repr(&BigInt::from(*v))),
+                _ => {
+                    let magnitude = v.abs() as u64;
+                    let magnitude_bytes = binary::uint::magnitude_to_bytes(magnitude);
+                    Some(magnitude_bytes.as_ref().into())
+                }
             },
-            Some(AnyInt::BigInt(b)) => Some(bigint_repr(b)),
+            Some(AnyInt::BigInt(b)) => Some(b.magnitude().to_bytes_be()),
             None => None,
         }
     }
--- a/ion-hash/src/representation.rs
+++ b/ion-hash/src/representation.rs
@@ -44,8 +44,8 @@ impl Representation for Option<&AnyInt> {
                 0 => None,
                 _ => {
                     let magnitude = v.abs() as u64;
-                    let magnitude_bytes = binary::uint::magnitude_to_bytes(magnitude);
-                    Some(magnitude_bytes.as_ref().into())
+                    let encoded = binary::uint::encode_uint(magnitude);
+                    Some(encoded.as_ref().into())
                 }
             },
             Some(AnyInt::BigInt(b)) => Some(b.magnitude().to_bytes_be()),
--- a/ion-hash/src/representation.rs
+++ b/ion-hash/src/representation.rs
@@ -45,7 +45,7 @@ impl Representation for Option<&AnyInt> {
                 _ => {
                     let magnitude = v.abs() as u64;
                     let encoded = binary::uint::encode_uint(magnitude);
-                    Some(encoded.as_ref().into())
+                    Some(encoded.as_bytes().into())
                 }
             },
             Some(AnyInt::BigInt(b)) => Some(b.magnitude().to_bytes_be()),
--- a/src/binary/mod.rs
+++ b/src/binary/mod.rs
@@ -7,7 +7,7 @@ mod header;
 mod int;
 mod nibbles;
 mod type_code;
-mod uint;
+pub mod uint;
 mod var_int;
 mod var_uint;
 pub mod writer;
--- a/src/binary/uint.rs
+++ b/src/binary/uint.rs
@@ -48,12 +48,8 @@ impl UInt {
 
     /// Encodes the provided `magnitude` as a UInt and writes it to the provided `sink`.
     pub fn write_u64<W: Write>(sink: &mut W, magnitude: u64) -> IonResult<usize> {
-        // leading_zeros() uses an intrinsic to calculate this quickly
-        let empty_leading_bytes: u32 = magnitude.leading_zeros() >> 3; // Divide by 8 to get byte count
-        let first_occupied_byte = empty_leading_bytes as usize;
-
-        let magnitude_bytes: [u8; mem::size_of::<u64>()] = magnitude.to_be_bytes();
-        let bytes_to_write: &[u8] = &magnitude_bytes[first_occupied_byte..];
+        let magnitude_bytes = magnitude_to_bytes(magnitude);
+        let bytes_to_write = magnitude_bytes.as_ref();
 
         sink.write_all(bytes_to_write)?;
         Ok(bytes_to_write.len())
@@ -70,6 +66,45 @@ impl UInt {
     #[inline(always)]
     pub fn size_in_bytes(&self) -> usize {
         self.size_in_bytes
+    }
+}
+
+/// The result of the [`magnitude_to_bytes`] function. The fields are private;
+/// users must go through `as_ref()` to see the slice that ignores leading
+/// zeros.
+///
+/// This type exists to avoid heap allocation.
+pub struct U64Magnitude {
+    magnitude_bytes: [u8; mem::size_of::<u64>()],
+    first_occupied_byte: usize,
+}
+
+impl AsRef<[u8]> for U64Magnitude {
+    fn as_ref(&self) -> &[u8] {
+        &self.magnitude_bytes[self.first_occupied_byte..]
+    }
+}
+
+/// Returns the magnitude as big-endian bytes.
+///
+/// ```
+/// use ion_rs::binary::uint;
+/// use std::convert::AsRef;
+///
+/// let repr = uint::magnitude_to_bytes(5u64);
+/// assert_eq!(&[0x05], repr.as_ref());
+/// ```
+pub fn magnitude_to_bytes(magnitude: u64) -> U64Magnitude {
+    // We can divide the number of leading zero bits by 8
+    // to to get the number of leading zero bytes.
+    let empty_leading_bytes: u32 = magnitude.leading_zeros() >> 3;
+    let first_occupied_byte = empty_leading_bytes as usize;
+
+    let magnitude_bytes: [u8; mem::size_of::<u64>()] = magnitude.to_be_bytes();
+
+    U64Magnitude {
+        magnitude_bytes,
+        first_occupied_byte,
     }
 }
 
--- a/src/binary/uint.rs
+++ b/src/binary/uint.rs
@@ -11,14 +11,14 @@ const MAX_UINT_SIZE_IN_BYTES: usize = mem::size_of::<UIntStorage>();
 /// [UInt and Int Fields](http://amzn.github.io/ion-docs/docs/binary.html#uint-and-int-fields)
 /// section of the binary Ion spec for more details.
 #[derive(Debug)]
-pub struct UInt {
+pub struct DecodedUInt {
     size_in_bytes: usize,
     value: UIntStorage,
 }
 
-impl UInt {
+impl DecodedUInt {
     /// Reads a UInt with `length` bytes from the provided data source.
-    pub fn read<R: IonDataSource>(data_source: &mut R, length: usize) -> IonResult<UInt> {
+    pub fn read<R: IonDataSource>(data_source: &mut R, length: usize) -> IonResult<DecodedUInt> {
         if length > MAX_UINT_SIZE_IN_BYTES {
             return decoding_error(format!(
                 "Found a {}-byte UInt. Max supported size is {} bytes.",
@@ -40,7 +40,7 @@ impl UInt {
             magnitude |= byte;
         }
 
-        Ok(UInt {
+        Ok(DecodedUInt {
             size_in_bytes: length,
             value: magnitude,
         })
@@ -48,8 +48,8 @@ impl UInt {
 
     /// Encodes the provided `magnitude` as a UInt and writes it to the provided `sink`.
     pub fn write_u64<W: Write>(sink: &mut W, magnitude: u64) -> IonResult<usize> {
-        let magnitude_bytes = magnitude_to_bytes(magnitude);
-        let bytes_to_write = magnitude_bytes.as_ref();
+        let encoded = encode_uint(magnitude);
+        let bytes_to_write = encoded.as_ref();
 
         sink.write_all(bytes_to_write)?;
         Ok(bytes_to_write.len())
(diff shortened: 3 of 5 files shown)


src/binary/writer.rs
```diff
@@ -868,6 +860,76 @@ impl<W: Write> BinarySystemWriter<W> {
...
+            &self.magnitude_bytes[self.first_occupied_byte..]
+        }
+    }
+
+    /// Returns the magnitude of an i64 as big-endian bytes.
+    ///
+    /// The caller must slice the returned array, ignoring the bits up until the
+    /// indicated usize. This is neccessary because the magnitude bytes is
+    /// dynamically sized (leading zero bits must be removed), but we want to
+    /// avoid allocations (for performance reasons).
```
zslayton, line 913: Minor quibble here: while accurate, this doc comment requires the user to be familiar with the inner workings of the `I64Magnitude`.

src/binary/writer.rs
```diff
@@ -868,6 +860,76 @@ impl<W: Write> BinarySystemWriter<W> {
...
+    ///
+    /// See the [`i64_magnitude`] macro.
```
zslayton, line 915: I think this is outdated?

ion-hash/src/representation.rs
```diff
@@ -7,10 +7,10 @@
 //! and not speed.
 
 use ion_rs::{
+    binary::writer::raw,
```
zslayton, line 10: There are already some modules for encoding primitive types like `VarUInt`, `Int`, etc. This probably belongs with those? You could make the case that those should be moved to this module.

zslayton (commented):
I think this type should probably also be to support going the other direction: turning an encoded magnitude byte array from a stream into an `i64`? If you agree we can just open an issue to track that.

marcbowes:
> I think this type should probably also be to support going the other direction: turning an encoded magnitude byte array from a stream into an `i64`? If you agree we can just open an issue to track that.

What do you think about moving this into your `Int` type? That might make things clearer.

src/binary/writer.rs
```diff
@@ -868,6 +860,76 @@ impl<W: Write> BinarySystemWriter<W> {
...
+    pub struct I64Magnitude {
+        magnitude_bytes: [u8; mem::size_of::<u64>()],
+        first_occupied_byte: usize,
+    }
+
+    impl AsRef<[u8]> for I64Magnitude {
+        fn as_ref(&self) -> &[u8] {
+            &self.magnitude_bytes[self.first_occupied_byte..]
+        }
+    }
```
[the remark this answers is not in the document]
marcbowes: I agree with the point around autocomplete, but *in my opinion* I don't think it's as important as being idiomatic. Now, I'm not stating that what I've done **is** idiomatic - rather, that I **think** I am being idiomatic! Part of the benefit of being idiomatic is the tooling will catch up to the idioms.

I'm fine with having two methods (one being the idiomatic one, the other being the one that works with autocomplete), but honestly I don't think it's worth it - especially not for this "low level" stuff that we really don't expect the majority of users to use.

src/binary/writer.rs
marcbowes: yeah it's actually a little stale too - was more relevant before I made I64Magnitude only have one accessor method that couldn't be used incorrectly. I'll rework it.

src/binary/writer.rs
marcbowes: wups

ion-hash/src/representation.rs
marcbowes: agreed

src/binary/writer.rs
```diff
@@ -868,6 +860,76 @@ impl<W: Write> BinarySystemWriter<W> {
...
+/// include an i64 in an Ion hash computation, the length is not relevant!
+/// That's why this factoring is useful: it enables code sharing of the
+/// individual steps.
+///
+/// [spec]: https://amzn.github.io/ion-docs/docs/binary.html#2-and-3-int
+pub mod raw {
+    use super::*;
+
+    /// The result of `i64_magnitude`. The fields are private; users must go
+    /// through `as_ref()` to see the slice that ignores leading zeros.
```
almann:
```suggestion
    /// Provides the *compact* slice over the big-endian representation of the magnitude of
    /// an `i64`.  That is the representation of the magnitude of an `i64` without leading zero
    /// bytes.
```

src/binary/writer.rs
```diff
@@ -868,6 +860,76 @@ impl<W: Write> BinarySystemWriter<W> {
...
+    ///
+    /// This type exists to avoid heap allocation.
+    pub struct I64Magnitude {
```
almann: Should this `#[derive(Copy, Clone, Debug)]` at a minimum?

src/binary/writer.rs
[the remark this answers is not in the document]
almann: I don't have a strong opinion, but if I think of `I64Magnitude` as a slice, then the `AsRef<[u8]>` makes sense because it is essentially a container for producing that slice reference.

That said, I do like also having an explicit method that gives the option to make the call site more obvious when not using the trait, this is a common pattern in a lot of types in the Rust library but those are typically in types that implement `AsRef<...>` more than once (e.g. `String` that has `as_str()`/`as_bytes()` and implements `AsRef<str>` and `AsRef<[u8]>`.

src/binary/uint.rs
```diff
@@ -73,6 +69,45 @@ impl UInt {
     }
 }
 
+/// The result of the [`magnitude_to_bytes`] function. The fields are private;
+/// users must go through `as_ref()` to see the slice that ignores leading
+/// zeros.
```
almann, line 74: This doc comment reads as a bit too "how".  Maybe:

```suggestion
/// The big-endian, compact slice of bytes for a `u64`.  Leading zero octets are
/// not part of the representation.
```

src/binary/uint.rs
```diff
@@ -73,6 +69,45 @@ impl UInt {
...
+///
+/// This type exists to avoid heap allocation.
+pub struct U64Magnitude {
```
almann, line 77: Should this `#[derive(Copy, Clone, Debug)]` at a minimum since it is effectively a smallish data type.

src/binary/uint.rs
```diff
@@ -73,6 +69,45 @@ impl UInt {
...
+    magnitude_bytes: [u8; mem::size_of::<u64>()],
+    first_occupied_byte: usize,
```
almann, line 79: Probably not worth optimizing, but `usize` is overkill for sure, `u8` is all that is needed, but even `u32` probably aligns better (`u8` probably doesn't matter because of padding).

src/binary/uint.rs
```diff
@@ -73,6 +69,45 @@ impl UInt {
...
+}
+
+impl AsRef<[u8]> for U64Magnitude {
+    fn as_ref(&self) -> &[u8] {
+        &self.magnitude_bytes[self.first_occupied_byte..]
+    }
+}
+
+/// Returns the magnitude as big-endian bytes.
```
almann, line 88: I would just say:

```suggestion
/// Converts a `u64` to a [`U64Magnitude`.]
```

The rework of the doc comment makes me wonder if it would be more idiomatic as just `impl From<u64> for U64Magnitude`.

src/binary/uint.rs
```diff
@@ -73,6 +69,45 @@ impl UInt {
...
+/// use ion_rs::binary::uint;
+/// use std::convert::AsRef;
+///
+/// let repr = uint::magnitude_to_bytes(5u64);
+/// assert_eq!(&[0x05], repr.as_ref());
+/// ```
+pub fn magnitude_to_bytes(magnitude: u64) -> U64Magnitude {
+    // We can divide the number of leading zero bits by 8
+    // to to get the number of leading zero bytes.
+    let empty_leading_bytes: u32 = magnitude.leading_zeros() >> 3;
```
almann, line 100: *Minor*:  the comment talks about divide by 8, and then you left shift 3 which I grant is the same thing, but if `rustc` can't figure that out on its own from `zeros() / 8`, I would be floored (no math pun `int`ended).  :rofl: *I'll see myself out...*

src/binary/uint.rs
```diff
@@ -73,6 +69,45 @@ impl UInt {
...
+
+/// Returns the magnitude as big-endian bytes.
+///
+/// ```
+/// use ion_rs::binary::uint;
+/// use std::convert::AsRef;
+///
+/// let repr = uint::magnitude_to_bytes(5u64);
+/// assert_eq!(&[0x05], repr.as_ref());
+/// ```
```
almann, line 96: Doc test is nice, but some unit boundary tests for the slices that you produce would be good.  They are probably covered by integration, but I'd still say its worth it.

almann (commented):
Looks pretty good--I only have a few questions.

src/binary/writer.rs
marcbowes: I think not - I literally want to hide anything other than the slice view.

src/binary/writer.rs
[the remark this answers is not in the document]
zslayton: > this is a common pattern in a lot of types in the Rust library but those are typically in types that implement AsRef<...> more than once (e.g. String that has as_str()/as_bytes() and implements AsRef<str> and AsRef<[u8]>.

This was exactly the example I had in mind, thanks for adding this. I should've been more explicit.

zslayton:
> > I think this type should probably also be to support going the other direction: turning an encoded magnitude byte array from a stream into an `i64`? If you agree we can just open an issue to track that.
> 
> What do you think about moving this into your `Int` type? That might make things clearer.

The `magnitude` field [is actually a `UInt`](https://amzn.github.io/ion-docs/docs/binary.html#2-and-3-int), so it would be more at home there. However, in the process of confirming this I realized that this code already exists in `UInt`:

https://github.com/amzn/ion-rust/blob/2965c941de2581402d6b1bd9600d5564178f435e/src/binary/uint.rs#L50-L60

For some reason the writer was never updated to use this method. 🙃

src/binary/uint.rs
```diff
@@ -73,6 +69,44 @@ impl UInt {
     }
 }
 
+/// The big-endian, compact slice of bytes for a `u64`.  Leading zero octets are
+/// not part of the representation.
+///
+/// This type exists to avoid heap allocation.
+pub struct U64Magnitude {
```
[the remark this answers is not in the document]
marcbowes: Nice suggestions, thanks

src/binary/mod.rs
```diff
@@ -7,7 +7,7 @@ mod header;
 mod int;
 mod nibbles;
 mod type_code;
-mod uint;
+pub mod uint;
```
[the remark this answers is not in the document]
marcbowes: Just from an api stability point of view, or something else?

src/binary/uint.rs
marcbowes: I added one more test to the doctest to show two bytes. Otherwise, I played around with the implementation a bit (e.g. changed the division by 8) and coverage is excellent by tests that are already in this file.

src/binary/uint.rs
marcbowes: I love the convert traits, but I think this one is too unintuitive. Zack suggested `encode_uint`, which I like, so I went with that.

src/binary/uint.rs
marcbowes: It has to become a usize at some point because of the `RangeFrom` implementation. As an aside, I didn't change this at all in this PR, I just de-duplicated the existing implementation.

src/binary/uint.rs
marcbowes: Added

src/binary/writer.rs
marcbowes: I landed up adding this, but I'm not sure it has a use.

src/binary/uint.rs
```diff
@@ -11,14 +11,14 @@ const MAX_UINT_SIZE_IN_BYTES: usize = mem::size_of::<UIntStorage>();
 /// [UInt and Int Fields](http://amzn.github.io/ion-docs/docs/binary.html#uint-and-int-fields)
 /// section of the binary Ion spec for more details.
 #[derive(Debug)]
-pub struct UInt {
+pub struct DecodedUInt {
     size_in_bytes: usize,
     value: UIntStorage,
```
almann, line 16: This is not your change, but I find the type alias really meaningless (it literally puzzled me until I went looking for it.  Maybe a question for @zslayton, but why bother with the alias here?

src/binary/uint.rs
```diff
@@ -11,14 +11,14 @@ const MAX_UINT_SIZE_IN_BYTES: usize = mem::size_of::<UIntStorage>();
 /// [UInt and Int Fields](http://amzn.github.io/ion-docs/docs/binary.html#uint-and-int-fields)
 /// section of the binary Ion spec for more details.
 #[derive(Debug)]
-pub struct UInt {
+pub struct DecodedUInt {
```
almann, line 14: This name is confusing to me--but if you and @zslayton had agreed on this, I am fine with it.

src/binary/uint.rs
```diff
@@ -40,16 +40,16 @@ impl UInt {
             magnitude |= byte;
         }
 
-        Ok(UInt {
+        Ok(DecodedUInt {
             size_in_bytes: length,
             value: magnitude,
         })
```
almann, line 46: Not your change, but I really think we need to note why this type is what it is, since this length might be padded or not directly relevant to the decoded value...

src/binary/uint.rs
```diff
@@ -119,39 +127,39 @@ mod tests {
...
     }
 
     #[test]
     fn test_read_three_byte_uint() {
         let data = &[0b0011_1100, 0b1000_0111, 0b1000_0001];
-        let uint = UInt::read(&mut Cursor::new(data), data.len()).expect(READ_ERROR_MESSAGE);
+        let uint = DecodedUInt::read(&mut Cursor::new(data), data.len()).expect(READ_ERROR_MESSAGE);
         assert_eq!(uint.size_in_bytes(), 3);
         assert_eq!(uint.value(), 3_966_849);
     }
```
almann, line 149: Not really related to your change, but I think it odd that I could have the following:

```rust
let data = &[0x00, 0x00, 0x00, 0x01];
let uint = DecodedUInt::read(...);
assert_eq(uint.size_in_bytes(), 4);
assert_eq(uint.value(), 1);
```

@zslayton, is this intentional?

almann (approved):
Minor questions around naming, and questions mostly for @zslayton, but I think this is good to go.

marcbowes:
> Minor questions around naming, and questions mostly for @zslayton, but I think this is good to go.

Thanks. I'm going to wait on @zslayton before merging.

src/binary/mod.rs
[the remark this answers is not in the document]
zslayton: > Just from an api stability point of view [...]?

Yep.

src/binary/uint.rs
zslayton: When I first wrote this years ago, I vacillated between using a few types for storing magnitudes. `usize` was convenient because I could use the magnitude for various `std::collections` operations without casting. Then I decided I wanted it to be 8 bytes regardless of the platform, so I switched it to `u64`, which required refactoring a lot of API names in an era before CLion made that easy. THEN I heard rumblings that they were going to eventually add a `u128` primitive type (which they did [in Sept of 2020](https://github.com/rust-lang/rust/releases/tag/1.26.0)!) and wondered if I might some day think that was a better choice. I decided to use a type alias to simplify refactoring if/when I inevitably changed my mind again. :)

In short: we don't need it now. If we DO decide to change the storage type, refactoring is a lot easier in 2021.

src/binary/uint.rs
zslayton: Yes. The `DecodedUInt` represents a value that was read from the stream; the `size_in_bytes` describes what was decoded. This API mirrors that of the `VarUInt` type, where the purpose for this is more obvious. 
```rust
let var_uint = VarUInt::read(some_input);
// How many bytes did we read from input?
let representation_size = var_uint.size_in_bytes();
// And what was the numeric value of the VarUInt?
let length = var_uint.value();
```
We don't _need_ this field/method per se in the `UInt` type because the reader had to specify how many bytes to read to begin with, it's mostly here for consistency with the other encoding primitives.

src/binary/uint.rs
```diff
@@ -73,9 +69,56 @@ impl UInt {
...
 }
 
+/// The big-endian, compact slice of bytes for a `u64`. Leading zero octets are
+/// not part of the representation.
```
zslayton, line 73: I think this should be described in terms of the `UInt` encoding primitive (with a link to [that section of the spec](https://amzn.github.io/ion-docs/docs/binary.html#uint-and-int-fields)) rather than in terms of `u64`.

src/binary/uint.rs
zslayton: Naming is hard. 🙂 I think we need a name pair for `Uint`/`U64Magnitude` that communicates which one was read from a stream (and so is an integer) and which one is being written to a stream (and so is a byte sequence). Open to suggestions.

src/binary/writer.rs
```diff
@@ -406,29 +408,21 @@ impl<W: Write> BinarySystemWriter<W> {
...
-            // We can divide the number of leading zero bits by 8
-            // to to get the number of leading zero bytes.
-            let empty_leading_bytes: u32 = magnitude.leading_zeros() >> 3;
-            let first_occupied_byte = empty_leading_bytes as usize;
-
-            let magnitude_bytes: [u8; mem::size_of::<u64>()] = magnitude.to_be_bytes();
-            let bytes_to_write: &[u8] = &magnitude_bytes[first_occupied_byte..];
+            let magnitude = value.abs() as u64;
+            let encoded = uint::encode_uint(magnitude);
+            let bytes_to_write = encoded.as_ref();
```
zslayton, line 413:
```suggestion
            let bytes_to_write = encoded.as_bytes();
```

zslayton (approved):
Looking good, just a few minor things below.

src/binary/uint.rs
marcbowes: Hehe, this is @almann 's suggested change from earlier!

Latest commit should be a good compromise.

zslayton (approved):

PR #268, opened 2021-05-26
Rework ion-hash representations

marcbowes:
Before this commit, every `Element` visited during ion hash computation
would need to allocate a vector: because each element type can be
represented by a variable number of bytes! For example, a boolean has no
representation while a string is representated by its UTF8 bytes.

This commit fixes that by having the representation implementation write
to the digester directly. For some types (like a BigInt) we will still
need a variable length vector. But in other cases, we now no longer need
to allocate.

While implementing this, I wanted to make the escape method also not
require allocation. This is a simple case of wrapping the `update`
method to maybe write two bytes per byte written. However, because we
were dependent on the `Digest` trait, writing the wrapper required
wrapping all of those methods. In this commit, we move away from the
`Digest` trait and rather use the traits that *make up* that trait. This
means the escaping wrapper just needs to worry about `update`.

### DIFF
--- a/ion-hash/src/lib.rs
+++ b/ion-hash/src/lib.rs
@@ -85,8 +85,9 @@ where
             | IonType::Symbol
             | IonType::String
             | IonType::Clob
-            | IonType::Blob => self.hash_scalar(elem),
-            IonType::List | IonType::SExpression | IonType::Struct => todo!(),
+            | IonType::Blob
+            | IonType::Struct => self.hash(elem),
+            IonType::List | IonType::SExpression => todo!(),
         };
         self.mark_end();
 
@@ -110,8 +111,8 @@ where
         self.hasher.update(tq.as_bytes());
     }
 
-    fn hash_scalar<E: Element + ?Sized>(&mut self, elem: &E) {
+    fn hash<E: Element + ?Sized>(&mut self, elem: &E) {
         self.hash_no_repr(elem);
-        representation::write_repr(elem, &mut self.hasher);
+        representation::update_with_representation(elem, &mut self.hasher);
     }
 }
--- a/ion-hash/src/lib.rs
+++ b/ion-hash/src/lib.rs
@@ -16,7 +16,7 @@
 //! # }
 //! ```
 
-use digest::{Digest, Output};
+use digest::{Digest, FixedOutput, Output, Reset, Update};
 use ion_rs::result::IonResult;
 use ion_rs::{value::Element, IonType};
 
@@ -51,16 +51,20 @@ impl Markers {
 // hash. For now, we provide a standalone hasher using the Element API.
 /// Provides Ion hash over arbitrary [`Element`] instances with a given
 /// [`Digest`] algorithm.
+///
+/// The trait bounds are over the various subtraits that make up the `Digest`
+/// trait. The reason for this is we can't go from `Digest` to subtrait (e.g.
+/// `Update`) but can go the other way around!
 pub struct IonHasher<D>
 where
-    D: Digest,
+    D: Update + FixedOutput + Reset + Clone + Default,
 {
     hasher: D,
 }
 
 impl<D> IonHasher<D>
 where
-    D: Digest,
+    D: Update + FixedOutput + Reset + Clone + Default,
 {
     pub fn new(hasher: D) -> Self {
         Self { hasher }
@@ -86,7 +90,7 @@ where
 
         // TODO: Annotations
 
-        Ok(self.hasher.finalize())
+        Ok(self.hasher.finalize_fixed())
     }
 
     #[inline]
@@ -106,21 +110,6 @@ where
 
     fn hash_scalar<E: Element + ?Sized>(&mut self, elem: &E) {
         self.hash_no_repr(elem);
-        if let Some(repr) = representation::binary_repr(elem) {
-            self.hasher.update(ion_hash_escape(&repr[..]));
-        }
+        representation::write_repr(elem, &mut self.hasher);
     }
-}
-
-/// Replaces each marker byte M with ESC || M.
-fn ion_hash_escape(representation: &[u8]) -> Vec<u8> {
-    let mut out = Vec::with_capacity(representation.len());
-    for byte in representation {
-        if let Markers::B | Markers::E | Markers::ESC = *byte {
-            out.push(0x0C);
-        }
-        out.push(*byte);
-    }
-
-    out
 }
--- a/ion-hash/src/lib.rs
+++ b/ion-hash/src/lib.rs
@@ -52,9 +52,11 @@ impl Markers {
 /// Provides Ion hash over arbitrary [`Element`] instances with a given
 /// [`Digest`] algorithm.
 ///
-/// The trait bounds are over the various subtraits that make up the `Digest`
-/// trait. The reason for this is we can't go from `Digest` to subtrait (e.g.
-/// `Update`) but can go the other way around!
+/// Note that [`Digest`] *does not imply* the following traits directly, but is
+/// a *convenience* trait for these other traits with a blanked implementation.
+/// This is the reason for this is we can't go from [`Digest`] to these
+/// traits (e.g. [`Update`]) but because of a blanket implementation, we can go
+/// the other way.
 pub struct IonHasher<D>
 where
     D: Update + FixedOutput + Reset + Clone + Default,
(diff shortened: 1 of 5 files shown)


ion-hash/src/representation.rs
```diff
@@ -15,83 +17,87 @@ use ion_rs::{
...
+            if v == 0.0 {
+                return if !v.is_sign_positive() {
+                    hasher.update(&[0x80, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00])
+                };
+            }
+            if v.is_infinite() {
+                return if v.is_sign_positive() {
+                    hasher.update(&[0x7F, 0xF0, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00])
+                } else {
+                    hasher.update(&[0xFF, 0xF0, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00])
```
almann, line 85: *Minor*, it would be nice to have this in constants (magic numbers and all...)

ion-hash/src/lib.rs
```diff
@@ -106,21 +110,6 @@ where
 
     fn hash_scalar<E: Element + ?Sized>(&mut self, elem: &E) {
         self.hash_no_repr(elem);
-        if let Some(repr) = representation::binary_repr(elem) {
-            self.hasher.update(ion_hash_escape(&repr[..]));
-        }
+        representation::write_repr(elem, &mut self.hasher);
```
almann, line 113: I got confused by this terminology

`no_repr` is confusing, I might spell this as `hash_type` or something, because IIRC, you need to do this for all values, right?  `write_repr` might be better spelled as `scalar_repr` or something like that.

Channeling my inner @zslayton also I would suggest we spell out the terms to be consistent with the style of the other crates (e.g. `representation`).

src/types/mod.rs
```diff
@@ -33,6 +33,30 @@ pub enum IonType {
...
+                IonType::String => "string",
+                IonType::Clob => "clob",
+                IonType::Blob => "blob",
+                IonType::List => "list",
+                IonType::SExpression => "sexp",
+                IonType::Struct => "struct",
+            }
+        )
+    }
+}
```
almann, line 58: Out of curiosity, where do you need this?  Is the `Debug` not sufficient for some use case?

ion-hash/src/lib.rs
```diff
@@ -51,16 +51,20 @@ impl Markers {
...
 /// Provides Ion hash over arbitrary [`Element`] instances with a given
 /// [`Digest`] algorithm.
+///
+/// The trait bounds are over the various subtraits that make up the `Digest`
+/// trait. The reason for this is we can't go from `Digest` to subtrait (e.g.
+/// `Update`) but can go the other way around!
 pub struct IonHasher<D>
 where
-    D: Digest,
+    D: Update + FixedOutput + Reset + Clone + Default,
```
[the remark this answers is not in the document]
marcbowes: What would the value be? I kind of like enumerating the APIs we need (that we need a thing that goes quack; it need not be a duck).

ion-hash/src/representation.rs
```diff
@@ -15,83 +17,87 @@ use ion_rs::{
...
+        IonType::Null | IonType::Boolean => {} // these types have no representation
+        IonType::Integer => write_repr_integer(elem.as_any_int(), &mut hasher),
+        IonType::Float => write_repr_float(elem.as_f64(), &mut hasher),
         IonType::Decimal | IonType::Timestamp | IonType::Symbol => todo!(),
-        IonType::String => elem.as_str().repr(),
+        IonType::String => write_repr_string(elem.as_str(), &mut hasher),
         IonType::Clob | IonType::Blob | IonType::List | IonType::SExpression | IonType::Struct => {
-            todo!()
+            panic!("type {} is not yet supported", elem.ion_type())
         }
```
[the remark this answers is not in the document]
marcbowes: I just don't really see the value in it. The trait just adds more lines of codes, and no safety.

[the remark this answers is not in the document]
marcbowes: And I quite liked that only the struct one needs FixedOutput+Default.

ion-hash/src/representation.rs
```diff
@@ -15,83 +17,87 @@ use ion_rs::{
...
-                0 => None,
-                _ => {
-                    let magnitude = v.abs() as u64;
-                    let encoded = binary::uint::encode_uint(magnitude);
-                    Some(encoded.as_bytes().into())
-                }
-            },
-            Some(AnyInt::BigInt(b)) => Some(b.magnitude().to_bytes_be()),
-            None => None,
+struct UpdateEscaping<'a, U: Update>(&'a mut U);
```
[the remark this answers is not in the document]
marcbowes: I really like that newer languages like Rust and Kotlin don't force you to have methods on objects - you're free to organize functions however you want! The current structure in this PR spends no effort trying to tease out or rationalize a new concept here. We have a method that writes out representations, it dispatches with a match statement, and we have an escaping wrapper.

ion-hash/src/lib.rs
marcbowes: Ok!

src/types/mod.rs
marcbowes: I keep doing println-debugging and `{}` is just natural. I can remove it if you want..

ion-hash/src/lib.rs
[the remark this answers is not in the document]
almann: Fair, I can't think of it--the `Digest` trait and its factoring is a bit weird IMO.

ion-hash/src/representation.rs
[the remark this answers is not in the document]
almann: Yeah, I don't feel strongly other than it really does look like `hasher.write_repr_string(elem.as_str())` to me.

ion-hash/src/representation.rs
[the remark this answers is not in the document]
almann: Again, I don't feel super strong about it, it just did trigger a code smell to me.

src/types/mod.rs
almann: No biggie--I was just wondering.

ion-hash/src/representation.rs
marcbowes: Done

ion-hash/src/representation.rs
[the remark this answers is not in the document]
marcbowes: The upcoming struct variant might make this more clear..

ion-hash/src/representation.rs
```diff
@@ -66,36 +82,45 @@ fn write_repr_integer<U: Update>(value: Option<&AnyInt>, hasher: &mut UpdateEsca
...
 }
 
+/// Ion hash defines representations for some special values.
+struct Floats;
+impl Floats {
+    const NEGATIVE_ZERO: [u8; 8] = [0x80, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00];
+    const POSITIVE_INFINITY: [u8; 8] = [0x7F, 0xF0, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00];
+    const NEGATIVE_INFINITY: [u8; 8] = [0xFF, 0xF0, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00];
+    const NAN: [u8; 8] = [0x7F, 0xF8, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00];
+}
```
almann, line 92: I don't mind the namespacing, but I'd be cool with just keeping them flat in the module.

ion-hash/src/type_qualifier.rs
```diff
@@ -76,8 +76,8 @@ impl TypeQualifier {
             /*IonType::Clob => Clob,
             IonType::Blob => Blob,
             IonType::List => List,
-            IonType::SExpression => SExpression,
-            IonType::Struct => Struct,*/
+            IonType::SExpression => SExpression,*/
+            IonType::Struct => combine(Struct, 0),
```
almann, line 80: Shouldn't this have a `Some`/`None` type of check too for the `null` case (is this tested or am I missing something)?  Also to be consistent with the other code `0x00` as the literal is probably warranted.

almann (approved):
Minor concern around some `struct` related concern, but otherwise :ship:

ion-hash/src/type_qualifier.rs
marcbowes: I meant to back this out.

PR #270, opened 2021-05-28
Implement IonHash support for structs

marcbowes:
Quite a lot going on in this commit! Structs are different from the
scalars we've been hashing so far in that they create **new hashers**:

> escape(concat(sort(H(field1), H(field2), ..., H(fieldn))))

The big `H` there is an actual hash computation.

This immediately means that `IonHasher` not be a useful abstraction
since the requirement to be able to make new hashers means that asking
for the initial hasher is a little silly. So, in this commit I stripped
that struct away. This creates a bit of weirdness in lib.rs where there
are a bunch of functions that take a mutable hasher as their second
argument. My intention is to add some other abstraction (maybe a trait)
in a future commit. [fn:1]

The next bump in the road was that all our abstractions coupled
"dispatch" and "implementation", where the former means "look at the Ion
Type then do something specific for that type". This posed a problem
when computing the field hash: the key (of a field) is a Symbol and a
Symbol is *not an Element*. To fix this, I decoupled dispatch and
implementation, allowing the field hash computation to go directly to
the TQ/repr infrastructure for symbols.

The biggest bump by *far* was yet more trouble with the ion-hash-test
suite. Using a real hasher (sha256) became extremely tricky with the
field hashers because the expectations don't show that. For example:

```ion
{
  ion:{'':1},              // empty symbol
    // ^^^^ NOTE: this field requires a *new hasher*
  expect:{ identity:(
    update::(0x0b) update::(0xd0)
    // --------LOOK DOWN! -------------------
        update::(0x0b) update::(0x70) update::(0x0e)
        update::(0x0b) update::(0x20) update::(0x01) update::(0x0e)
        digest::(0x0b 0x70 0x0e 0x0b 0x20 0x01 0x0e)
    // ^ this idented sequence is on a new instance of the hasher
    // --------LOOK UP! ---------------------
      update::(0x0c 0x0b 0x70 0x0c 0x0e 0x0c 0x0b 0x20 0x01 0x0c 0x0e)
    update::(0x0e)
    digest::(0x0b 0xd0 0x0c 0x0b 0x70 0x0c 0x0e 0x0c 0x0b 0x20 0x01 0x0c 0x0e 0x0e)
  ) }
}
```

So, I've gone back to an `IdentityHasher` with fixed size output and
worked around the trailing zeros.

A few (3) of the struct tests fail because structs also test other types
that we haven't implemented. I added an ignore list so that we can have
all the struct tests in this file and clearly show which aren't yet
expected to pass.

[fn:1]: The struct representation has some duplicate sequence code that
I plan to fix at the same time.

-------

*Issue #, if available:*

*Description of changes:*


ion-hash/src/lib.rs
```diff
@@ -57,62 +55,53 @@ impl Markers {
 /// This is the reason for this is we can't go from [`Digest`] to these
 /// traits (e.g. [`Update`]) but because of a blanket implementation, we can go
 /// the other way.
-pub struct IonHasher<D>
+pub fn hash_element<E, D>(elem: &E) -> IonResult<Output<D>>
```
marcbowes, line 58: Note this can fail because we couldn't find the text for a symbol. Even though I haven't actually implemented symbols, I think this is the right API.

ion-hash/src/representation.rs
```diff
@@ -57,13 +67,20 @@ impl<'a, D> Update for EscapingDigest<'a, D>
...
             if let Markers::B | Markers::E | Markers::ESC = *byte {
-                self.0.update([0x0C]);
+                escaped.extend(&[Markers::ESC, *byte]);
+            } else {
+                escaped.extend(&[*byte]);
             }
-            self.0.update([*byte]);
         }
+
+        self.0.update(escaped);
```
marcbowes, line 83: I'm not very happy with this. It's not obvious, but the tests made me do this (so I can see the full update sequence and skip trailing zeros).

ion-hash/tests/ion_hash_tests.rs
```diff
@@ -1,12 +1,109 @@
...
+        // does call update with trailing zeros (e.g. float negative zero). So
+        // we have a problem: when should we use the data verbatim and when
+        // should we strip the trailing zeros that are *only there because of
+        // the fixed output trait*? The hack is to strip trailing zeroes if we
+        // get called with something that looks like it came from the result of
+        // another IdentityDigest. Note that `>=` is used here due to escaping
+        // (which ironically, means the output isn't fixed size).
+        let data = data.as_ref();
+        let actual_update = if data.len() >= IdentityDigest::output_size() {
+            without_trailing_zeros(data)
```
marcbowes, line 34: 😅

ion-hash/src/type_qualifier.rs
```diff
@@ -7,11 +7,16 @@ use std::slice;
...
-    value::{AnyInt, Element},
+    value::{AnyInt, Element, Sequence, Struct, SymbolToken},
     IonType,
 };
 use num_bigint::Sign;
 
+// For many types, the qualifier is either 'null' or 'not null'. That's what
+// these constants are for!
+const QUALIFIER_NOT_NULL: u8 = 0x00;
+const QUALIFIER_NULL: u8 = 0x0F;
```
almann, line 18: I think you need `0x01` for the `$0` case, also `bool` special cases these (`0x0` for `false` and `0x1` for `true`)

ion-hash/src/type_qualifier.rs
```diff
@@ -30,55 +35,27 @@ impl TypeQualifier {
...
-            _ => todo!(),
+            IonType::Timestamp => Timestamp,*/
+            IonType::Symbol => type_qualifier_symbol(elem.as_sym()),
+            IonType::String => type_qualifier_string(elem.as_str()),
+            IonType::Clob => type_qualifier_clob(elem.as_bytes()),
+            IonType::Blob => type_qualifier_blob(elem.as_bytes()),
+            IonType::List => type_qualifier_list(elem.as_sequence()),
+            IonType::SExpression => type_qualifier_sexp(elem.as_sequence()),
+            IonType::Struct => type_qualifier_struct(elem.as_struct()),
+            other => unimplemented!("TypeQualifier is not implemented for {}", other),
```
almann, line 58: Would it make sense to handle the dominant cases more explicitly?  E.g. if [`elem.is_null()`][1] the `Q` nibble is `0xF` full stop.  For the boolean case, deal with `true`/`false`, for the `$0` case you have the special `Q` of `0x1`, and for everything else it is `Q` of `0xF`.

Put another way, I think you can do a `const fn` to table derive the type to `0xT0` where `T` is the type code and more simply generate the low nibble `0x0Q` where `Q` is the qualifier and just bitwise-or them together.

I *think* this simplifies the logic, but maybe I am missing something.

[1]: https://docs.rs/ion-rs/0.6.0/ion_rs/value/trait.Element.html#tymethod.is_null

ion-hash/tests/ion_hash_tests.rs
almann: Oof.. yeah this is, pretty special.  I thought you were going to use SHA2 to derive the tests from identity?  Is that in a later PR?  If so doesn't that solve this problem all together?

almann (commented):
Some questions below.

ion-hash/tests/ion_hash_tests.rs
marcbowes: I had to undo that because it's nearly impossible to get right with subhashers

ion-hash/src/representation.rs
```diff
@@ -6,40 +6,50 @@
...
 // representations can be written by the "raw" writer (ref. the Java
 // implementation).
-pub(crate) fn update_with_representation<E: Element + ?Sized, D: Update + FixedOutput + Default>(
-    elem: &E,
-    hasher: &mut D,
-) {
+pub(crate) fn update_with_representation<E, D>(elem: &E, hasher: &mut D) -> IonResult<()>
+where
+    E: Element + ?Sized,
+    D: Update + FixedOutput + Reset + Clone + Default,
```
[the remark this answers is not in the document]
marcbowes: The `D: ...` sequence is just the subtraits of the `Digest` trait.

ion-hash/src/representation.rs
```diff
@@ -120,9 +137,99 @@ fn write_repr_float<D: Update>(value: Option<f64>, hasher: &mut EscapingDigest<'
...
+    S: SymbolToken + ?Sized,
+{
+    match value {
+        Some(s) => {
+            match s.text() {
+                Some(s) => write_repr_string(Some(s), hasher),
+                None => {
+                    // TODO: Throw an error if this is not $0
+                    {}
+                }
```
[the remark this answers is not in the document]
marcbowes: Will change to a FIXME for this PR.

ion-hash/src/representation.rs
```diff
@@ -120,9 +137,99 @@ fn write_repr_float<D: Update>(value: Option<f64>, hasher: &mut EscapingDigest<'
...
-        Some(s) => hasher.update(s.as_bytes()),
+        Some(s) if s.len() > 0 => hasher.update(s.as_bytes()),
+        _ => {}
+    }
+}
+
+fn write_repr_blob<D: Update>(value: Option<&[u8]>, hasher: &mut EscapingDigest<'_, D>) {
+    match value {
+        Some(bytes) => hasher.update(bytes),
         None => {}
```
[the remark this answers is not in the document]
marcbowes: It actually doesn't matter in terms of the value, but I'll add it anyways.

The reason for it was simply debugging calls to `update` - the empty slice calls were annoying me.

ion-hash/src/type_qualifier.rs
marcbowes: Ok

ion-hash/src/type_qualifier.rs
marcbowes: That's kind of how I started and the requirement to call into specific variants (like tq-symbol for the struct case) made me rethink it. I quite like that this method "is just dispatch and no implementation".

ion-hash/tests/ion_hash_tests.rs
marcbowes: btw I explained this is the commit message :) not sure if you saw that.

ion-hash/src/type_qualifier.rs
almann: I guess--but it just leads to a bunch of seemingly similar code paths that are not particularly differentiated nor illuminating, but I don't feel particularly strong about it if you find this code to be more readable that the (seemingly) more concise alternative.

ion-hash/tests/ion_hash_tests.rs
almann: Re-reading the PR summary, yes, that explains it...

I think we could have the test data be more structural to make this doable, but that is a larger change so I get trying to make this work.

marcbowes: So does that mean you're OK with the code as is, or should I make some changes?

ion-hash/src/lib.rs
```diff
@@ -57,62 +55,53 @@ impl Markers {
...
+}
 
-    fn hash_type<E: Element + ?Sized>(&mut self, elem: &E) {
-        let tq = TypeQualifier::from_element(elem);
-        self.hasher.update(tq.as_bytes());
-    }
+fn update_type_qualifier<E: Element + ?Sized, U: Update>(elem: &E, hasher: &mut U) {
+    let tq = TypeQualifier::from_element(elem);
+    hasher.update(tq.as_bytes());
+}
```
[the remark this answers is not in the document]
marcbowes: I will do that as soon as I'm done fleshing out the type support. Otherwise the multiple inflight PRs become a nightmare to rebase.

ion-hash/src/type_qualifier.rs
marcbowes: I made this change and don't like it. It doesn't enable code-reuse and I find the indirection annoying. Happy to actually do this, but I personally don't like it.

marcbowes: Constants are useful to give things names (make intention clear) or for re-use. I don't think defining single-use constants here helps with either. The intentional is already clear.

ion-hash/src/type_qualifier.rs
marcbowes: I'm going to leave it as is. The ability to call into specific representations is required for structs, and I don't think we should special case symbols.

ion-hash/tests/ion_hash_tests.rs
almann: I am okay, but not particularly happy with it.  I mean really we're fighting the testing file and the way the `Digest` trait works.  I think this warrants a TODO around making these tests not wag the dog, but I don't think that belongs in this PR.

almann (approved):
:ship:

ion-hash/src/representation.rs
```diff
@@ -120,9 +137,99 @@ fn write_repr_float<D: Update>(value: Option<f64>, hasher: &mut EscapingDigest<'
...
+    S: SymbolToken + ?Sized,
+{
     match value {
-        Some(s) => hasher.update(s.as_bytes()),
+        Some(s) => {
+            match s.text() {
+                Some(s) => write_repr_string(Some(s), hasher),
+                None => {
+                    // FIXME: Throw an error if this is not $0
+                    {}
```
zslayton, line 151: Since this handles neither `$0` nor non-`$0`, this might be a good place to add a `todo!()` so any tests that exercise this code path will fail noisily.
```suggestion
                    todo!("hash SymbolToken without text")
```

ion-hash/src/representation.rs
```diff
@@ -120,9 +137,99 @@ fn write_repr_float<D: Update>(value: Option<f64>, hasher: &mut EscapingDigest<'
...
+
+/// Iterates over a `Struct`, computing the "field hash" of each field
+/// (key-value pair). The field hash is defined in the spec as:
+///
+/// ```text
+/// H(field) -> h(s(fieldname) || s(fieldvalue))
+/// ```
+///
+/// The resulting `Vec` is not sorted (i.e. is in the same order as the field
+/// iterator).
```
zslayton, line 207: I think some amount of this comment actually applies to `write_repr_struct` above.

ion-hash/src/representation.rs
```diff
@@ -120,9 +137,99 @@ fn write_repr_float<D: Update>(value: Option<f64>, hasher: &mut EscapingDigest<'
...
+fn struct_field_hash<D, F, E>(key: &F, value: &E) -> IonResult<Output<D>>
+where
+    D: Update + FixedOutput + Reset + Clone + Default,
+    F: SymbolToken + ?Sized,
+    E: Element + ?Sized,
+{
+    let mut hasher = D::default();
```
zslayton, line 214: I notice the `Reset` trait requirement in the signature above. Could `write_repr_struct` create a single instance of `D`, pass it into `struct_field_hash` and then reset it between fields/calls?

marcbowes: I think it could.

My old manager used to tell me "make it work, make it fast". I love approaching work like that. So, my overall thinking is that once we have the features fleshed out, we can do some benchmarking and make sure that the 1.0 release is both feature and performance ready.

ion-hash/src/representation.rs
marcbowes: Hah, woops! Fixed.

ion-hash/src/representation.rs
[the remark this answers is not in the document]
almann: Have we come full circle here and proposed [`Digest`](https://docs.rs/digest/0.9.0/digest/trait.Digest.html).  I realize it is slightly different than `Digest` (using the bounds on `Self`) rather than a parallel API, but the blanket implementation is the same, so by definition we've kind of gone that route...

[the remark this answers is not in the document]
marcbowes: I think it's different because the current decomposition allows for impls of just the bits we need. Either way, I will clean this up in a future PR.

PR #272, opened 2021-05-29
Refactor binary timestamps

marcbowes:
*Description of changes:*

These commits refactor binary timestamp encoding out of writer.rs so that they can be used by IonHash. While we're at it, we make the coder precision-aware.


src/binary/timestamp.rs
```diff
@@ -0,0 +1,98 @@
...
+    fn write_timestamp(&mut self, value: &Timestamp) -> IonResult<()>;
+}
+
+impl<W> TimestampBinaryExt for W
+where
+    W: Write,
+{
+    /// Write a timestamp. Encodes the timestamp and prepends an appropriate
+    /// type decriptor.
+    fn write_timestamp(&mut self, value: &Timestamp) -> IonResult<()> {
```
almann, line 78: I think returning `IonResult<usize>` would be more helpful.

Should this take `Option<&Timestamp>` to capture the null case more explicitly?

almann (changes requested):
Thanks for starting to factor these APIs out.  I think they are exposing some interesting API design elements, and I would like to get @zslayton's take just to be sure I am not missing anything in the original factoring.

The `binary::timestamp` which has a separate inherent `impl` block for serializing Timestamp into an owned buffer (using a `SmallVec` as the owned wrapper for a slice that holds the encoded data).  We either need a cheap conversion from `DateTime<FixedOffset>` to `Timestamp` struct, but I wonder if this serializer is just a generic function that takes a [`DateLike`][1] `+` [`TimeLike`][2] and we implement those traits on our `Timestamp` structure (I am okay with making this a TODO with an issue if you agree, but don't want to conflate it in this PR, though I think it germane).  Because of Ion semantics, these two traits might not be enough, and we might need a third trait to capture Ion related semantics around precision, and negative zero offset (`IonDateTimeLike`?).

I like the `TimestampBinaryExt` trait idea (Maybe call this `TimestampWrite` or something), given this trait, do we even need even need the above separate function?  We could have `write_timestamp_value` that does the encoding, and `write_timestamp` that does the type + value encoding.  If those APIs returned length, given a blanket implementation on `Write`, they would just work on `Vec` and `&[u8]` inherently.  Furthermore, if they can be implemented on `DateLine + TimeLike + IonDateTimeLike` that makes them quite flexible.

Anyhow, maybe this is excessive, but I think the factoring will be the template for the public API of these primitives so I think it pretty important.

[1]: https://docs.rs/chrono/0.4.19/chrono/trait.Datelike.html
[2]: https://docs.rs/chrono/0.4.19/chrono/trait.Timelike.html

src/binary/timestamp.rs
```diff
@@ -0,0 +1,98 @@
...
+};
+
+/// Extends [`Timestamp`] to support [Ion binary].
+///
+/// [Ion binary]: https://amzn.github.io/ion-docs/docs/binary.html#6-timestamp
+impl Timestamp {
+    const BUFFER_SIZE: usize = 16;
+
+    /// NOTE: Currently, this function always encodes with nanosecond precision.
+    pub fn encode_binary(&self) -> IonResult<SmallVec<[u8; Timestamp::BUFFER_SIZE]>> {
```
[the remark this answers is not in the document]
almann: FWIW, these small stack allocated buffers is *maybe* better modeled as [`arrayvec`](https://crates.io/crates/arrayvec).  To be honest, I don't think it is going to make a sizeable difference, but the one benefit is that we don't actually need the heap fallback code so in theory it is a bit faster and closer to the intent.  I definitely think either this or `smallvec` can be used to eliminate the `u64` to compact bytes view code we have.

src/binary/writer.rs
```diff
@@ -519,54 +520,11 @@ impl<W: Write> BinarySystemWriter<W> {
...
-                type_descriptor = 0x6E;
-                enc_buffer.push(type_descriptor);
-                VarUInt::write_u64(enc_buffer, encoded_length as u64)?;
-            }
-            let raw_buffer = writer.into_inner().into_inner();
-            enc_buffer.extend_from_slice(&raw_buffer[..encoded_length]);
-            Ok(())
+            // TODO: Avoid the requirement to clone to make a `Timestamp`.
+            let timestamp = value.clone().into();
+            enc_buffer.write_timestamp(&timestamp)
```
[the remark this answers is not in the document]
zslayton: I agree in principle, but do note [that `DateTime<T>` implements `Copy`](https://docs.rs/chrono/0.4.19/chrono/struct.DateTime.html#impl-Copy). This clone should be trivial.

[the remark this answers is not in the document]
marcbowes: Ah, thanks! I'll leave a comment.

src/binary/timestamp.rs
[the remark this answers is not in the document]
marcbowes: Ok! I'll try out arrayvec.

src/binary/timestamp.rs
```diff
@@ -0,0 +1,98 @@
...
+use crate::{
+    binary::{uint::DecodedUInt, var_int::VarInt, var_uint::VarUInt, writer::MAX_INLINE_LENGTH},
+    result::IonResult,
+    types::timestamp::{Precision, Timestamp},
+};
+
+/// Extends [`Timestamp`] to support [Ion binary].
+///
+/// [Ion binary]: https://amzn.github.io/ion-docs/docs/binary.html#6-timestamp
+impl Timestamp {
```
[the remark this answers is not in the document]
marcbowes: I can move it all into the single trait and try come up with a new name :)

src/binary/timestamp.rs
marcbowes: Ok!

src/binary/timestamp.rs
```diff
@@ -0,0 +1,98 @@
...
+
+        // Ion encodes offsets in minutes while DateTime stores it in seconds.
+        let offset_seconds = self.offset.map(|o| o.utc_minus_local()).unwrap_or(0);
+        let offset_minutes = (offset_seconds as f32 / SECONDS_PER_MINUTE).round() as i64;
+        VarInt::write_i64(&mut writer, offset_minutes)?;
+
+        VarUInt::write_u64(&mut writer, utc.year() as u64)?;
+        VarUInt::write_u64(&mut writer, utc.month() as u64)?;
+
+        // So far, we've written required fields. The rest are optional!
```
zslayton, line 39: Month [is an optional field](https://amzn.github.io/ion-docs/docs/binary.html#6-timestamp) as well:

> Timestamp representations have 7 components, where 5 of these components are optional depending on the precision of the timestamp. The 2 non-optional components are offset and year. The 5 optional components are (from least precise to most precise): month, day, hour and minute, second, fraction_exponent and fraction_coefficient. All of these 7 components are in Universal Coordinated Time (UTC).

src/binary/timestamp.rs
```diff
@@ -0,0 +1,98 @@
...
+    types::timestamp::{Precision, Timestamp},
+};
+
+/// Extends [`Timestamp`] to support [Ion binary].
+///
+/// [Ion binary]: https://amzn.github.io/ion-docs/docs/binary.html#6-timestamp
+impl Timestamp {
+    const BUFFER_SIZE: usize = 16;
+
+    /// NOTE: Currently, this function always encodes with nanosecond precision.
```
zslayton, line 21: This comment originated on the `write_datetime` method:

https://github.com/amzn/ion-rust/blob/0e3d46ee0b115c496a5192a0b2bb124c8717a9d7/src/binary/writer.rs#L521-L524

It made a bit more sense there because a `DateTime<FixedOffset>` always has all of the time unit fields up to and including fractional seconds.

However, `Timestamp`s can have a precision of `Year` or `Month`, so I think we need a bit of clarification:

```suggestion
    /// NOTE: If the Timestamp being encoded has fractional seconds, they will be written as nanoseconds.
```

src/binary/timestamp.rs
```diff
@@ -0,0 +1,98 @@
...
+        if self.precision > Precision::Month {
+            VarUInt::write_u64(&mut writer, utc.day() as u64)?;
+            if self.precision > Precision::Day {
+                VarUInt::write_u64(&mut writer, utc.hour() as u64)?;
+                VarUInt::write_u64(&mut writer, utc.minute() as u64)?;
+                if self.precision > Precision::HourAndMinute {
+                    VarUInt::write_u64(&mut writer, utc.second() as u64)?;
+                }
+                if self.precision > Precision::Second {
+                    // This implementation always writes the datetime out with nanosecond precision.
```
zslayton, line 49:
```suggestion
                    // This implementation always writes the fractional seconds out with nanosecond precision.
```

src/binary/timestamp.rs
zslayton: > Should this take Option<&Timestamp> to capture the null case more explicitly?

`ion-rust` hasn't been using this pattern elsewhere. You'd write a `null.timestamp` with `write_null(IonType::Timestamp)`.

src/binary/timestamp.rs
```diff
@@ -0,0 +1,98 @@
...
+    }
+}
+
+pub trait TimestampBinaryExt {
+    fn write_timestamp(&mut self, value: &Timestamp) -> IonResult<()>;
+}
+
+impl<W> TimestampBinaryExt for W
+where
+    W: Write,
```
zslayton, line 74: Whatever documentation we produce for this trait should mention that writing a single value to a type that implements `Write` isn't guaranteed to produce a legal Ion stream. (e.g. Encoding a `Timestamp` won't emit a leading IVM; encoding a `Symbol` won't produce a local symbol table with the corresponding text.)

src/types/timestamp.rs
```diff
@@ -75,10 +75,10 @@ fn number_of_digits(value: u32) -> u32 {
...
 #[derive(Debug, Clone)]
 pub struct Timestamp {
-    date_time: NaiveDateTime,
-    offset: Option<FixedOffset>,
-    precision: Precision,
-    fractional_seconds: Option<Mantissa>,
+    pub(crate) date_time: NaiveDateTime,
+    pub(crate) offset: Option<FixedOffset>,
+    pub(crate) precision: Precision,
+    pub(crate) fractional_seconds: Option<Mantissa>,
```
zslayton, line 81: I think these are only being made visible because you added an inherent impl to a different module. Is that right?

src/binary/timestamp.rs
marcbowes: Whoops. Thanks.

src/types/timestamp.rs
marcbowes: Yeah. Even if it's not an inherent impl (e.g. an ext trait) we'd need access. Open to other ideas..

almann: This suggests that implementing `TimeLike` and `DateLike` (as well as some hypothetical `IonTimestampLike` or some such) are probably in order for the API.

Temporarily to work around the fact that we do not expose a useful API, this seems okay.

src/binary/timestamp.rs
```diff
@@ -18,7 +18,7 @@ use crate::{
 impl Timestamp {
     const BUFFER_SIZE: usize = 16;
 
-    /// NOTE: Currently, this function always encodes with nanosecond precision.
+    /// NOTE: If the Timestamp being encoded has fractional seconds, they will be written as nanoseconds.
```
almann, line 21: So this means that if there exists sub-second precision, that precision is always nine digits (i.e. nanoseconds)?

src/types/timestamp.rs
marcbowes: I can open an issue to track that?

src/binary/timestamp.rs
marcbowes: Yeah, for now.

My overall philosophy right now is I'm trying to extract/de-duplicate code between ion-rust and ion-hash. (So, the nanosecond precision caveat was already there.) Happy to open and fix these issues later, but given how little time I have to work on this (and likewise, your review time) I'm trying to take babysteps.

almann: No worries--just making sure I understood the comment.

src/types/timestamp.rs
almann: Please :D

almann: Did you one better and added #273.

src/binary/timestamp.rs
almann: Added #274.

src/binary/writer.rs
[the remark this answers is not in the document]
almann: I think #273 can resolve this and avoid the copy altogether.

src/binary/timestamp.rs
marcbowes: Ok! I'll try add something. Would be nice if we could come up with some terminology here so it's kind of well understood what this thing is (and therefore, hopefully, what it isn't). Something like "binary codec"?

src/binary/timestamp.rs
```diff
@@ -0,0 +1,98 @@
...
+                    const SCALE: i64 = -9; // "Scale" (used by BigDecimal) is -1 * exponent
+                    VarInt::write_i64(&mut writer, SCALE)?;
+                    DecodedUInt::write_u64(&mut writer, nanoseconds as u64)?;
+                }
+            }
+        }
+
+        let encoded_length = writer.get_ref().position() as usize;
+        let slice = &writer.into_inner().into_inner()[..encoded_length];
+        let encoded = SmallVec::from(slice);
```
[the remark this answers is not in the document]
marcbowes: Fixed by using arrayvec as per your other suggestion.

src/binary/timestamp.rs
```diff
@@ -1,95 +1,85 @@
...
+
 /// Extends [`Timestamp`] to support [Ion binary].
 ///
 /// [Ion binary]: https://amzn.github.io/ion-docs/docs/binary.html#6-timestamp
-impl Timestamp {
-    const BUFFER_SIZE: usize = 16;
+pub trait TimestampBinaryExt {
+    fn encode_binary(&self) -> IonResult<ArrayVec<u8, { ENCODED_BUFFER_SIZE }>>;
+    fn write_binary<W: Write>(&self, sink: &mut W) -> IonResult<()>;
+}
```
almann, line 22: There are two things going on in this trait that seem like a strange conflation of responsibilities.

One is the ability to encode a `Timestamp` into some octets, and the other is to write the `Timestamp` with its type header to some `Write` sink.

I might frame this a little differently:

```rust
pub const MAX_TIMESTAMP_LENGTH = 16;

/// Provides the ability to serialize [`Timestamp`] to some kind of sink.
pub trait TimestampBinaryEncoder {
    /// Encodes the content of a [`Timestamp`] as per the Ion binary encoding.
    /// Returns the length of the encoded bytes.
    ///
    /// This does not encode the type descriptor nor the associated length.
    /// Prefer [`TimestampBinaryEncoder::encode_timestamp_value`] for that.
    fn encode_timestamp(&mut self, timestamp: &Timestamp) -> IonResult<usize>;

    /// Encodes a [`Timestamp`] as an Ion value with the type descriptor and length.
    /// Returns the length of the encoded bytes.
    fn encode_timestamp_value(&mut self, timestamp: &Timestamp) -> IonResult<usize>;
}

impl<T> TimestampBinaryEncoder for T where T: Write {
    ...
}
```

What this factoring gives us is the ability to write to a buffer or sink of our choosing without requiring an intermediate:

```rust
let mut sink: ArrayVec<u8, MAX_TIMESTAMP_LENGTH> = ArrayVec::new_const();
sink.encode_timestamp(&some_ts);
...
```

I realize the intermediate is necessary for the writing of the full value, but that is in some ways an implementation detail and this API hides that.

This also will work nicely as a pattern for the other types we'll expose, so we can model them each as their own traits and pull them together via a trait that is constrained by all the typed encoders and blankets over any type that implements all the specific encoders which all `Write` types will.

Note also that arbitrary precision timestamps will probably make `MAX_TIMESTAMP_LENGTH` into something like `TYPICAL_TIMESTAMP_LENGTH` and necessitate the use of `SmallVec` or some such, but that is an exercise for another day.

marcbowes: Very nice. Thanks. I made those changes.

Note that the usize return isn't yet possible due to other changes that need to happen. For now, I left a TODO (in this PR). If we can settle on these APIs, I will also open an issue to track that.

src/binary/timestamp.rs
```diff
@@ -11,75 +11,89 @@ use crate::{
     types::timestamp::{Precision, Timestamp},
 };
 
-const ENCODED_BUFFER_SIZE: usize = 16;
+const MAX_TIMESTAMP_LENGTH: usize = 16;
 
 /// Extends [`Timestamp`] to support [Ion binary].
```
almann, line 16: Minor doc comment suggestion:

```suggestion
/// Provides support to write [`Timestamp`] into [Ion binary].
```

src/binary/timestamp.rs
```diff
@@ -11,75 +11,89 @@ use crate::{
...
+        let utc = timestamp.date_time;
 
         // Ion encodes offsets in minutes while DateTime stores it in seconds.
-        let offset_seconds = self.offset.map(|o| o.utc_minus_local()).unwrap_or(0);
+        let offset_seconds = timestamp.offset.map(|o| o.utc_minus_local()).unwrap_or(0);
         let offset_minutes = (offset_seconds as f32 / SECONDS_PER_MINUTE).round() as i64;
-        VarInt::write_i64(&mut sink, offset_minutes)?;
-        VarUInt::write_u64(&mut sink, utc.year() as u64)?;
+        VarInt::write_i64(self, offset_minutes)?;
+        VarUInt::write_u64(self, utc.year() as u64)?;
```
almann, line 49: Not related to this change, but I think we should use a similar API pattern for these "encoders".  This is a reminder for me to cut an issue on that.

almann (approved):
Looks great!  :ship:

src/binary/timestamp.rs
almann: Cut #277 to track this.

marcbowes: Thanks. I left a note in it too re: their return value.

PR #275, opened 2021-06-02
Use delegate-attr to replace delegate

upsuper:
*Issue #, if available:*

*Description of changes:* use `delegate-attr` macro to replace `delegate`, which should have a nicer syntax.


zslayton:
Hi, thanks for the PR! At a glance, it seems like these libraries are a wash for our use case; the bulk of this PR removes a level of indentation. Are there other benefits to using the [`delegate-attr`](https://docs.rs/delegate-attr/0.2.9/delegate_attr/) crate over [`delegate`](https://docs.rs/delegate/0.5.1/delegate/) that we should consider?

zslayton:
Our `cargo fmt` check is failing for the current commit.

upsuper:
> Are there other benefits to using the [`delegate-attr`](https://docs.rs/delegate-attr/0.2.9/delegate_attr/) crate over [`delegate`](https://docs.rs/delegate/0.5.1/delegate/) that we should consider?

I think this way the code looks cleaner, and more friendlier to things like code formatter, as it uses attribute macro rather than function-style macro, for which the formatter can't really know what to do.

> Our `cargo fmt` check [is failing for the current commit](https://github.com/amzn/ion-rust/pull/275/checks?check_run_id=2729367716).

Hmmm, it's very interesting. I think this is a bug of rustfmt, and I'm raising a [PR](https://github.com/rust-lang/rustfmt/pull/4858) to fix it. In the mean time I think replacing `;` with `{}` would work around this bug. I'll push a new commit to fix it.

zslayton:
As this achieves the same thing as our current code but with slightly different syntax, I'm inclined to leave our code as-is. If `delegate-attr` eventually offers other more objective benefits (for example: better performance), I'd be happy to revisit it.

upsuper:
I don't think there will be much performance difference as it's just a proc macro... I would note that I think delegate-attr should have a better test coverage, though.

Issue #190, opened 2021-04-17
Timestamp Builder API Mismatch


almann:
The [code][1] in the builder that builds a `Timestamp` seems to contradict the `build_at_offset` builder API [here][2]:

```rust
    pub fn build_at_offset(mut self, offset_minutes: i32) -> IonResult<Timestamp> {
        self.builder.offset = Some(offset_minutes);
        self.into_builder().build()
    }
```

The parameter above is minutes, but the corresponding code that builds the `FixedOffset`:

```rust
            let offset = FixedOffset::west_opt(offset).ok_or_else(|| {
                illegal_operation_raw(format!("specified offset ('{}') is invalid", offset))
            })?;
```

So for the `UTC-08:00` offset, one would need to pass `28800` to the `build_at_offset` which doesn't seem right since the parameter indicates minutes.  My expectation would be that `-480` would be the argument--at a minimum we should be clarifying what this parameter means.

[1]: https://github.com/amzn/ion-rust/blob/1347af2b31fd43b0b6adff99a15c693ab75d1253/src/types/timestamp.rs#L337-L340
[2]: https://github.com/amzn/ion-rust/blob/1347af2b31fd43b0b6adff99a15c693ab75d1253/src/types/timestamp.rs#L470-L473

zslayton:
As discussed offline, this was indeed a mistake. I started writing the API with `minutes` (as are used in the Ion spec) and then drifted into using `seconds` because that's what `chron::FixedOffset` used. I also meant to use `east_opt` rather than `west_opt` so that timezones in the western hemisphere would take a negative number of minutes.

PR #279, opened 2021-06-03
Wire up timestamp support

marcbowes:
This commit wires up ion-hash to the newly factored binary/timestamp.rs
encoder. We copy over the tests from the suite. They fail:

```
thread 'ion_hash_tests' panicked at 'assertion failed: `(left == right)`
  left: `[11, 96, 192, 15, 225, 14]`,
 right: `[11, 96, 99, 128, 15, 225, 14]`: test case 2017T failed', ion-hash\tests\ion_hash_tests.rs:152:5
```

Turns out there were two bugs related to offsets in ion-rust. Those are
now fixed!

The binary timestamp tests (10n) are not part of this commit because
they require a bit more machinery.

-------

*Issue #, if available:*

*Description of changes:*


ion-hash/tests/ion_hash_tests.rs
```diff
@@ -150,8 +150,8 @@ fn test_case<E: Element>(ion: &E, strukt: &E) -> IonResult<()> {
     // variable sized result. Without this, any test failure will write lots of
     // stuff to your console which can be annoying since it takes forever.
     assert_eq!(
-        without_trailing_zeros(&expected[..]),
-        without_trailing_zeros(&result[..]),
+        format!("{:02x?}", without_trailing_zeros(&expected[..])),
+        format!("{:02x?}", without_trailing_zeros(&result[..])),
```
marcbowes, line 154: This change makes debugging easier because the failures look like the hex in the test files.

src/binary/timestamp.rs
```diff
@@ -42,10 +42,18 @@ where
...
 
-        // Ion encodes offsets in minutes while DateTime stores it in seconds.
-        let offset_seconds = timestamp.offset.map(|o| o.utc_minus_local()).unwrap_or(0);
-        let offset_minutes = (offset_seconds as f32 / SECONDS_PER_MINUTE).round() as i64;
-        VarInt::write_i64(self, offset_minutes)?;
+        // Write out the offset (minutes difference from UTC). If the offset is
+        // unknown, negative zero is used.
+        if let Some(offset) = timestamp.offset {
+            // Ion encodes offsets in minutes while DateTime stores it in seconds.
+            let offset_seconds = offset.local_minus_utc();
```
marcbowes, line 49: ⚠️⚠️⚠️

I changed from `utc_minus_local()` to `local_minus_utc()`. I think this is right, but please check me.

ion-hash/tests/ion_hash_tests.rs
```diff
@@ -85,6 +85,9 @@ const IGNORE_LIST: &[&'static str] = &[
     r#"{a:{d:[1,2,3],b:{c:5}},e:6}"#,
     // Uses md5 (not identity)
     r#"{Metrics:{'Event.Catchup':[{Value:0,Unit:ms}],'FanoutCache.Time':[{Value:1,Unit:ms}]}}"#,
+    // timestamps, broken because we always write out fractional seconds as nanoseconds
+    r#"2000-01-01T00:00:00.0Z"#,
```
marcbowes, line 89: I will fix these in a future PR.

ion-hash/src/representation.rs
```diff
@@ -145,6 +167,21 @@ where
...
+fn write_repr_timestamp<D>(
+    value: Option<&Timestamp>,
+    hasher: &mut EscapingDigest<'_, D>,
+) -> IonResult<()>
+where
+    D: Update + FixedOutput + Reset + Clone + Default,
+{
+    match value {
+        None => {}
+        Some(timestamp) => hasher.encode_timestamp(timestamp)?,
```
almann, line 179: Wow, this worked out very nicely!

ion-hash/tests/ion_hash_tests.rs
almann: Also there is [`pretty_hex`](https://docs.rs/pretty-hex/0.2.1/pretty_hex/index.html) which I use in the element round tripping tests IIRC, which is handy (no need to change this, just an FYI).

src/binary/timestamp.rs
almann: I think this is related to #190 it is encoded the "wrong way" in our `Timestamp` type.

almann (approved):

ion-hash/tests/ion_hash_tests.rs
```diff
@@ -119,19 +122,50 @@ fn test_file(file_name: &str) -> IonResult<()> {
...
+                let value = seq_to_bytes(binary);
+                // TODO: the binary IVM is pub(crate) in ion-rust
+                let mut bytes = vec![0xE0, 0x01, 0x00, 0xEA];
+                bytes.extend(value);
+
+                let loaded = element_reader().read_all(&bytes)?;
+                let elem = loaded
+                    .into_iter()
+                    .nth(0)
+                    .expect("10n test case should have a single element");
```
zslayton, line 152: Nit: if I encountered this error message, I would assume the test case had unexpectedly had more than one element. In reality, there weren't any elements at all.

ion-hash/tests/ion_hash_tests.rs
```diff
@@ -119,19 +122,50 @@ fn test_file(file_name: &str) -> IonResult<()> {
...
+                test_case(&elem, expect)?;
+            }
+            _ => unreachable!("test cases have one of `ion` or `10n`"),
```
zslayton, line 155: Phrasing suggestion to be a bit more explicit:

```suggestion
            _ => unreachable!("test case structs must have either an `ion` field or  an `10n` field"),
```

src/binary/var_int.rs
```diff
@@ -132,6 +134,11 @@ impl VarInt {
         Ok(())
     }
 
+    /// Effectively the same as `write_i64` if a i64 could represent -0.
```
zslayton, line 137: Nit:
```suggestion
    /// Effectively the same as `write_i64` if an i64 could represent -0.
```

zslayton (approved):
Looks good! Some superficial notes below.

src/binary/timestamp.rs
marcbowes: I see. Can I leave this as is in this PR and fix #190 as a follow up? I'm comfortable doing that since we have test coverage that ensures we don't get into an inconsistent state.

ion-hash/tests/ion_hash_tests.rs
marcbowes: I'll update the message.

src/binary/timestamp.rs
almann: Sure--I am just pointing out that your change is reasonable given #190, I might put a reference to that in the comment here as a TODO, but I am not too worried because, as you point out, a fix here will break a bunch of tests anyway.

PR #282, opened 2021-06-10
Clarify TimestampBuilder.build_at_offset

marcbowes:
*Issue #, if available:* #190

*Description of changes:*

Before this commit, `TimestampBuilder.build_at_offset` claimed to take
minutes but actually took "seconds west of UTC". Now, we take "minutes
east of UTC" which is more inline with both intuition (at least, my
own!) and how Ion represents the value in binary.

Note that the `Timestamp` type uses the chrono types internally (such as
an offset in seconds). This creates some annoying multiply/divide by 60.
However, given that Ion doesn't actually support representing offset in
seconds, it feels right to me to use minutes in our API. Maybe, in
future work, we should more clearly restrict offsets to minutes and not
use the chrono type.

-------

### DIFF
--- a/src/types/timestamp.rs
+++ b/src/types/timestamp.rs
@@ -337,7 +337,10 @@ impl TimestampBuilder {
     /// (like those bypassed by daylight saving time), this method will return an `Err(IonError)`.
     fn build(mut self) -> IonResult<Timestamp> {
         if let Some(offset) = self.offset {
-            let offset = FixedOffset::west_opt(offset).ok_or_else(|| {
+            // Our builder stores the offset in minutes difference from UTC.
+            // We're about to convert to the chrono type which uses seconds.
+            // Both values consider positive to mean Eastern Hemisphere.
+            let offset = FixedOffset::east_opt(offset * 60).ok_or_else(|| {
                 illegal_operation_raw(format!("specified offset ('{}') is invalid", offset))
             })?;
             let mut datetime: DateTime<FixedOffset> = offset.ymd(0, 1, 1).and_hms_nano(0, 0, 0, 0);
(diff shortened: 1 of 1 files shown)


src/types/timestamp.rs
```diff
@@ -337,7 +337,10 @@ impl TimestampBuilder {
     /// (like those bypassed by daylight saving time), this method will return an `Err(IonError)`.
     fn build(mut self) -> IonResult<Timestamp> {
         if let Some(offset) = self.offset {
-            let offset = FixedOffset::west_opt(offset).ok_or_else(|| {
+            // Our builder stores the offset in minutes difference from UTC.
+            // We're about to convert to the chrono type which uses seconds.
+            // Both values consider positive to mean Eastern Hemisphere.
+            let offset = FixedOffset::east_opt(offset * 60).ok_or_else(|| {
```
marcbowes, line 343: Note west -> east.

src/types/timestamp.rs
```diff
@@ -675,8 +683,8 @@ mod timestamp_tests {
     #[test]
     fn test_timestamps_with_same_ymd_hms_millis_at_known_offset_are_equal() -> IonResult<()> {
         let builder = Timestamp::with_ymd_hms_millis(2021, 2, 5, 16, 43, 51, 192);
-        let timestamp1 = builder.clone().build_at_offset(5 * 60 * 60)?;
-        let timestamp2 = builder.clone().build_at_offset(5 * 60 * 60)?;
+        let timestamp1 = builder.clone().build_at_offset(5 * 60)?;
+        let timestamp2 = builder.clone().build_at_offset(5 * 60)?;
```
marcbowes, line 687: All these tests lost a minutes -> seconds conversion.

src/types/timestamp.rs
```diff
@@ -1012,7 +1014,7 @@ mod ionc_tests {
         ),
         Timestamp::with_ymd(2020, 1, 1)
             .with_hour_and_minute(0, 1)
-            .build_at_offset(8 * 60 * 60)
+            .build_at_offset(-8 * 60)
```
marcbowes, line 1017: This (and the next) test needed a sign update.

almann (approved):
This looks right to me, thanks for doing it.

zslayton (approved):
Excellent, thanks for taking care of this!

Issue #83, opened 2020-08-17
[ion-c-sys] Support for Decimal Signed Zero in High-Level APIs


almann:
`bigdecimal::BigDecimal` relies on `num_bigint::BigInt` to provide the signed coefficient.  `BigInt` does not support maintaining sign when constructing it ([`bigint.rs` L2774-2786][1]).  This information is lost when converting from `ION_DECIMAL` to `BigDecimal`, so we should have some kind of API to preserve this if so desired.

One can always use `ION_DECIMAL` and `IonDecimalPtr` to get at the low-level Ion C APIs, but they shouldn't have to.

[1]: https://github.com/rust-num/num-bigint/blob/7562ab24330792817e42b808f60b0cac51ca261a/src/bigint.rs#L2774-L2786

marcbowes:
I think the issue might be deeper than that. Loading `-0d0` and stepping through the code in `try_to_bigdecimal`, I see that the sign is lost.. `signum` is set to 0 and calling `ion_decimal_is_negative` returns false.

zslayton:
I believe we opened this issue before [the `Decimal` type](https://github.com/amzn/ion-rust/blob/a23433aa270122171b429d123e0c66ddfa2e00e4/src/types/decimal.rs#L13-L17) existed. `Decimal` maintains the sign information and [provides conversions to/from `BigDecimal`](https://github.com/amzn/ion-rust/blob/a23433aa270122171b429d123e0c66ddfa2e00e4/src/types/decimal.rs#L153-L181). (Like `Timestamp`, it currently serves to preserve fidelity and not very useful on its own.) `ion-c-sys` could provide a `try_to_decimal` alongside `try_to_bigdecimal`.

zslayton:
In retrospect, this would create a circular dependency between `ion-rust` and `ion-c-sys`. We'd need to do some refactoring. (A new `ion-rust-types` crate? Bleh.)

almann:
@zslayton is correct, this all predates the higher-level APIs in `ion-rs`.  I don't think we need to refactor the types here, what we can do is preserve the sign information by returning it in an API like `try_to_bigdecimal() -> Result<IonCDecimal>` or some such.

Just to spit-ball, something that looks like this might be sufficient:

```rust
/// Wrapper over [`BigDecimal`] to preserve signed zero cases
pub enum IonCDecimal {
    Normal(BigDecimal),
    /// Negative zero with an exponent
    NegativeZero(i64)
}

[...]

Issue #220, opened 2021-04-25
Add an `IonEq` Trait for Ion Data Model Equivalence


almann:
Things like `PartialEq` cannot be implemented against the Ion data model for things like `SymbolToken` because of symmetry concerns, but also structural equality around numerics (e.g. `NaN`) cannot be done with `PartialEq` as well.  We should add a trait to support these kinds of comparisons.

zslayton:
We also need `IonEq` for `Decimal`. Currently, `cmp` considers `-0` to not be equal to `0` because they're not Ion-equivalent. This isn't correct mathematically, however. `cmp` should consider them equal while `IonEq` would consider them different.

zslayton:
See also: #127, which I have closed as a dup of this.

PR #285, opened 2021-06-11
Wire up ion-hash for decimals

marcbowes:
*Issue #, if available:*

*Description of changes:*

This commit uses the newly extracted binary decimal encoder to support
ion-hash for timestamps. We fix two bugs related to
representation (positive zero & eliding +0 coefficients).

Unfortunately, negative zero coefficients had to be commented out. The
issue is described in #83. The bug is such that I couldn't even skip it
since the loader mangles the "name".

-------


ion-hash/tests/ion_hash_tests.ion
```diff
@@ -109,6 +109,45 @@ intLength512::{
...
+{ ion: 0d-1,     expect:{ identity:(update::(0x0b) update::(0x50) update::(                0xc1) update::(0x0e) digest::(0x0b 0x50                 0xc1 0x0e)) } }
+{ ion: 0d5,      expect:{ identity:(update::(0x0b) update::(0x50) update::(                0x85) update::(0x0e) digest::(0x0b 0x50                 0x85 0x0e)) } }
+{ ion: 0d-5,     expect:{ identity:(update::(0x0b) update::(0x50) update::(                0xc5) update::(0x0e) digest::(0x0b 0x50                 0xc5 0x0e)) } }
+/* #83: Negative zero doesn't work yet.
+   These are commented out (vs. skipped in code) because the skipping has the same bug! (the test name presents as 0d0).
+{ ion:-0d0,      expect:{ identity:(update::(0x0b) update::(0x50) update::(0x80            0x80) update::(0x0e) digest::(0x0b 0x50 0x80            0x80 0x0e)) } }
+{ ion:-0d-0,     expect:{ identity:(update::(0x0b) update::(0x50) update::(0x80            0x80) update::(0x0e) digest::(0x0b 0x50 0x80            0x80 0x0e)) } }
+{ ion:-0d1,      expect:{ identity:(update::(0x0b) update::(0x50) update::(0x81            0x80) update::(0x0e) digest::(0x0b 0x50 0x81            0x80 0x0e)) } }
+/*
```
marcbowes, line 124: Related to #83.

src/binary/decimal.rs
```diff
@@ -34,12 +46,21 @@ where
     W: Write,
 {
     fn encode_decimal(&mut self, decimal: &Decimal) -> IonResult<()> {
+        if decimal == &DECIMAL_POSITIVE_ZERO {
+            return Ok(());
+        }
```
marcbowes, line 51: https://amzn.github.io/ion-docs/docs/binary.html#5-decimal:
> If the value is 0. (aka 0d0) then L is zero, there are no length or representation fields, and the entire value is encoded as the single byte 0x50.

src/types/coefficient.rs
```diff
@@ -20,8 +20,8 @@ pub enum Sign {
 // [Magnitude].
 #[derive(Debug, Clone, Ord, PartialOrd, Eq, PartialEq)]
 pub struct Coefficient {
-    sign: Sign,
-    magnitude: Magnitude,
+    pub(crate) sign: Sign,
+    pub(crate) magnitude: Magnitude,
```
marcbowes, line 24: Same as with timestamps, we should probably think about defining the encoding over traits so that we can do zerocopy encoding of BigDecimal.

src/types/magnitude.rs
```diff
@@ -71,7 +71,10 @@ impl Magnitude {
 
 impl From<BigUint> for Magnitude {
     fn from(value: BigUint) -> Self {
-        Magnitude::BigUInt(value)
+        match value.to_u64() {
+            Some(unsigned) => Magnitude::U64(unsigned),
+            None => Magnitude::BigUInt(value),
+        }
```
marcbowes, line 77: May be contentious, but I think this is right. The other place to do this is at the call-site (conversion from BigDecimal to Decimal). I think this is more robust.

src/binary/decimal.rs
```diff
@@ -3,14 +3,26 @@
...
 };
 
 const DECIMAL_BUFFER_SIZE: usize = 16;
+const DECIMAL_POSITIVE_ZERO: Decimal = Decimal {
+    coefficient: Coefficient {
+        sign: Sign::Positive,
+        magnitude: Magnitude::U64(0),
+    },
+    exponent: 0,
+};
```
almann, line 25: This is a specific positive zero--is this constant modeling for that specific case?

src/types/coefficient.rs
almann: Just to clarify, I think you mean the "number"-ness for our decimal and `bigdecimal::BigDecimal`, yes?  In which case I very much agree.

src/types/magnitude.rs
almann: This seems reasonable to me--we are making sure that the conversion is more compact, right?

src/types/magnitude.rs
```diff
@@ -71,7 +71,10 @@ impl Magnitude {
 
 impl From<BigUint> for Magnitude {
     fn from(value: BigUint) -> Self {
-        Magnitude::BigUInt(value)
+        match value.to_u64() {
```
almann, line 74: Minor, if my comment below is correct, I would put a small intent comment to make this more obvious.

```suggestion
        // prefer a compact representation for the magnitude
        match value.to_u64() {
```

almann (approved):
Looks good to me, very minor comments below.

src/binary/decimal.rs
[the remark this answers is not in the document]
marcbowes: My intention was to implement this from the spec:

> If the value is 0. (aka 0d0) then L is zero, there are no length or representation fields, and the entire value is encoded as the single byte 0x50.

I took that to be structural, i.e. 0dE only for E = 0. There aren't test cases for this, which seems to be a miss.

In terms of implementation.. 

```rust
impl PartialEq for Decimal {
    fn eq(&self, other: &Self) -> bool {
        self.cmp(other) == Ordering::Equal
    }
}
```

and, the relevant section in `cmp` is:

```rust
        if d1.exponent < d2.exponent {
            Self::compare_scaled_magnitudes(d1, d2)
        } else {
            Self::compare_scaled_magnitudes(d2, d1)
        }
```

I believe that will disambiguate 0d100 from 0d0 from 0d10.

src/binary/decimal.rs
marcbowes: Responded in other comment.

src/binary/decimal.rs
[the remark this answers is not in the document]
marcbowes: Pushed a commit with a test to show how I understood it.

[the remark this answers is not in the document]
almann: Yes, that verifies what I was expecting, we have a structural comparison here so that is what we want for what you're doing.  What is interesting (and out of scope for your PR) is that most people will not expect this behavior as they will expect value comparison.  I think this is basically related to #220, we should really push this structural stuff into its own trait.

almann (approved):
:shipit:

src/types/coefficient.rs
almann: Added #286 to track this.

Issue #273, opened 2021-06-02
Implement Accessor APIs for `Timestamp`


almann:
#272 illustrates that we need APIs to access the `Timestamp` properties even within the crate.  My suggestion is to implement `TimeLike` and `DateLike` from `chrono` as well as define some kind of `IonTimestamp` trait to fill in the precision/negative zero offset gaps that the other two traits don't have.

This allows us to define APIs in terms of these traits, but more importantly make it possible to be generic against our `Timestamp` representation and `DateTime<FixedOffset>` for serializers and other things as we can implement the `IonTimestamp` trait for `DateTime<FixedOffset>` (which has fixed precision to nanoseconds and never has negative zero offset).

---

This suggests that implementing `TimeLike` and `DateLike` (as well as some hypothetical `IonTimestampLike` or some such) are probably in order for the API.

Temporarily to work around the fact that we do not expose a useful API, this seems okay.

_Originally posted by @almann in https://github.com/amzn/ion-rust/pull/272#discussion_r643642733_

almann:
This is related to #209, though that issue has other implementation goals.

popematt:
We won't implement `TimeLike` and `DateLike` from `chrono` because of https://github.com/chronotope/chrono/issues/1768, but I'm leaving this issue open because a `TimestampLike` trait is still a good idea.

PR #288, opened 2021-06-14
Use new decimal encoding to fix #274

### LINKED ISSUE #274: `Timestamp` Serialization Should Not Be Fixed to Nanoseconds
#272 ports some internal code to APIs, part of this is a TODO to make `Timestamp` serialization to be not fixed to nanoseconds.

When searching for prior issues, I found #17, but I am not sure if that is different or already resolved (since it indicates *skipping* fractional seconds).  @zslayton can probably confirm this.

---

So this means that if there exists sub-second precision, that precision is always nine digits (i.e. nanoseconds)?

_Originally posted by @almann in https://github.com/amzn/ion-rust/pull/272#discussion_r643643079_

marcbowes:
*Issue #, if available:* #273

*Description of changes:*

This commit uses the new decimal encoder to encode fractional timestamps
as a decimal. Two timestamp tests that were failing now pass!

-------

### DIFF
--- a/src/binary/timestamp.rs
+++ b/src/binary/timestamp.rs
@@ -1,14 +1,20 @@
 // Copyright Amazon.com, Inc. or its affiliates.
 
-use std::io::Write;
+use std::{io::Write, ops::Neg};
 
 use arrayvec::ArrayVec;
 use chrono::{Datelike, Timelike};
 
 use crate::{
-    binary::{uint::DecodedUInt, var_int::VarInt, var_uint::VarUInt, writer::MAX_INLINE_LENGTH},
+    binary::{
+        decimal::DecimalBinaryEncoder, var_int::VarInt, var_uint::VarUInt,
+        writer::MAX_INLINE_LENGTH,
+    },
     result::IonResult,
-    types::timestamp::{Precision, Timestamp},
+    types::{
+        decimal::Decimal,
+        timestamp::{Mantissa, Precision, Timestamp},
+    },
 };
 
 const MAX_TIMESTAMP_LENGTH: usize = 16;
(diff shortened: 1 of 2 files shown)


src/binary/timestamp.rs
```diff
@@ -67,13 +73,25 @@ where
...
                     if timestamp.precision > Precision::Second {
-                        // This implementation always writes the datetime out with nanosecond precision.
-                        // The nanoseconds are an integer, which means that our exponent is always 9
-                        // and our coefficient is the number of nanoseconds.
-                        let nanoseconds = utc.nanosecond();
-                        const SCALE: i64 = -9; // "Scale" (used by BigDecimal) is -1 * exponent
-                        VarInt::write_i64(self, SCALE)?;
-                        DecodedUInt::write_u64(self, nanoseconds as u64)?;
+                        // It's not possible to have a precision of fractional
+                        // seconds and then not have any!
```
zslayton, line 77: Consider turning this into an [`unreachable!`](https://doc.rust-lang.org/std/macro.unreachable.html):
```
else {
  unreachable!("timestamps with fractional second precision must have fractional seconds");
}
```

src/binary/timestamp.rs
```diff
@@ -67,13 +73,25 @@ where
...
+                            // branch owning vs borrowing the decimal
+                            // representation. #286 should provide a fix.
+                            match mantissa {
+                                Mantissa::Digits(precision) => {
+                                    // Consider the following case: `2000-01-01T00:00:00.123Z`.
+                                    // That's 123 millis, or 123,000,000 nanos.
+                                    // Our mantissa is 0.123, or 123d-3.
+                                    let scaled = utc.nanosecond() / 10u32.pow(9 - *precision); // 123,000,000 -> 123
+                                    let exponent = (*precision as i64).neg(); // -3
+                                    let fractional = Decimal::new(scaled, exponent); // 123d-3
```
zslayton, line 89: These comments are very helpful! 👍

zslayton (approved):
LGTM! One suggestion below that you can take or leave.

src/binary/timestamp.rs
marcbowes: I think we want to have the property that this crate _never_ panics. So even though this code is actually unreachable, I don't want to make our auditing difficult. I was thinking about adding a new `IonError` variant for non-panic unreachable, wdyt?

PR #291, opened 2021-06-16
Move ion-hash implementation behind traits

marcbowes:
Before this commit, we were passing around a sink (that implements
digest or write) as a `&mut` "out" parameter. Ugly in general, but even
more so given we had to repeat the `Digest` subtraits over and over.

Now, we have two traits (split simply for code organization). One trait
handles the overall hashing orchestration, while another handles the
representation. Other than some funny business with calling syntax, this
worked out pretty nicely!

Note that I absolutely have _not_ thought through the traits
nicely (i.e. "are we happy with the public/private split?" was not a
consideration). Rather, I focused on the mechanical refactoring to see
if this was worth it. There are loads of doc/comments that don't make
sense anymore that I'll fix up shortly!

-------

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/ion-hash/src/element_hasher.rs
+++ b/ion-hash/src/element_hasher.rs
@@ -1,0 +1,106 @@
+// Copyright Amazon.com, Inc. or its affiliates.
+
+//! Provides [`ElementHasher`], a single-use IonHash implementation for an
+//! [`Element`].
+
+use std::io;
+
+use digest::{FixedOutput, Output, Reset, Update};
+use ion_rs::result::IonResult;
+use ion_rs::value::Element;
+
+use crate::representation::RepresentationEncoder;
+use crate::type_qualifier::TypeQualifier;
+use crate::Markers;
+
+pub(crate) struct ElementHasher<D>
+where
+    D: Update + FixedOutput + Reset + Clone + Default,
+{
+    pub(crate) digest: D,
+}
+
+impl<D> ElementHasher<D>
+where
+    D: Update + FixedOutput + Reset + Clone + Default,
+{
+    pub(crate) fn new(digest: D) -> ElementHasher<D> {
+        ElementHasher { digest }
+    }
+
+    pub(crate) fn hash_element<E>(mut self, elem: &E) -> IonResult<Output<D>>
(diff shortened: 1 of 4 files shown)


marcbowes:
Part of https://github.com/amzn/ion-rust/issues/287

ion-hash/src/element_hasher.rs
```diff
@@ -0,0 +1,106 @@
...
+        ElementHasher { digest }
+    }
+
+    pub(crate) fn hash_element<E>(mut self, elem: &E) -> IonResult<Output<D>>
+    where
+        E: Element + ?Sized,
+    {
+        self.update_serialized_bytes(elem)?;
+        Ok(self.digest.finalize_fixed())
+    }
```
almann, line 37: To line up our traits, I think this could be  `impl IonHasher for ElementHasher`, though this is a very minor factoring change as this is an internal wrapper.

ion-hash/src/element_hasher.rs
```diff
@@ -0,0 +1,106 @@
...
+
+        for byte in data.as_ref() {
+            if let Markers::B | Markers::E | Markers::ESC = *byte {
+                escaped.extend(&[Markers::ESC, *byte]);
+            } else {
+                escaped.extend(&[*byte]);
+            }
+        }
+
+        self.digest.update(escaped);
```
almann, line 86: I might be missing something, but is the intermediate buffer even necessary?  Could we not update directly?

almann (approved):
Looks good for the refactor--minor comments below.

ion-hash/src/element_hasher.rs
marcbowes: Whoops, yeah.

ion-hash/src/element_hasher.rs
marcbowes: not anymore, I think. needed them for the awful test setup we had in the past.

marcbowes: https://github.com/amzn/ion-rust/issues/295

Issue #299, opened 2021-08-31
Introduce an IonError::IncompleteData variant


zslayton:
Currenly `IonError` (and therefore `IonResult`) doesn't have a way to cleanly communicate that the underlying data stream terminated unexpectedly. We should add an `IonError::IncompleteData` variant.

This will be non-trivial; the places where an `Eof` would be detected do not have access to information that would be valuable to include in the resulting `Err`, such as:
* The byte (or code unit) offset in the stream where the error occurred
* What the cursor was trying to read at that moment (e.g. Were we in the middle of a struct? A list? An int?)

Some refactoring will be required.

zslayton:
This will probably involve rejiggering the `IonDataSource::next_byte` method, which currently returns an `IonResult<Option<u8>>`.

https://github.com/amzn/ion-rust/blob/256f164be5453df30bd1a40fda95f95af877b4c7/src/data_source.rs#L70-L78

Once we have an `IncompleteData` error variant, this signature could be simplified to `IonResult<u8>`.

PR #301, opened 2021-08-31
Makes IonDataSource report EOF reliably

### LINKED ISSUE #300: Unexpected EOF results in infinite loop in binary cursor
Various methods in the `IonDataSource` trait call [`fill_buf()?`](https://doc.rust-lang.org/std/io/trait.BufRead.html#tymethod.fill_buf) expecting EOF to be an error that's early-returned by `?`. However, `fill_buf()` returns an empty slice on EOF, not an error. If the stream ends when more data is expected, these methods will call `fill_buf()` in an infinite loop. This is most often seen when reading `VarUInt`s.

While we _should_ have a distinct `IonError::IncompleteData` variant, this will take a bit of effort to implement (see #299). In the meantime, we can simply return an `IonError::DecodingError` as a stopgap fix.

zslayton:
Prior to this commit, multiple methods in the IonDataSource trait would
either loop infinitely or return an `IoError` in the event of an
unexpected EOF.

This commit causes all of the methods in question to detect EOFs and
report them as a decoding error. Work needed to add a proper
IonError::IncompleteData enum variant is being tracked in #299.

Fixes #300.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -23,7 +23,7 @@ jobs:
         with:
           profile: minimal
           # nightly can be very volatile--pin this to a version we know works well...
-          toolchain: nightly-2021-05-07
+          toolchain: nightly-2021-08-30
           override: true
       - name: Cargo Test
         uses: actions-rs/cargo@v1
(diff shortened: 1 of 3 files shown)


desaikd (approved):
LGTM! Although the CI build shows failed test.

zslayton:
Build failure was a formatting issue discovered by `cargo fmt`. I've pushed a fix.

zslayton:
Looks like there's an issue with our test coverage checker.

PR #310, opened 2021-09-16
Fixes panic in read_datetime

### LINKED ISSUE #306: timestamp `1990T` fails to parse (and similarly incomplete dates)
Example test that fails:

    #[test]
    fn test_parse_incomplete_dates() {
        let t1990 = "4AEA6mPAD8Y="; // date "1990T" as binary ion
        let t1990 = base64::decode(t1990).unwrap();

        let cursor = IoCursor::new(t1990);
        let cursor = BinaryIonCursor::new(cursor);
        let mut reader = Reader::new(cursor);
        
        reader.next().unwrap();
        reader.read_datetime().unwrap(); // panics
    }

thread 'tests::test_parse_incomplete_dates' panicked at 'invalid or out-of-range date',

These are valid values according to the java implementation and https://amzn.github.io/ion-docs/docs/spec.html#timestamp

zslayton:
The binary cursor's `read_datetime` method incorrectly 0-indexes
its default month and day values. If a timestamp is encountered
that only specifies a year or year+month, the default value of
zero causes the construction of a DateTime to panic.

This patch sets the default month and day values to 1.

Fixes #306.

Note that long-term, we [intend to replace `read_datetime` with a `read_timestamp` method](https://github.com/amzn/ion-rust/issues/308).

### DIFF
--- a/src/binary/cursor.rs
+++ b/src/binary/cursor.rs
@@ -523,6 +523,10 @@ impl<R: IonDataSource> Cursor for BinaryIonCursor<R> {
         self.clob_ref_map(|c| c.into())
     }
 
+    // This method will return year-, month-, and day-precision timestamps as UTC DateTimes
+    // with smaller time units (hour, minute, etc) zeroed out. Longer term, this method
+    // will be replaced by a `read_timestamp` method that returns a proper Timestamp.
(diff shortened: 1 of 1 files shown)


desaikd (approved):
LGTM!

marcbowes (approved):

PR #312, opened 2021-09-17
    Adds `read_timestamp` method to the binary reader

zslayton:
This PR adds a `read_timestamp` method alongside the existing
`read_datetime` method, which is now marked deprecated. This
allows users to read an Ion timestamp with full fidelity.

It also adds implementations of `TryInto` to allow users to
convert a Timestamp into the corresponding chrono type:
* Timestamps with unknown offsets can be turned into
  `NaiveDateTime`s.
* Timestamps with known offsets can be turned into
  `DateTime<FixedOffset>`s.

In order to report whether a given Timestamp has a known or
unknown offset, the `VarInt` type now stores an extra bool
field to track whether it is actually negative zero. Prior
to this commit, the value of the `VarInt` was stored as an
i64, which cannot represent the -0 case on its own.

Finally, this PR fixes a bug in the way that Timestamps with
different Mantissa representations (Decimal vs number-of-digits)
were being tested for equality.


src/binary/cursor.rs
```diff
@@ -523,6 +525,96 @@ impl<R: IonDataSource> Cursor for BinaryIonCursor<R> {
         self.clob_ref_map(|c| c.into())
     }
 
+    fn read_timestamp(&mut self) -> IonResult<Option<Timestamp>> {
```
zslayton, line 528: This method feels a bit long to me. I could make it a bit more concise by either:
* Defining some traits for the various Timestamp builder API types to allow `build()`, `build_at_offset()`, and `build_at_unknown_offset()` to be called generically. This would let me write a couple of helper methods.
* Defining a macro for the 
```rust
if self.finished_reading_value() {
  let timestamp = // ...
  return Ok(Some(timestamp));
}
```
boilerplate repeated for each field.

The main drawback is that this is the only function where those artifacts would be very useful. I'm not sure how much net benefit we'd get for the investment. Open to other opinions, though!

desaikd (approved):
Looks good! We might want to change the `examples/read_all_values.rs` to use this newly implemented `read_timestamp()` method instead.
https://github.com/amzn/ion-rust/blob/d8b30a95e0131db6b075bdd65e56487243ee0755/examples/read_all_values.rs#L54-L56

jobarr-amzn (approved):

marcbowes (commented):
Very nice!

PR #315, opened 2021-09-21
 Adds a `read_decimal` method to the binary reader

### LINKED ISSUE #9: Int and VarInt lose sign information when reading -0
`Int` and `VarInt` represent the value of the signed integer being read in as an `i64`, which cannot represent `-0` on its own. `Int` and `VarInt` are already structs, so we can add a `sign` field to these and check for it when necessary.

### LINKED ISSUE #309: Replace read_big_decimal with read_decimal
The reader API was created before a lossless `Decimal` type existed. We should return `Decimal` values (which can be turned into a `BigDecimal` as needed) to guarantee high fidelity.

zslayton:
This patch adds a `read_decimal` method to the binary reader
alongside the existing `read_big_decimal` method, which is now
deprecated. Fixes #309.

It also adds an `is_negative` flag to the `Int` encoding primitive,
allowing it to represent negative zero when necessary. Fixes #9.

### DIFF
--- a/src/binary/cursor.rs
+++ b/src/binary/cursor.rs
@@ -433,6 +433,27 @@ impl<R: IonDataSource> Cursor for BinaryIonCursor<R> {
         })
     }
 
+    fn read_decimal(&mut self) -> IonResult<Option<Decimal>> {
+        read_safety_checks!(self, IonType::Decimal);
+
+        if self.cursor.value.value_length == 0 {
+            return Ok(Some(Decimal::new(0, 0)));
+        }
+
+        let exponent_var_int = self.read_var_int()?;
+        let coefficient_size_in_bytes =
+            self.cursor.value.value_length - exponent_var_int.size_in_bytes();
+
+        let exponent = exponent_var_int.value() as i64;
+        let coefficient = self.read_int(coefficient_size_in_bytes)?;
+
+        if coefficient.is_negative_zero() {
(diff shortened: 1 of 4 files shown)


src/binary/cursor.rs
```diff
@@ -1216,8 +1238,41 @@ mod tests {
...
+        assert_eq!(cursor.next()?, Some(Value(IonType::Decimal, false)));
+        assert_eq!(cursor.read_decimal()?, Some(Decimal::negative_zero()));
+        Ok(())
+    }
+
+    #[test]
+    fn test_read_decimal_positive_exponent() -> IonResult<()> {
+        let mut cursor = ion_cursor_for(&[0x52, 0x81, 0x02]);
+        assert_eq!(cursor.next()?, Some(Value(IonType::Decimal, false)));
+        assert_eq!(cursor.read_big_decimal()?, Some(20.into()));
```
desaikd, line 1261: Allow deprecated missing here:
```suggestion
        #![allow(deprecated)] // `read_big_decimal` is deprecated
        assert_eq!(cursor.read_big_decimal()?, Some(20.into()));
```

desaikd (approved):
one minor change otherwise looks good!

src/binary/cursor.rs
zslayton: Good catch! This was a copy/paste error; it should be `read_decimal`. Pushing a fix now.

PR #316, opened 2021-09-23
 Adds `write_decimal` method to binary writer

zslayton:
* Adds a `write_decimal` method to the binary writer alongside
  the existing `write_big_decimal` method, which is now deprecated.
* Adds a `write_negative_zero` method to the Int encoding primitive.
  This method was not needed previously as `BigDecimal` was not
  capable of representing -0. Users could not write a decimal -0
  prior to this patch.
* Adds an `is_negative_zero` method to the Int encoding primitive,
  allowing the binary reader to test for that case.
* Adds an implementation of `TryFrom<f64>` for `Decimal`. This produces
  a `Decimal` approximation of the input `f64`'s value (unless the `f64` is
  `inf`, `-inf`, or `nan`). It's helpful largely for writing tests.

### DIFF
--- a/src/binary/var_int.rs
+++ b/src/binary/var_int.rs
@@ -144,9 +144,14 @@ impl VarInt {
         Ok(encoded_bytes.len())
     }
 
-    /// Effectively the same as `write_i64` if an i64 could represent -0.
+    /// Encodes a negative zero as an `VarInt` and writes it to the privided `sink`.
+    /// Returns the number of bytes written.
+    ///
+    /// This method is similar to [write_i64]. However, because an i64 cannot represent a negative
+    /// zero, a separate method is required.
     pub fn write_negative_zero<W: Write>(sink: &mut W) -> IonResult<usize> {
-        Ok(sink.write(&[VARINT_NEGATIVE_ZERO])?)
+        sink.write_all(&[VARINT_NEGATIVE_ZERO])?;
+        Ok(1)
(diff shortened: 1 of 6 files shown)


src/binary/var_int.rs
```diff
@@ -144,9 +144,14 @@ impl VarInt {
...
-    /// Effectively the same as `write_i64` if an i64 could represent -0.
+    /// Encodes a negative zero as an `VarInt` and writes it to the privided `sink`.
+    /// Returns the number of bytes written.
+    ///
+    /// This method is similar to [write_i64]. However, because an i64 cannot represent a negative
+    /// zero, a separate method is required.
     pub fn write_negative_zero<W: Write>(sink: &mut W) -> IonResult<usize> {
-        Ok(sink.write(&[VARINT_NEGATIVE_ZERO])?)
+        sink.write_all(&[VARINT_NEGATIVE_ZERO])?;
+        Ok(1)
```
zslayton, line 154: Minor cleanup to this method that already existed. In particular:
* Changed the doc comment to be more descriptive/helpful.
* Replaced the call to `write` (which could choose to write zero bytes and report success) with a call to `write_all`, which guarantees that it will not return until the provided bytes have been written in full or an error has occurred.

desaikd (approved):
Looks good!

jobarr-amzn (approved):

PR #317, opened 2021-09-25
Adds streaming support to the text reader

zslayton:
This PR allows a `TextReader` to be created from a variety of text
data sources, including `String`, `&str`, `&[u8]`, and `BufReader`.
Support for other types can be addede by implementing the
`TextIonDataSource` trait.

It also introduces the `TextBuffer`, a type which can pull text
into a buffer `n` lines at a time and discard bytes as they're
processed. This enables the reader to process its input in a
streaming fashion, holding a modest amount of data in memory at
a time.

The patch adds parsing logic to detect the beginnings and endings of
the various container types. However, it does not add logic for
stepping into/out of those containers.

Finally, by leveraging the above features, the existing
proto-`TextReader` can now iterate over a text Ion stream containing
multiple top-level values. If it encounters an error, it can report
the line of input that caused it.

### DIFF
--- a/src/text/parsers/containers.rs
+++ b/src/text/parsers/containers.rs
@@ -1,0 +1,117 @@
+use nom::IResult;
+use crate::text::TextStreamItem;
+use nom::combinator::{recognize, map};
+use nom::bytes::streaming::tag;
+use nom::branch::alt;
+
+/// Matches the beginning of a container and returns a [TextStreamItem] indicating its type.
+pub(crate) fn parse_container_start(input: &str) -> IResult<&str, TextStreamItem> {
+    alt((
+        recognize_struct_start,
+        recognize_list_start,
+        recognize_s_expression_start
+    ))(input)
+}
+
+/// Matches the beginning of a struct and returns a [TextStreamItem::StructStart].
+pub(crate) fn recognize_struct_start(input: &str) -> IResult<&str, TextStreamItem> {
+    map(
+        recognize(tag("{")),
+        |_matched_text| TextStreamItem::StructStart
+    )(input)
+}
+
+/// Matches the beginning of a list and returns a [TextStreamItem::ListStart].
+pub(crate) fn recognize_list_start(input: &str) -> IResult<&str, TextStreamItem> {
+    map(
+        recognize(tag("[")),
+        |_matched_text| TextStreamItem::ListStart
+    )(input)
+}
+
+/// Matches the beginning of an s-expression and returns a [TextStreamItem::SExpressionStart].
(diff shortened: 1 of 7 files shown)


src/text/parsers/containers.rs
```diff
@@ -0,0 +1,109 @@
+use crate::text::TextStreamItem;
+use nom::branch::alt;
+use nom::bytes::streaming::tag;
+use nom::combinator::{map, recognize};
+use nom::IResult;
```
zslayton, line 5: This file defines `nom` parsing rules to detect the beginnings and endings of the various container types.

src/text/parsers/top_level.rs
```diff
@@ -0,0 +1,42 @@
...
+use crate::text::parsers::containers::{parse_container_end, parse_container_start};
+use crate::text::parsers::decimal::parse_decimal;
+use crate::text::parsers::float::parse_float;
+use crate::text::parsers::integer::parse_integer;
+use crate::text::parsers::null::parse_null;
+use crate::text::parsers::string::parse_string;
+use crate::text::parsers::symbol::parse_symbol;
+use crate::text::parsers::timestamp::parse_timestamp;
+use crate::text::parsers::whitespace;
+use crate::text::TextStreamItem;
```
zslayton, line 18: The rules defined in this file originally lived alongside the `TextReader` in `text::reader.rs`. They've been promoted to their own module.

desaikd (approved):

src/text/reader.rs
```diff
@@ -1,105 +1,148 @@
...
+            // in the stream and return a &str slice containing the remaining, not-yet-parsed text.
+            match top_level::top_level_value(self.buffer.remaining_text()) {
+                // If `top_level_value` returns 'Incomplete', there wasn't enough text in the buffer
+                // to match the next item. No syntax errors have been encountered (yet?), but we
+                // need to load more text into the buffer before we try to parse it again.
+                Err(Incomplete(_needed)) => {
+                    // Ask the buffer to load another line of text.
+                    // TODO: Currently this loads a single line at a time for easier testing.
+                    //       We may wish to bump it to a higher number of lines at a time (8?)
+                    //       for efficiency once we're confident in the correctness.
```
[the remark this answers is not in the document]
zslayton: Right. The only multi-line entities that the parser will deal with are (I think?):
* Long-form strings
* Long-form clobs
* Generously spaced blobs
* `/* Block comments */`
* and pathologically written annotations:
```json
foo
  ::
        bar ::
::  baz ::
7
```

PR #319, opened 2021-09-29
Adds support for reading text annotations

zslayton:
* `parse_symbol` can now match text symbol IDs (e.g. `$23`)
* Symbols read from the stream are now represented as
  `OwnedSymbolToken`s instead of `String`s to allow for the case
  in which a given symbol's ID was found and the text has not yet
  been looked up.
* The reader's `next()` method now returns an `(annotations, item)`
  tuple instead of just an `item`. If there are no annotations on
  an item, the annotations Vec is empty.

### DIFF
--- a/src/text/reader.rs
+++ b/src/text/reader.rs
@@ -5,6 +5,7 @@ use crate::text::parsers::top_level;
 use crate::text::text_buffer::TextBuffer;
 use crate::text::text_data_source::TextIonDataSource;
 use crate::text::TextStreamItem;
+use crate::value::owned::OwnedSymbolToken;
 
 // TODO: Implement a full text reader.
 //       This implementation is a placeholder. It is currently capable of reading a stream of top-level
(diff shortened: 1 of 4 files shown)


desaikd (approved):
Looks good!

src/text/reader.rs
```diff
@@ -141,9 +145,79 @@ mod reader_tests {
...
+    #[test]
+    fn test_text_read_multiple_annotated_top_level_values() {
+        let ion_data = r#"
+            mercury::null
+            venus::'earth'::true
+            $17::mars::5
+            jupiter::5e0
+            'saturn'::5.5
+            $100::$200::$300::2021-09-25T
+            'uranus'::foo
```
jobarr-amzn, line 157: A test that randomly combines words from astronomy and computing, eh?

jobarr-amzn (approved):

PR #320, opened 2021-09-30
Resolves ambiguous text parsing at EOF

### LINKED ISSUE #318: Handle incomplete values at the end of an input stream
The parser combinator library we're using ([nom](https://github.com/Geal/nom)) provides streaming functions that will helpfully report `Incomplete` in any situation where the input buffer ends with bytes that might be a partial value.

Some examples of buffer contents that lead to `Incomplete`:
* `7` might be the first digit in a longer int
* `'''long-form string'''\n` might be the first chunk of a multi-chunk, long-form string
* `foo` could be the beginning of a longer identifier like `foobarbaz`
* `foo\n\n\n` could be an identifier OR it could be the first annotation in a series of annotations that will continue after several blanks lines / comments.

Normally, this would be resolved by telling the parser that the end of the buffer is `EOF`--the end of the input stream. With the knowledge that no more text will follow, a parser can handle each of the above edge cases successfully.

However, in `nom`'s case, there isn't a simple way to say "do what you normally do, but this time it's EOF". `nom` does provide a near-identical set of parsing primitives in a module called `complete` that have the desired effect, but using them requires that parser rules for various elements be written in two flavors: `streaming` and `complete`. We'd end duplicating a good chunk of the parser code; one copy would call the `streaming` functions and the other the `complete` functions. Seems like a lot of overhead for a single corner case (`EOF`).

I'm planning to add a workaround: when the low-level reader runs into EOF, it will append a newline and a text IVM (`\n$ion_1_0`) to the end of buffer. This will allow the streaming parser functions to look ahead and see another element coming, signaling that their parse of the current method is complete. When the reader encounters the IVM at the end, it will quietly discard it.

zslayton:
This PR modifies the text reader to be able to unambiguously
parse data found at the end of the stream. Because the
parser always operates on a fixed buffer, it could not easily
distinguish between the end of the buffer (more data to come)
and the end of the stream being loaded into the buffer (EOF).

When EOF is detected, the reader will now append a sentinel
value to the end of the stream and re-attempt parsing. If the
sentinel value is found, `EndOfStream` is reported. If a different
value is found, that value is returned instead and the sentinel
is discarded.

Fixes #318. See that issue for more details.

This PR is based on the `read-text-annotations` branch, used in PR #319. The diff below is with that branch. Once #319 is merged, I'll rebase this with `main`.

### DIFF
--- a/src/text/mod.rs
+++ b/src/text/mod.rs
@@ -31,7 +31,7 @@ pub(crate) enum TextStreamItem {
     SExpressionEnd,
     StructStart,
     StructEnd,
-    EndOfStream
+    EndOfStream,
 }
 
 impl TextStreamItem {
--- a/src/text/mod.rs
+++ b/src/text/mod.rs
@@ -6,6 +6,7 @@ pub mod writer;
 
 use crate::types::decimal::Decimal;
 use crate::types::timestamp::Timestamp;
+use crate::value::owned::OwnedSymbolToken;
 use crate::IonType;
 
 /// Represents a single item encountered in a text Ion stream. The enum includes variants for each
@@ -20,8 +21,7 @@ pub(crate) enum TextStreamItem {
     Timestamp(Timestamp),
     // TODO: String(&str) will be possible if/when we add reusable buffers to the TextReader.
     String(String),
-    // TODO: This will be a Symbol(SymbolToken) when the SymbolToken API is implemented.
-    Symbol(String),
+    Symbol(OwnedSymbolToken),
     // TODO: [BC]lob(&[u8]) will be possible if/when we add reusable buffers to the TextReader.
     Blob(Vec<u8>),
     Clob(Vec<u8>),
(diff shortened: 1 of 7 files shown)


jobarr-amzn (approved):
Looks good!

jobarr-amzn:
It occurs to me to ask though- #318 proposes a text IVM as the sentinel value but this PR uses `0`. What's the reasoning behind the selection of the sentinel value? Any value will work so it might as well be shorter than the IVM?

zslayton:
> It occurs to me to ask though- #318 proposes a text IVM as the sentinel value but this PR uses `0`. What's the reasoning behind the selection of the sentinel value? Any value will work so it might as well be shorter than the IVM?

Good eye! Two reasons:
* I realized that if I'm discarding the sentinel value anyway, I might has well make it the cheapest possible thing to parse. Reading a 1-digit integer is cheaper than reading an 8-character `$ion_1_0`.
* I haven't written the parser rule for reading an IVM yet. 😛

PR #325, opened 2021-10-06
Support NOP-padded data

jobarr-amzn:
*Description of changes:* 

Adds support for NOP padding in the binary cursor, as described in the [binary spec](https://amzn.github.io/ion-docs/docs/binary.html). 

### DIFF
--- a/src/binary/cursor.rs
+++ b/src/binary/cursor.rs
@@ -1112,7 +1112,7 @@ where
         if self.cursor.index_at_depth > 0 {
             // NOPs are not values, so bytes_read can be past the last consumed value (`end`).
             // This is an artifact of consuming NOP sleds with recursive `next()`, and could be
-            // fixed with refactoring. 
+            // fixed with refactoring.
             // `saturating_sub` avoids underflow by returning 0 if `position` > `end`.
             let bytes_to_skip = end.saturating_sub(position);
             self.skip_bytes(bytes_to_skip)
--- a/src/binary/cursor.rs
+++ b/src/binary/cursor.rs
@@ -333,7 +333,7 @@ impl<R: IonDataSource> Cursor for BinaryIonCursor<R> {
                         bytes = number_of_bytes,
                         nop_range = nop_range,
                         over = nop_range.end - container_range.end,
-                        s = if number_of_bytes == 1 {""} else {"s"},
+                        s = if number_of_bytes == 1 { "" } else { "s" },
                         container_range = container_range,
                     ));
                 }
--- a/src/binary/cursor.rs
+++ b/src/binary/cursor.rs
@@ -195,7 +195,7 @@ impl Default for EncodedValue {
             ion_type: IonType::Null,
             header: Header {
                 ion_type: None,
-                ion_type_code: IonTypeCode::NullOrWhitespace,
+                ion_type_code: IonTypeCode::NullOrNop,
                 length_code: length_codes::NULL,
             },
             field_id: None,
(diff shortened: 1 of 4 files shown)


src/binary/cursor.rs
```diff
@@ -310,11 +310,49 @@ impl<R: IonDataSource> Cursor for BinaryIonCursor<R> {
...
+                        "{bytes}-byte NOP padding on byte range [{nop_start}, {nop_end}) is {over} \
+                        byte{s} past container content range [{container_start}, {container_end})",
+                        bytes = number_of_bytes,
+                        nop_start = nop_range.start,
+                        nop_end = nop_range.end,
+                        over = nop_range.end - container_range.end,
+                        s = if number_of_bytes == 1 {""} else {"s"},
+                        container_start = container_range.start,
+                        container_end = container_range.end,
+                    ));
```
jobarr-amzn, line 340: This creates error messages like:

```
Error: DecodingError { description: "1-byte NOP padding on byte range [7, 8) is 1 byte past container content range [5, 7)" }
```

```
Error: DecodingError { description: "4-byte NOP padding on byte range [10, 14) is 2 bytes past container content range [5, 12)" }
```

src/binary/cursor.rs
```diff
@@ -310,11 +310,49 @@ impl<R: IonDataSource> Cursor for BinaryIonCursor<R> {
...
+                if nop_range.end > container_range.end {
+                    // This NOP is malformed, let's assemble data for error reporting
+                    return decoding_error(&format!(
+                        "{bytes}-byte NOP padding on byte range [{nop_start}, {nop_end}) is {over} \
+                        byte{s} past container content range [{container_start}, {container_end})",
+                        bytes = number_of_bytes,
+                        nop_start = nop_range.start,
+                        nop_end = nop_range.end,
+                        over = nop_range.end - container_range.end,
+                        s = if number_of_bytes == 1 {""} else {"s"},
```
zslayton, line 337: ha, nice. I would've just put `byte(s)` in the format string, but you went above and beyond. 👏

src/binary/cursor.rs
```diff
@@ -327,6 +365,12 @@ impl<R: IonDataSource> Cursor for BinaryIonCursor<R> {
                 Some(header) => header,
                 None => return Ok(None),
             };
+            if header.is_nop() {
+                return decoding_error(&format!(
+                    // What helpful context could we include here?
+                    "Can't have NOP padding inside an annotation wrapper"
```
zslayton, line 371: What do you think of:
```suggestion
                    "The annotation wrapper starting at byte {} contains NOP padding, which is illegal.",
                    self.cursor.bytes_read
```
?

src/binary/cursor.rs
```diff
@@ -1063,11 +1107,13 @@ where
     }
 
     fn skip_current_value(&mut self) -> IonResult<()> {
-        if self.cursor.index_at_depth == 0 {
-            Ok(())
-        } else {
+        if self.cursor.index_at_depth > 0
+            && self.cursor.bytes_read < self.cursor.value.value_end_exclusive()
```
zslayton, line 1111: I think this check is superfluous? `skip_bytes` will early return if `bytes_to_skip` is `0`.

src/binary/cursor.rs
```diff
@@ -1802,4 +1848,130 @@ mod tests {
...
+
+        assert!(matches!(cursor.next(), Err(IonError::DecodingError { .. })));
+
+        cursor.step_out()?;
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_nop_pad_larger_than_list() -> IonResult<()> {
```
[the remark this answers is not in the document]
jobarr-amzn: Ah, oops. I put an extra `cursor.next()?;` in there so that I could capture the generated message. Forgot to remove it!

src/binary/cursor.rs
jobarr-amzn: We dove on this offline- this check was due to recursive use of `next()` to handle NOP sleds. The method has been clarified.

src/binary/cursor.rs
```diff
@@ -1063,11 +1105,18 @@ where
...
-            let bytes_to_skip = self.cursor.value.value_end_exclusive() - self.cursor.bytes_read;
+        let pos = self.cursor.bytes_read;
+        let end = self.cursor.value.value_end_exclusive();
+
+        // Don't skip a value if we haven't called `next()` at the current level
+        if self.cursor.index_at_depth > 0 {
+            // NOPs are not values, so bytes_read can be past the last consumed value (`end`).
+            // This is an artifact of consuming NOP sleds with recursive `next()`, and could be
+            // fixed with refactoring.
+            let bytes_to_skip = end.saturating_sub(pos);
```
zslayton, line 1116: Two thoughts here:
* I'd prefer to spell `pos` out as `position` if you don't mind
* `saturating_sub` isn't super common, so I added a little explanation
```suggestion
        let position = self.cursor.bytes_read;
        let end = self.cursor.value.value_end_exclusive();

        // Don't skip a value if we haven't called `next()` at the current level
        if self.cursor.index_at_depth > 0 {
            // NOPs are not values, so bytes_read can be past the last consumed value (`end`).
            // This is an artifact of consuming NOP sleds with recursive `next()`, and could be
            // fixed with refactoring. 
            // `saturating_sub` avoids underflow by returning 0 if `position` > `end`.
            let bytes_to_skip = end.saturating_sub(position);
```

zslayton (approved):
Great to have this change, thanks! One suggestion below.

Issue #327, opened 2021-10-07
Use a consistent naming convention for text parser functions


zslayton:
The `nom` parser functions defined in the `text::parsers` module initially used this naming convention:
* `recognize_THING` would identify a `&str` slice from input that represented a `THING`. (The term `recognize` being borrowed from a `nom` [function of the same name](https://docs.rs/nom/7.0.0/nom/combinator/fn.recognize.html)).
* `parse_THING` would recognize text representing a `THING` and then perform additional steps to convert it into a more meaningful data type. (For example, `parse_float` would identify the text representation of an Ion `float` and convert it into a `TextStreamItem::Float(f64)`.

Over time, functions started to just be named `THING` (dropping both prefixes) regardless of their return type. This may be sufficient; however, there is now a mix of `THING`, `parse_THING`, and `recognize_THING` throughout that module. We should settle on a naming convention and be consistent. This will become more important if/when we get around to implementing #322.

PR #331, opened 2021-10-10
Introduces `AnnotatedTextValue` type

### LINKED ISSUE #330: Replace the text reader's (annotations, item) tuple with a struct
When the text reader gained the ability to parse annotation sequences, a number of functions that worked with `TextStreamItem` (effectively: an un-annotated Ion value) started working with `(Vec<OwnedSymbolToken>, TextStreamItem)` instead, representing an item and its associated annotations.

This new type is pretty unwieldy; it gets even worse when wrapped in other types like `IResult<Option<(Vec<OwnedSymbolToken>, TextStreamItem)>>`.

We should make a proper two-field struct called `TextValue` that we can use instead. (Real name TBD. This type is not going to be part of the public API, so the stakes are low.)

zslayton:
NOTES:
* This PR does not include any logic changes. It's entirely renames and mechanical refactoring.
* It depends on the changes in PR #329. When 329 is merged, this PR will be rebased against it.
---------

* Now that `TextStreamItem` only represents Ion values (and not
  comments, container ends, or EOF), it has been renamed to
  `TextValue`.
* The pervasive tuple type `(Vec<OwnedSymbolToken>, TextStreamItem)`
  has been replaced by an `AnnotatedTextValue` struct that combines
  a `TextValue` with an `Vec<OwnedSymbolToken>`. It also provides
  helper methods for construction and conversions.
* Several parser names have been simplified by removing `recognize_`
  and `parse_` as prefixes. This is part of the work needed to
  address #327.
* Replaced `stream item` in variable/method names with `value`.

Fixes #330.

### DIFF
--- a/src/text/mod.rs
+++ b/src/text/mod.rs
@@ -4,4 +4,3 @@ mod text_buffer;
 mod text_data_source;
 mod text_value;
 pub mod writer;
-


src/text/mod.rs
```diff
@@ -2,53 +2,5 @@ pub(in crate::text) mod parsers;
...
-
-use crate::types::decimal::Decimal;
-use crate::types::timestamp::Timestamp;
-use crate::value::owned::OwnedSymbolToken;
-use crate::IonType;
-
-/// Represents a single item encountered in a text Ion stream. The enum includes variants for each
-/// scalar type as well as variants for the beginning and end of each container type.
-#[derive(Debug, Clone, PartialEq)]
-pub(crate) enum TextStreamItem {
```
zslayton, line 15: This type has been renamed `TextValue` and moved to its own module.

src/text/reader.rs
```diff
@@ -393,149 +404,89 @@ mod reader_tests {
...
-    }
-
-    fn top_level_value_test(ion_text: &str, expected: TextStreamItem) {
-        let mut reader = TextReader::new(ion_text);
-        let item = reader.next().unwrap().unwrap().1;
-        assert_eq!(item, expected);
-    }
-
-    #[test]
-    fn test_read_single_top_level_values() -> IonResult<()> {
```
zslayton, line 516: This test was moved to the top of the module and refactored to use `#[rstest]`.

desaikd (approved):
Looks good!

PR #336, opened 2021-10-13
Implements the `RawReader` trait for the `RawTextReader`

zslayton:
This PR depends on the work in #332; when that PR is merged, this one will be rebased against `main`.

-----

This PR renames a number of constructs:

* The `Cursor` trait is now the `RawReader` trait to better align with
   terminology used in other libraries.
* The `TextReader` is now the `RawTextReader` as it does not provide
   any symbol management.
* The `BinaryIonCursor` (which implements `RawReader`) is now the
  `RawBinaryReader`.

Prior to this patch, the `RawReader` trait used `SymbolId` as the
return type for methods that read annotations, struct field names,
and symbols because that was/is the only kind of symbol token
that appears in binary. Now that we have a `RawTextReader`, we
also need to be able to read symbol tokens that define their text
inline. This change set introduces a `RawSymbolToken` enum that
can store either a `SymbolId` or a `String`. The `RawReader` trait
has been updated to use `RawSymbolToken` throughout.

Finally, this PR implements `RawReader` for the `RawTextReader`,
making it possible to navigate an Ion text stream using the expected
`next()`, `step_in()`, `step_out()`, and `read_TYPE()` method calls.


src/binary/raw_binary_reader.rs
```diff
@@ -250,23 +250,24 @@ pub struct CursorState {
     // All of the annotations on values in `parents` and the current value.
     // Having a single, reusable Vec reduces allocations and keeps the size of
     // the EncodedValue type (which is frequently moved) small.
-    annotations: Vec<SymbolId>,
+    annotations: Vec<RawSymbolToken>,
 }
 
 /// Verifies that the current value is of the expected type and that the bytes representing that
 /// value have not yet been consumed from the data source. This macro is called by the
 /// BinaryCursor#read_{typename} methods.
```
desaikd, line 258: Over here and some other places in comments there is mention of `BinaryCursor`/ `Cursor` which can now be changed to `RawBinaryReader`/`RawReader`.

src/reader.rs
```diff
@@ -30,7 +30,7 @@ pub struct Reader<C: Cursor> {
 // macro, making it impossible to apply this #[allow(deprecated)] more narrowly. When `read_datetime`
 // is removed, this annotation should be removed too.
 #[allow(deprecated)]
-impl<C: Cursor> Reader<C> {
+impl<C: RawReader> Reader<C> {
     pub fn new(cursor: C) -> Reader<C> {
```
desaikd, line 34: Same as above this can be renamed to `raw_reader` instead.

PR #341, opened 2021-11-16
reexport bigdecimal

### LINKED ISSUE #302: Re-export Dependencies for Public APIs
As indicated on [this issue][1], we should be not exporting data types in our public APIs of our dependencies.  This will allow for multiple versions of a dependency to more easily co-exist in down stream dependents.

For example, instead of directly returning `bigdecimal::BigDecimal` in our APIs, we should return something like `ion_rs::bigdecimal::BigDecimal` which in our `lib.rs`, might look something like:

```rust
pub use bigdecimal;
```

We may want to put these re-exports in a sub-module to emphasize that we are just re-exporting them, but other candidates for this are `chrono`, `num-bigint`, `num-traits`, and even `ion-c-sys` when such types are part of our public API.

Note that this would require us to actually change any of the APIs that use these types to use the re-exported ones instead, but that should be a fairly mechanical change.

[1]: https://github.com/awslabs/amazon-qldb-shell/pull/135#issuecomment-908631701

marcbowes:
This commit pub-uses bigdecimal so that consumers can use the bigdecimal
type without have explicit exact-version dependency.

bigdecimal is a <1.0 library, and so cargo will pull in additional
copies of it by default, as semver says each release of a <1.0 crate is
backwards incompatible. So, if a consumer of ion-rust didn't have the
exact right version, then the `TryFrom<Decimal> for BigDecimal`
implementation wouldn't be accessible.

Fixes #302

-------

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -35,3 +35,7 @@ pub use reader::Reader;
 pub use symbol_table::SymbolTable;
 pub use system_event_handler::SystemEventHandler;
 pub use types::IonType;
+
+pub mod reexport {
+    pub use bigdecimal;
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -36,6 +36,9 @@ pub use symbol_table::SymbolTable;
 pub use system_event_handler::SystemEventHandler;
 pub use types::IonType;
 
-pub mod reexport {
+/// Re-exports of third party dependencies that are part of our public API.
+///
+/// See also: https://github.com/amzn/ion-rust/issues/302.
+pub mod external {
     pub use bigdecimal;
 }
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -35,3 +35,7 @@ pub use reader::Reader;
 pub use symbol_table::SymbolTable;
 pub use system_event_handler::SystemEventHandler;
 pub use types::IonType;
+
+pub mod reexport {
```
almann, line 39: I would just spell this a `ext` or if we really want to be verbose `external`.  Mentioned that spelling in #302, but I am not going to veto if we prefer `reexport`.

marcbowes: Yeah, I left some thoughts in that issue :)

I'm not in ❤️ with `reexport`, but `ext` is more confusing.

marcbowes: per slack, going with external. we agree with top-level..

almann (approved):

zslayton (approved):

PR #354, opened 2022-02-10
Experimenting with a "packed" VarUInt header

zslayton:
This is an encoding/performance experiment and should NOT be merged.

The `VarUInt` encoding primitive requires applications to perform bitwise operations on each byte of the value being read/written; it is inherently handled byte-at-a-time. @jobarr-amzn observed that they might be faster/easier to work with if all of the continuation bits were at the beginning of the value. This branch contains an implementation of that idea. 

The resulting data size is identical. However, read and write speeds were only modestly better than the original format.

image

(Ignore the notes about 'regression' or 'no change'; what matters is the comparison between `Classic write VarUInt`/`New write VarUInt` and `Classic read VarUInt`/`New read VarUInt`.)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,11 @@ ion-c-sys = { path = "./ion-c-sys", version = "0.4" }
 
 [dev-dependencies]
 rstest = "0.9"
+criterion = "0.3"
+
+[[bench]]
+name = "var_uint_experiment"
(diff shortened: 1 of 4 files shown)


zslayton:
Closing this PR. The results of the experiment will still be visible down the line if we want to revisit it.

PR #360, opened 2022-04-18
Adds user-level `TextWriter` and `BinaryWriter`

zslayton:
This PR introduces user-level wrappers for the `RawBinaryWriter`
and the `RawTextWriter`. These higher-level types manage a
symbol table and so can convert between text and symbol IDs,
adding new symbols when appropriate.

This PR also renames the `RawWriter` trait to `Writer`, as the
methods in its interface are applicable to both the raw and user
views of data.

While this PR adds basic unit test for the new wrapper types, it
stops short of providing exhaustive correctness checking. One or
more follow-on PRs will add full integration with ion-tests,
demonstrating spec conformance.

### DIFF
--- a/src/text/text_writer.rs
+++ b/src/text/text_writer.rs
@@ -1,0 +1,161 @@
+use crate::raw_symbol_token_ref::{AsRawSymbolTokenRef, RawSymbolTokenRef};
+use crate::result::{illegal_operation, IonResult};
+use crate::text::raw_text_writer::RawTextWriter;
+use crate::types::decimal::Decimal;
+use crate::types::timestamp::Timestamp;
+use crate::writer::Writer;
+use crate::IonType;
+use crate::SymbolTable;
+use delegate::delegate;
+use std::io::Write;
+
+/**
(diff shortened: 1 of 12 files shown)


zslayton:
The CI build failed because `cargo clippy` is complaining about code produced by `bindgen` in the `ion-c-sys` subcrate. It is not related to this change.

image

I'll investigate that issue separately.

src/text/text_writer.rs
```diff
@@ -0,0 +1,161 @@
...
+        TextWriter {
+            raw_writer,
+            symbol_table: SymbolTable::new(),
+        }
+    }
+}
+
+impl<W: Write> Writer for TextWriter<W> {
+    fn supports_text_symbol_tokens(&self) -> bool {
+        true
```
desaikd, line 33: Should we add a comment around the return value here same as the `RawBinaryWriter`?

desaikd (approved):
Looks good! Thanks for putting all the detailed comments, those were helpful.

src/text/text_writer.rs
zslayton: Added. 👍

Issue #335, opened 2021-10-12
Replace closure-based Reader methods like `string_ref_map`


zslayton:
Several methods in the `Cursor`/`SystemReader` trait were implemented before [Non-Lexical Lifetimes (NLL)](https://blog.rust-lang.org/2018/12/06/Rust-1.31-and-rust-2018.html#non-lexical-lifetimes) were stabilized in Rust 1.31. Back then, borrowing a reference to something could really complicate downstream code, requiring the programmer to add extra scopes to clarify when a reference would stop being used.

The methods below allowed the user to specify a closure that take a reference as a parameter; the user could read/copy/analyze the data and return it without scoping issues.

https://github.com/amzn/ion-rust/blob/739d74c57347746fdde3ab5a8844ed824c2b5662/src/cursor.rs#L75-L91

https://github.com/amzn/ion-rust/blob/739d74c57347746fdde3ab5a8844ed824c2b5662/src/cursor.rs#L99-L104

https://github.com/amzn/ion-rust/blob/739d74c57347746fdde3ab5a8844ed824c2b5662/src/cursor.rs#L109-L114

Now that we have NLL, this approach is no longer necessary. It'd be simpler to just give out a reference to the data.

zslayton:
In my work on #345 I found that NLL was not the only reason we had created the `TYPE_ref_map` methods.

The raw binary reader wraps a provided `BufRead` implementation. When it goes to read the next Ion value from the stream, it needs all of that value's bytes to be in a contiguous buffer. It uses [two different approaches to achieve this](https://github.com/amzn/ion-rust/blob/65087108c0f2f5615d645d6e1bc0315372069294/src/data_source.rs#L155-L199):
1. First, it checks to see if all of that value's bytes are already in the `ReadBuf`'s input buffer. If so, we're ready to parse those bytes.
2. On the other hand, if only some of the bytes are in the input buffer (which is relatively rare), it must load all of the bytes into a separate `Vec<u8>` that it owns. This can involve several calls to `read()` to get more bytes.

This second step is necessary because the Ion reader doesn't control the input buffer. If it's not big enough to hold the next input value, the reader can't reallocate the buffer to fix that. Instead, it manages its own backup.

If we rewrite the `IonDataSource` trait to be a struct that controls its own buffer, we'll be able to replace methods like `string_ref_map()` with methods like `read_str(&mut self) -> &str`. This is also a prerequisite for addressing #351.

zslayton:
[...]

PR #366, opened 2022-04-27
Adds `Symbol` type, `StreamReader` trait

### LINKED ISSUE #345: Create a unifying Reader trait for the Raw/System/User levels
The observations below are @almann's thoughts from PR #344.

-----

**TL;DR:** These are some API ergonomics that the addition of the `SystemReader` has brought to light for me.  I don't think this should block merging, but serve as some observations for future work around the API.  I will cut an issue from this missive.

----

I noticed our APIs around this are slightly non-orthogonal.   For example:

```rust
// in trait RawReader
fn next(&mut self) -> IonResult<Option<RawStreamItem>>;

// in SystemReader
pub fn next(&mut self) -> IonResult<Option<SystemStreamItem>> { ... }

// in Reader
pub fn next(&mut self) -> IonResult<Option<(IonType, bool)>> { ... }
```

These enum definitions also are related semantically but not in code:

```rust
pub enum RawStreamItem {
    VersionMarker(u8, u8),
    Value(IonType, bool),
}

pub enum SystemStreamItem {
    VersionMarker(u8, u8),
    SymbolTableData(IonType, bool),
    Value(IonType, bool),
}

// not actually modeled as enum, but as a tuple in the actual API
// added here because it is effectively modeling a part of the above
pub enum StreamItem {
    Value(IonType, bool)
}
```

So my question is this--should we eliminate `Option` and align these enums?  E.g. (spelling nonwithstanding),

```rust
pub enum RawStreamItem {
    VersionMarker(u8, u8),
    Value(IonType, bool),
    Nothing,
}

pub enum SystemStreamItem {
    VersionMarker(u8, u8),
    SymbolTableData(IonType, bool),
    Value(IonType, bool),
    Nothing,
}

pub enum UserStreamItem {
    Value(IonType, bool),
    Nothing,
}

pub trait StreamItem {
    // either marker trait or nice ways to normalize the variants above...
    ...
}
```

It is debatable to me if we want to model `Nothing` as a value that is either `End`/`Before` or something--I think that state difference _could_ be useful.

This could make it so that we can define what is now the `RawReader` trait as just `Reader` with an associated type `type Item: StreamItem;` or some such and make `next(&mut self) -> Self::Item`.  This way the higher level implementations (`Reader`/`SystemReader`) and the underlying implementations (`RawBinaryReader`/`RawTextReader`) can all implement the trait.  I think we'd probably need an associated type for the symbol token as well.

Technically the generalization of the trait doesn't required my flattening above (nor the bounds on the associated type), but I think it might make the API a bit cleaner and less wrapped in nested enums.

Another alternative would be to wrap the enums in common, which keeps the definitions associated with one another, but I think that is uglier in net.

All of this is a long winded way of saying we should reconsider how we factor the stream item enum and factor the `RawReader` trait such that it can be easily used by all implementations.  The benefits are pretty clear on the latter (the enum flattening is optional for a more generic trait), we can actually now write generic code against the system reader and the regular reader and the raw readers (which allows someone else to do decoration or other interesting things).

_Originally posted by @almann in https://github.com/amzn/ion-rust/pull/344#discussion_r779139401_

zslayton:
Fixes #345.

Note: this diff looks very large, but it's comprised mostly of superficial 
changes. Many methods that used to return `IonResult<Option<T>>`
now simply return `IonResult<T>`, which had sweeping (but simple) effects.

-------

This PR makes the following high-level changes:

1. The `SymbolTable` has been modified to store `Rc<str>` text
values instead of `String`s, allowing the text they represent
to be shared cheaply.
2. It introduces a new `Symbol` type that represents the resolved
text of a field name, annotation, or symbol value.
3. It adds a unifying `StreamReader` trait implemented by all
reader types.
4. It removes methods that worked with the external `DateTime` and
`BigDecimal` types. These methods were deprecated in v0.6.1.

### Changes to the `SymbolTable`

The `SymbolTable` has been modified to store `Rc<str>` entities
that can be cheaply cloned and passed around rather than `String`s
that require more heap allocations to copy/share.

### The `Symbol` data type

A `Symbol` represents the text of a fully resolved Ion symbol, field
name, or annotation. It is distinct from the various 'token' types in
that it does not expose information about how the symbol was originally
encoded in the stream. It aims to provide an ergonomic,
application-level text type with a static lifetime.

If the reader encounters a text symbol token in the stream, a `Symbol`
can be constructed with that token's text stored as a `String`.

If the reader encounters a symbol ID token in the stream, a `Symbol`
can be constructed with an `Rc<str>` pointing to that ID's corresponding
text in the `SymbolTable`.

### The `StreamReader` trait

`StreamReader` is a trait that captures the format-agnostic
functionality needed to navigate through an Ion stream at any level of
abstraction.

`StreamReader` has two associated types:
1. `Item`: a type that represents the next entity found in the
   stream being read. This varies according to the implementor's
   level of abstraction.
2. `Symbol`: a type that represents a field name, annotation, or
   Ion symbol value at this level of abstraction. A `RawReader`
   would return a `RawSymbolToken` (a `SymbolId` OR `String`),
   allowing for the possibility that the symbol's text still needs
   to be resolved. In contrast, a `Reader` would return a fully
   resolved `Symbol` (a `String` or `Rc<str>` from the symbol table).

`StreamReader` is implemented by all of the following readers:
   * `RawBinaryReader`
   * `RawTextReader`
   * `SystemReader`
   * `Reader`

What was previously the `RawReader` trait is now an alias for
`StreamReader<Item=RawStreamItem, Symbol=RawSymbolToken>`.

The `StreamReader` changes the return type of most methods in the old
reader implementations from `IonResult<Option<T>>` to simply
`IonResult<T>`. This simplifies usage in the 'happy path' substantially.

We previously used `Option::None` to represent no-more-values or to
indicate that the current value could not be read with a given `read_*`
method. (e.g. An `int` could not be read with `read_string`.)
No-more-values is now represented by a new enum variant called `Nothing`
in the `StreamItem`, `SystemStreamItem`, `RawStreamItem` enums. Calling
a `read_*` method with an expected type that does not match the current
value's type will result in an `IonError::IllegalOperation`.

All `StreamReader` methods that require symbol resolution (`field_name()`,
`annotations()`) now return an `IonResult<Self::Symbol>` to allow for
the possibility of an `IonError::DecodingError` caused by undefined
text. Previously, such an error would cause a `panic!`.


src/binary/raw_binary_reader.rs
```diff
@@ -228,6 +228,8 @@ where
  */
 #[derive(Clone, Debug)]
 pub struct CursorState {
+    // Indicates whether the reader is positioned over an IVM, a value, or nothing.
+    current_item: RawStreamItem,
```
zslayton, line 232: PR tour: the raw binary reader now tracks its current `RawStreamItem` (discovered during calls to `next()`) in case the user later requests it with `reader.current()`.

This is analogous to the way ion-java's user-level `IonReader` has a [`IonType next()`](https://www.javadoc.io/doc/com.amazon.ion/ion-java/latest/com/amazon/ion/IonReader.html#next--) and an [`IonType getType()`](https://www.javadoc.io/doc/com.amazon.ion/ion-java/latest/com/amazon/ion/IonReader.html#getType--), but allows for stream item types other than `Value`.

src/binary/raw_binary_reader.rs
```diff
@@ -259,7 +261,7 @@ macro_rules! read_safety_checks {
         if $raw_binary_reader.cursor.value.ion_type != $ion_type
             || $raw_binary_reader.cursor.value.is_null
         {
-            return Ok(None);
+            return Err($raw_binary_reader.expected(format!("{} value", $ion_type)));
```
zslayton, line 264: Calling `read_string()` on an `int` (or other type-mismatched operations) now returns `IonError::IllegalOperation` instead of `Ok(None)`.

src/binary/raw_binary_reader.rs
```diff
@@ -365,12 +371,16 @@ impl<R: IonDataSource> RawReader for RawBinaryReader<R> {
...
         self.cursor.value.index_at_depth = self.cursor.index_at_depth;
 
-        Ok(Some(RawStreamItem::Value(
+        Ok(self.set_current_item(RawStreamItem::Value(
             self.cursor.value.ion_type,
             self.is_null(),
         )))
     }
 
+    fn current(&self) -> Self::Item {
```
zslayton, line 380: PR tour: doc comments for the new methods live on the trait and not on the impls.

src/binary/raw_binary_reader.rs
```diff
@@ -379,36 +389,53 @@ impl<R: IonDataSource> RawReader for RawBinaryReader<R> {
         self.cursor.value.is_null
     }
 
-    fn annotations(&self) -> &[RawSymbolToken] {
+    fn annotations<'a>(&'a self) -> Box<dyn Iterator<Item = IonResult<Self::Symbol>> + 'a> {
```
zslayton, line 392: PR tour: when [GATs land](https://blog.rust-lang.org/2021/08/03/GATs-stabilization-push.html) in May/June, we can replace this with an associated type whose lifetime is bound to `'a`. Then it won't require heap allocation and virtual dispatch.

src/binary/raw_binary_reader.rs
```diff
@@ -478,105 +504,78 @@ impl<R: IonDataSource> RawReader for RawBinaryReader<R> {
         let coefficient = self.read_int(coefficient_size_in_bytes)?;
 
         if coefficient.is_negative_zero() {
-            return Ok(Some(Decimal::negative_zero_with_exponent(exponent)));
-        }
-
-        Ok(Some(Decimal::new(coefficient.value(), exponent)))
-    }
-
-    fn read_big_decimal(&mut self) -> IonResult<Option<BigDecimal>> {
```
zslayton, line 487: PR tour: the read/write methods for the external `BigDecimal` type were deprecated in v0.6.1 in favor of the `Decimal` type.

src/binary/raw_binary_reader.rs
```diff
@@ -478,105 +504,78 @@ impl<R: IonDataSource> RawReader for RawBinaryReader<R> {
...
     }
 
-    fn read_string(&mut self) -> IonResult<Option<String>> {
-        self.string_ref_map(|s: &str| s.into())
+    fn read_string(&mut self) -> IonResult<String> {
+        self.map_string(|s| s.to_owned())
     }
 
-    fn string_ref_map<F, T>(&mut self, f: F) -> IonResult<Option<T>>
+    fn map_string<F, U>(&mut self, f: F) -> IonResult<U>
```
zslayton, line 517: PR tour: ultimately the `map_TYPE` methods will be removed in favor of more straightforward ones. They're an artifact of the way the reader's internal buffering works and lifetime/bookkeeping rules that were a bigger deal before Non-Lexical Lifetimes (NLL) became a thing.

See: #335

src/binary/raw_binary_reader.rs
```diff
@@ -663,96 +662,7 @@ impl<R: IonDataSource> RawReader for RawBinaryReader<R> {
...
         }?;
 
-        Ok(Some(timestamp))
-    }
-
-    // This method will return year-, month-, and day-precision timestamps as UTC DateTimes
-    // with smaller time units (hour, minute, etc) zeroed out. Longer term, this method
-    // will be replaced by a `read_timestamp` method that returns a proper Timestamp.
-    // TODO: https://github.com/amzn/ion-rust/issues/308
-    fn read_datetime(&mut self) -> IonResult<Option<DateTime<FixedOffset>>> {
```
zslayton, line 673: PR tour: the read/write methods for the external `DateTime` type were deprecated in v0.6.1 in favor of the `Timestamp` type.

src/binary/raw_binary_reader.rs
```diff
@@ -1176,24 +1098,32 @@ where
...
+            self.current()
+        ))
+    }
 }
 
 #[cfg(test)]
 mod tests {
     use std::io;
 
-    use bigdecimal::BigDecimal;
```
zslayton, line 1185: PR tour: the unit tests were updated to use the new `StreamReader` API, so lots of `Option` wrappers were removed. Tests related to `read_datetime` and `read_big_decimal` were also removed.

src/binary/raw_binary_writer.rs
```diff
@@ -369,34 +367,6 @@ impl<W: Write> RawBinaryWriter<W> {
         self.num_annotations_current_value > 0
     }
 
-    /// Writes an Ion decimal with the specified value.
```
zslayton, line 372: PR tour: removing deprecated methods.

src/raw_reader.rs
```diff
@@ -1,152 +1,47 @@
...
-use crate::result::IonResult;
-use crate::types::decimal::Decimal;
-use crate::types::timestamp::Timestamp;
+use crate::stream_reader::StreamReader;
 use crate::types::IonType;
-use bigdecimal::BigDecimal;
-use chrono::{DateTime, FixedOffset};
+use std::fmt::{Display, Formatter};
 
-/**
```
zslayton, line 9: PR tour: this trait has been subsumed by the `StreamReader` trait. See just below this large deletion for the new `RawReader` trait, which is a simple alias for `StreamReader<Item=RawStreamItem, Symbol=RawSymbolToken>`.

src/reader.rs
```diff
@@ -1,79 +1,57 @@
...
 ///
 /// Reader itself is format-agnostic; all format-specific logic is handled by the
-/// wrapped Cursor implementation.
+/// wrapped [RawReader] implementation.
 pub struct Reader<R: RawReader> {
     raw_reader: R,
     symbol_table: SymbolTable,
 }
 
-// FIXME: The `read_datetime` method is deprecated. However, this function body is generated by a
```
zslayton, line 26: PR tour: the application-level `Reader` has been changed from a standalone struct to a struct that implements `StreamReader`.

src/stream_reader.rs
```diff
@@ -0,0 +1,172 @@
...
+pub trait StreamReader {
+    /// The type returned by calls to [next], indicating the next entity in the stream.
+    /// Reader implementations representing different levels of abstraction will surface
+    /// different sets of encoding artifacts. While an application-level Reader would only surface
+    /// Ion values, a lower level Reader might surface symbol tables, Ion version markers, etc.
+    type Item;
+
+    /// The types used to represent field names, annotations, and symbol values at this Reader's
+    /// level of abstraction.
+    type Symbol;
```
zslayton, line 22: PR tour: when GATs land, we can make this generic over the lifetime of the method that constructs it. This will allow us to return (e.g.) `&str` comprised of bytes in the input buffer rather than a full `String`.

src/symbol_table.rs
```diff
@@ -1,14 +1,118 @@
...
+    fn eq(&self, other: &Symbol) -> bool {
+        self.as_str() == other.as_ref()
+    }
+}
+
+impl PartialEq<Symbol> for &str {
+    fn eq(&self, other: &Symbol) -> bool {
+        self == &other.as_ref()
+    }
+}
```
zslayton, line 74: PR tour: these impls allow `Symbol` instances to be compared to other text types in (e.g.) an `assert_eq!`, which streamlines a lot of unit tests.

src/symbol_table.rs
```diff
@@ -1,14 +1,118 @@
...
+
 /// Stores mappings from Symbol IDs to text and vice-versa.
 // SymbolTable instances always have at least system symbols; they are never empty.
 #[allow(clippy::len_without_is_empty)]
 pub struct SymbolTable {
-    symbols_by_id: Vec<Option<String>>,
-    ids_by_text: HashMap<String, SymbolId>,
+    symbols_by_id: Vec<Option<Symbol>>,
```
zslayton, line 115: PR tour: the rest of the changes in this file modify the `SymbolTable` to store `Rc<str>` (by way of `Symbol`) instead of `String`s. This makes it cheap/easy to hand out `Symbol` clones without a lifetime or requiring new heap allocations.

zslayton (commented):
I've added my own review comments as a sort of 'PR tour' -- hopefully this makes the diff easier to digest.

src/symbol_table.rs
```diff
@@ -1,14 +1,119 @@
...
+use std::ops::Deref;
+use std::rc::Rc;
 
 use crate::constants::v1_0;
 use crate::types::SymbolId;
 
+/// Stores or points to the text of a given [Symbol].
+#[derive(Debug, Eq)]
+enum SymbolText {
+    // This Symbol refers to a string in the symbol table
```
desaikd, line 13: Should we specify here whether it points to local symbol table or shared symbol tables or both? thoughts?

src/text/raw_text_reader.rs
```diff
@@ -371,6 +374,14 @@ impl<T: TextIonDataSource> RawTextReader<T> {
 
         value
     }
+
+    fn expected<E: Display>(&self, expected: E) -> IonError {
```
desaikd, line 378: Maybe a developer comment on how this is being used?

desaikd (commented):
Some minor suggestions on adding comments around the significance of some methods(use this only for the once that doesn't have a trait comment already),
Otherwise looks good!

src/raw_reader.rs
```diff
@@ -1,152 +1,47 @@
...
+    /// * after the reader has read the last item in a container
+    Nothing,
+}
+
+impl Display for RawStreamItem {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        use RawStreamItem::*;
+        match self {
+            VersionMarker(major, minor) => write!(f, "ion version marker (v{}.{})", major, minor),
+            Value(ion_type, true) => write!(f, "null.{}", ion_type),
```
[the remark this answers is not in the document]
zslayton: Great question. I've opened #368 to look into alternatives to `bool`.

src/binary/raw_binary_reader.rs
```diff
@@ -822,6 +732,10 @@ impl<R: IonDataSource> RawReader for RawBinaryReader<R> {
         Ok(())
     }
 
+    fn parent_type(&self) -> Option<IonType> {
```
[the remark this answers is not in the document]
zslayton: There's a doc comment in the trait:

https://github.com/amzn/ion-rust/blob/e0442b8acfe90bf567dcc62a858ddb0c6c3f09ad/src/stream_reader.rs#L164-L166

src/symbol_table.rs
zslayton: A recent clarification to the Ion spec says that at the application level, symbols must be represented as text. If a symbol ID is encountered that cannot be resolved to text, the reader can raise an error. (Readers can _try_ to keep processing the data without the relevant text if the task at hand doesn't require it, but are not required to do so by the spec. That's an API decision.)

In light of this distinction, I think we should either:
* Ask folks who are interested in surfacing lower-level encoding artifacts like symbol tables and IDs to use a different interface (like `SystemReader`, maybe?)
  AND/OR
* Wait until someone has a use case for storing provenance in the `Symbol` itself

In either case, I'd like to hold off on trying to add that now.

desaikd (approved):
🚀

Issue #364, opened 2022-04-21
Implement `precision` and `offset` methods for `Timestamp`


desaikd:
Currently `Timestamp` type doesn't have any functions returning the `precision` and `offset` of the `Timestamp` value. These method will be useful inside `ion-schema-rust` to implement [precision](https://amzn.github.io/ion-schema/docs/spec.html#timestamp_precision) and [offset](https://amzn.github.io/ion-schema/docs/spec.html#timestamp_offset) constraints.

PR #369, opened 2022-04-28
adds precision and offset methods to Timestamp type

desaikd:
*Issue #364*

*Description of changes:*
This PR works on adding precision and offset methods to Timestamp type

*Test:*
Adds unit tests for precision and offset.

### DIFF
--- a/src/types/timestamp.rs
+++ b/src/types/timestamp.rs
@@ -111,13 +111,10 @@ impl Timestamp {
         // This function is used when comparing two Timestamps with different Mantissa representations.
         use Mantissa::*;
         match self.fractional_seconds.as_ref() {
-            // This timestamp stores its fractional seconds in its `date_time` field.
-            // We'll need to convert the date_time's nanoseconds to a Decimal and return it's scale.
-            Some(Digits(number_of_digits)) => {
-                let coefficient = first_n_digits_of(*number_of_digits, self.date_time.nanosecond());
-                let scale = (coefficient as f64).log10().ceil() as i64;
-                Some(scale)
-            }
+            // number_of_digits represent number of digits of precision in the Timestamp's fractional seconds.
+            // this is equivalent to the decimal scale when we convert the fractional seconds into a decimal
+            // and return its scale
+            Some(Digits(number_of_digits)) => Some(*number_of_digits as i64),
             // This timestamp already stores its fractional seconds as a Decimal; return the scale of this Decimal.
             Some(Arbitrary(decimal)) => Some(decimal.scale()),
             // This Timestamp's precision is too low to have a fractional seconds field.
@@ -273,9 +270,10 @@ impl Timestamp {
         FractionalSecondSetter { builder }
     }
 
-    /// Returns the offset that has been specified in the [Timestamp].
-    pub fn offset(&self) -> Option<FixedOffset> {
-        self.offset
+    /// Returns the offset in minutes that has been specified in the [Timestamp].
+    /// A positive value indicates Eastern Hemisphere, while a negative value indicates Western Hemisphere.
+    pub fn offset(&self) -> Option<i32> {
+        self.offset.map(|offset| offset.local_minus_utc() / 60)
     }
 
     /// Returns the precision that has been specified in the [Timestamp].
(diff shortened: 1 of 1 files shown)


src/types/timestamp.rs
```diff
@@ -103,6 +103,28 @@ impl Timestamp {
...
+    /// If the precision is [Precision::FractionalSeconds], returns a Decimal scale
+    /// of this Timestamp's fractional seconds; otherwise, returns None.
+    ///
+    /// For example, a Timestamp with 553 milliseconds would return a Decimal scale of 3.
+    pub fn fractional_seconds_scale(&self) -> Option<i64> {
+        // This function is used when comparing two Timestamps with different Mantissa representations.
+        use Mantissa::*;
+        match self.fractional_seconds.as_ref() {
+            // This timestamp stores its fractional seconds in its `date_time` field.
+            // We'll need to convert the date_time's nanoseconds to a Decimal and return it's scale.
```
zslayton, line 115: I don't think this converts it to a Decimal, does it?

src/types/timestamp.rs
```diff
@@ -250,6 +272,16 @@ impl Timestamp {
             .into_builder();
         FractionalSecondSetter { builder }
     }
+
+    /// Returns the offset that has been specified in the [Timestamp].
+    pub fn offset(&self) -> Option<FixedOffset> {
+        self.offset
```
zslayton, line 278: I'd like to avoid returning an external type as part of the public API. (`FixedOffset` lives in `chrono`.) I think this should be:
```
pub fn offset_minutes(&self) -> Option<i32>;
```
(Be sure to document what positive and negative minutes mean.)

src/types/timestamp.rs
```diff
@@ -1169,6 +1202,54 @@ mod timestamp_tests {
...
+    fn test_timestamp_fractional_seconds_scale() -> IonResult<()> {
+        // Set fractional seconds as Decimal
+        let timestamp_with_micro_seconds = Timestamp::with_ymd_hms(2021, 4, 6, 10, 15, 0)
+            .with_fractional_seconds(Decimal::new(553u64, -6))
+            .build_at_offset(-5 * 60)?;
+
+        assert_eq!(
+            timestamp_with_micro_seconds
+                .fractional_seconds_scale()
+                .unwrap(),
```
zslayton, line 1235: Could you add one that checks for `None` when the precision is too low?

src/types/timestamp.rs
desaikd: Sure. That makes sense. I will change this.

src/types/timestamp.rs
```diff
@@ -1247,6 +1245,16 @@ mod timestamp_tests {
                 .unwrap(),
             3
         );
+
+        // Set a fractional seconds as Decimal with low precision
+        let timestamp_with_micro_seconds =
```
zslayton, line 1250:
```suggestion
        let timestamp_with_seconds =
```

zslayton (approved):

src/types/timestamp.rs
```diff
@@ -1169,6 +1200,64 @@ mod timestamp_tests {
...
+            3
+        );
+
+        // Set a fractional seconds as Decimal with low precision
+        let timestamp_with_seconds =
+            Timestamp::with_ymd_hms(2021, 4, 6, 10, 15, 0).build_at_offset(-5 * 60)?;
+
+        // For low precision fractional_seconds_scale should return a None
+        assert_eq!(
+            timestamp_with_micro_seconds.fractional_seconds_scale(),
```
zslayton, line 1255:
```suggestion
            timestamp_with_seconds.fractional_seconds_scale(),
```

src/types/timestamp.rs
```diff
@@ -1169,6 +1200,64 @@ mod timestamp_tests {
...
+
+        // Set a fractional seconds as Decimal with low precision
+        let timestamp_with_seconds =
+            Timestamp::with_ymd_hms(2021, 4, 6, 10, 15, 0).build_at_offset(-5 * 60)?;
+
+        // For low precision fractional_seconds_scale should return a None
+        assert_eq!(
+            timestamp_with_seconds.fractional_seconds_scale(),
+            None
+        );
```
zslayton, line 1257:
```suggestion
        assert_eq!(timestamp_with_seconds.fractional_seconds_scale(), None);
```

PR #370, opened 2022-04-29
Null stream item

### LINKED ISSUE #368: Find a clearer way to model typed Null values
Currently the `StreamItem`, `SystemStreamItem`, and `RawStreamItem` enums have a variant that looks like this:

```rust
//        v--- the value's ion type (int, float, timestamp, etc)
Value(IonType, bool)
//              ^--- whether the value is a null
```

In situations where you're writing code to match on it, sometimes you'll see code that looks like this:
```rust
match reader.next()? {
  Value(ion_type, true) => {},
  Value(ion_type, false) => {},
}
```

[@desaikd pointed out ](https://github.com/amzn/ion-rust/pull/366#discussion_r860147556) that in code like this, the meaning of the boolean values is not readily apparent.

Some alternatives:

**Option 1: Use an `enum` instead of `bool`**

Make a new enum like this (names notwithstanding):
```rust
enum Nullness {
  Null,
  NonNull
}
```
change the `StreamItem::Value` variant to look like this:
```rust
Value(IonType, Nullness)
```
and write destructuring code like this:
```rust
match reader.next()? {
  Value(ion_type, Null) => {},
  Value(ion_type, NonNull) => {},
}
```

**Option 2: Break `Value` into two enum variants**

Make `StreamItem::Value` into two variants:
```rust
Value(IonType),
Null(IonType),
```
and write destructuring code like this:
```rust
match reader.next()? {
  Value(ion_type) => {},
  Null(ion_type) => {},
}
```

I think I prefer the latter, though dislike that it suggests `null` is not a value. We'll need to see how this refactoring will impact the rest of the codebase.

zslayton:
*Issue #, if available:* Fixes #368.

*Description of changes:*

Each implementation of `StreamReader` has an associated type called
`Item` that represents the set of possible entities that the reader may
encounter in its stream:

| Reader       |      Item        |
|--------------|------------------|
| RawReader    | RawStreamItem    |
| SystemReader | SystemStreamItem |
| Reader       | StreamItem       |

While the entities enumerated by `Item` vary according to the reader's
level of abstraction, all of them include a `Value` variant that
indicates that the reader is currently positioned over an Ion value.

Prior to this PR, each `Value` enum variant was modeled as:

      Value(IonType, bool)

in which the `IonType` represented the value's corresponding Ion data
type and the `bool` indicated whether the value was a `null`. This
lead to user code that looked like:

```rust
     match reader.next()? {
        Value(IonType::String, true)  => /* ... */,
        Value(IonType::String, false) => /* ... */,
        _ => {}
     }
```

In such situations, the meaning of the `true` or `false` was not
obvious.

This PR splits `Value(IonType, bool)` out into two variants for
each enum:

```rust
enum $NAME {
     // ...
     Value(IonType),
     Null(IonType),
     // ...
}
````

This makes most interactions with the enum both more self-documenting
and more concise.

This PR also modifies `Reader` and `SymbolTable` to be able to handle
non-`string` values in a local symbol table.

### DIFF
--- a/examples/read_all_values.rs
+++ b/examples/read_all_values.rs
@@ -25,15 +25,15 @@ fn main() -> IonResult<()> {
 // Prints the total number of values read upon completion.
 fn read_all_values<R: IonDataSource>(cursor: &mut RawBinaryReader<R>) -> IonResult<usize> {
     use IonType::*;
-    use RawStreamItem::{VersionMarker, Value, Null as NullValue, Nothing};
+    use RawStreamItem::{Nothing, Null as NullValue, Value, VersionMarker};
     let mut count: usize = 0;
     loop {
         match cursor.next()? {
             VersionMarker(_major, _minor) => {}
             NullValue(_ion_type) => {
                 count += 1;
-                continue
-            },
+                continue;
+            }
             Value(ion_type) => {
                 count += 1;
                 match ion_type {
--- a/examples/read_all_values.rs
+++ b/examples/read_all_values.rs
@@ -25,16 +25,17 @@ fn main() -> IonResult<()> {
 // Prints the total number of values read upon completion.
 fn read_all_values<R: IonDataSource>(cursor: &mut RawBinaryReader<R>) -> IonResult<usize> {
     use IonType::*;
-    use RawStreamItem::*;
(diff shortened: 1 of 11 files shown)


desaikd (approved):
Looks good 🚢

PR #372, opened 2022-05-04
Adds `NativeElementReader`, lifts int size restriction

zslayton:
This PR:
* Adds `NativeElementReader`, a pure-Rust implementation of the
`ElementReader` API. Iterates over a given byte slice and returns
materialized `OwnedElement` representations of the data.
* Modifies `tests/element_test_vectors.rs` to be able to run the
same `ion-tests` logic using both the `IonCElementReader` and the
new `NativeElementReader`. This includes the creation of a (very
long) skip list for the `NativeElementReader`.
* Renames `AnyInt` to `Integer` and moves it to the 'types' module
alongside `Decimal` and `Timestamp`.
* Renames `AnyUInt` to `UInteger` and moves it to the `types` module.
* Modifies the `UInt` decoder to be able to return `UInt`s of any
size, not just `u64`s.
* Modifies the `StreamReader` trait (and all implementations) to include
a `fn read_integer(&mut self) -> IonResult<Integer>` method that can
return an integer of any size.
* Fixes a bug in the `RawTextReader` that caused it to try and step into
container-typed `null` values.
* Adds a check to the `Timestamp` builder to make sure that the year
provided is in the range `1-9999` inclusive.

### DIFF
--- a/src/binary/uint.rs
+++ b/src/binary/uint.rs
@@ -1,11 +1,15 @@
+use num_bigint::BigUint;
 use std::io::Write;
 use std::mem;
 
 use crate::data_source::IonDataSource;
 use crate::result::{decoding_error, IonResult};
+use crate::types::integer::{Integer, UInteger};
 
-type UIntStorage = u64;
-const MAX_UINT_SIZE_IN_BYTES: usize = mem::size_of::<UIntStorage>();
+// This limit is used for stack-allocating buffer space to encode/decode UInts.
+const UINT_STACK_BUFFER_SIZE: usize = 16;
+// This number was chosen somewhat arbitrarily and could be lifted if a use case demands it.
+const MAX_UINT_SIZE_IN_BYTES: usize = 2048;
 
 /// Represents a fixed-length unsigned integer. See the
 /// [UInt and Int Fields](http://amzn.github.io/ion-docs/docs/binary.html#uint-and-int-fields)
@@ -13,7 +17,7 @@ const MAX_UINT_SIZE_IN_BYTES: usize = mem::size_of::<UIntStorage>();
 #[derive(Debug)]
 pub struct DecodedUInt {
     size_in_bytes: usize,
-    value: UIntStorage,
+    value: UInteger,
 }
 
 impl DecodedUInt {
@@ -26,23 +30,48 @@ impl DecodedUInt {
             ));
         }
 
-        // Create a stack-allocated buffer to hold the data we're going to read in.
-        let mut buffer = [0u8; MAX_UINT_SIZE_IN_BYTES];
+        if length <= UINT_STACK_BUFFER_SIZE {
+            let buffer = &mut [0u8; UINT_STACK_BUFFER_SIZE];
+            DecodedUInt::read_using_buffer(data_source, length, buffer)
+        } else {
+            // We're reading an enormous int. Heap-allocate a Vec to use as storage.
+            let mut buffer = Vec::with_capacity(length);
+            for _ in 0..length {
+                buffer.push(0u8);
+            }
+            DecodedUInt::read_using_buffer(data_source, length, buffer.as_mut_slice())
+        }
+    }
+
+    fn read_using_buffer<R: IonDataSource>(
+        data_source: &mut R,
+        length: usize,
+        buffer: &mut [u8],
+    ) -> IonResult<DecodedUInt> {
         // Get a mutable reference to a portion of the buffer just big enough to fit
         // the requested number of bytes.
         let buffer = &mut buffer[0..length];
 
-        let mut magnitude: UIntStorage = 0;
         data_source.read_exact(buffer)?;
-        for &byte in buffer.iter() {
-            let byte = u64::from(byte);
-            magnitude <<= 8;
-            magnitude |= byte;
-        }
+
+        let value = if length <= mem::size_of::<u64>() {
+            // The UInt is small enough to fit in a u64.
+            let mut magnitude: u64 = 0;
+            for &byte in buffer.iter() {
+                let byte = u64::from(byte);
+                magnitude <<= 8;
+                magnitude |= byte;
+            }
+            UInteger::U64(magnitude)
+        } else {
+            // The UInt is too large to fit in a u64; read it as a BigUInt instead
+            let magnitude = BigUint::from_bytes_be(buffer);
+            UInteger::BigUInt(magnitude)
+        };
 
         Ok(DecodedUInt {
             size_in_bytes: length,
-            value: magnitude,
+            value,
         })
     }
 
@@ -57,8 +86,8 @@ impl DecodedUInt {
 
     /// Returns the magnitude of the unsigned integer.
     #[inline(always)]
-    pub fn value(&self) -> UIntStorage {
-        self.value
+    pub fn value(&self) -> &UInteger {
+        &self.value
     }
 
     /// Returns the number of bytes that were read from the data source to construct this
@@ -66,6 +95,16 @@ impl DecodedUInt {
     #[inline(always)]
     pub fn size_in_bytes(&self) -> usize {
         self.size_in_bytes
+    }
+}
+
+impl From<DecodedUInt> for Integer {
+    fn from(uint: DecodedUInt) -> Self {
+        let DecodedUInt {
+            value,
+            .. // Ignore 'size_in_bytes'
+        } = uint;
+        Integer::from(value)
     }
 }
 
@@ -121,7 +160,7 @@ pub fn encode_uint(magnitude: u64) -> EncodedUInt {
 
 #[cfg(test)]
 mod tests {
-    use super::DecodedUInt;
+    use super::*;
     use std::io::Cursor;
 
     const READ_ERROR_MESSAGE: &str = "Failed to read a UInt from the provided cursor.";
@@ -132,7 +171,7 @@ mod tests {
         let data = &[0b1000_0000];
         let uint = DecodedUInt::read(&mut Cursor::new(data), data.len()).expect(READ_ERROR_MESSAGE);
         assert_eq!(uint.size_in_bytes(), 1);
-        assert_eq!(uint.value(), 128);
+        assert_eq!(uint.value(), &UInteger::U64(128));
     }
 
     #[test]
@@ -140,7 +179,7 @@ mod tests {
         let data = &[0b0111_1111, 0b1111_1111];
         let uint = DecodedUInt::read(&mut Cursor::new(data), data.len()).expect(READ_ERROR_MESSAGE);
         assert_eq!(uint.size_in_bytes(), 2);
-        assert_eq!(uint.value(), 32_767);
+        assert_eq!(uint.value(), &UInteger::U64(32_767));
     }
 
     #[test]
@@ -148,14 +187,18 @@ mod tests {
         let data = &[0b0011_1100, 0b1000_0111, 0b1000_0001];
         let uint = DecodedUInt::read(&mut Cursor::new(data), data.len()).expect(READ_ERROR_MESSAGE);
         assert_eq!(uint.size_in_bytes(), 3);
-        assert_eq!(uint.value(), 3_966_849);
+        assert_eq!(uint.value(), &UInteger::U64(3_966_849));
     }
 
     #[test]
-    fn test_read_uint_overflow() {
-        let data = &[0x01, 0x23, 0x45, 0x67, 0x89, 0xAB, 0xCD, 0xEF, 0x01];
+    fn test_read_uint_too_large() {
+        let mut buffer = Vec::with_capacity(MAX_UINT_SIZE_IN_BYTES + 1);
+        for _ in 0..(MAX_UINT_SIZE_IN_BYTES + 1) {
+            buffer.push(1);
+        }
+        let data = buffer.as_slice();
         let _uint = DecodedUInt::read(&mut Cursor::new(data), data.len())
-            .expect_err("This should have failed due to overflow.");
+            .expect_err("This exceeded the configured max UInt size.");
     }
 
     #[test]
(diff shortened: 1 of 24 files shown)


src/binary/uint.rs
```diff
@@ -26,23 +30,45 @@ impl DecodedUInt {
...
-        let mut buffer = [0u8; MAX_UINT_SIZE_IN_BYTES];
+        if length <= UINT_STACK_BUFFER_SIZE {
+            let buffer = &mut [0u8; UINT_STACK_BUFFER_SIZE];
+            DecodedUInt::read_using_buffer(data_source, length, buffer)
+        } else {
+            // We're reading an enormous int. Heap-allocate a Vec to use as storage.
+            let mut buffer = vec![0u8; length];
+            DecodedUInt::read_using_buffer(data_source, length, buffer.as_mut_slice())
+        }
+    }
```
zslayton, line 41: PR tour: when decoding a `UInt`, the reader will use a stack-allocated buffer for values that are `<= 16` bytes. If a particularly large `UInt` is encountered, the reader will allocate an appropriately sized `Vec` to decode it. This should be very rare indeed; however, it does occur in several of the `ion-tests`. The largest integer in `ion-tests` is some 1200+ bytes at the moment. At the top of this method, we enforce an overall maximum allowable size, currently set to 2048 bytes. This is to prevent this from becoming a denial-of-service vector.

src/binary/uint.rs
```diff
@@ -26,23 +30,45 @@ impl DecodedUInt {
...
+
+    fn read_using_buffer<R: IonDataSource>(
```
zslayton, line 43: PR tour: this helper method was broken out so the core reading logic would work with either a stack- or heap-allocated buffer.

src/binary/uint.rs
```diff
@@ -132,30 +168,34 @@ mod tests {
...
         assert_eq!(uint.size_in_bytes(), 3);
-        assert_eq!(uint.value(), 3_966_849);
+        assert_eq!(uint.value(), &UInteger::U64(3_966_849));
     }
 
     #[test]
-    fn test_read_uint_overflow() {
-        let data = &[0x01, 0x23, 0x45, 0x67, 0x89, 0xAB, 0xCD, 0xEF, 0x01];
+    fn test_read_uint_too_large() {
+        let mut buffer = Vec::with_capacity(MAX_UINT_SIZE_IN_BYTES + 1);
```
zslayton, line 192: PR tour: now that we can read an integer of (practically) any size, this test had to be rewritten to pass a `UInt` buffer of 2049 bytes, one byte in length beyond the configured maximum allowable.

src/text/parsers/integer.rs
```diff
@@ -88,28 +92,63 @@ fn base_2_integer_digits(input: &str) -> IResult<&str, &str> {
...
         recognize(preceded(opt(char('-')), base_10_integer_digits)),
-        |text| parse_i64_with_radix(text, 10).map(TextValue::Integer),
+        |text| parse_integer_with_radix(text, 10).map(TextValue::Integer),
     )(input)
 }
 
 /// Strips any underscores out of the provided text and then parses it according to the specified
 /// radix.
-fn parse_i64_with_radix(text: &str, radix: u32) -> Result<i64, ParseIntError> {
+fn parse_integer_with_radix(text: &str, radix: u32) -> IonResult<Integer> {
```
zslayton, line 101: PR tour: the `RawTextReader` had to be modified to return a `BigInt` if the int being processed was too large to fit in an `i64`.

src/text/raw_text_reader.rs
```diff
@@ -62,11 +63,12 @@ impl<T: TextIonDataSource> RawTextReader<T> {
         // If the reader's current value is the beginning of a container and the user calls `next()`,
         // we need to skip the entire container. We can do this by stepping into and then out of
         // that container; `step_out()` has logic that will exhaust the remaining values.
-        let need_to_skip_container = self
-            .current_value
-            .as_ref()
-            .map(|v| v.value().ion_type().is_container())
-            .unwrap_or(false);
+        let need_to_skip_container = !self.is_null()
```
zslayton, line 66: PR tour: this addressed a bug that caused the `RawTextReader` to try to step into `null` values with container types (`null.struct`, `null.list`, etc).

src/text/raw_text_reader.rs
```diff
@@ -1044,4 +1057,58 @@ mod reader_tests {
 
         Ok(())
     }
+
+    #[test]
```
zslayton, line 1061: PR tour: these unit tests overlap heavily with the `ion-tests` integration but do not use the `Element` (DOM) API. They were helpful in exploring some errors surfaced by `element_test_vectors.rs`.

src/text/text_data_source.rs
```diff
@@ -25,7 +25,7 @@ impl<'a> TextIonDataSource for &'a str {
     }
 }
 
-impl<'a> TextIonDataSource for &'a &[u8] {
+impl<'a> TextIonDataSource for &'a [u8] {
```
zslayton, line 28: PR tour: the second `&` was a typo that I didn't notice until the compiler brought my attention to the crazy lifetimes it was introducing.

src/types/integer.rs
```diff
@@ -0,0 +1,201 @@
+use crate::result::{decoding_error, IonError};
```
zslayton, line 1: PR tour: everything in this file was moved wholesale from the `value` module when I promoted `AnyInt` to a public-facing type. There are no logic changes.

src/types/timestamp.rs
```diff
@@ -354,6 +354,12 @@ impl TimestampBuilder {
     where
         D: Datelike + Timelike + Debug,
     {
+        if self.year == 0 || self.year > 9999 {
+            return illegal_operation(format!(
+                "Timestamp year '{}' out of range (1-9999)",
+                self.year
+            ));
+        }
```
zslayton, line 362: PR tour: this fixed several failures that were surfaced by `ion-tests`. I probably should've waited and done it in a follow-on PR, but I got excited.

src/value/ion_c_reader.rs
```diff
@@ -0,0 +1,220 @@
+use crate::result::IonResult;
```
zslayton, line 1: PR tour: the contents of this file were moved wholesale from the `value::reader` module when I introduced the `NativeElementReader`. Now the trait lives in `value::reader` while the implementations live in `value::ion_c_reader` and `value::native_reader` respectively.

tests/element_test_vectors.rs
```diff
@@ -13,56 +13,6 @@ use test_generator::test_resources;
 /// Buffer size for our writing tests around round-tripping
```
zslayton, line 13: PR tour: this file has been heavily refactored so it can run the same tests for both `IonCElementReader` and `NativeElementReader`. The logic is unchanged.

tests/element_test_vectors.rs
```diff
@@ -385,3 +365,560 @@ fn non_equivs(file_name: &str) {
...
+    use ion_rs::value::native_reader::NativeElementReader;
+    use ion_rs::value::reader::native_element_reader;
+
+    struct NativeElementApi {}
+
+    impl ElementApi for NativeElementApi {
+        type ElementReader = NativeElementReader;
+
+        fn global_skip_list() -> SkipList {
+            &[
```
zslayton, line 536: PR tour: this is the skip list for the native reader. Glancing at the failures, there seem to be a few bugs that are responsible for the lion's share. I'll start to address them in follow-on PRs.

For the moment, the `NativeElementApi` is using the native reader but the ion-c writer. Once the native reader is passing most of its `ion-tests`, I'll switch it over to a native writer. In this way I hope to minimize the research required for each bug; there should only be one untested thing being used at a time.

src/text/raw_text_reader.rs
```diff
@@ -1044,4 +1057,58 @@ mod reader_tests {
...
+        // The reader will reject the unquoted float keyword 'nan' when used as an annotation
+        let pretty_ion = br#"
+            nan::23
+        "#;
+        let mut reader = RawTextReader::new(&pretty_ion[..]);
+        let result = reader.next();
+        println!("{:?}", result);
+        assert!(result.is_err());
+        Ok(())
+    }
```
jobarr-amzn, line 1113: I found these two mildly surprising, so I went to go read first the [Type Annotations](https://amzn.github.io/ion-docs/docs/spec.html#annot) section of the Ion specification and then the [text ANTLR grammar](https://amzn.github.io/ion-docs/grammar/IonText.g4.txt).  The spec didn't call out this kind of limitation, and my read of the ANTLR grammar (disclaimer I could be bad at this, relevant fragment reproduced below) doesn't seem to support this either. 


```plain
top_level_value
    : annotation+ top_level_value
    | delimiting_entity
    // numeric literals (if followed by something), need to be followed by
    // whitespace or a token that is either quoted (e.g. string) or
    // starts with punctuation (e.g. clob, struct, list)
    | numeric_entity ws
    | numeric_entity quoted_annotation value
    | numeric_entity delimiting_entity
    // literals that are unquoted symbols or keywords have a similar requirement
    // as the numerics above, they have different productions because the
    // rules for numerics are the same in s-expressions, but keywords
    // have different rules between top-level and s-expressions.
    | keyword_entity ws
    | keyword_entity quoted_annotation value
    | keyword_entity keyword_delimiting_entity
    ;

annotation
    : symbol ws* COLON COLON ws*
    ;

symbol
    : IDENTIFIER_SYMBOL
    // note that this is because we recognize the type names for null
    // they are ordinary symbols on their own
    | TYPE
    | QUOTED_SYMBOL
    ;

IDENTIFIER_SYMBOL
    : [$_a-zA-Z] ([$_a-zA-Z] | DEC_DIGIT)*
    ;

numeric_entity
    : BIN_INTEGER
    | DEC_INTEGER
    | HEX_INTEGER
    | TIMESTAMP
    | FLOAT
    | DECIMAL
    ;
```

*Is* this how annotations *should* work?

src/types/timestamp.rs
jobarr-amzn: The annotation question above drove me to read through the ANTLR grammar for text encoding, and I noted the construct used there to enforce this limit: 
```plain

fragment
YEAR
    : '000'                     [1-9]
    | '00'            [1-9]     DEC_DIGIT
    | '0'   [1-9]     DEC_DIGIT DEC_DIGIT
    | [1-9] DEC_DIGIT DEC_DIGIT DEC_DIGIT
    ;
```

src/value/native_reader.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+impl ElementReader for NativeElementReader {
+    fn iterate_over<'a, 'b>(
+        &'a self,
+        data: &'b [u8],
+    ) -> IonResult<Box<dyn Iterator<Item = IonResult<OwnedElement>> + 'b>> {
+        // Match against the slice to see if it starts with the binary IVM
+        match data {
+            &[0xe0, 0x01, 0x00, 0xea, ..] => {
```
jobarr-amzn, line 31: Is there any way to get re-use here between this format detection and the IVM parsing elsewhere in the library?

tests/element_test_vectors.rs
jobarr-amzn: That's very reassuring given that the diff is so large GitHub won't load it by default.

jobarr-amzn (approved):
Very exciting! I have some questions to clarify about `false`/`true`/`nan` as annotations but no reason I see not to ship this. My Rust comprehension is still, well, not yet oxidized but it looks good to me.

src/binary/uint.rs
zslayton: > I wonder what the largest integer we've encountered or heard of in real data is?

Excellent question. It's not hard to find ion-java client code using `IonReader#bigIntegerValue()`, but it's impossible to know whether their data requires that or it's just defensive programming.

A bit of code archeology: back when it was a hobby project,`ion-rust` originally only supported `i64` representations of integer. When Rust v1.26 was released, it introduced [a stable `i128` data type](https://doc.rust-lang.org/std/primitive.i128.html). I briefly planned to make that the go-to representation for integer; it was stack allocated and had fast intrinsic operations relative to `BigInt`. Ultimately though, I went with `BigInt` because I knew there were [spec conformance tests that featured enormous integers](https://github.com/amzn/ion-tests/blob/master/iontestdata/good/intBigSize1201.10n). 

We could have just said "we don't support arbitrarily sized ints" and added those test files to a skip list, but I concluded that when 99.999% of integer use cases will fit in an `i64`, we don't lose much by using a `BigInt` instead of an `i128` for the remaining .001%.

src/text/raw_text_reader.rs
```diff
@@ -1044,4 +1057,58 @@ mod reader_tests {
...
+        assert_eq!(reader.read_i64()?, 1);
+        assert_eq!(reader.next()?, RawStreamItem::Value(IonType::Integer));
+        assert_eq!(reader.field_name()?, RawSymbolToken::Text("b".to_string()));
+        assert_eq!(reader.read_i64()?, 2);
+        reader.step_out()?;
+
+        assert_eq!(reader.next()?, RawStreamItem::Value(IonType::Struct));
+        reader.step_out()?;
+        Ok(())
+    }
```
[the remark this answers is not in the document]
zslayton: I agree that that would be helpful. 

The current unit tests in `ion-rust` are a product of the library's bootstrapping; a lot of them were written back when the library was missing a lot of features/abstractions and have only been updated to keep the project compiling over time. Now that we have some ergonomic features (like text Ion support!), a lot of the tests could be rewritten for improved readability.

src/text/raw_text_reader.rs
zslayton: In cases like this, I tend to defer to `ion-tests`. Here are some relevant tests from [the `bad` examples directory](https://github.com/amzn/ion-tests/tree/master/iontestdata/bad):
* [`false` as an annotation](https://github.com/amzn/ion-tests/blob/master/iontestdata/bad/annotationFalse.ion)
* [`nan` as an annotation](https://github.com/amzn/ion-tests/blob/master/iontestdata/bad/annotationNan.ion)

`ion-java` does not allow unquoted keywords as annotations, which jibes with these `ion-tests`.

PR #373, opened 2022-05-06
Fixes most `bad` native reader tests from `ion-tests`

zslayton:
* The `RawBinaryReader` now surfaces an error if it encounters...
   * an annotated value whose length doesn't match that of the
   annotations wrapper.
   * an annotations wrapper with no annotations
   * an annotations wrapper with no value
   * a negative integer (typecode=3) with a magnitude of 0.
   * a malformed IVM.
   * an IVM at any depth other than 0.
   * a value with with typecode=15. (Previously it panicked.)
   * a timestamp with a length <= 1 (missing required fields)
   * an empty ordered struct
* The `RawBinaryWriter` now correctly encodes the annotations wrapper
  length of annotated containers. Previously the reader would
  ignore this (often redundant) field, allowing tests to pass
  erroneously. The binary writer would omit the container's header
  bytes from its size calculations.
* The `RawTextReader` will now surface an error if it encounters...
   * a text value with invalid unicode escapes. (Previously it would
   fail to match the invalid escape and fall back to treating the
   escapes as uninterpreted text.)
   * a timestamp offset with `hours >= 24` or `minutes >= 60`.
   * a timestamp with a `Decimal` fractional seconds `< 0` or
   `>= 1`.
* The `Reader` will now surface an error if it encounters a symbol
  table with more than one `symbols` field or more than one
  `imports` field.

This patch also adds comments to the remaining `bad` tests in the
native reader's skip list describing what needs to be done to address
them in the future.

### DIFF
--- a/src/binary/raw_binary_reader.rs
+++ b/src/binary/raw_binary_reader.rs
@@ -962,7 +962,7 @@ where
         // TODO: Is cursor.value.ion_type redundant?
         self.cursor.value.ion_type = header.ion_type.ok_or_else(|| {
             decoding_error_raw(format!(
-                "Found an invalid type code: {:?}",
+                "found an invalid type code: {:?}",
                 header.ion_type_code
             ))
         })?;
@@ -1104,9 +1104,6 @@ where
         let num_annotations_before = self.cursor.annotations.len();
         // The first byte of the annotations envelope is now behind the cursor
         let annotations_offset = self.cursor.bytes_read - 1;
-        // The encoding allows us to skip over the annotations list and the value, but in practice
-        // we won't know if we want to skip this value until we've read the type descriptor byte.
-        // That means we need to read the length even though we have no intent to use it.
         let annotations_and_value_length = self.read_standard_length()?;
         let annotations_length = self.read_var_uint()?;
         if annotations_length.value() == 0 {
--- a/src/binary/raw_binary_reader.rs
+++ b/src/binary/raw_binary_reader.rs
@@ -975,9 +975,8 @@ where
         use IonTypeCode::*;
         let length = match header.ion_type_code {
             NullOrNop | Boolean => 0,
-            PositiveInteger | NegativeInteger | Decimal | String | Symbol | List | SExpression | Clob | Blob => {
-                self.read_standard_length()?
-            }
+            PositiveInteger | NegativeInteger | Decimal | String | Symbol | List | SExpression
+            | Clob | Blob => self.read_standard_length()?,
             Timestamp => {
                 let length = self.read_standard_length()?;
                 if length <= 1 && !self.cursor.value.is_null {
--- a/src/binary/raw_binary_reader.rs
+++ b/src/binary/raw_binary_reader.rs
@@ -975,31 +975,22 @@ where
         use IonTypeCode::*;
         let length = match header.ion_type_code {
             NullOrNop | Boolean => 0,
-            PositiveInteger | Decimal | String | Symbol | List | SExpression | Clob | Blob => {
+            PositiveInteger | NegativeInteger | Decimal | String | Symbol | List | SExpression | Clob | Blob => {
                 self.read_standard_length()?
-            }
-            NegativeInteger => {
-                let length = self.read_standard_length()?;
-                if length == 0 && !self.cursor.value.is_null {
-                    return decoding_error(
-                        "Found a non-null negative integer (typecode=3) with a length of 0.",
-                    );
-                }
-                length
             }
             Timestamp => {
                 let length = self.read_standard_length()?;
                 if length <= 1 && !self.cursor.value.is_null {
                     return decoding_error(
-                        "Found a non-null timestamp (typecode=6) with a length <= 1.",
+                        "found a non-null timestamp (typecode=6) with a length <= 1",
                     );
                 }
                 length
             }
             Float => self.read_float_length()?,
             Struct => self.read_struct_length()?,
-            Annotation => return decoding_error("Found an annotation wrapping an annotation."),
-            Reserved => return decoding_error("Found an Ion Value with a Reserved type code."),
+            Annotation => return decoding_error("found an annotation wrapping an annotation"),
+            Reserved => return decoding_error("found an Ion Value with a Reserved type code"),
         };
 
         self.cursor.value.header_length =
--- a/src/binary/raw_binary_reader.rs
+++ b/src/binary/raw_binary_reader.rs
@@ -26,6 +26,7 @@ use crate::stream_reader::StreamReader;
 use crate::types::decimal::Decimal;
 use crate::types::integer::{IntAccess, Integer};
 use crate::types::timestamp::Timestamp;
+use num_traits::Zero;
 use std::ops::Range;
 
 /// Information about the value over which the RawBinaryReader is currently positioned.
@@ -96,14 +97,15 @@ impl EncodedValue {
     /// Returns the length of this value's header, including the type descriptor byte and any
     /// additional bytes used to encode the value's length.
     fn header_length(&self) -> usize {
-        self.header_length as usize
+        // The `header_length` field does not include the type descriptor byte, so add 1.
+        self.header_length as usize + 1
     }
 
     /// Returns an offset Range containing this value's type descriptor
     /// byte and any additional bytes used to encode the `length`.
     fn header_range(&self) -> Range<usize> {
         let start = self.header_offset;
-        let end = start + self.header_length as usize + 1;
+        let end = start + self.header_length();
         start..end
     }
 
@@ -344,6 +346,7 @@ impl<R: IonDataSource> StreamReader for RawBinaryReader<R> {
         }
 
         self.clear_annotations();
+        let mut expected_annotated_value_length = None;
         if header.ion_type_code == IonTypeCode::Annotation {
             if header.length_code == 0 {
                 // This is actually the first byte in an Ion Version Marker
@@ -352,7 +355,7 @@ impl<R: IonDataSource> StreamReader for RawBinaryReader<R> {
             }
             // We've found an annotated value. Read all of the annotation symbols leading
             // up to the value
-            let _ = self.read_annotations()?;
+            expected_annotated_value_length = Some(self.read_annotations()?);
             // Now read the next header representing the value itself.
             header = match self.read_next_value_header()? {
                 Some(header) => header,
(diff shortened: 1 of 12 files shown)


src/reader.rs
```diff
@@ -60,6 +60,11 @@ impl<R: RawReader> Reader<R> {
         let mut is_append = false;
         let mut new_symbols = vec![];
 
+        // It's illegal for a symbol table to have multiple `symbols` or `imports` fields.
+        // Keep track of whethe we've already encountered them.
```
desaikd, line 64:
```suggestion
        // Keep track of whether we've already encountered them.
```

src/binary/raw_binary_reader.rs
```diff
@@ -947,12 +975,22 @@ where
...
-            PositiveInteger | NegativeInteger | Decimal | Timestamp | String | Symbol | List
-            | SExpression | Clob | Blob => self.read_standard_length()?,
+            PositiveInteger | NegativeInteger | Decimal | String | Symbol | List | SExpression | Clob | Blob => {
+                self.read_standard_length()?
+            }
+            Timestamp => {
+                let length = self.read_standard_length()?;
+                if length <= 1 && !self.cursor.value.is_null {
+                    return decoding_error(
+                        "found a non-null timestamp (typecode=6) with a length <= 1",
```
zslayton, line 985: 🗺️ Binary `timestamp`s [must specify a `year` and `offset`](https://amzn.github.io/ion-docs/docs/binary.html#0x6-timestamp), making their minimum legal length `2`.

src/binary/raw_binary_reader.rs
```diff
@@ -1049,16 +1098,25 @@ where
...
+    /// wrapped value that will follow.
+    fn read_annotations(&mut self) -> IonResult<usize> {
         let num_annotations_before = self.cursor.annotations.len();
         // The first byte of the annotations envelope is now behind the cursor
         let annotations_offset = self.cursor.bytes_read - 1;
-        // The encoding allows us to skip over the annotations list and the value, but in practice
-        // we won't know if we want to skip this value until we've read the type descriptor byte.
-        // That means we need to read the length even though we have no intent to use it.
-        let _annotations_and_value_length = self.read_standard_length()?;
+        let annotations_and_value_length = self.read_standard_length()?;
```
zslayton, line 1107: 🗺️ The total annotations wrapper length (`annotations` length + `value` length) is often redundant for the 'happy path'. However, it's useful for detecting malformed streams. We now use it to confirm that the wrapper and the wrapped value's lengths agree.

src/binary/raw_binary_writer.rs
```diff
@@ -777,6 +777,10 @@ impl<'a, W: Write> Writer for RawBinaryWriter<W> {
             Ok(())
         })?;
 
+        // Now that we know how large the container's header is, add its length to the
+        // calculate container size.
+        let container_size = container_size + header_io_range.len();
```
zslayton, line 782: 🗺️ Prior to this patch, the binary reader would ignore the annotations wrappers' combined length fields as they were largely redundant. It turns out the binary writer had been calculating that field incorrectly; because the `ion-rust` reader was usually (always?) the one consuming the `ion-rust` binary writer's output, the problem went unnoticed.

src/reader.rs
```diff
@@ -60,6 +60,11 @@ impl<R: RawReader> Reader<R> {
         let mut is_append = false;
         let mut new_symbols = vec![];
 
+        // It's illegal for a symbol table to have multiple `symbols` or `imports` fields.
+        // Keep track of whether we've already encountered them.
+        let mut has_found_symbols_field = false;
+        let mut has_found_imports_field = false;
```
zslayton, line 66: 🗺️ Here we detect and reject repeated `imports` and `symbols` in a symbol table.

src/reader.rs
```diff
@@ -372,20 +393,18 @@ mod tests {
     const EXAMPLE_STREAM: &[u8] = &[
         // $ion_symbol_table::{imports: $ion_symbol_table, symbols: ["foo", "bar", "baz"]}
         0xEE, // Var len annotations
-        0x94, // Annotations + Value length: 20 bytes
+        0x92, // Annotations + Value length: 21 bytes
```
zslayton, line 396: 🗺️ Correcting bad example data originally produced by the binary writer.

src/text/parsers/text_support.rs
```diff
@@ -62,28 +62,43 @@ pub(crate) fn escaped_char_literal(input: &str) -> IResult<&str, char> {
...
+    // If this step fails, the Ion data stream is malformed and we need to bail out completely.
+    // We can't simply return an error as we did above; if we did that, the parser would go on to
+    // treat the input as a String literal without escapes, which is the incorrect behavior.
+    // Instead, we need to return a nom `Err::Failure`, indicating that we cannot proceed.
+    let number_value = match u32::from_str_radix(hex_digits, 16) {
+        Ok(number_value) => number_value,
+        Err(_parse_int_error) => {
+            // TODO: A custom error type would be required to bubble up specific information
+            //       about the failure.
+            return Err(Err::Failure(Error::from_error_kind(
```
zslayton, line 84: 🗺️ Typically, when a `nom` parser returns an `Err` it's a signal that it didn't find what it was looking for in the input. This is both common and expected; the reader will move on to another parser and see if that one matches instead. However, there are times when the reader is definitely using the right parser but it has found a fatal problem. In this case, we've found a Unicode escape sequence that's correctly formatted (e.g. `\xE0`) but when we tried to turn it into a codepoint, we realized it wasn't valid. We need to communicate to the reader that the stream is malformed and it shouldn't continue on trying other parsers. In cases like this, we want to return `nom`'s `Err::Failure(_)` instead of `Err::Error(_)`, which is what this change does.

Unfortunately, the default `nom` error/failure type doesn't provide a way to bubble up contextual information about the problem it encountered to the reader. As a future task, we'll need to make a custom parsing error type that allows us to provide (e.g.) a `String` explaining what went wrong.

src/text/parsers/timestamp.rs
```diff
@@ -270,13 +269,21 @@ fn timezone_offset(input: &str) -> IResult<&str, Option<i32>> {
...
-    )(input)
+    )(input)?;
+    let hour_int = trim_zeros_expect_i32(hours, "offset hours");
+    let minute_int = trim_zeros_expect_i32(minutes, "offset minutes");
+    if hour_int >= 24 {
+        return Err(Err::Failure(make_error(hours, ErrorKind::TooLarge)));
+    }
+    if minute_int >= 60 {
+        return Err(Err::Failure(make_error(minutes, ErrorKind::TooLarge)));
+    }
```
zslayton, line 285: 🗺️ Previously we only checked that the number of minutes represented by the `offset` we found (`hours * 60 + minutes`) was reasonable. This check occurred in the `Timestamp` constructor. Now we are also checking that the values of the individual `hours` and `minutes` substrings make sense.

src/text/raw_text_reader.rs
```diff
@@ -91,10 +91,16 @@ impl<T: TextIonDataSource> RawTextReader<T> {
...
-                // We found an IVM; we currently only support Ion 1.0.
-                self.current_ivm = Some((1, 0));
-                return Ok(());
+            match self.parse_next(ion_1_0_version_marker) {
+                Ok(Some(_)) => {
+                    // We found an IVM; we currently only support Ion 1.0.
+                    self.current_ivm = Some((1, 0));
+                    return Ok(());
+                }
+                Err(e @ IonError::IoError { .. }) => return Err(e),
```
zslayton, line 100: 🗺️ Previously we would ignore any kind of `Err(_)` here. In most cases this was fine because we expect parsers like `ion_1_0_version_marker` to return an `Err` if they don't match the input. That's both common and expected.

If we encountered an `IoError`, however, it means there was a problem caused by something other than parsing. It could be a failure to read from the disk or a socket, or it could mean that the input contained invalid utf-8 data. We need to surface this to the application.

src/types/decimal.rs
```diff
@@ -62,6 +63,38 @@ impl Decimal {
         }
     }
 
+    /// Returns true if this Decimal's coefficient has a negative sign AND a magnitude greater than
+    /// zero. Otherwise, returns false. (Negative zero returns false.)
+    pub fn is_less_than_zero(&self) -> bool {
```
zslayton, line 68: 🗺️ These two helper methods are used by the `Timestamp` type to see if a given `Decimal` representing a fractional number of seconds is in the range `[0, 1)`. `is_less_than_zero` is actually more broadly useful because applications will often want to know if a value is negative but _don't_ care if it's `-0` (which is negative, but not less than zero).

src/types/integer.rs
```diff
@@ -181,6 +181,40 @@ impl Neg for Integer {
...
+                }
+            }
+            (I64(this), Big(that)) => Big(BigInt::from(this).add(that)),
+            (Big(this), I64(that)) => Big(this.add(&BigInt::from(that))),
+            (Big(this), Big(that)) => Big(this.add(&that)),
+        }
+    }
+}
+
+impl Zero for Integer {
```
zslayton, line 205: 🗺️ There are a couple of places where we read an `Integer` and then want to find out if it's zero. Having an implementation of the `Zero` trait means that we don't have to do this check manually everywhere. The `Add` trait, implemented above, is a prerequisite trait for implementing `Zero`.

Longer-term, I think it would make sense for us to provide implementations of all the common `num` traits for `Integer` so users don't have to manually destructure each value and then do their own math operations.

src/types/magnitude.rs
```diff
@@ -95,11 +95,22 @@ impl Magnitude {
         digits
     }
 
-    fn calculate_u64_digits(value: &u64) -> u64 {
-        match value {
-            0 => 1,
-            1 => 1,
-            i => (*i as f64).log10().ceil() as u64,
+    fn calculate_u64_digits(value: u64) -> u64 {
```
zslayton, line 98: 🗺️ This helper method is used indirectly to determine if a given `Decimal` is in the range `[0, 1)`. It had a bug that caused the digit count for powers of ten to be off by one.

tests/element_test_vectors.rs
```diff
@@ -534,77 +534,38 @@ mod native_element_tests {
 
         fn global_skip_list() -> SkipList {
             &[
```
zslayton, line 536: 🗺️ Here I've removed the now-passing `bad` test files from the skip list and added some comments explaining what needs to be done to address the remaining ones.

zslayton (commented):
🗺️ PR tour comments are prefixed with the world map emoji.

desaikd (approved):
Its good to see the reduced skip-list in tests. Thanks for all the PR tour 🗺️ comments!

PR #376, opened 2022-05-09
Fixes a variety of ion-tests failures

zslayton:
Fixes include:
* Blobs can have whitespace between chunks of base64 characters.
  Example: `{{aGVsbG8=}}` is the same as `{{  a G Vsb G8 = }}`.
* Clobs can have whitespace before and after their quoted text.
  Example: `{{"hello"}}` is the same as `{{  "hello"  }}`.
* Clobs (which are text of an unspecified encoding) no longer
  erroneously support Unicode escape sequences (\u and \U).
* S-expressions no longer require whitespace between self-delimiting
  values. Example: `(+++foo)` parses successfully as an s-expression
  with an operator, `+++`, and a symbol, `foo`.
* S-expression operators followed by negative numbers no longer
  fail to parse. Examples:
   * `-3` is a negative int
   * `--3` is an operator, `--`, followed by an int, `3`.
* Struct field names can be long strings. `{'''a''': foo}`
  is the same as `{a: foo}`.
* Decimal and float exponents can start with a `+`.
* Decimal and float exponents can start with multiple leading zeros.
* All escaped newlines (`\` followed by unescaped \r or \r\n) are
  now discarded. (Previously, only `\` followed by \n was discarded).
* Skip lists have been updated to reflect the current functionality.


src/text/parsers/clob.rs
```diff
@@ -75,7 +82,7 @@ fn long_clob_body(input: &str) -> IResult<&str, Vec<u8>> {
 fn long_clob_fragment(input: &str) -> IResult<&str, StringFragment> {
     alt((
         escaped_newline,
-        escaped_char,
+        escaped_char_no_unicode,
```
zslayton, line 85: 🗺️ Clobs aren't (necessarily) Unicode, so they don't recognize Unicode escape sequences. (`\u` and `\U`)

src/text/parsers/containers.rs
```diff
@@ -111,23 +110,28 @@ pub(crate) fn s_expression_value(input: &str) -> IResult<&str, AnnotatedTextValu
...
             // ...an annotated operator (`foo::++`)...
             pair(parse_annotations, parse_operator)
                 .map(|(annotations, value)| AnnotatedTextValue::new(annotations, value)),
-            // ...an un-annotated operator (`++`) paired with an empty annotations Vec...
-            parse_operator.map(|op| op.without_annotations()),
-            // ...or some other kind of value (`5`, `"hello"`, etc).
+            // ...a non-operator value, with or without annotations (`5`, `foo::5`, `"hello"`, etc)...
             top_level_value,
+            // ...or an un-annotated operator (`++`).
+            parse_operator.map(|op| op.without_annotations()),
```
zslayton, line 123: 🗺️ See the comment starting on line 125 for an explanation of why this was shuffled.

src/text/parsers/float.rs
```diff
@@ -33,8 +33,10 @@ fn float_special_value(input: &str) -> IResult<&str, TextValue> {
 fn float_numeric_value(input: &str) -> IResult<&str, TextValue> {
     map_res::<_, _, _, _, ParseFloatError, _, _>(
         recognize(tuple((
-            opt(tag("-")),
-            alt((floating_point_number, digits_before_dot)),
+            alt((
+                floating_point_number,
+                recognize(pair(opt(tag("-")), digits_before_dot)),
```
zslayton, line 38: 🗺️ `floating_point_number` already accepts an optional `-`, so having `opt(tag("-"))` outside the `alt(...)` meant that `float_numeric_value` could accept _two_ optional `-` signs. 😬 Now the `opt(tag("-"))` only applies to `digits_before_dot`.

src/text/parsers/string.rs
```diff
@@ -66,11 +69,71 @@ fn long_string_fragment(input: &str) -> IResult<&str, StringFragment> {
     ))(input)
 }
 
-/// Matches the next string fragment while respecting the long string delimiter (`'''`).
+/// Matches the next string fragment without escape sequences while also respecting the long
+/// string delimiter (`'''`).
 fn long_string_fragment_without_escaped_text(input: &str) -> IResult<&str, StringFragment> {
```
zslayton, line 74: 🗺️ The original implementation could overlook escape sequences in long string fragments (`'''foo'''`), treating them as literals. Implementing this in `nom` wasn't straightforward, so I opted to just write out the logic by hand. The comment goes into some detail.

src/text/parsers/symbol.rs
```diff
@@ -102,26 +101,7 @@ fn identifier_trailing_characters(input: &str) -> IResult<&str, &str> {
...
-        // `is_a` matches the longest leading string comprised of one or more of the given characters
-        is_a("!#%&*+-./;<=>?@^`|~"),
-        // The operator must be followed either by whitespace or the end of the s-expression.
-        alt((
-            peek(recognize(one_of(" \r\n\t"))),
-            peek(recognize(s_expression_end)),
-        )),
-    );
-    // The above `recognizer` outputs a &str; this operation turns that &str into a Symbol.
+    let recognizer = preceded(multispace0, is_a("!#%&*+-./;<=>?@^`|~"));
```
zslayton, line 104: 🗺️ Some earlier restructuring in `src/text/parsers/containers.rs` eliminated the (wrong/unnecessary) rule that s-expression scalars have whitespace or the closing `)` after them.

src/text/parsers/symbol.rs
```diff
@@ -253,4 +233,14 @@ mod symbol_parsing_tests {
             TextValue::Symbol(text_token(expected)),
         )
     }
+
+    #[test]
+    fn operators_and_negative_numbers() {
+        parse_test_ok(parse_operator, "--3", TextValue::Symbol(text_token("--")));
```
zslayton, line 239: 🗺️ I confirmed with `ion-java` and `ion-c` that the correct parse for the children of `(--3)` is `--` and `3`.

src/text/parsers/text_support.rs
```diff
@@ -24,7 +24,10 @@ pub(crate) enum StringFragment<'a> {
 
 /// Matches an escaped newline, returning [StringFragment::EscapedNewline].
 pub(crate) fn escaped_newline(input: &str) -> IResult<&str, StringFragment> {
-    value(StringFragment::EscapedNewline, tag("\\\n"))(input)
+    value(
+        StringFragment::EscapedNewline,
+        alt((tag("\\\n"), tag("\\\r\n"), tag("\\\r"))),
```
zslayton, line 29: 🗺️ Support for stripping newlines that use the `\r` and `\r\n` encodings.

src/text/parsers/text_support.rs
```diff
@@ -39,6 +42,16 @@ pub(crate) fn escaped_char(input: &str) -> IResult<&str, StringFragment> {
...
 
+/// Matches an escaped literal (like '\n') or a hex escape (like `\x`), returning the appropriate
+/// substitute character as a [StringFragment::EscapedChar]. Does NOT match Unicode escapes
+/// ('\u' or '\U').
+pub(crate) fn escaped_char_no_unicode(input: &str) -> IResult<&str, StringFragment> {
+    map(
+        preceded(char('\\'), alt((escaped_hex_char, escaped_char_literal))),
+        StringFragment::EscapedChar,
+    )(input)
+}
```
zslayton, line 53: 🗺️ Clobs aren't supposed to support Unicode escapes, so they can use this function instead of `escaped_char_unicode` below.

zslayton (commented):
🗺️ PR tour!

PR #378, opened 2022-05-11
Ivm fixes

zslayton:
This PR depends on #377, which in turn depends on #376. Once those are both merged, I'll rebase this on `main`.

-----

Previously, the text reader would detect the Ion version marker
by simply looking for the literal `$ion_1_0`. This worked fine
insofar as `$ion_1_0` was (and is) the only legal version marker.

This PR modifies the parser to match on `$ion_MAJOR_MINOR`, in which
`MAJOR` and `MINOR` are each decimal digit sequences representing
the major and minor version of the stream's encoding, respectively.

With this change in place, the reader can now identify unsupported
versions of Ion and return an appropriate `IonError`.


desaikd (approved):
🚀

PR #379, opened 2022-05-13
Adds custom `IonParseResult` type

zslayton:
The `nom` crate's out-of-the-box default error type is intentionally
minimalist. Constructing and processing a `nom::error::Error` is
trivial; unfortunately, it is also limited in what it information it
can communicate to the user.

With this in mind, all of `nom`'s parsers are generic over not only
their input and output types, but also their error type. This allows
users to replace `nom::error::Error` with the type of their choice.

This PR introduces a new type called `IonParseError` that can store
additional information about problems that were encountered during
parsing. In the typical use case (unmatched text), it is as cheap to
construct and process as `nom::error::Error`. However, when a fatal
error is encountered (for example: an I/O error, overflow, etc), it can
store additional, heap-allocated context to report to the end user.
For now this additional information is simply an `Option<String>`;
we can expand it in the future as we deem necessary.

To simplify type signatures in the `text::parsers` module, a new
type alias for `IResult<&str, T, IonParseError>` has also been
added: `IonParseResult<T>`. This eliminates substantial visual
noise from the parsers.

Finally, two extension traits have been added to the library that
make it easier to mix `nom` parsers that work with `IResult` with
`ion-rs` parsers that work with `IonParseResult`.

Now that our parsers return `IonParseResult`, it is possible for us to
surface fatal errors to the application with valuable context. Many
situations that would previously lead to a panic now result in a
user-manageable `IonParseResult::Err`.

See the documentation for each type and trait in
`src/text/parsers/parse_result.rs` for more information.


src/text/parsers/decimal.rs
```diff
@@ -19,59 +19,53 @@ use crate::types::magnitude::Magnitude;
...
-    digits_before_dot: &str,
-    digits_after_dot: Option<&str>,
-    exponent_text: &str,
-) -> Decimal {
+fn decimal_from_text_components<'a>(
+    sign_text: Option<&'a str>,
+    digits_before_dot: &'a str,
+    digits_after_dot: Option<&'a str>,
+    exponent_text: &'a str,
+) -> IonParseResult<'a, Decimal> {
```
zslayton, line 71: 🗺️ `IonParseResult<'a>` has a lifetime. Because this function takes several parameters that have their own lifetimes, we have to explicitly say which lifetime the returned `IonParseResult` is using. In this case, they all have the same lifetime so it doesn't matter.

src/text/parsers/decimal.rs
```diff
@@ -91,13 +88,21 @@ fn decimal_from_text_components(
...
     // succeed.
     let magnitude: Magnitude = if magnitude_text.len() < MAX_U64_DIGITS {
-        u64::from_str(&magnitude_text)
-            .expect("parsing coefficient magnitude as u64 failed")
-            .into()
+        let value = u64::from_str(&magnitude_text)
+            .or_fatal_parse_error(
+                input,
+                "parsing coefficient magnitude as u64 failed",
+            )?
```
zslayton: 🗺️ Bubble up information about the parse failure instead of panicking. `or_fatal_parse_error` is provided by the extension trait `OrFatalParseError` and is defined for all configurations of `std::result::Result`.

src/text/parsers/decimal.rs
```diff
@@ -107,11 +112,11 @@ fn decimal_from_text_components(
     let mut exponent = i64::from_str(&sanitized).expect("parsing exponent as i64 failed");
     // Reduce the exponent by the number of digits that follow the decimal point
     exponent -= digits_after_dot.chars().filter(|c| c.is_digit(10)).count() as i64;
-    Decimal::new(coefficient, exponent)
+    Ok(("", Decimal::new(coefficient, exponent)))
```
zslayton, line 109: 🗺️ Now that we're returning `IonParseResult` everywhere, we can't just return `Decimal`. We have to use `nom`'s success data type: `Ok(...)` containing a tuple of the remaining input (a `&str`) and the value we parsed (a `Decimal`).

src/text/parsers/float.rs
```diff
@@ -6,51 +7,50 @@ use crate::text::text_value::TextValue;
...
     map(tag("nan"), |_| TextValue::Float(f64::NAN))
         .or(map(tag("+inf"), |_| TextValue::Float(f64::INFINITY)))
         .or(map(tag("-inf"), |_| TextValue::Float(f64::NEG_INFINITY)))
         .parse(input)
+        .upgrade()
 }
 
 /// Matches numeric floating point values. (e.g. `7e0`, `7.1e0` or `71e-1`)
-fn float_numeric_value(input: &str) -> IResult<&str, TextValue> {
-    map_res::<_, _, _, _, ParseFloatError, _, _>(
```
zslayton, line 34: 🗺️ Replacing a `map_res` with the two-phase approach.

src/text/parsers/integer.rs
```diff
@@ -29,23 +31,22 @@ pub(crate) fn parse_integer(input: &str) -> IResult<&str, TextValue> {
 
 /// Matches a base-16 notation integer (e.g. `0xCAFE`, `0Xcafe`, or `-0xCa_Fe`) and returns the
 /// resulting [i64] as a [TextValue::Integer].
-fn base_16_integer(input: &str) -> IResult<&str, TextValue> {
-    map_res(
```
zslayton, line 33: 🗺️ Replacing a `map_res` with the two-phase approach.

src/text/parsers/float.rs
```diff
@@ -6,51 +7,50 @@ use crate::text::text_value::TextValue;
...
+        )),
+        recognize(float_exponent_marker_followed_by_digits),
+    )))(input)?;
+    // TODO: Reusable buffer for sanitization
+    let mut sanitized = text.replace('_', "");
+    if sanitized.ends_with('e') || sanitized.ends_with('E') {
+        sanitized.push('0');
+    }
+    let float = f64::from_str(&sanitized)
+        .or_fatal_parse_error(input, "could not parse float as f64")?
```
zslayton, line 48: 🗺️ Bubble up the `f64` conversion error.

src/text/parsers/integer.rs
```diff
@@ -29,23 +31,22 @@ pub(crate) fn parse_integer(input: &str) -> IResult<&str, TextValue> {
...
+fn base_16_integer(input: &str) -> IonParseResult<TextValue> {
+    let (remaining, (maybe_sign, text_digits)) = separated_pair(
+        opt(char('-')),
+        alt((tag("0x"), tag("0X"))),
+        base_16_integer_digits,
+    )(input)?;
+    let integer = parse_integer_with_radix(text_digits, 16)
+        .map(|(_, i)| if maybe_sign.is_some() { -i } else { i })
+        .map(TextValue::Integer)
+        .or_fatal_parse_error(input, "could not parse hex integer")?
```
zslayton, line 43: 🗺️ Bubble up the hex parsing error.

src/text/parsers/integer.rs
```diff
@@ -56,82 +57,82 @@ fn base_16_integer_digits(input: &str) -> IResult<&str, &str> {
...
+fn base_2_integer(input: &str) -> IonParseResult<TextValue> {
+    let (remaining, (maybe_sign, text_digits)) = separated_pair(
+        opt(char('-')),
+        alt((tag("0b"), tag("0B"))),
+        base_2_integer_digits,
+    )(input)?;
+    let integer = parse_integer_with_radix(text_digits, 2)
+        .map(|(_, i)| if maybe_sign.is_some() { -i } else { i })
+        .map(TextValue::Integer)
+        .or_fatal_parse_error(input, "could not parse binary integer")?
```
zslayton, line 75: 🗺️ Bubble up the binary parsing error.

src/text/parsers/integer.rs
```diff
@@ -56,82 +57,82 @@ fn base_16_integer_digits(input: &str) -> IResult<&str, &str> {
...
-fn base_10_integer(input: &str) -> IResult<&str, TextValue> {
-    map_res(
-        recognize(preceded(opt(char('-')), base_10_integer_digits)),
-        |text| parse_integer_with_radix(text, 10).map(TextValue::Integer),
-    )(input)
+fn base_10_integer(input: &str) -> IonParseResult<TextValue> {
+    let (remaining, int_text) = recognize(preceded(opt(char('-')), base_10_integer_digits))(input)?;
+    let integer = parse_integer_with_radix(int_text, 10)
+        .map(|(_, i)| TextValue::Integer(i))
+        .or_fatal_parse_error(input, "could not parse decimal integer")?
```
zslayton, line 97: 🗺️ Bubble up the decimal parsing error.

src/text/parsers/decimal.rs
```diff
@@ -91,13 +88,21 @@ fn decimal_from_text_components(
...
     let magnitude: Magnitude = if magnitude_text.len() < MAX_U64_DIGITS {
-        u64::from_str(&magnitude_text)
-            .expect("parsing coefficient magnitude as u64 failed")
-            .into()
+        let value = u64::from_str(&magnitude_text)
+            .or_fatal_parse_error(
+                input,
+                "parsing coefficient magnitude as u64 failed",
+            )?
+            .1;
```
zslayton, line 93: 🗺️ `.or_fatal_parse_error` returns an `IonParseError`. We've handled the error case with a `?` on line 95. The success case is an `(remaining_input: &str, value)` tuple. If `from_str` succeeded, there's no more input so the `remaining_input` is `""`. We can discard that and take the `value` using the tuple field access syntax: `.1`.

src/text/parsers/timestamp.rs
```diff
@@ -32,93 +33,88 @@ pub(crate) fn parse_timestamp(input: &str) -> IResult<&str, TextValue> {
 
 /// Matches the text representation of a timestamp value with year precision (e.g. `2018T`) and
 /// returns the resulting Timestamp as a [TextValue::Timestamp].
-fn timestamp_precision_y(input: &str) -> IResult<&str, TextValue> {
-    map_res::<_, _, _, _, IonError, _, _>(
```
zslayton, line 36: 🗺️ This module has (lots of) the same kinds of changes we've already seen in the other modules. Namely: replacing `map`/`map_res` with the two-phase approach, bubbling up failures, and dealing with `Ok((remaining_input, value))` success tuples with `.1`.

src/text/parsers/timestamp.rs
```diff
@@ -132,38 +128,45 @@ fn timestamp_precision_ymd_hms_fractional(input: &str) -> IResult<&str, TextValu
...
-                builder.build_at_offset(minutes)?
-            } else {
-                builder.build_at_unknown_offset()?
-            };
-            Ok(TextValue::Timestamp(timestamp))
-        },
-    )(input)
+        )(input)?;
+    let builder = Timestamp::with_ymd(year, month, day).with_hms(hour, minute, second);
+    let (_, builder) = assign_fractional_seconds(fractional_text, builder)?;
```
zslayton, line 133: 🗺️ This is an alternative to the `.1` syntax: pattern match on the tuple and use `_` to throw away the first value (in position `0`).

src/text/parse_result.rs
```diff
@@ -0,0 +1,286 @@
...
+impl<T, E> OrFatalParseError<T> for Result<T, E>
+where
+    E: Debug,
+{
+    fn or_fatal_parse_error<'a>(self, input: &'a str, label: &str) -> IonParseResult<'a, T> {
+        match self {
+            Ok(value) => Ok(("", value)),
+            Err(error) => fatal_parse_error(input, format!("{}: {:?}", label, error)),
+        }
+    }
```
zslayton, line 220: 🗺️ Note that this method includes both the label and the error itself in the `IonParseError` it returns. Additionally, the `RawTextReader` adds location information to it in the `IonError` that it returns to the user.

All of that to say: when you see something like:
```rust
let x = do_something()
    .or_fatal_parse_error("couldn't read an int")?;
```

it looks like it doesn't tell the user much, but there's plenty of information there that isn't visible at the callsite.

src/text/parsers/timestamp.rs
```diff
@@ -454,3 +456,9 @@ mod reader_tests {
         Ok(())
     }
 }
+
+/// Matches the next input character if it is a base-10 digit. This wraps [char::is_digit] in the
+/// nom parsing function signature.
+pub(crate) fn digit(input: &str) -> IResult<&str, char> {
```
zslayton, line 458: 🗺️ This low-level parser is used with `nom` parsers so widely in the `timestamp` module that I decided it wasn't worth it to convert its return type to `IonParseResult`. (It is not new; it used to live in `numeric_support.rs` but I realized it only used in `text/parsers/timestamp`).

src/text/parsers/top_level.rs
```diff
@@ -47,18 +50,12 @@ pub(crate) fn ion_version_marker(input: &str) -> IResult<&str, (u32, u32)> {
     // If the text matched but parsing that into a major and minor version fails, it's a fatal
     // error.
 
-    // XXX: We cannot signal a fatal error by returning `nom::Err::Failure` here as we would in any
-    // other case. Due to API limitations, it is currently not possible for the RawTextReader to
-    // distinguish between a fatal error and a run-of-the-mill mismatch. For now, in the unlikely
-    // event that the reader encounters an IVM with a major or minor version so large that they
-    // cannot fit in a u32, we simply panic.
-    // TODO: Create a custom parsing error type that allows us to bubble up more information.
```
zslayton, line 55: 🗺️ 🎉

src/text/raw_text_reader.rs
```diff
@@ -91,22 +105,25 @@ impl<T: TextIonDataSource> RawTextReader<T> {
...
                     return decoding_error(format!(
                         "Unsupported Ion version: v{}.{}. Only 1.0 is supported.",
                         major, minor
                     ));
                 }
-                // I/O errors are fatal; we cannot keep parsing.
-                Err(e @ IonError::IoError { .. }) => return Err(e),
+                RootParseResult::Failure(error_message) => {
+                    return decoding_error(error_message);
+                }
```
zslayton, line 122: 🗺️ This entire PR was effectively to allow this change: when the reader is checking to see if the next item in the stream is an IVM, it can distinguish between a parsing error ("text did not match") and a fatal error (IO error, invalid utf8, invalid value, etc) and report the latter to the user.

src/text/raw_text_reader.rs
```diff
@@ -249,25 +266,52 @@ impl<T: TextIonDataSource> RawTextReader<T> {
         }
     }
 
+    fn parse_next<P, O>(&mut self, parser: P) -> IonResult<Option<O>>
+    where
+        P: Fn(&str) -> IonParseResult<O>,
+    {
+        match self.parse_next_nom(parser) {
```
zslayton, line 273: 🗺️ `parse_next` is now an adapter around `parse_next_nom` below. It takes the parse result and maps it to a user-facing `IonResult`.

src/text/raw_text_reader.rs
```diff
@@ -249,25 +266,52 @@ impl<T: TextIonDataSource> RawTextReader<T> {
...
+            RootParseResult::Failure(error_message) => decoding_error(error_message),
+        }
+    }
+
     /// Attempts to parse the next entity from the stream using the provided parser.
     /// If there isn't enough data in the buffer for the parser to match its input conclusively,
     /// more data will be loaded into the buffer and the parser will be called again.
     /// If EOF is encountered, returns `Ok(None)`.
-    fn parse_next<P, O>(&mut self, parser: P) -> IonResult<Option<O>>
+    fn parse_next_nom<P, O>(&mut self, parser: P) -> RootParseResult<O>
```
zslayton, line 294: 🗺️ Unlike `parse_next`, `parse_next_nom` distinguishes between fatal and non-fatal errors.

zslayton (commented):
🗺️ PR tour

desaikd (approved):
Looks good!

src/text/parsers/boolean.rs
```diff
@@ -1,16 +1,17 @@
...
+use nom::Parser;
 
 use crate::text::parsers::stop_character;
 use crate::text::text_value::TextValue;
 
 /// Matches the text representation of a boolean value and returns the resulting `true` or `false`
 /// as a [TextValue::Boolean].
-pub(crate) fn parse_boolean(input: &str) -> IResult<&str, TextValue> {
+pub(crate) fn parse_boolean(input: &str) -> IonParseResult<TextValue> {
     map(
```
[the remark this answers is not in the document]
zslayton: Great question! All of `nom`'s parsers implement [the `Parser` trait](https://docs.rs/nom/latest/nom/trait.Parser.html), which looks like this:

```rust
pub trait Parser<I, O, E> {
  // Input -----^  ^  ^
  // Output -------+  |
  // Error -----------+

  // Takes an input (`I`) and returns a `Result` with either:
  // * an `Ok((I, O))` tuple with the remaining (unmatched) input and the `O` output value it parsed
  // * an `Err(I, nom::Err<E>)` with the input it couldn't match and the error `E` wrapped in an `Err`
  fn parse(&mut self, input: I) -> IResult<I, O, E>;
  
  /// ...other stuff
}
```

For any type that implements `Parser`, they'll need to pick their `I`, `O`, and `E` types. 

For us, `I` is always `&str`. `O` depends on what the parser's trying to do, and `E` is either `nom::error::Error` (the default error type for every core `nom` parser) or `IonParseError` (for parser functions we wrote ourselves).

The compiler tries to figure out which `E` to use based on how the parser is called. For [`map`](https://docs.rs/nom/latest/nom/combinator/fn.map.html), this is easy. Here's its function signature:
```rust
pub fn map<I, O1, O2, E, F, G>(mut parser: F, mut f: G) -> impl FnMut(I) -> IResult<I, O2, E>
where                  //                                                                  ^
  F: Parser<I, O1, E>, // <--- This is the only restriction on `E`,                        |
                       //      and the same `E` is used in the function's return type -----+
  G: FnMut(O1) -> O2,
```

The compiler will say that `map`'s `E` type is whatever `parser`'s `E` type is. Other parsers are tougher because they take _multiple_ parsers as parameters. For example, consider [the `pair` parser](https://docs.rs/nom/latest/nom/sequence/fn.pair.html):

```rust
pub fn pair<I, O1, O2, E: ParseError<I>, F, G>(
  mut first: F,
  mut second: G,
) -> impl FnMut(I) -> IResult<I, (O1, O2), E>
where
  F: Parser<I, O1, E>, // These parsers can produce different output types (`O1` and `O2`),
  G: Parser<I, O2, E>, // but they must return the *same* error type.
```

In situations like this, we can use `upgrade()` on the core `nom` parsers to change their `E` type to be `IonParseError`. In that way, all of the parsers being passed as parameters will have the same error type.

Other links:
* [the `IResult` type alias](https://docs.rs/nom/latest/nom/type.IResult.html)
* [the `nom::Err<E>` enum](https://docs.rs/nom/latest/nom/enum.Err.html)

PR #380, opened 2022-05-15
Text reader fixes: escapes, whitespace, IVMs

zslayton:
Fixes include:
* Extending the parser's definition of whitespace to include less
common control characters like vertical tab and form feed.
* Recognizing IVMs that are followed by '::' as annotations
* Escape slashes `\` that are not followed by a complete escape
sequence result in a fatal error.

This PR also consolidates the `RawTextReader`'s I/O and error
handling for reading IVMs and values, which were largely
duplicates of one another.

### DIFF
--- a/src/text/parsers/comments.rs
+++ b/src/text/parsers/comments.rs
@@ -1,7 +1,7 @@
 use crate::text::parse_result::{IonParseResult, UpgradeIResult};
+use crate::text::parsers::whitespace;
 use nom::branch::alt;
 use nom::bytes::streaming::{is_not, tag, take_until};
-use nom::character::streaming::multispace1;
 use nom::combinator::recognize;
 use nom::multi::many0_count;
 use nom::sequence::{delimited, preceded};
@@ -11,8 +11,7 @@ use nom::sequence::{delimited, preceded};
 pub(crate) fn whitespace_or_comments(input: &str) -> IonParseResult<&str> {
     recognize(many0_count(alt((
         // At least one character of whitespace...
-        multispace1,
-        // ...or a comment of any format.
+        whitespace, // ...or a comment of any format.
         comment,
     ))))(input)
 }
--- a/src/text/parsers/mod.rs
+++ b/src/text/parsers/mod.rs
@@ -26,8 +26,18 @@ pub(crate) mod timestamp;
 pub(crate) mod top_level;
 pub(crate) mod value;
 
-const WHITESPACE_CHARACTERS: &[char] = &[' ', '\t', '\r', '\n'];
-const WHITESPACE_CHARACTERS_AS_STR: &str = " \t\r\n";
+const WHITESPACE_CHARACTERS: &[char] = &[
+    ' ',        // Space
+    '\t',       // Tab
+    '\r',       // Carriage return
+    '\n',       // Newline
+    '\u{0009}', // Horizontal tab
+    '\u{000B}', // Vertical tab
+    '\u{000C}', // Form feed
+];
+
+/// Same as [WHITESPACE_CHARACTERS], but formatted as a string for use in some `nom` APIs
+const WHITESPACE_CHARACTERS_AS_STR: &str = " \t\r\n\u{0009}\u{000B}\u{000C}";
 
 // ===== The functions below are used by several modules and live here for common access. =====
 
--- a/src/text/parsers/text_support.rs
+++ b/src/text/parsers/text_support.rs
@@ -1,12 +1,12 @@
 // Parsing functions that are common to textual types
 
-use crate::text::parse_result::{fatal_parse_error, IonParseResult, UpgradeIResult};
+use crate::text::parse_result::{fatal_parse_error, IonParseError, IonParseResult, UpgradeIResult};
 use nom::branch::alt;
 use nom::bytes::streaming::tag;
 use nom::character::streaming::{char, satisfy};
 use nom::combinator::{map, recognize, value};
 use nom::sequence::{preceded, tuple};
-use nom::{AsChar, IResult};
+use nom::{AsChar, IResult, Parser};
 
 /// The text Ion types each need to be able to read strings that contain escaped characters.
 /// This type represents the possible types of substring that make up any given piece of text from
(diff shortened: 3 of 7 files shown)


src/text/parsers/comments.rs
```diff
@@ -11,8 +11,7 @@ use nom::sequence::{delimited, preceded};
 pub(crate) fn whitespace_or_comments(input: &str) -> IonParseResult<&str> {
     recognize(many0_count(alt((
         // At least one character of whitespace...
-        multispace1,
-        // ...or a comment of any format.
+        whitespace, // ...or a comment of any format.
```
zslayton, line 14: :world_map: `multispace1` uses `nom`'s definition of whitespace. Now that the reader recognizes esoteric whitespace like vertical tab, this needs to use our custom `whitespace` parser.

src/text/parsers/text_support.rs
```diff
@@ -22,6 +22,26 @@ pub(crate) enum StringFragment<'a> {
...
+{
+    // If it doesn't start with a slash, it's not a match. Return a non-fatal error.
+    let (remaining, _slash) = char('\\')(input).upgrade()?;
+    // If the provided parser doesn't match what comes next, it's not a valid escape.
+    // Return a fatal error.
+    match parser.parse(remaining) {
+        Ok((remaining, string_fragment)) => Ok((remaining, string_fragment)),
+        Err(e) => fatal_parse_error(remaining, format!("could not parse {}: {}", label, e)),
+    }
+}
```
zslayton, line 43: :world_map: This helper function enforces that if the stream has a `\` in it, the thing that comes next must be a complete escape sequence. For example, `\u000B` is a valid unicode escape, but `\u00` is too short and should result in a fatal error.

Prior to this PR, if the escape sequence was incomplete it would be considered literal text instead. (`\u00` would appear in the string as `\00`.)

src/text/parsers/top_level.rs
```diff
@@ -1,17 +1,39 @@
...
+pub(crate) fn stream_item(input: &str) -> IonParseResult<RawTextStreamItem> {
+    alt((
+        map(ion_version_marker, |(major, minor)| {
+            RawTextStreamItem::IonVersionMarker(major, minor)
+        }),
+        map(top_level_value, |value| {
+            RawTextStreamItem::AnnotatedTextValue(value)
+        }),
+    ))(input)
+}
```
zslayton, line 35: :world_map: Rather than having the `RawTextReader` check for an `ion_version_marker` and then check for a `top_level_value`, it can now use the `stream_item` parser to consolidate those processes.

src/text/parsers/top_level.rs
```diff
@@ -40,9 +62,10 @@ pub(crate) fn version_int(input: &str) -> IonParseResult<&str> {
 /// https://amzn.github.io/ion-docs/docs/symbols.html#ion-version-markers
 pub(crate) fn ion_version_marker(input: &str) -> IonParseResult<(u32, u32)> {
     // See if the input text matches an IVM. If not, return a non-fatal error.
-    let (remaining_input, (_, major_text, _, minor_text)) = preceded(
+    let (remaining_input, (_, major_text, _, minor_text)) = delimited(
         whitespace_or_comments,
         tuple((tag("$ion_"), version_int, tag("_"), version_int)),
+        not(annotation_delimiter),
```
zslayton, line 68: If an IVM is followed by a `::`, it's an annotation and not an IVM. (e.g. `$ion_1_0::5`)

zslayton (commented):
:world_map: PR tour

src/text/parsers/mod.rs
```diff
@@ -26,8 +26,18 @@ pub(crate) mod timestamp;
...
-const WHITESPACE_CHARACTERS: &[char] = &[' ', '\t', '\r', '\n'];
-const WHITESPACE_CHARACTERS_AS_STR: &str = " \t\r\n";
+const WHITESPACE_CHARACTERS: &[char] = &[
+    ' ',        // Space
+    '\t',       // Tab
+    '\r',       // Carriage return
+    '\n',       // Newline
+    '\u{0009}', // Horizontal tab
+    '\u{000B}', // Vertical tab
+    '\u{000C}', // Form feed
```
jobarr-amzn, line 36: Is there a reason to prefer e.g. `\u{0009}` to `\x09`?

I was curious as the the complete set of built-in escapes like `\n` for Rust since some of these have built-in forms in other languages, so I looked up [ASCII escapes](https://doc.rust-lang.org/beta/reference/tokens.html#ascii-escapes) in the Rust docs.

src/text/parsers/top_level.rs
jobarr-amzn: 🔥

src/text/parsers/top_level.rs
```diff
@@ -136,26 +160,34 @@ mod parse_top_level_values_tests {
...
-    #[case("$ion_2_0 ", (2, 0))]
-    #[case("\n \n $ion_2_0 ", (2, 0))]
-    #[case("$ion_5_8 ", (5, 8))]
-    #[case("$ion_21_99 ", (21, 99))]
+    #[case("$ion_1_1 1", (1, 1))]
+    #[case("/*hello!*/ $ion_1_1 1", (1, 1))]
+    #[case("$ion_2_0 1", (2, 0))]
+    #[case("\n \n $ion_2_0 1", (2, 0))]
+    #[case("$ion_5_8 1", (5, 8))]
+    #[case("$ion_21_99 1", (21, 99))]
```
jobarr-amzn, line 181: Quibble, but as I read these I wonder whether it would be clearer to use a symbol like `some_symbol` instead of the number `1` as more immediately self-documenting?

jobarr-amzn (approved):
LGTM! Thanks as always for the detailed comments and PR tour comments.

src/text/parsers/mod.rs
zslayton: `\x09` would probably be better. I copied these values directly from an `ion-tests` example that was failing.

Incidentally, the reader's handling of Unicode escapes in general isn't quite right. It doesn't handle multi-codepoint escapes properly. When I fix that I'll revisit this.

src/text/parsers/top_level.rs
zslayton: I'd be ok with that, though the ` 1` is already being used in a lot of parser tests not visible in this diff. It'd need a broader change.

PR #384, opened 2022-05-18
Adds `NativeElementWriter`, string/clob validation, big `DecodedInt`s

zslayton:

* A native Rust `ElementWriter` implementation (`NativeElementWriter`)
has been added with basic testing; it is not yet being used in
`ion-tests`.
* ion-tests integration has been refactored to facilitate running
roundtrip tests with either the `IonCSliceElementWriter` or the
`NativeElementWriter` as desired.
* Fixed a bug in the ion-tests integration that caused the ion-c reader
to be used in some tests even when the native reader had been requested.
This change surfaced some additional failing ion-tests that have been
addressed in the changes below.
* The 64-bit size limit on the binary encoding primitive `Int` has been
lifted. This allows Timestamps and Decimals of arbitrary precision to be
read; write support for big `Int`s has not been added yet.
* Clobs now disallow unescaped non-ASCII characters and unescaped ASCII
control characters.
* Strings now disallow unescaped ASCII control characters.
* Strings and clobs now correctly decode utf-16 surrogate pairs.
* Long-form clobs and strings now normalize unescaped newlines.
* Fixed a bug in the `TextBuffer` that could cause a panic if restacking
the buffer's contents caused `truncate` to evaluate garbage data beyond
the truncation point.
* Fixed a bug in the way that Timestamp fractional seconds were being
converted to a Decimal.

### DIFF
--- a/src/binary/int.rs
+++ b/src/binary/int.rs
@@ -2,31 +2,40 @@ use std::mem;
 
 use crate::data_source::IonDataSource;
 use crate::result::{decoding_error, IonResult};
+use crate::types::coefficient;
+use crate::types::coefficient::Coefficient;
+use crate::Integer;
+use num_bigint::{BigInt, Sign};
+use num_traits::Zero;
 use std::io::Write;
 
 type IntStorage = i64;
-const MAX_INT_SIZE_IN_BYTES: usize = mem::size_of::<IntStorage>();
 const INT_NEGATIVE_ZERO: u8 = 0x80;
+
+// This limit is used for stack-allocating buffer space to encode/decode Ints.
+const INT_STACK_BUFFER_SIZE: usize = 16;
+// This number was chosen somewhat arbitrarily and could be lifted if a use case demands it.
+const MAX_INT_SIZE_IN_BYTES: usize = 2048;
 
 /// Represents a fixed-length signed integer. See the
 /// [UInt and Int Fields](http://amzn.github.io/ion-docs/docs/binary.html#uint-and-int-fields)
 /// section of the binary Ion spec for more details.
 #[derive(Debug)]
-pub struct Int {
+pub struct DecodedInt {
     size_in_bytes: usize,
-    value: IntStorage,
-    // [IntStorage] is not capable of natively representing negative zero. We track the sign
+    value: Integer,
+    // [Integer] is not capable of natively representing negative zero. We track the sign
     // of the value separately so we can distinguish between 0 and -0.
     is_negative: bool,
 }
 
-impl Int {
+impl DecodedInt {
     /// Reads an Int with `length` bytes from the provided data source.
-    pub fn read<R: IonDataSource>(data_source: &mut R, length: usize) -> IonResult<Int> {
+    pub fn read<R: IonDataSource>(data_source: &mut R, length: usize) -> IonResult<DecodedInt> {
         if length == 0 {
-            return Ok(Int {
+            return Ok(DecodedInt {
                 size_in_bytes: 0,
-                value: 0,
+                value: Integer::I64(0),
                 is_negative: false,
             });
         } else if length > MAX_INT_SIZE_IN_BYTES {
@@ -36,29 +45,65 @@ impl Int {
             ));
         }
 
-        // Create a stack-allocated buffer to hold the data we're going to read in.
-        let mut buffer = [0u8; MAX_INT_SIZE_IN_BYTES];
+        if length <= INT_STACK_BUFFER_SIZE {
+            let buffer = &mut [0u8; INT_STACK_BUFFER_SIZE];
+            DecodedInt::read_using_buffer(data_source, length, buffer)
+        } else {
+            // We're reading an enormous int. Heap-allocate a Vec to use as storage.
+            let mut buffer = vec![0u8; length];
+            DecodedInt::read_using_buffer(data_source, length, buffer.as_mut_slice())
+        }
+    }
+
+    pub fn read_using_buffer<R: IonDataSource>(
+        data_source: &mut R,
+        length: usize,
+        buffer: &mut [u8],
+    ) -> IonResult<DecodedInt> {
         // Get a mutable reference to a portion of the buffer just big enough to fit
         // the requested number of bytes.
         let buffer = &mut buffer[0..length];
 
         data_source.read_exact(buffer)?;
         let mut byte_iter = buffer.iter();
+        let mut is_negative: bool = false;
 
-        let first_byte: i64 = i64::from(byte_iter.next().copied().unwrap());
-        let sign: IntStorage = if first_byte & 0b1000_0000 == 0 { 1 } else { -1 };
-        let mut magnitude: IntStorage = first_byte & 0b0111_1111;
+        let value = if length <= mem::size_of::<i64>() {
+            // This Int will fit in an i64.
+            let first_byte: i64 = i64::from(byte_iter.next().copied().unwrap());
+            let sign: i64 = if first_byte & 0b1000_0000 == 0 {
+                1
+            } else {
+                is_negative = true;
+                -1
+            };
+            let mut magnitude: i64 = first_byte & 0b0111_1111;
+            for &byte in byte_iter {
+                let byte = i64::from(byte);
+                magnitude <<= 8;
+                magnitude |= byte;
+            }
+            Integer::I64(sign * magnitude)
+        } else {
+            // This Int is too big for an i64, we'll need to use a BigInt
+            let sign: num_bigint::Sign = if buffer[0] & 0b1000_0000 == 0 {
+                Sign::Plus
+            } else {
+                is_negative = true;
+                Sign::Minus
+            };
+            // We're going to treat the buffer's contents like the big-endian bytes of an
+            // unsigned integer. Now that we've made a note of the sign, set the sign bit
+            // in the buffer to zero.
+            buffer[0] &= 0b0111_1111;
+            let value = BigInt::from_bytes_be(sign, buffer);
+            Integer::BigInt(value)
+        };
 
-        for &byte in byte_iter {
-            let byte = i64::from(byte);
-            magnitude <<= 8;
-            magnitude |= byte;
-        }
-
-        Ok(Int {
+        Ok(DecodedInt {
             size_in_bytes: length,
-            value: magnitude * sign,
-            is_negative: sign == -1,
+            value,
+            is_negative,
         })
     }
 
@@ -97,13 +142,13 @@ impl Int {
         // `self.value` can natively represent any negative integer _except_ -0.
         // To check for negative zero, we need to also look at the sign bit that was encoded
         // in the stream.
-        self.value == 0 && self.is_negative
+        self.value.is_zero() && self.is_negative
     }
 
     /// Returns the value of the signed integer.
     #[inline(always)]
-    pub fn value(&self) -> IntStorage {
-        self.value
+    pub fn value(&self) -> &Integer {
+        &self.value
     }
 
     /// Returns the number of bytes that were read from the data source to construct this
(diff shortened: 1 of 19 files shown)


src/text/text_buffer.rs
```diff
@@ -122,6 +122,14 @@ impl<R: BufRead> TextBuffer<R> {
             // The invariants enforced by the `consume` method guarantee that this will behave safely.
```
zslayton, line 122: 🗺️ 

> // The `copy_within()` method below is unsafe.
> // ....
> // The invariants enforced by the `consume` method guarantee that this will behave safely.

Famous last words. 🤦

Once in a blue moon, this method would lead to a panic. Here's what happened.

`TextBuffer` has a `String` that acts as its reusable buffer. Any time we need more data from STDIN, it calls `read_line()` some number of times and appends the new text to the end of the string. When the reader finds an incomplete value at the end of the buffer, it shifts that remaining text to the beginning of the buffer and then reads more in. This works fine.

What I hadn't anticipated was the way that `String::truncate` works. When you truncate a `String` to a new length (`new_len`), [it checks to make sure that the byte at `new_len` is a character boundary](https://doc.rust-lang.org/src/alloc/string.rs.html#1202). What's surprising about this (at least to me) is that `new_len` is _beyond_ the end of the buffer after truncation. `String` is effectively looking at here-be-dragons data that used to have meaning. 🐉 

I was doing the `is_char_boundary` check _before_ shifting the remaining text backwards in the buffer and also each time that the reader called `consume` to flag bytes as having been read. This was functionally the same thing that `truncate` was doing. However, I never imagined that it would be checking the value of a byte that was not in the `String`'s readable range.

On rare occasions (like when reading in data that used surrogate pairs), the buffer would be filled with multi-byte code points. My `is_char_boundary` checks worked great, but after shifting the remaining text backwards in the buffer, the byte at `new_len` would be in the middle of what used to be a multi-byte code point. Kaboom.

This fixes the issue by setting the byte at `new_len` to a zero.

Anyway, that's what I get for using an `unsafe` method. 😞

src/types/magnitude.rs
```diff
@@ -94,25 +95,6 @@ impl Magnitude {
         }
         digits
     }
-
-    fn calculate_u64_digits(value: u64) -> u64 {
```
zslayton, line 98: 🗺️ This method was being used in a couple of different places, so I moved it to the module root for common access.

src/types/mod.rs
```diff
@@ -65,6 +65,26 @@ impl IonType {
     }
 }
 
+/// Returns the number of base-10 digits needed to represent `value`.
+fn num_decimal_digits_in_u64(value: u64) -> u64 {
```
zslayton, line 69: 🗺️ As mentioned earlier, this was moved to the root of the module for common access.

src/types/timestamp.rs
```diff
@@ -56,23 +56,14 @@ pub enum Mantissa {
...
 }
 
-/// Returns the number of base-10 digits needed to represent `value`.
-fn number_of_digits(value: u32) -> u32 {
-    match value {
-        0 => 1,
-        1 => 1,
-        i => (i as f64).log10().ceil() as u32,
-    }
-}
```
zslayton, line 74: 🗺️ This method was buggy and its usages have been replaced by calls to `num_decimal_digits_in_u64` above.

src/types/timestamp.rs
```diff
@@ -134,8 +125,16 @@ impl Timestamp {
...
+                const MAX_NANOSECOND_DIGITS: u32 = 9; // If it were 10, it'd be > a second
+                let nanoseconds = self.date_time.nanosecond();
+                let leading_zeros = MAX_NANOSECOND_DIGITS
+                    - super::num_decimal_digits_in_u64(nanoseconds as u64) as u32;
+                let coefficient = if leading_zeros >= *number_of_digits {
+                    0
+                } else {
+                    first_n_digits_of(*number_of_digits - leading_zeros, nanoseconds)
+                };
+                let exponent = -i64::from(*number_of_digits);
```
zslayton, line 137: 🗺️ When comparing two `Timestamp` values: if one timestamp has a `Decimal` fractional seconds and the other stores its fractional seconds as nanoseconds on the `DateTime`, the `eq` method will upconvert the nanoseconds fractional seconds to a `Decimal` for the sake of comparison.

The original implementation was overly simple; `ion-tests` shed light on a few corner cases that weren't handled properly.

src/value/native_writer.rs
```diff
@@ -0,0 +1,114 @@
...
+impl<W: Writer> ElementWriter for NativeElementWriter<W> {
+    type Output = W;
+
+    fn write<E: Element>(&mut self, element: &E) -> IonResult<()> {
+        self.write_element(None, element)
+    }
+
+    fn finish(mut self) -> IonResult<Self::Output> {
+        self.writer.flush()?;
+        Ok(self.writer)
```
zslayton, line 20: 🗺️ Calling `finish()` on the `ElementWriter` consumes it and returns the wrapped `Writer`.

tests/element_test_vectors.rs
```diff
@@ -34,60 +37,129 @@ fn contains_path(paths: &[&str], file_name: &str) -> bool {
...
+// Once all of the tests are passing, we should work to unify their APIs.
+trait RoundTrip {
+    /// Encodes `elements` to a buffer in the specified Ion `format` and then reads them back into
+    /// a `Vec<OwnedElement>` using the provided reader.
+    fn roundtrip<R: ElementReader>(
+        elements: &Vec<OwnedElement>,
+        format: Format,
+        reader: R,
+    ) -> IonResult<Vec<OwnedElement>>;
+}
```
zslayton, line 51: 🗺️ The `ion-c-sys`-powered writer has some lifetime constraints that make it kind of tough to work with. This trait will enable us to choose which writing implementation we want to run in each test.

tests/element_test_vectors.rs
```diff
@@ -34,60 +37,129 @@ fn contains_path(paths: &[&str], file_name: &str) -> bool {
...
+    fn roundtrip<R: ElementReader>(
+        elements: &Vec<OwnedElement>,
+        format: Format,
+        reader: R,
+    ) -> IonResult<Vec<OwnedElement>> {
+        // Unlike the C writer, the Rust writer can write into a growing Vec.
+        // No need for an aggressive preallocation.
+        let mut buffer = Vec::with_capacity(2048);
+        match format {
+            // TODO: Handle `TextKind`: Pretty
```
zslayton, line 86: 🗺️ `ion-rust` does not yet have a pretty printer. It should be relatively easy to implement, however, now that we have a `Writer` trait.

tests/element_test_vectors.rs
```diff
@@ -34,60 +37,129 @@ fn contains_path(paths: &[&str], file_name: &str) -> bool {
...
+        };
+
+        reader.read_all(buffer.as_slice())
+    }
+}
+
 trait ElementApi {
-    type ElementReader: ElementReader;
+    type ReaderApi: ElementReader;
+    type RoundTripper: RoundTrip;
```
zslayton, line 107: 🗺️ Now that each API can specify an implementation of `RoundTrip`, there's no need for the individual `make_text_writer`/`make_pretty_text_writer`/`make_binary_writer` methods below.

tests/element_test_vectors.rs
```diff
@@ -34,60 +37,129 @@ fn contains_path(paths: &[&str], file_name: &str) -> bool {
...
-            output.hex_dump()
+            "Roundtrip via {:?} failed: {}",
+            format,
+            Self::not_eq_error_message(&source_elements, &new_elements)
         );
         Ok(new_elements)
     }
 
-    fn assert_three_way_round_trip<F1, F2>(
+    fn not_eq_error_message(e1: &Vec<OwnedElement>, e2: &Vec<OwnedElement>) -> String {
```
zslayton, line 133: 🗺️ This method adds helpful context to round trip failures.

tests/element_test_vectors.rs
```diff
@@ -163,17 +235,18 @@ trait ElementApi {
...
-        reader: Self::ElementReader,
+        reader: Self::ReaderApi,
         file_name: &str,
         value_assert: F1,
         group_assert: F2,
     ) -> IonResult<()>
     where
-        F1: Fn(&OwnedElement, &OwnedElement),
+        // group index, value 1 index, value 1, value 2 index, value 2
+        F1: Fn(usize, usize, &OwnedElement, usize, &OwnedElement),
```
zslayton, line 242: 🗺️ This closure now takes indexes for the group being evaluated and for each of the elements being compared. This made debugging failures much easier.

tests/element_test_vectors.rs
```diff
@@ -34,60 +37,126 @@ fn contains_path(paths: &[&str], file_name: &str) -> bool {
...
-    {
-        let mut buf = vec![0u8; WRITE_BUF_LENGTH];
-        let mut writer = make_writer(&mut buf)?;
-        writer.write_all(source_elements)?;
-        let output = writer.finish()?;
-        let new_elements = element_reader().read_all(output)?;
+        format: Format,
+    ) -> IonResult<Vec<OwnedElement>> {
+        let new_elements =
+            Self::RoundTripper::roundtrip(source_elements, format, Self::make_reader())?;
```
zslayton, line 123: 🗺️ Blink and you'll miss it! I know because _I_ missed it. 😞 

The `Self::make_reader()` at the end of this line used to be a call to `element_reader()`, which meant that for lots of tests the native Rust reader wasn't being used; `element_reader()` always returns an `IonCSliceElementWriter`.

This PR fixed that, bringing a handful of other test failures to light. This PR also fixed most of them.

tests/element_test_vectors.rs
```diff
@@ -341,7 +408,17 @@ fn equivs<E: ElementApi>(_element_api: E, file_name: &str) {
         E::read_group(
             E::make_reader(),
             file_name,
-            |this, that| assert!(this.ion_eq(that)),
+            |group_index, this_index, this, that_index, that| {
+                assert!(
+                    this.ion_eq(that),
+                    "in group {}, index {} ({:?}) was not ion_eq to index {} ({:?})",
```
zslayton, line 414: 🗺️ Here and below: lots of additional context around test failures.

tests/element_test_vectors.rs
```diff
@@ -616,6 +674,14 @@ mod native_element_tests {
...
         fn round_trip_skip_list() -> SkipList {
             &[
+                // These tests fail because the ion-c writer doesn't fully step out of the last
+                // container it writes in some circumstances, leading to a panic.
+                "ion-tests/iontestdata/good/testfile25.ion",
+                "ion-tests/iontestdata/good/testfile33.ion",
+                "ion-tests/iontestdata/good/non-equivs/timestamps.ion",
+                "ion-tests/iontestdata/good/equivs/timestamps.ion",
+                "ion-tests/iontestdata/good/equivs/timestampFractions.ion",
+                // -----
```
zslayton, line 684: 🗺️ I plan to leave the `ion-c`-writer-related failures on the skip list for now and move on to wiring up the `NativeElementWriter`. I assume it also has bugs, but that's a higher priority for investing time in fixes.

tests/element_test_vectors.rs
```diff
@@ -664,79 +740,64 @@ mod native_element_tests {
...
-
-        fn make_binary_writer(buffer: &mut [u8]) -> IonResult<SliceElementWriter> {
-            Format::Binary.element_writer_for_slice(buffer)
-        }
     }
 
     #[test_resources("ion-tests/iontestdata/good/**/*.ion")]
     #[test_resources("ion-tests/iontestdata/good/**/*.10n")]
     fn native_good_roundtrip_text_binary(file_name: &str) {
-        good_roundtrip_text_binary(NativeElementApi {}, file_name)
```
zslayton, line 690: 🗺️ Originally, `NativeElementApi` was defined like this:

```rust
struct NativeElementApi {}
```

This PR changed it to:
```rust
struct NativeElementApi;
```

which means that all usages can now drop their `{}`s.

zslayton (commented):
🗺️ PR tour

src/types/timestamp.rs
```diff
@@ -1431,4 +1431,38 @@ mod ionc_tests {
         assert_eq!(source, converted_source);
         Ok(())
     }
+
+    #[test]
+    fn ion_eq_fraction_seconds_mixed_mantissa() {
+        let t1 = Timestamp {
+            date_time: NaiveDateTime::from_str("1857-05-29T19:25:59.100").unwrap(),
```
jobarr-amzn, line 1438: What is this date? Arbitrary? Closest thing I can find in history is the founding of the precursor bank to BBVA, on May 28, 1857.

tests/element_test_vectors.rs
jobarr-amzn: Yes, I like this plan.

jobarr-amzn (approved):

jobarr-amzn:
I still find the conversation/threading model in GitHub PRs to be just bizarre.

src/binary/int.rs
```diff
@@ -36,29 +45,65 @@ impl Int {
...
+                Sign::Plus
+            } else {
+                is_negative = true;
+                Sign::Minus
+            };
+            // We're going to treat the buffer's contents like the big-endian bytes of an
+            // unsigned integer. Now that we've made a note of the sign, set the sign bit
+            // in the buffer to zero.
+            buffer[0] &= 0b0111_1111;
+            let value = BigInt::from_bytes_be(sign, buffer);
```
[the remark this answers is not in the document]
zslayton: We could! The `i64` implementation was one of the first bits of `ion-rust` I wrote, and I wrote it before `from_bytes_be` was a feature--[it wasn't stabilized until 2019](https://blog.rust-lang.org/2019/01/17/Rust-1.32.0.html).

src/text/parsers/clob.rs
```diff
@@ -112,12 +142,43 @@ fn short_clob_fragment(input: &str) -> IonParseResult<StringFragment> {
...
+    Err(Incomplete(Needed::Unknown))
+}
+
+fn char_is_legal_clob_ascii(c: char) -> bool {
+    // Depending on where you look in the spec and/or `ion-tests`, you'll find conflicting
+    // information about which ASCII characters can appear unescaped in a clob. Some say
+    // "characters >= 0x20", but that excludes lots of whitespace characters that are < 0x20.
+    // Some say "displayable ASCII", but DEL (0x7F) is shown to be legal in one of the ion-tests.
+    // The definition used here has largely been inferred from the contents of `ion-tests`.
+    c.is_ascii() && (u32::from(c) >= 0x20 || WHITESPACE_CHARACTERS.contains(&c))
```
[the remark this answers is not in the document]
zslayton: 👍 https://github.com/amzn/ion-tests/issues/76

src/types/timestamp.rs
zslayton: I pulled it from a `timestamp` test that was failing for closer inspection.

Issue #381, opened 2022-05-15
Implement `IonEq` for `Decimal` and `Timestamp`


zslayton:
`Decimal` and `Timestamp` both have `PartialEq` implementations based on Ion's definition of equivalence. This is overly strict for many use cases. For example:
* Decimal `0.0` and `-0.0` are considered unequal according to Ion, but would be considered mathematically equivalent.
* Timestamps representing the same point in time but at different precisions or offsets would be considered unequal according to Ion, but would be considered mathematically equivalent.

Now that we have the `IonEq` trait (discussed in issue #220 and added in PR #382), we should move these Ion-centric equality comparisons to `IonEq` implementations for those types and provide value-based equivalence definitions of `PartialEq`.

PR #386, opened 2022-05-22
Writer tests; fixes:symbols, timestamps, big Ints

zslayton:

* Enables `ion-tests` integration for the `NativeElementWriter`,
the pure-Rust DOM writing API. The `IonCSliceElementWriter` is now only
used when paired with the `IonCElementReader`.
* Adds a `Writer::write_integer` method that can encode an `Integer` of
any size.
* Lifts the size restriction on the `Int` encoding primitive, allowing
`ion-rs` to support `Timestamp`s and `Decimal`s of truly arbitrary
precision when necessary.
* Modifies the `RawTextWriter` to:
   * escape literals like `\n` and `\r`
   * detect symbols whose text resembles a symbol ID (for example:
`'$99'`) and wrap them in quotes to keep them from being interpreted
     after roundtripping
* Removes the `FractionalSeconds` variant of `Precision`, making it more
straightforward to compare `Timestamp`s that specify 'empty' fractional
seconds values like `0d0` with `Timestamp`s that specify no fractional
seconds at all.
* Makes it possible to downgrade a high-precision `Timestamp` to a
`DateTime` with nanoseconds (potentially losing precision in the process).

### DIFF
--- a/src/binary/raw_binary_writer.rs
+++ b/src/binary/raw_binary_writer.rs
@@ -3,6 +3,8 @@ use std::ops::Range;
 use std::{io, mem};
 
 use bytes::BufMut;
+use num_bigint::Sign;
+use num_traits::Zero;
 
 use crate::binary::constants::v1_0::IVM;
 use crate::binary::uint::DecodedUInt;
@@ -13,7 +15,7 @@ use crate::types::decimal::Decimal;
 use crate::types::timestamp::Timestamp;
 use crate::types::SymbolId;
 use crate::writer::Writer;
-use crate::IonType;
+use crate::{Integer, IonType};
 
 use super::decimal::DecimalBinaryEncoder;
 use super::timestamp::TimestampBinaryEncoder;
@@ -566,8 +568,9 @@ impl<'a, W: Write> Writer for RawBinaryWriter<W> {
     /// Writes an Ion integer with the specified value.
     fn write_i64(&mut self, value: i64) -> IonResult<()> {
         self.write_scalar(|enc_buffer| {
-            let magnitude = value.abs() as u64;
-            let encoded = uint::encode_uint(magnitude);
+            // Get the absolute value of the i64 and store it in a u64.
+            let magnitude: u64 = value.unsigned_abs();
+            let encoded = uint::encode_u64(magnitude);
             let bytes_to_write = encoded.as_bytes();
 
             // The encoded length will never be larger than 8 bytes, so it will
@@ -580,6 +583,44 @@ impl<'a, W: Write> Writer for RawBinaryWriter<W> {
             };
             enc_buffer.push(type_descriptor);
             enc_buffer.extend_from_slice(bytes_to_write);
+
+            Ok(())
+        })
+    }
+
+    /// Writes an Ion integer with the specified value.
+    fn write_integer(&mut self, value: &Integer) -> IonResult<()> {
+        // If the `value` is an `i64`, use `write_i64` and return.
+        let value = match value {
+            Integer::I64(i) => return self.write_i64(*i),
+            Integer::BigInt(i) => i,
+        };
+
+        // From here on, `value` is a `BigInt`.
+        self.write_scalar(|enc_buffer| {
+            if value.is_zero() {
+                enc_buffer.push(0x20);
+                return Ok(());
+            }
+
+            let (sign, magnitude_be_bytes) = value.to_bytes_be();
+
+            let mut type_descriptor: u8 = match sign {
+                Sign::Plus | Sign::NoSign => 0x20,
+                Sign::Minus => 0x30,
+            };
+
+            let encoded_length = magnitude_be_bytes.len();
+            if encoded_length <= 13 {
+                type_descriptor |= encoded_length as u8;
+                enc_buffer.push(type_descriptor);
+            } else {
+                type_descriptor |= 0xEu8;
+                enc_buffer.push(type_descriptor);
+                VarUInt::write_u64(enc_buffer, encoded_length as u64)?;
+            }
+
+            enc_buffer.extend_from_slice(magnitude_be_bytes.as_slice());
 
             Ok(())
         })
@@ -602,7 +643,7 @@ impl<'a, W: Write> Writer for RawBinaryWriter<W> {
     /// Writes an Ion float with the specified value.
     fn write_f64(&mut self, value: f64) -> IonResult<()> {
         self.write_scalar(|enc_buffer| {
-            if value == 0f64 {
+            if value == 0f64 && !value.is_sign_negative() {
                 enc_buffer.push(0x40);
                 return Ok(());
             }
@@ -854,8 +895,10 @@ mod writer_tests {
     use crate::raw_symbol_token::{local_sid_token, RawSymbolToken};
     use crate::symbol::Symbol;
     use crate::StreamReader;
+    use num_bigint::BigInt;
     use num_traits::Float;
     use std::convert::TryInto;
+    use std::str::FromStr;
 
     type TestWriter<'a> = RawBinaryWriter<&'a mut Vec<u8>>;
     type TestReader<'a> = Reader<RawBinaryReader<std::io::Cursor<&'a [u8]>>>;
@@ -1091,6 +1134,15 @@ mod writer_tests {
         expect_scalar(reader, IonType::Integer, |r| r.read_i64(), value);
     }
 
+    fn expect_big_integer(reader: &mut TestReader, value: &BigInt) {
+        expect_scalar(
+            reader,
+            IonType::Integer,
+            |r| r.read_integer(),
+            Integer::BigInt(value.clone()),
+        );
+    }
+
     fn expect_float(reader: &mut TestReader, value: f64) {
         expect_scalar(reader, IonType::Float, |r| r.read_f64(), value);
     }
@@ -1162,6 +1214,35 @@ mod writer_tests {
         writer.step_out()?;
         writer.step_out()?;
         Ok(())
+    }
+
+    #[test]
+    fn binary_writer_large_integers() -> IonResult<()> {
+        // 11 byte UInt
+        let big_positive = BigInt::from_str("123456789123456789123456789").unwrap();
+        // 19 byte UInt
+        let very_big_positive =
+            BigInt::from_str("123456789123456789123456789123456789123456789").unwrap();
+        let big_negative = -big_positive.clone();
+        let very_big_negative = -very_big_positive.clone();
+        binary_writer_test(
+            |writer| {
+                writer.write_integer(&Integer::BigInt(BigInt::zero()))?;
+                writer.write_integer(&Integer::BigInt(big_positive.clone()))?;
+                writer.write_integer(&Integer::BigInt(very_big_positive.clone()))?;
+                writer.write_integer(&Integer::BigInt(big_negative.clone()))?;
+                writer.write_integer(&Integer::BigInt(very_big_negative.clone()))?;
+                Ok(())
+            },
+            |reader| {
+                expect_big_integer(reader, &BigInt::zero());
+                expect_big_integer(reader, &big_positive);
+                expect_big_integer(reader, &very_big_positive);
+                expect_big_integer(reader, &big_negative);
+                expect_big_integer(reader, &very_big_negative);
+                Ok(())
+            },
+        )
     }
 
     #[test]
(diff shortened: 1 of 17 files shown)


src/text/raw_text_writer.rs
```diff
@@ -373,20 +449,8 @@ impl<'a, W: Write> Writer for RawTextWriter<W> {
 
             write!(output, ":{:0>2}", datetime.second())?;
             //                   ^-- delimiting colon, formatted second
-            if value.precision == Precision::Second {
```
zslayton, line 376: 🗺️ `Second` and `FractionalSeconds` are no longer distinct variants.

src/text/raw_text_writer.rs
```diff
@@ -373,20 +449,8 @@ impl<'a, W: Write> Writer for RawTextWriter<W> {
...
-                return Ok(());
-            }
-
-            // We want the fractional seconds to show as many decimal places as are interesting.
-            // Here we format the string as nanoseconds and then eliminate any trailing zeros.
-            //TODO: The initial conversion to datetime will discard precision greater than nanoseconds.
-            //TODO: Eliminate this allocation.
-            let fractional_text = format!("{:0>9}", datetime.nanosecond());
-            let fractional_text = fractional_text.trim_end_matches('0');
+            value.fmt_fractional_seconds(&mut *output)?;
```
zslayton, line 452: 🗺️ A more robust implementation of text fractional seconds serialization is now provided by `Timestamp` directly. It is `pub(crate)`, keeping it out of the public API for now.

src/text/raw_text_writer.rs
```diff
@@ -403,7 +467,9 @@ impl<'a, W: Write> Writer for RawTextWriter<W> {
     /// Writes the provided &str value as an Ion string.
     fn write_string<S: AsRef<str>>(&mut self, value: S) -> IonResult<()> {
         self.write_scalar(|output| {
-            write!(output, "\"{}\"", value.as_ref())?;
+            write!(output, "\"")?;
+            RawTextWriter::write_escaped_text_body(output, value)?;
+            write!(output, "\"")?;
```
zslayton, line 472: 🗺️ The `RawTextWriter` now escapes characters that require it in the body of a `string`.

src/types/coefficient.rs
```diff
@@ -59,6 +61,24 @@ impl Coefficient {
...
+
+    /// Returns true if the Coefficient represents a zero of any sign.
+    pub(crate) fn is_zero(&self) -> bool {
+        match (self.sign, &self.magnitude) {
+            (_, Magnitude::U64(0)) => true,
+            (_, Magnitude::BigUInt(b)) if b.is_zero() => true,
+            _ => false,
+        }
+    }
+
```
zslayton, line 81: 🗺️ These convenience methods are helpful when comparing timestamp `Mantissa`s.

src/types/decimal.rs
```diff
@@ -170,7 +170,7 @@ impl Decimal {
         // We do this by multiplying it times 10^exponent_delta, which is 1 in this case.
         // This lets us compare 80 and 80, determining that the decimals are equal.
         let mut scaled_coefficient: BigUint = d1.coefficient.magnitude().to_biguint().unwrap();
-        scaled_coefficient *= 10u64.pow(exponent_delta as u32);
+        scaled_coefficient *= BigUint::from(10u64).pow(exponent_delta as u32);
```
zslayton, line 173: 🗺️ An `ion-test` contained a large enough value that using `u64` led to overflow.

src/types/timestamp.rs
```diff
@@ -19,10 +23,8 @@ pub enum Precision {
     Day,
     /// Minute-level precision (e.g. `2020-08-01T12:34Z`)
     HourAndMinute,
-    /// Second-level precision. (e.g. `2020-08-01T12:34:56Z`)
+    /// Second-level precision or greater. (e.g. `2020-08-01T12:34:56Z` or `2020-08-01T12:34:56.123456789Z`)
     Second,
-    /// Sub-second precision (e.g. `2020-08-01T12:34:56.123456789Z`)
-    FractionalSeconds,
```
zslayton, line 25: 🗺️ This file contains most of the changes related to removing the distinct `FractionalSeconds` variant of `Precision`.

src/types/timestamp.rs
```diff
@@ -307,23 +418,18 @@ impl PartialEq for Timestamp {
...
     }
 }
 
-// We cannot provide an implementation of [Ord] for [Timestamp] because many instances cannot be
-// meaningfully compared to each other to derive an ordering. For example, a Timestamp with
-// Precision::Year cannot be compared to a timestamp with Precision::Day, and a Timestamp with
-// an unknown offset cannot be compared to one with a known offset.
-//TODO: It would be possible to provide a PartialOrd for Timestamps with the same precision and
-//      either known or unknown offsets. It's unclear how valuable that would be.
-
```
zslayton, line 326: 🗺️ This comment is outdated/incorrect. It's possible to provide friendly definitions of `PartialEq` and `PartialOrd` and relegate the precision-aware definition of equality to `IonEq`. See #381.

src/types/timestamp.rs
```diff
@@ -1296,6 +1390,40 @@ mod timestamp_tests {
         assert_eq!(12345678, super::first_n_digits_of(8, 123456789));
         assert_eq!(123456789, super::first_n_digits_of(9, 123456789));
     }
+
```
zslayton, line 1393: 🗺️ These unit tests had accidentally been placed in the testing module for `ion-c`'s timestamp code. They've been moved to the native impl testing module. You'll see the corresponding deletion immediately following.

src/value/native_writer.rs
```diff
@@ -56,9 +55,7 @@ impl<W: Writer> NativeElementWriter<W> {
         match element.ion_type() {
             IonType::Null => unreachable!("element has IonType::Null but is_null() was false"),
             IonType::Boolean => self.writer.write_bool(element.as_bool().unwrap()),
-            IonType::Integer => self
-                .writer
-                .write_i64(element.as_integer().unwrap().as_i64().unwrap()),
+            IonType::Integer => self.writer.write_integer(element.as_integer().unwrap()),
```
zslayton, line 58: 🗺️ The native element writer can now write an integer of arbitrary size; no need for an `as_i64()`.

tests/element_test_vectors.rs
```diff
@@ -649,20 +652,12 @@ mod native_element_tests {
...
-                // DecodedInt to store an Integer instead of an i64.
-                "ion-tests/iontestdata/good/typecodes/T5.10n",
                 // These files are encoded in utf16 and utf32; the reader currently assumes utf8.
                 "ion-tests/iontestdata/good/utf16.ion",
                 "ion-tests/iontestdata/good/utf32.ion",
                 // NON-EQUIVS
-                // These tests fail because the ion-c writer bindings don't support writing -0.0
-                "ion-tests/iontestdata/good/non-equivs/decimals.ion",
-                "ion-tests/iontestdata/good/non-equivs/floatsVsDecimals.ion",
-                // -----
```
zslayton, line 665: 🗺️ Now that we're not using the `ion-c` writer for the native tests, the failures caused by the `ion-c` writer are no longer a problem.

zslayton (commented):
🗺️ PR tour

src/types/timestamp.rs
```diff
@@ -197,12 +247,68 @@ impl Timestamp {
...
+            return Ok(());
+        }
+        let mantissa = self.fractional_seconds.as_ref().unwrap();
+        if mantissa.is_empty() {
+            // No need to write anything.
+            return Ok(());
+        }
+        match mantissa {
+            Mantissa::Digits(num_digits) => {
+                let scaled = self.date_time.nanosecond() / 10u32.pow(9 - *num_digits); // 123,000,000 -> 123
```
jobarr-amzn, line 275:
```suggestion
                // e.g. 123,000,000 -> 123 when num_digits = 6 denoting microsecond precison
                let scaled = self.date_time.nanosecond() / 10u32.pow(9 - *num_digits); 
```

tests/element_test_vectors.rs
jobarr-amzn: 🎊

jobarr-amzn (approved):
Once again thanks for the PR tour comments, and thanks for diving on all of these issues.

src/binary/raw_binary_writer.rs
```diff
@@ -585,6 +588,44 @@ impl<'a, W: Write> Writer for RawBinaryWriter<W> {
...
+        let value = match value {
+            Integer::I64(i) => return self.write_i64(*i),
+            Integer::BigInt(i) => i,
+        };
+
+        // From here on, `value` is a `BigInt`.
+        self.write_scalar(|enc_buffer| {
+            if value.is_zero() {
+                enc_buffer.push(0x20);
+                return Ok(());
```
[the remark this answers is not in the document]
zslayton: I'm broadly for it. However, there are some weird corner cases where it's not obvious what the behavior should be. For example, if I step into a struct, a text writer would return `1` for the opening `{`. In a binary writer, I won't know what the length of the header is until I step out. Similarly, methods like `set_field_name` or `set_annotations` don't necessarily write anything immediately, but they could. I think the easiest place to return the number of bytes written with a consistent behavior would be the `flush()` method.

src/text/raw_text_writer.rs
```diff
@@ -112,16 +120,47 @@ impl<W: Write> RawTextWriter<W> {
             && chars.all(|c| c == '$' || c == '_' || c.is_ascii_alphanumeric())
     }
 
+    /// Returns `true` if the provided text is an Ion keyword. Keywords like `true` or `null`
+    /// resemble identifiers, but writers must wrap them in quotes when using them as symbol text.
+    fn token_is_keyword(token: &str) -> bool {
+        const KEYWORDS: &[&str] = &["true", "false", "nan", "null"];
+        KEYWORDS.contains(&token)
+    }
```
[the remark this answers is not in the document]
zslayton: In this case it actually is trivial. Check out [the optimized output in the compiler explorer](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,selection:(endColumn:9,endLineNumber:2,positionColumn:9,positionLineNumber:2,selectionStartColumn:9,selectionStartLineNumber:2,startColumn:9,startLineNumber:2),source:'//+Type+your+code+here,+or+load+an+example.%0A++++pub+fn+token_is_keyword(token:+%26str)+-%3E+bool+%7B%0A++++++++const+KEYWORDS:+%26%5B%26str%5D+%3D+%26%5B%22true%22,+%22false%22,+%22nan%22,+%22null%22%5D%3B%0A++++++++KEYWORDS.contains(%26token)%0A++++%7D%0A%0A//+If+you+use+%60main()%60,+declare+it+as+%60pub%60+to+see+it+in+the+output:%0A//+pub+fn+main()+%7B+...+%7D%0A'),l:'5',n:'0',o:'Rust+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:r1610,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,libs:!(),options:'-C++++++++++++++++opt-level%3D3',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1,tree:'1'),l:'5',n:'0',o:'rustc+1.61.0+(Rust,+Editor+%231,+Compiler+%231)',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4). The function is inlined and the loop is unrolled, which is possible because the array is `const`. It boils down to a few `cmp`/`je`s:

```
example::token_is_keyword:
        cmp     rsi, 3
        je      .LBB0_7
        cmp     rsi, 5
        je      .LBB0_6
        cmp     rsi, 4
        jne     .LBB0_5
        mov     al, 1
        cmp     dword ptr [rdi], 1702195828
        je      .LBB0_9
        cmp     dword ptr [rdi], 1819047278
        jne     .LBB0_5
```

In any case, I haven't done any performance work on the text writer. Once we start profiling, I'm sure there will be plenty of lower-hanging fruit to choose from. :stuck_out_tongue:

src/types/timestamp.rs
zslayton: I like this, I'm going to add it in #387 to simplify the merge.

PR #387, opened 2022-05-23
Implements IonEq for Timestamp, Decimal

### LINKED ISSUE #381: Implement `IonEq` for `Decimal` and `Timestamp`

zslayton:
This PR depends on #386; when that one is merged, this one will be rebased on top of `main`.

------

Timestamp equality (using `PartialEq` or `==`) now checks to see if the
two `Timestamp`s represent the same instant in time after applying their
respective offsets. Precision is ignored. The original field-by-field
comparison that considers precision has been moved to an implementation
of the `IonEq` trait.

Decimal equality (using `PartialEq` or `==`) now checks to see if the
two `Decimal`s represent the same mathematical value, ignoring
precision. In particular, positive and negative zeros of any exponent
are considered equal. The original comparison that requires the sign,
magnitude, and exponent to all match exactly has been moved to an
implementation of the `IonEq` trait.

Fixes #381.

### DIFF
--- a/src/binary/decimal.rs
+++ b/src/binary/decimal.rs
@@ -5,6 +5,7 @@ use std::io::Write;
 use arrayvec::ArrayVec;
 use bigdecimal::Zero;
 
+use crate::ion_eq::IonEq;
 use crate::{
     binary::{
         int::DecodedInt, raw_binary_writer::MAX_INLINE_LENGTH, var_int::VarInt, var_uint::VarUInt,
@@ -48,7 +49,7 @@ where
 {
     fn encode_decimal(&mut self, decimal: &Decimal) -> IonResult<usize> {
         // 0d0 has no representation, as per the spec.
-        if decimal == &DECIMAL_POSITIVE_ZERO {
+        if decimal.ion_eq(&DECIMAL_POSITIVE_ZERO) {
             return Ok(0);
         }
 
@@ -137,13 +138,18 @@ mod binary_decimal_tests {
     use super::*;
     use rstest::*;
 
-    /// This test ensures that we implement PartialEq correctly for the special
+    /// This test ensures that we implement [PartialEq] and [IonEq] correctly for the special
     /// decimal value 0d0.
     #[test]
     fn decimal_0d0_is_a_special_zero_value() {
         assert_eq!(DECIMAL_POSITIVE_ZERO, Decimal::new(0, 0));
-        assert_ne!(DECIMAL_POSITIVE_ZERO, Decimal::new(0, 10));
-        assert_ne!(DECIMAL_POSITIVE_ZERO, Decimal::new(0, 100));
+        assert!(DECIMAL_POSITIVE_ZERO.ion_eq(&Decimal::new(0, 0)));
+
+        assert_eq!(DECIMAL_POSITIVE_ZERO, Decimal::new(0, 10));
+        assert!(!DECIMAL_POSITIVE_ZERO.ion_eq(&Decimal::new(0, 10)));
+
+        assert_eq!(DECIMAL_POSITIVE_ZERO, Decimal::new(0, 100));
+        assert!(!DECIMAL_POSITIVE_ZERO.ion_eq(&Decimal::new(0, 100)));
     }
 
     #[rstest]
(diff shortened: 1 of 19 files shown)


src/binary/decimal.rs
```diff
@@ -48,7 +49,7 @@ where
 {
     fn encode_decimal(&mut self, decimal: &Decimal) -> IonResult<usize> {
         // 0d0 has no representation, as per the spec.
-        if decimal == &DECIMAL_POSITIVE_ZERO {
+        if decimal.ion_eq(&DECIMAL_POSITIVE_ZERO) {
             return Ok(0);
         }
```
zslayton, line 54: 🗺️ `IonEq` is stricter than `PartialEq`; `==` will ignore differences in sign/exponent when comparing two `Decimal` zeros.

src/binary/decimal.rs
```diff
@@ -137,13 +138,18 @@ mod binary_decimal_tests {
...
         assert_eq!(DECIMAL_POSITIVE_ZERO, Decimal::new(0, 0));
-        assert_ne!(DECIMAL_POSITIVE_ZERO, Decimal::new(0, 10));
-        assert_ne!(DECIMAL_POSITIVE_ZERO, Decimal::new(0, 100));
+        assert!(DECIMAL_POSITIVE_ZERO.ion_eq(&Decimal::new(0, 0)));
+
+        assert_eq!(DECIMAL_POSITIVE_ZERO, Decimal::new(0, 10));
+        assert!(!DECIMAL_POSITIVE_ZERO.ion_eq(&Decimal::new(0, 10)));
+
+        assert_eq!(DECIMAL_POSITIVE_ZERO, Decimal::new(0, 100));
+        assert!(!DECIMAL_POSITIVE_ZERO.ion_eq(&Decimal::new(0, 100)));
```
zslayton, line 152: 🗺️ Unit tests that previously relied on `==`/`eq()` now use `ion_eq()`.

`ion_eq()` is stricter than `eq`, so if we're making an assertion about two values being `IonEq` then we can also assert that they are `PartialEq`. Likewise, if we say they're _not_ `IonEq`, we can also assert that they are not `PartialEq`.

src/types/coefficient.rs
```diff
@@ -191,7 +192,12 @@ mod coefficient_tests {
         let neg_zero = Decimal::new(Coefficient::negative_zero(), 0);
         let pos_zero = Decimal::new(0, 0);
         assert_eq!(neg_zero, neg_zero);
-        assert_ne!(neg_zero, pos_zero);
+        assert!(neg_zero.ion_eq(&neg_zero));
+
+        assert_eq!(neg_zero, pos_zero);
+        assert!(!neg_zero.ion_eq(&pos_zero));
```
zslayton, line 198: 🗺️ `-0.` and `0.` are `PartialEq` but not `IonEq`.

src/types/timestamp.rs
```diff
@@ -1176,32 +1233,36 @@ mod timestamp_tests {
...
         Ok(())
     }
 
     #[test]
     fn test_timestamps_with_different_days_are_not_equal() -> IonResult<()> {
         let builder = Timestamp::with_year(2021).with_month(2);
-        let timestamp1 = builder.clone().with_day(5).build();
-        let timestamp2 = builder.with_day(6).build();
+        let timestamp1 = builder.clone().with_day(5).build()?;
+        let timestamp2 = builder.with_day(6).build()?;
```
zslayton, line 1244: 🗺️ A few of these unit tests accidentally forgot to unwrap the `Result<_>` using `?`. This worked when using `assert_eq` because it can compare two `Result` values via `PartialEq`. However, when I added calls to `ion_eq` it rightly complained that `IonEq` is not implemented for `Result`.

jobarr-amzn (approved):
LGTM, thanks

PR #388, opened 2022-05-27
Reader/Writer builders, pretty text writer

zslayton:
This PR introduces (barebones) builder APIs for readers and
writers of different abstraction levels. This will allow us to add
new configuration elements over time in a backwards-compatible way.

It also adds a way to construct a "pretty" `TextWriter` that adds
generous, human-friendly whitespace to its output. This new version of
the `TextWriter` has been added to the `ion-tests` alongside the
standard one.

Finally, the `Reader` type has been renamed to `UserReader` to better
align with the names for the `Raw*Reader` and `SystemReader` types. The
name `Reader` is now an alias for a `UserReader<Box<dyn RawReader>>`,
which is the type of `Reader` we expect users to create the most often.
A `ReaderBuilder` will detect whether the provided data is text or
binary and return a `Reader` that abstracts over the format
differences via `Box<dyn RawReader>`.


src/text/ion_data_source.rs
```diff
@@ -0,0 +1,67 @@
+use std::fs::File;
+use std::io;
+use std::io::{BufRead, BufReader, Read};
+
+/// Types that implement this trait can be converted into an implementation of [io::BufRead],
+/// allowing users to build a [Reader] from a variety of types that might not define I/O operations
+/// on their own.
+pub trait ToIonDataSource {
```
zslayton, line 8: 🗺️ This trait was previously called `TextIonDataSource`, but it's actually useful for any format. I've renamed it to be more general.

A `Reader` can be built from any type that implements `ToIonDataSource`. I can call `ReaderBuilder::new().build(...)` with a `&str` or `String` of Ion text, a `Vec` of bytes, a file to read from, etc.

src/text/raw_text_writer.rs
```diff
@@ -9,15 +9,134 @@ use crate::raw_symbol_token_ref::{AsRawSymbolTokenRef, RawSymbolTokenRef};
...
+use crate::types::ContainerType;
 use crate::writer::Writer;
 use crate::{Integer, IonType};
 
+pub struct RawTextWriterBuilder {
+    space_between_values: String,
+    indentation: String,
+    space_after_field_name: String,
+    space_after_container_start: String,
+}
```
zslayton, line 21: 🗺️ There are two big changes in this file. First, I've added a builder API that allows whitespace in the raw text writer to be customized. Second, I've used this functionality to allows applications to make a 'pretty' writer.

src/text/raw_text_writer.rs
```diff
@@ -9,15 +9,134 @@ use crate::raw_symbol_token_ref::{AsRawSymbolTokenRef, RawSymbolTokenRef};
...
+        // other builder APIs and to allow for fallible setup operations in the future.
+        Ok(raw_text_writer)
+    }
+}
+
+#[derive(Debug, PartialEq, Default)]
+struct EncodingLevel {
+    container_type: ContainerType,
+    child_count: usize,
+}
```
zslayton, line 128: 🗺️ The writer needs to track how many values have been written in each container so it knows whether to write a delimiter (`,`) before the next one.

src/text/raw_text_writer.rs
```diff
@@ -81,27 +188,68 @@ impl<W: Write> RawTextWriter<W> {
...
-        let delimiter = match self.containers.last() {
-            Some(Struct) | Some(List) => ",",
-            Some(SExpression) => " ",
-            Some(scalar) => unreachable!("Inside a non-container type: {:?}", scalar),
-            None => "\n", // Top-level values appear on their own line
+        use ContainerType::*;
+        let delimiter = match self.parent_level().container_type {
+            TopLevel => "",
+            Struct | List => ",",
+            SExpression => "",
```
zslayton, line 221: 🗺️ Using a `ContainerType` instead of `Option<IonType>` means that I don't have to handle `unreachable!` arms of the `match` any more.

src/text/text_data_source.rs
```diff
@@ -1,42 +0,0 @@
-use std::io;
-use std::io::{BufRead, BufReader};
-
-/// Types that implement this trait can be converted into an implementation of [io::BufRead],
-/// allowing them to be passed to the [TextReader::new] constructor. This allows [TextReader::new]
-/// to support reading Strings, &str slices, &[u8] slices, files, etc.
-pub trait TextIonDataSource {
```
zslayton, line 7: 🗺️ The renamed version of this trait/file appeared earlier in the diff: `ToIonDataSource`.

src/types/mod.rs
```diff
@@ -173,3 +173,19 @@ mod type_test {
...
+
+// Represents a level into which the writer has stepped.
+// A writer that has not yet called step_in() is at the top level.
+#[derive(Debug, PartialEq)]
+pub(crate) enum ContainerType {
+    TopLevel,
+    SExpression,
+    List,
+    Struct,
+}
```
zslayton, line 185: 🗺️ This type was moved out of the raw binary writer earlier in the diff.

src/value/native_reader.rs
```diff
@@ -26,23 +26,9 @@ impl ElementReader for NativeElementReader {
         &'a self,
         data: &'b [u8],
     ) -> IonResult<Box<dyn Iterator<Item = IonResult<OwnedElement>> + 'b>> {
-        // Match against the slice to see if it starts with the binary IVM
```
zslayton, line 29: 🗺️ This code was moved to the body of `ReaderBuilder::build` earlier in the diff. (It would probably make sense to offer it as a utility method, but I'll take care of that in a future PR.)

tests/element_test_vectors.rs
```diff
@@ -696,12 +699,7 @@ mod native_element_tests {
         }
 
         fn non_equivs_skip_list() -> SkipList {
-            &[
-                // Decimal's `eq` considers 123. and 123.0 to be equal (which they are)
-                // but we expect it to test for Ion equality, not value equality.
-                // This is related to https://github.com/amzn/ion-rust/issues/381
-                "ion-tests/iontestdata/good/non-equivs/decimals.ion",
-            ]
```
zslayton, line 704: 🗺️ I'm unsure why this shows up in the diff; I fixed it in a previous PR.

tests/good_test_vectors.rs
```diff
@@ -141,7 +141,7 @@ fn read_all_values(reader: &mut Reader<RawBinaryReader<BufReader<File>>>) -> Ion
                 reader.step_out()?;
             }
             String => {
-                let _text = reader.map_string(|_s| ())?;
+                let _text = reader.read_string()?;
```
zslayton, line 144: 🗺️ This test used the non-object-safe methods we plan to remove. I've switched it to use the safe-but-more-expensive equivalent methods for now.

zslayton (commented):
🗺️ PR tour

desaikd (approved):
LGTM 🚀

PR #389, opened 2022-05-27
Makes ion-c-sys an opt-in feature

zslayton:
`ion-c-sys` provides bindings to the `ion-c` library, allowing Rust
applications to use `ion-c`'s readers and writers to work with Ion data.
Now that the native Rust readers and writers are passing the `ion-tests`
conformance suite, `ion-c-sys` is no longer mandatory.

This PR introduces a new feature, "ion-c-sys", that allows users to opt
into the functionality and dependencies that come with `ion-c-sys`.

For users that do not require `ion-c-sys`, this eliminates the need to
install a C toolchain on their development machine (cmake, clang, etc).
It also cuts build times and test runtimes by more than half.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,9 @@ members = [
   "ion-hash"
 ]
 
+[features]
+ion_c = ["dep:ion-c-sys"]
+
 [dependencies]
 base64 = "0.12"
 bigdecimal = "0.2"
@@ -42,11 +45,10 @@ arrayvec = "0.7"
 # NB: We use the tree dependency here for development and CI.
 #     Note that when publishing you should update the version
 #     so that users can get the correct underlying ion-c-sys version.
-ion-c-sys = { path = "./ion-c-sys", version = "0.4" }
+ion-c-sys = { path = "./ion-c-sys", version = "0.4", optional = true }
 
 [dev-dependencies]
 rstest = "0.9"
-
 # Used by ion-tests integration
 walkdir = "2.3"
 test-generator = "0.3"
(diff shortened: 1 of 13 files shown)


tests/element_test_vectors.rs
```diff
@@ -51,23 +48,8 @@ trait RoundTrip {
 
 /// These unit structs implement the [RoundTrip] trait and can be passed throughout the
 /// integration tests to dictate which readers/writers should be used in each test.
-struct IonCElementWriterApi;
 struct NativeElementWriterApi;
 
-impl RoundTrip for IonCElementWriterApi {
```
zslayton, line 57: 🗺️ This code was moved into an `ion_c`-centric module later in the file.

tests/element_test_vectors.rs
```diff
@@ -13,9 +13,6 @@ use std::fs::read;
 use std::path::MAIN_SEPARATOR as PATH_SEPARATOR;
 use test_generator::test_resources;
 
-/// Buffer size for our writing tests around round-tripping
-const WRITE_BUF_LENGTH: usize = 16 * 1024 * 1024;
```
zslayton, line 17: 🗺️ This code was moved into an `ion_c`-centric module later in the file.

Cargo.toml
```diff
@@ -26,6 +26,9 @@ members = [
   "ion-hash"
 ]
 
+[features]
+ion_c = ["dep:ion-c-sys"]
```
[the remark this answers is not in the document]
jobarr-amzn: I [wondered](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features) how one enabled this, if one were an existing `ion-rs` user and wanted to keep existing ion-c behaviors. It should be like this:

```toml
[dependencies]
# Enables the `ion_c` feature of ion-rs.
ion-rs = { version = "0.10.0", features = ["ion_c"] }
```

Right?

jobarr-amzn (approved):

PR #391, opened 2022-06-03
adds text formatter and `Display` of `OwnedElement`

desaikd:
*Description of changes:*
This PR works on implementing a text formatter that will be used for `Display` of `OwnedElement` and for raw text writer. 

*List of changes:*
- adds changes for text formatter
- adds `Display` implementation of `OwnedElement`

*Test:*
adds unit tests for text formatter

*Notes:*
I will open a subsequent PR to remove some of the method from raw text writer to use this text formatter.


src/text/raw_text_writer.rs
```diff
@@ -148,7 +148,7 @@ pub struct RawTextWriter<W: Write> {
 /**
  * String escape codes, for Ion Clob.
  */
-fn string_escape_code_init() -> Vec<String> {
+pub(crate) fn string_escape_code_init() -> Vec<String> {
```
zslayton, line 151: Couple of thoughts here:
* This is now only used in the `IonValueFormatter`, so I think it makes sense for it to live in that module instead.
* I implemented this as a `Vec<String>` because it would only be constructed once per text writer. Now that it needs to exist for every `IonValueFormatter` (which we want to be very cheap to create), I think we should take the time to convert it to a `const STRING_ESCAPE_CODES: &'static [&'static str]`. It'll be a bit ugly as a literal, but then it'll be free to use/reference everywhere.

src/text/raw_text_writer.rs
```diff
@@ -406,7 +406,7 @@ impl<W: Write> RawTextWriter<W> {
 
     /// Writes the body (i.e. no start or end delimiters) of a string or symbol with any illegal
     /// characters escaped.
-    fn write_escaped_text_body<S: AsRef<str>>(
+    pub(crate) fn write_escaped_text_body<S: AsRef<str>>(
```
zslayton, line 409: If this is only used in `IonValueFormatter`, let's move it to that module.

src/text/text_formatter.rs
```diff
@@ -0,0 +1,401 @@
+use crate::raw_symbol_token_ref::AsRawSymbolTokenRef;
+use crate::types::timestamp::Precision;
+use crate::value::owned::{OwnedSequence, OwnedStruct, OwnedSymbolToken};
+use crate::value::{Sequence, Struct, SymbolToken};
+use crate::{Decimal, Integer, IonResult, IonType, RawTextWriter, Timestamp};
+use chrono::{DateTime, Datelike, FixedOffset, NaiveDateTime, TimeZone, Timelike};
+use std::convert::TryInto;
+use std::io::BufWriter;
+
+pub struct IonValueFormatter<'a, W: std::io::Write> {
```
zslayton, line 10: This could use a doc comment.

src/text/text_formatter.rs
```diff
@@ -0,0 +1,401 @@
...
+    pub(crate) string_escape_codes: Vec<String>,
+}
+
+impl<'a, W: std::io::Write> IonValueFormatter<'a, W> {
+    pub(crate) fn format_annotations(
+        &mut self,
+        annotations: &Vec<OwnedSymbolToken>,
+    ) -> IonResult<()> {
+        for annotation in annotations {
+            self.format_symbol(annotation.text().unwrap())?;
```
zslayton, line 21: Rather than calling `text().unwrap()` (which could panic), I think we should implement `Display` for `OwnedSymbolToken` and have it write out either `foo` or `$23` as its able to.

src/text/text_formatter.rs
```diff
@@ -0,0 +1,401 @@
...
+            return Ok(());
+        }
+
+        write!(
+            self.output,
+            "T{:0>2}:{:0>2}",
+            datetime.hour(),
+            datetime.minute()
+        )?;
+        //                   ^-- delimiting T, formatted hour, delimiting colon, formatted minute
```
zslayton, line 139: `cargo fmt` has thrown off the spacing for this comment. Could you move it up a couple of lines and line up the `^--` to the `T`?

src/text/text_formatter.rs
```diff
@@ -0,0 +1,401 @@
...
+        let sign = if offset_minutes >= 0 { "+" } else { "-" };
+        let offset_minutes = offset_minutes.abs();
+        let hours = offset_minutes / MINUTES_PER_HOUR;
+        let minutes = offset_minutes % MINUTES_PER_HOUR;
+        write!(self.output, "{}{:0>2}:{:0>2}", sign, hours, minutes)?;
+        Ok(())
+    }
+
+    pub(crate) fn format_symbol<A: AsRawSymbolTokenRef>(&mut self, value: A) -> IonResult<()> {
+        RawTextWriter::write_symbol_token(&mut BufWriter::new(&mut *self.output), value)?;
```
zslayton, line 171: Constructing a `BufWriter` is pretty expensive. We can probably make `write_symbol_token` accept a `W: io::Write` instead.

Should `write_symbol_token` be moved to this module?

src/text/text_formatter.rs
```diff
@@ -0,0 +1,401 @@
...
+        Ok(())
+    }
+
+    pub(crate) fn format_string<S: AsRef<str>>(&mut self, value: S) -> IonResult<()> {
+        write!(self.output, "\"")?;
+        RawTextWriter::write_escaped_text_body(&mut BufWriter::new(&mut *self.output), value)?;
```
zslayton, line 177: Same thought here RE: `BufWriter`.

src/text/text_formatter.rs
```diff
@@ -0,0 +1,401 @@
...
+    }
+
+    pub(crate) fn format_struct(&mut self, value: &OwnedStruct) -> IonResult<()> {
+        write!(self.output, "{{ ")?;
+        let mut peekable_itr = value.iter().peekable();
+        while peekable_itr.peek() != None {
+            let (field_name, field_value) = peekable_itr.next().unwrap();
+            self.format_symbol(field_name.text().unwrap())?;
+            write!(self.output, ": ")?;
+            write!(self.output, "{}", field_value)?;
```
zslayton, line 218: We can merge these to make:
```suggestion
            write!(self.output, ": {}", field_value)?;
```

src/text/text_formatter.rs
```diff
@@ -0,0 +1,401 @@
...
+
+    pub(crate) fn format_blob<A: AsRef<[u8]>>(&mut self, value: A) -> IonResult<()> {
+        write!(self.output, "{{{{{}}}}}", base64::encode(value))?;
+        Ok(())
+    }
+
+    pub(crate) fn format_struct(&mut self, value: &OwnedStruct) -> IonResult<()> {
+        write!(self.output, "{{ ")?;
+        let mut peekable_itr = value.iter().peekable();
+        while peekable_itr.peek() != None {
```
zslayton, line 214: For the container formatters, consider rearranging your loop a bit:

```rust
// Iterate normally...
for (field_name, field_value) in iter {
    self.format_symbol(field_name.text().unwrap())?;
    write!(self.output, ": {}", field_value)?;
    // ...but before moving to the next iteration, check and see if this is the last one.
    if iter.peek().is_some() {
        write!(self.output, ", ")?;
    }
}
```

src/value/owned.rs
```diff
@@ -464,6 +467,39 @@ impl OwnedElement {
     }
 }
 
+impl Display for OwnedElement {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), std::fmt::Error> {
+        let mut ivf = IonValueFormatter {
+            output: &mut Vec::new(),
+            string_escape_codes: string_escape_code_init(),
+        };
```
zslayton, line 475: Ah, shoot. I hadn't thought about this bit. 😞 

We don't want to have to allocate a new `Vec` and buffer all the output each time someone uses `Display` -- we'd like to be able to write directly to `f`. Unfortunately, I think this means we have to use `fmt::Write` throughout `IonValueFormatter` instead of `io::Write`. (If we genuinely need to format text to an `io::Write` in some other context, I can think of a way to achieve that.)

Let me know if you'd like to discuss this offline. The `io::Write`/`fmt::Write` divide is definitely one of the stranger parts of the standard library that I've encountered. The

src/value/owned.rs
```diff
@@ -470,8 +469,7 @@ impl OwnedElement {
 impl Display for OwnedElement {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), std::fmt::Error> {
         let mut ivf = IonValueFormatter {
-            output: &mut Vec::new(),
-            string_escape_codes: string_escape_code_init(),
+            output: &mut String::new(),
```
zslayton, line 472: No need for a temporary `String`, this can be the parameter `f`.

src/value/owned.rs
```diff
@@ -495,7 +493,7 @@ impl Display for OwnedElement {
         }
         .map_err(|_| std::fmt::Error)?;
 
-        f.write_str(String::from_utf8(ivf.output.to_vec()).unwrap().as_str())?;
+        f.write_str(ivf.output)?;
```
zslayton, line 496: You can remove this after setting `ivf.output` to `f`.

zslayton (approved):
Looks good! One substantive comment below. Please also address the clippy warning (not necessarily the gcov warnings) before merging.

Issue #351, opened 2022-01-29
Support stream reading/parsing


mlartz:
Given an Ion file of a series of records (e.g. an object in S3), we'd like to be able to stream read/parse this file, ultimately creating an async (Tokio) `Stream` of Ion records.  Due to memory constraints (Lambda, Fargate, etc), we cannot read the entire file into memory and then iterate through the records.

zslayton:
To clarify a bit: the `Reader` is already capable of reading a file without pulling the entire thing into memory. However, if the input source it is reading from contains incomplete data, the `Reader` would either block (e.g. waiting on a socket) or fail (e.g. at the end of a `Vec<u8>`).

We'd like to offer an API that allows the `Reader` to be resilient to such a use case. For example, calls to `next()` or `read_string()` might return an `IonError::InsufficientData` (or `IonError::WouldBlock`?), indicating that the user should try again once more data is available.

zslayton:
We now have non-blocking text and binary readers, but they can't be constructed with the `ReaderBuilder` API yet. Once that's done (see #484), we can close this issue.

PR #394, opened 2022-06-25
Adds a non-blocking raw binary reader

### LINKED ISSUE #4: Promote the header byte jump table to a constant
The `create_header_byte_jump_table` function in `binary/header.rs` will parse each possible byte value from 0 to 255 as though it were [the one-octet type descriptor](http://amzn.github.io/ion-docs/docs/binary.html#typed-value-formats) found at the beginning of all binary Ion values. The output from parsing each byte is stored in a newly allocated `Vec` that can be used as a jump table to avoid re-calculating the meaning of same byte values repeatedly.

This is sub-optimal:
* It requires a `Vec` to be allocated when a fixed-size array would do
* It requires entities that wish to use the jump table to maintain their own copy of the `Vec` despite the fact that its contents will always be identical.

Ideally, we would store a fixed-length array as a static constant in the `header` module. Unlike C, however, the Rust compiler tries very hard to prevent programs from mutating global state. This means that initializing a static array is currently somewhat difficult to do without sacrificing either speed or safety.

Here are some options I've explored:

### safe + slow

The [`lazy_static` crate](https://crates.io/crates/lazy_static) provides a macro that allows you to declare and initialize a global constant. It works by wrapping your constant in a custom smartpointer type--the first time the pointer is dereferenced, `lazy_static` will call a user-defined function to calculate the value of the constant. Every dereference that occurs from then on will receive the already-calculated value.

This is easy to implement, but profiling with `valgrind` has shown that the indirection introduced by the smartpointer indirection adds an unacceptable amount of overhead in terms of CPU time. Derefencing the jump table each time you begin parsing a value in a stream with millions of values adds up quickly.

`thread_local` copies of the jump table suffer from even worse indirection overhead, rendering them a non-starter.

If/when we add an `IonSystem` concept, we may be able to have `IonReader`s share a reference to a shared jump table living in the `IonSystem`, but this may not be worth the complexity.

### unsafe + fast

Initializing a fixed-size array (i.e. `[T]`, not a Vec<T>) [is currently surprisingly challenging](https://www.joshmcguigan.com/blog/array-initialization-rust/) for non-trivial types. The best methods available only work for types that implement [the `Copy` trait](https://doc.rust-lang.org/std/marker/trait.Copy.html), and types that implement [`Default`](https://doc.rust-lang.org/std/default/trait.Default.html) and where the array is <32 entries long.

We could choose to skirt this problem with `unsafe` code, but this has its own problems:
* The compiler cannot guarantee the memory safety of `unsafe` code, so we would need to scrutinize it very carefully. 
* Because the array would be initialized at runtime (not build time), we would need to design an appropriate synchronization mechanism to guarantee that readers didn't reference the jump table before it was initialized.
* The APIs for working with not-yet-initialized data are in a state of flux. The Rust ecosystem is migrating away from [`mem::uninitialized`](https://doc.rust-lang.org/std/mem/fn.uninitialized.html) (currently deprecated in nightly) and towards [`mem::MaybeUninit`](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html) (currently unstable).

### safe + fast + not possible yet
Rust is incrementally adding support for `const fn`s, functions whose outputs can be calculated entirely at compile time. This is perfect for our use case -- all of the inputs are statically known (`u8` values from 0 to 255) and processing them requires no information from the outside world. With this feature, we'll be able to simply write:

```rust
const HEADER_JUMP_TABLE: [IonResult<Option<IonValueHeader>>] = create_jump_table();
```

The compiler will calculate the definition of `HEADER_JUMP_TABLE` at build time and then all interested entities can refer to it freely -- no wasted memory, no additional overhead from indirection.

At the time of this writing, `const fn`s have severe restrictions on which language features may be used. Functionality like `match` statements, the `?` operator, calling non-`const` functions, looping, etc. are planned but not yet supported. These restrictions are so limiting that I have not been able to write an array initialization `const fn`.

zslayton:
The existing `RawBinaryReader` wraps an `io::BufRead` implementation
and pulls additional bytes from it as needed. This allows it to read
large files in a streaming fashion; however, it also means that the
file must be complete. If the data source is still growing (for
example: a file that is being appended to), the streaming reader will
treat the end of the stream as an unexpected EOF and error out.
Likewise, in async contexts where bytes can arrive in separate events,
trying to read chunk of Ion bytes can fail if the slice doesn't contain
a complete Ion stream.

This PR adds two new types: a `BinaryBuffer` and a `RawBinaryBufferReader`.

The `BinaryBuffer` is a stack-allocated wrapper around an implementation of
`AsRef<[u8]>`. It provides methods to read binary Ion encoding primitives like
`VarUInt`, `VarInt`, and `NOP` from the data source. It is very cheap to
create and does not modify the data source.

The `RawBinaryBufferReader` provides a full `Reader` interface over a
`BinaryBuffer`. If it encounters an unexpected EOF while reading, it returns
an `IonError::Incomplete`. Unlike the original `RawBinaryReader`, however,
it is guaranteed to still be in a good state afterwards. If more data
becomes available later on, it can be appended to the buffer and the
`RawBinaryBufferReader` can try parsing it again.

Finally, a new configuration for the `ion-tests` integration suite has been
added that exercises the `Reader` interface in the `RawBinaryBufferReader`,
demonstrating that it is spec-compliant.

This PR duplicates a fair amount of functionality that is available
elsewhere in the crate. Follow-on PRs will consolidate these implementations;
it is expected that the blocking raw binary reader will be re-implemented
as a wrapper around the non-blocking reader, making calls to `read` whenever
an `Incomplete` is encountered.

Fixes #4, related to #351.


src/binary/non_blocking/binary_buffer.rs
```diff
@@ -0,0 +1,856 @@
...
+// This number was chosen somewhat arbitrarily and could be lifted if a use case demands it.
+const MAX_INT_SIZE_IN_BYTES: usize = 2048;
+
+/// A stack-allocated wrapper around an `AsRef<[u8]>` that provides methods to read Ion's
+/// encoding primitives.
+///
+/// When the wrapped type is a `Vec<u8>`, data can be appended to the buffer between read
+/// operations.
+#[derive(Debug, PartialEq)]
+pub(crate) struct BinaryBuffer<A: AsRef<[u8]>> {
```
zslayton, line 32: 🗺️ This type allows you to read Ion encoding primitives from a `&[u8]`, `Vec<u8>`, `bytes::Buf`, or any other container of bytes.

There are many code blocks that are basically copy/pasted from the blocking implementation (for example: reading `VarUInt`, `VarInt`, etc). I intend to do a lot of consolidation in a follow-on PR.

zslayton (commented):
🗺️ PR tour

src/binary/non_blocking/raw_binary_reader.rs
```diff
@@ -0,0 +1,1824 @@
...
+    /// be [ReaderState::Skipping], indicating that it is mid-stream and awaiting more data.
+    fn advance_to_next_item(&mut self) -> IonResult<()> {
+        use ReaderState::*;
+        match &mut self.state {
+            // We're done advancing and are ready to try reading the current byte.
+            Ready => Ok(()),
+            // The `Skipping(n)` state means that we ran out of data on a previous call to `next()`.
+            // The only way to fix it is to add more input bytes to the buffer using the `append_bytes`
+            // or `read_from` methods. Adding input bytes to the buffer is only supported
+            // when the generic type `A` is `Vec<u8>`.
```
zslayton, line 359: 🗺️ There's a tradeoff lurking here. Moving to the `Skipping(n)` state means that we leave the reader in no man's land; no operation will succeed until more data has been added.

We could instead have the reader simply return `IonError::Incomplete` without changing its state. However:
* This would mean that the only time the reader could advance is when the entire current value is in the buffer. Given that some real-world use cases work with multi-GB values, this feels like an onerous requirement. Having  a `Skipping(n)` state allows the reader to incrementally skip over very large values.
* If a call to `next()` (which in turn calls `advance_to_next_item()`) fails as `Incomplete`, I don't think users would expect to be able to perform more read operations on the value they were on before calling `next()`.

If a use case for the alternative behavior surfaces down the road, I could see adding a config option for this (`disable_incremental_advancement`?).

src/result.rs
```diff
@@ -106,6 +133,14 @@ impl PartialEq for IonError {
     }
 }
 
+pub fn incomplete_data_error<T>(label: &'static str, offset: usize) -> IonResult<T> {
+    Err(incomplete_data_error_raw(label, offset))
+}
+
+pub fn incomplete_data_error_raw(label: &'static str, offset: usize) -> IonError {
```
desaikd, line 140: Why not `#[inline(never)]` this raw error function as well similar to other raw error functions.

src/binary/non_blocking/binary_buffer.rs
```diff
@@ -0,0 +1,808 @@
...
+            magnitude |= lower_seven;
+            encoded_size_in_bytes += 1;
+            if byte >= 0b1000_0000 {
+                terminated = true;
+                break;
+            }
+        }
+
+        if !terminated {
+            // panic!("incomplete (unterm) varint @ pos {}: {:x?}", self.total_consumed(), self.bytes());
```
desaikd, line 229: debug comment?
```suggestion

```

src/binary/non_blocking/binary_buffer.rs
```diff
@@ -0,0 +1,808 @@
...
+            return decoding_error(format!(
+                "Found a {}-byte VarInt. Max supported size is {} bytes.",
+                encoded_size_in_bytes, MAX_ENCODED_SIZE_IN_BYTES
+            ));
+        }
+
+        self.consume(encoded_size_in_bytes);
+        Ok(VarInt::new(
+            magnitude * sign,
+            !is_positive,
```
desaikd, line 246: Suggestion: Everywhere we use `!is_positive` with `VarInt::new`, why not just use `is_negative` in declaration above and use it as is. Only reason I suggest this is because I had to go back and look at the definition of `VarInt::new` to make sure the above statement is correct.

src/result.rs
zslayton: Great question! The other error methods each allocate a `String` via `format!`, which causes their assembly to include code for allocation, formatting, and deallocation. `incomplete_data_error_raw` holds a `&'static str`, so it doesn't need to allocate; its assembly is pretty small for now.

src/binary/non_blocking/binary_buffer.rs
zslayton: Done!

src/result.rs
zslayton: Here's an assembly dump for `decoding_error_raw` (which is behind `inline(never)`):

image

In contrast, here's the assembly for `incomplete_data_error_raw`, which usually gets inlined:

image

I'm using [`cargo asm`](https://github.com/gnzlbg/cargo-asm) to generate these.

src/binary/non_blocking/binary_buffer.rs
```diff
@@ -56,7 +56,7 @@ impl<A: AsRef<[u8]>> BinaryBuffer<A> {
             data: &self.data,
             start: self.start,
             end: self.end,
-            total_consumed: 0,
+            total_consumed: self.total_consumed,
```
zslayton, line 59: 🗺️ Previously, the transaction buffer would track its own total consumption. However, I realized this wasn't terribly useful in its existing role. Having it copy the original buffer's `total_consumed` means that if it encounters errors, the error position that it reports will be the absolute offset in the Ion stream which is what we want.

Incidentally, this piece of information was the only reason we stored a reference to the original buffer in the `TxReader`, so that's been removed as well.

src/binary/non_blocking/raw_binary_reader.rs
```diff
@@ -1030,9 +996,7 @@ impl<'a> Iterator for AnnotationsIterator<'a> {
 /// In this way, the RawBinaryBufferReader will never be in a bad state. It only updates when the
 /// TxReader has already found the next item.
 struct TxReader<'a, A: AsRef<[u8]>> {
-    ion_version: (u8, u8),
     state: &'a mut ReaderState,
-    main_buffer: &'a BinaryBuffer<A>,
```
zslayton: 🗺️ Now that the tx_buffer below has the original buffer's `total_consumed`, this field has been removed.

zslayton (commented):
These comments describe changes I introduced in the latest commit. If you hadn't looked at earlier commits, you can probably ignore them and just read the latest version.

src/binary/non_blocking/raw_binary_reader.rs
```diff
@@ -57,6 +59,9 @@ struct EncodedValue {
     // The number of bytes used to encode the field ID (if present) preceding the Ion value. If
     // `field_id` is undefined, `field_id_length` will be zero.
     field_id_length: u8,
+    // If this value is inside a struct, `field_id` will contain the SymbolId that represents
+    // its field name.
+    field_id: Option<SymbolId>,
     // The number of bytes used to encode the annotations wrapper (if present) preceding the Ion
     // value. If `annotations` is empty, `annotations_length` will be zero.
```
desaikd, line 66:
```suggestion
    // value. If `annotations` is empty, `annotations_header_length` will be zero.
```

src/binary/non_blocking/raw_binary_reader.rs
```diff
@@ -69,6 +74,12 @@ struct EncodedValue {
     // The number of bytes used to encode the value itself, not including the header byte
     // or length fields.
     value_length: usize,
+    // The sum total of:
+    //     field_id_length + annotations_header_length + self.header_length() + self.value_length()
```
desaikd, line 78:
```suggestion
     //     field_id_length + annotations_header_length + header_length + value_length
```

desaikd (approved):

nirosys (approved):

PR #396, opened 2022-06-30
Optimizes reading binary symbol and integer values

### LINKED ISSUE #395: Remove support for oversized SymbolIds
The binary readers (blocking and non-blocking) have nominal support for oversized `SymbolId`s. That is: `SymbolId`s that are encoded using more than `usize` bytes (8 on 64-bit platforms, 4 on 32-bit).

There are two reasons for this:
* `SymbolId`s are encoded as `UInt`s. Because integers are also encoded as `UInt`s, the primary decoding method for `UInt` supports reading `UInt`s of arbitrary size.
* The [`good/typecodes/T7-large.10n`](https://github.com/amzn/ion-tests/blob/master/iontestdata/good/typecodes/T7-large.10n) file in `ion-tests` has several symbols that are encoded using more than 8 bytes. In order for that test to pass, we have to support oversized symbols.

With that said, having this support doesn't really make sense:
* On a 64-bit machine, having a symbolId that doesn't fit in a `usize` means that you have over 18 quintillion symbols. A symbol table with that many symbols would not fit in memory.
* `good/typecodes/T7-large.10n` uses zero-padded symbol IDs to create symbols that large. We can make that test pass by simply ignoring leading zeros. Even that is of dubious value, however, as that adds code to support a single contrived use case.

In the non-blocking reader, replacing this method implementation:
```rust
    /// If the reader is currently positioned on a symbol value, parses that value into a `SymbolId`.
    pub fn read_symbol_id(&mut self) -> IonResult<SymbolId> {
        let (encoded_value, mut buffer) = self.value_and_buffer(IonType::Symbol)?;
        match buffer.read_uint(encoded_value.value_length())?.value() {
            UInteger::U64(symbol_id) => {
                // This will always succeed on 64-bit platforms where u64 and usize are the same.
                if let Ok(sid) = usize::try_from(*symbol_id) {
                    Ok(sid)
                } else {
                    decoding_error("found a u64 symbol ID that was too large to fit in a usize")
                }
            }
            UInteger::BigUInt(symbol_id) => Self::try_symbol_id_from_big_uint(symbol_id),
        }
    }
```

with something like this:
```rust
    /// If the reader is currently positioned on a symbol value, parses that value into a `SymbolId`.
    pub fn read_symbol_id(&mut self) -> IonResult<SymbolId> {
        let (_encoded_value, bytes) = self.value_and_bytes(IonType::Symbol)?;
        if bytes.len() > mem::size_of::<usize>() {
            return decoding_error("found a symbol Id that was too large to fit in a usize");
        }
        let mut magnitude: usize = 0;
        for &byte in bytes {
            let byte = usize::from(byte);
            magnitude <<= 8;
            magnitude |= byte;
        }
        Ok(magnitude)
}
```

shrunk the compiled assembly enough that it caused overall read speeds on a large file (1.3GB, containing a variety of data types) to increase by 4%.

zslayton:
This PR depends on the changes in PR #394. After that PR has been merged, this one will be rebased on top of `main`.

------

This PR:

* Eliminates redundant boundary checking that happens when
integer and symbol values are read. It does this by offering
unchecked UInt decoding methods that read a single value from a
provided slice.
* Removes support for reading symbol IDs that require more than 8
bytes to encode. This limits the size of the symbol table to about
18 quintillion symbol values, which seems reasonable.
* Adds Into<Integer> implementations for Rust's integer primitives.

Together, these reduced the time needed to read every value in a
1.3GB Ion log file by about 6%.

Fixes #395.


src/binary/non_blocking/binary_buffer.rs
```diff
@@ -93,7 +93,7 @@ impl<A: AsRef<[u8]>> BinaryBuffer<A> {
     /// If the buffer is not empty, returns `Some(_)` containing the next byte in the buffer.
     /// Otherwise, returns `None`.
     pub fn peek_next_byte(&self) -> Option<u8> {
-        self.bytes().get(0).copied()
+        self.data.as_ref().get(self.start).copied()
```
zslayton, line 96: 🗺️ The original code created a byte slice (`&[u8]`) even though we only needed the first byte in the data. Now it just gets the first byte without creating an intermediate slice.

src/binary/non_blocking/binary_buffer.rs
```diff
@@ -262,7 +262,7 @@ impl<A: AsRef<[u8]>> BinaryBuffer<A> {
     ///
     /// See: https://amzn.github.io/ion-docs/docs/binary.html#uint-and-int-fields
     pub fn read_uint(&mut self, length: usize) -> IonResult<DecodedUInt> {
-        if length <= mem::size_of::<usize>() {
+        if length <= mem::size_of::<u64>() {
```
zslayton, line 265: 🗺️ Here and below: the code was originally checking for `usize`, which had the potential to be overly conservative. On 32-bit systems, this would cause medium-sized values to be pushed into a `BigInt`.

In some cases (like reading a `UInt` that represents a `SymbolId`), limiting it to `usize` makes sense. However, that's not this method's responsibility.

src/binary/non_blocking/binary_buffer.rs
```diff
@@ -271,18 +271,13 @@ impl<A: AsRef<[u8]>> BinaryBuffer<A> {
...
         let uint_bytes = self
             .peek_n_bytes(length)
             .ok_or_else(|| incomplete_data_error_raw("a UInt", self.total_consumed()))?;
-        let mut magnitude: u64 = 0;
-        for &byte in uint_bytes {
-            let byte = u64::from(byte);
-            magnitude <<= 8;
-            magnitude |= byte;
-        }
+        let magnitude = DecodedUInt::small_uint_from_slice(uint_bytes);
```
zslayton, line 280: 🗺️ This bitshifting loop now lives in `DecodedUInt::small_uint_from_slice` so it can be called from multiple locations. I'll point out the other caller location below.

src/binary/non_blocking/raw_binary_reader.rs
```diff
@@ -475,18 +476,13 @@ impl<A: AsRef<[u8]>> RawBinaryBufferReader<A> {
...
-        match buffer.read_uint(encoded_value.value_length())?.value() {
-            UInteger::U64(symbol_id) => {
-                // This will always succeed on 64-bit platforms where u64 and usize are the same.
-                if let Ok(sid) = usize::try_from(*symbol_id) {
-                    Ok(sid)
-                } else {
-                    decoding_error("found a u64 symbol ID that was too large to fit in a usize")
-                }
-            }
-            UInteger::BigUInt(symbol_id) => Self::try_symbol_id_from_big_uint(symbol_id),
```
zslayton, line 488: 🗺️ This code path (`try_symbol_id_from_big_uint`) would only succeed if the `BigUInt` was small enough that it could have been encoded as a `usize` to begin with. This code path existed solely to satisfy one of the `ion-tests`:
```
ion-tests/iontestdata/good/typecodes/T7-large.10n
```
Text view of its contents:
```
---------------------------------------------------------------------------
 Offset   |  Length   |        Binary Ion        |         Text Ion
---------------------------------------------------------------------------
          |         4 | e0 01 00 ea              |  // Ion 1.0 Version Marker
        4 |         6 | 75 00 00 00 00 00        |   '$0' // $0
       10 |         7 | 76 00 00 00 00 00 00     |   '$0' // $0
       17 |         8 | 77 00 00 00 00 00 00 00  |   '$0' // $0
       25 |         9 | 78 00 00 00 00 00 00 00  |   '$0' // $0
          |           | 00                       |
       34 |        10 | 79 00 00 00 00 00 00 00  |   '$0' // $0
          |           | 00 00                    |
       44 |        11 | 7a 00 00 00 00 00 00 00  |   '$0' // $0
          |           | 00 00 00                 |
       55 |        12 | 7b 00 00 00 00 00 00 00  |   '$0' // $0
          |           | 00 00 00 00              |
       67 |        13 | 7c 00 00 00 00 00 00 00  |   '$0' // $0
          |           | 00 00 00 00 00           |
       80 |        14 | 7d 00 00 00 00 00 00 00  |   '$0' // $0
          |           | 00 00 00 00 00 00        |
       94 |        16 | 7e 8e 00 00 00 00 00 00  |   '$0' // $0
          |           | 00 00 00 00 00 00 00 00  |
```

I have removed that code path altogether and added the test to the skip list. I don't believe this represents a substantive loss of functionality.

src/binary/non_blocking/raw_binary_reader.rs
```diff
@@ -475,18 +476,13 @@ impl<A: AsRef<[u8]>> RawBinaryBufferReader<A> {
...
+        let (_encoded_value, bytes) = self.value_and_bytes(IonType::Symbol)?;
```
zslayton, line 479: 🗺️ We no longer construct a `BinaryBuffer`; `bytes` contains the value byte slice that we're interested in.

src/binary/non_blocking/raw_binary_reader.rs
```diff
@@ -475,18 +476,13 @@ impl<A: AsRef<[u8]>> RawBinaryBufferReader<A> {
...
+        if bytes.len() > mem::size_of::<usize>() {
+            return decoding_error("found a symbol Id that was too large to fit in a usize");
         }
+        let magnitude = DecodedUInt::small_uint_from_slice(bytes);
```
zslayton, line 483: 🗺️ We can read from `bytes` directly with no additional boundary or error checking using the helper method we previously created: `DecodedUInt::small_uint_from_slice`.

src/binary/non_blocking/raw_binary_reader.rs
```diff
@@ -628,6 +624,12 @@ impl<A: AsRef<[u8]>> StreamReader for RawBinaryBufferReader<A> {
         Box::new(self.annotations_iter())
     }
 
+    fn has_annotations(&self) -> bool {
```
zslayton, line 627: 🗺️ This method is incidental to this changeset, but is also an important optimization. Readers checking to see if the current value has an annotation can simply see if the encoding had an annotations sequence; this is much faster than the default implementation that constructs a boxed iterator and reports whether it is empty.

src/binary/non_blocking/raw_binary_reader.rs
```diff
@@ -1195,7 +1200,7 @@ impl<'a, A: AsRef<[u8]>> TxReader<'a, A> {
         // Read the length of the annotations sequence
         let annotations_length = self.tx_buffer.read_var_uint()?;
 
-        // Validate that neither the annotations sequence is not empty.
+        // Validate that the annotations sequence is not empty.
```
zslayton, line 1203: 🗺️ Fixing a typo.

src/binary/non_blocking/raw_binary_reader.rs
```diff
@@ -1204,7 +1209,7 @@ impl<'a, A: AsRef<[u8]>> TxReader<'a, A> {
         let expected_value_length = annotations_and_value_length
             - annotations_length.size_in_bytes()
             - annotations_length.value();
-        self.tx_buffer.total_consumed();
```
zslayton, line 1207: 🗺️ Removing a stray accessor function that had no impact on the method logic (and was optimized out of the assembly anyway).

src/binary/non_blocking/raw_binary_reader.rs
```diff
@@ -1749,7 +1754,7 @@ mod tests {
     fn debug() -> IonResult<()> {
         let data = &[
             0xE0, 0x01, 0x00, 0xEA, // IVM
-            0xc3, 0xd2, 0x84, 0x11, // {'name': true}
+            0xc3, 0xd2, 0x84, 0x11, // ({'name': true})
```
zslayton, line 1757: 🗺️ Correcting the text Ion in the comment for this unit test.

src/binary/uint.rs
```diff
@@ -28,6 +28,24 @@ impl DecodedUInt {
...
+    /// overflow may quietly occur.
+    pub(crate) fn small_uint_from_slice(uint_bytes: &[u8]) -> u64 {
+        let mut magnitude: u64 = 0;
+        for &byte in uint_bytes {
+            let byte = u64::from(byte);
+            magnitude <<= 8;
+            magnitude |= byte;
+        }
+        magnitude
+    }
```
zslayton, line 42: 🗺️ Our new helper methods that operate directly on the provided byte slice.

nirosys (approved):

PR #402, opened 2022-08-04
Serde feature and support for ion-rs

jmt-lab:
*Issue #, if available:*

*Description of changes:*
Implemented serde deserializer and serializer for ion-rs. This serializer can handle basic serialization of types from ion -> rust, rust -> ion. It even includes support for ion Timestamp objects. In addition you can use serde_with to have chrono datetimes work. The Timestamp solution was inspired by how toml-rs handles a similar problem. There is a single test for now provided that passes.

**Note:** The timestamp support currently only supports timestamps with a proper offset. This is due to how it is implemented as passthrough.

### DIFF
--- a/src/serde/datetime.rs
+++ b/src/serde/datetime.rs
@@ -1,0 +1,162 @@
+use crate::Timestamp;
+use chrono::{DateTime, FixedOffset, Utc};
+use serde::{self, de, ser, ser::SerializeStruct, Deserialize, Serialize};
+use serde_with::{DeserializeAs, SerializeAs};
+use std::fmt;
+use std::fmt::Write;
+
+impl Serialize for Timestamp {
+    #[inline]
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::ser::Serializer,
+    {
+        let datetime: DateTime<FixedOffset> = self.clone().try_into().unwrap();
+        let mut state = serializer.serialize_struct("$datetime", 1)?;
+        state.serialize_field("$datetime", datetime.to_rfc3339().as_str())?;
+        state.end()
+    }
+}
+
+impl<'de> Deserialize<'de> for Timestamp {
+    fn deserialize<D>(deserializer: D) -> Result<Timestamp, D::Error>
+    where
+        D: de::Deserializer<'de>,
+    {
+        struct TimestampVisitor;
+
+        impl<'de> de::Visitor<'de> for TimestampVisitor {
+            type Value = Timestamp;
+
+            fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+                formatter.write_str("a ION Timestamp")
+            }
+
+            fn visit_map<V>(self, mut visitor: V) -> Result<Timestamp, V::Error>
+            where
+                V: de::MapAccess<'de>,
+            {
+                let value = visitor.next_key::<TimestampKey>()?;
+                if value.is_none() {
+                    return Err(de::Error::custom("timestamp key not found"));
+                }
+                let v: TimestampFromString = visitor.next_value()?;
+                Ok(v.value)
+            }
+        }
+
+        static FIELDS: [&str; 1] = ["$datetime"];
+        deserializer.deserialize_struct("$datetime", &FIELDS, TimestampVisitor)
+    }
+}
+
+struct TimestampKey;
+
+impl<'de> de::Deserialize<'de> for TimestampKey {
+    fn deserialize<D>(deserializer: D) -> Result<TimestampKey, D::Error>
+    where
+        D: de::Deserializer<'de>,
+    {
+        struct FieldVisitor;
+
+        impl<'de> de::Visitor<'de> for FieldVisitor {
+            type Value = ();
+
+            fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+                formatter.write_str("a valid datetime field")
+            }
+
+            fn visit_str<E>(self, s: &str) -> Result<(), E>
+            where
+                E: de::Error,
+            {
+                if s == "$datetime" {
+                    Ok(())
+                } else {
+                    Err(de::Error::custom("expected field with custom name"))
+                }
+            }
+        }
+
+        deserializer.deserialize_identifier(FieldVisitor)?;
+        Ok(TimestampKey)
+    }
+}
+
+pub struct TimestampFromString {
+    pub value: Timestamp,
+}
+
+impl<'de> de::Deserialize<'de> for TimestampFromString {
+    fn deserialize<D>(deserializer: D) -> Result<TimestampFromString, D::Error>
+    where
+        D: de::Deserializer<'de>,
+    {
+        struct Visitor;
+
+        impl<'de> de::Visitor<'de> for Visitor {
+            type Value = TimestampFromString;
+
+            fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+                formatter.write_str("string containing a datetime")
+            }
+
+            fn visit_str<E>(self, s: &str) -> Result<TimestampFromString, E>
+            where
+                E: de::Error,
+            {
+                let naive: DateTime<FixedOffset> = DateTime::parse_from_rfc3339(s).unwrap();
+                Ok(TimestampFromString {
+                    value: Timestamp::from(naive),
+                })
+            }
+        }
+
+        deserializer.deserialize_str(Visitor)
+    }
+}
+
+impl SerializeAs<DateTime<Utc>> for Timestamp {
+    fn serialize_as<S>(source: &chrono::DateTime<Utc>, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: ser::Serializer,
+    {
+        let dt0: DateTime<FixedOffset> = source.clone().into();
+        let dt1 = Timestamp::from(dt0);
+        dt1.serialize(serializer)
+    }
+}
+
+impl SerializeAs<DateTime<FixedOffset>> for Timestamp {
+    fn serialize_as<S>(
+        source: &chrono::DateTime<FixedOffset>,
+        serializer: S,
+    ) -> Result<S::Ok, S::Error>
+    where
+        S: ser::Serializer,
+    {
+        let dt = Timestamp::from(source.clone());
+        dt.serialize(serializer)
+    }
+}
+
+impl<'de> DeserializeAs<'de, DateTime<Utc>> for Timestamp {
+    fn deserialize_as<D>(deserializer: D) -> Result<DateTime<Utc>, D::Error>
+    where
+        D: de::Deserializer<'de>,
+    {
+        let d0 = Timestamp::deserialize(deserializer)?;
+        let d1: DateTime<FixedOffset> = d0.try_into().unwrap();
+        Ok(d1.into())
+    }
+}
+
+impl<'de> DeserializeAs<'de, DateTime<FixedOffset>> for Timestamp {
+    fn deserialize_as<D>(deserializer: D) -> Result<DateTime<FixedOffset>, D::Error>
+    where
+        D: de::Deserializer<'de>,
+    {
+        let d0 = Timestamp::deserialize(deserializer)?;
+        Ok(d0.try_into().unwrap())
+    }
+}
--- a/src/serde/de.rs
+++ b/src/serde/de.rs
@@ -1,7 +1,9 @@
 use crate::{IonError, IonResult, IonType, ReaderBuilder, StreamItem, StreamReader, Symbol};
+use chrono::{DateTime, FixedOffset};
 use serde::de;
 use serde::de::{DeserializeSeed, EnumAccess, MapAccess, SeqAccess, Visitor};
 use serde::Deserialize;
+use std::borrow::Cow;
 
 pub fn from_slice<'a, T>(s: &'a [u8]) -> IonResult<T>
 where
@@ -45,7 +47,7 @@ where
 }
 
 pub struct IonDeserializer<R> {
-    reader: R,
+    pub(crate) reader: R,
 }
 
 impl<'de, 'a, R> serde::de::Deserializer<'de> for &'a mut IonDeserializer<R>
@@ -68,6 +70,10 @@ where
                 IonType::Boolean => self.deserialize_bool(visitor),
                 IonType::List => self.deserialize_seq(visitor),
                 IonType::Struct => self.deserialize_struct("", &[], visitor),
+                IonType::Timestamp => visitor.visit_map(TimestampDeserializer::<'a> {
+                    visited: false,
+                    deserializer: self,
+                }),
                 _ => Err(IonError::DecodingError {
                     description: "unexpected ion type".to_string(),
                 }),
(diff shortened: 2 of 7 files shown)


src/serde/de.rs
```diff
@@ -0,0 +1,955 @@
...
+    let t = T::deserialize(&mut deserializer)?;
+
+    if deserializer.reader.ion_type().is_none() {
+        Ok(t)
+    } else {
+        Err(IonError::DecodingError {
+            description: "expected eof".to_string(),
+        })
+    }
+}
```
jnicholls, line 47: Couple of points here.

1. Perhaps these implementations can be shared by a generic implementation that accepts anything implementing `ToIonDataSource`, and each marked with a `#[inline]` hint.
2. Does it really make sense to enforce that `T` deserializes the entirety of the Ion item stream?
4. Related to number 2, an `StreamDeserializer` that implements `Iterator` would be a good addition too, just an idea. A la [`serde_json::de::StreamDeserializer`](https://docs.rs/serde_json/latest/serde_json/de/struct.StreamDeserializer.html).

src/serde/de.rs
```diff
@@ -0,0 +1,955 @@
...
+
+pub struct IonDeserializer<R> {
```
jnicholls, line 49: Thinking this should just be named `Deserializer`. This is consistent to naming patterns in the serde ecosystem, e.g. [`serde_json::de::Deserializer`](https://docs.rs/serde_json/latest/serde_json/de/struct.Deserializer.html).

jnicholls (commented):
Nice work! Left a few comments.

src/serde/ser.rs
```diff
@@ -0,0 +1,804 @@
...
+        _variant_index: u32,
+        variant: &'static str,
+    ) -> Result<Self::Ok, Self::Error> {
+        let datetime =
+            DateTime::parse_from_rfc3339(variant).map_err(|e| IonError::EncodingError {
+                description: e.to_string(),
+            })?;
+        let timestamp = Timestamp::from(datetime);
+        self.serializer.writer.write_timestamp(&timestamp)
+    }
```
jnicholls, line 609: In what case would this be used, out of curiosity? Same goes for `serialize_newtype_struct` and `serialize_some`. Since we control the implementation of `Serialize` for `Timestamp` and we ensure we're passing a `&str` to `serialize_field`, these implementations seems unnecessary as `serialize_str` will certainly always be [the invoked method by `&str`'s implementation of `Serialize`](https://docs.rs/serde/latest/src/serde/ser/impls.rs.html#43-51), no?

src/serde/de.rs
jmt-lab: I'll see what i can do to combine them, i'll need to add a flag though for the different number of skips required from binary to string representation (from my testing it looks like i have to do two next calls to get to first object for binary, but only 1 for string). But this brings the problem 2 you mentioned. I'd need to somehow detect if i'm at the start of stream or not

src/serde/de.rs
jmt-lab: Ack renaming

src/serde/ser.rs
jmt-lab: Most likely true, I was just copying what toml-rs did, they had these implemented as well intriguingly enough. Not exactly sure why.

src/serde/de.rs
jmt-lab: Oops ignore the twice thing that was a brain miss

jnicholls: Ok 😌 I was going to say, this would mean [`NativeElementIterator` has a bug](https://docs.rs/ion-rs/latest/src/ion_rs/value/native_reader.rs.html#28). Good to hear it's not an issue.

src/serde/ser.rs
jnicholls: Ok roger that. Doesn't affect the implementation of course, but just for someone like me reading the code, I'd be yearning for a comment to understand in what case it was necessary. I think it's best to return an error, since these methods are not expected to be invoked.

src/serde/de.rs
jmt-lab: Done rename

src/serde/de.rs
jmt-lab: Adjusted the methods and also implemented a quick and easy StreamIterator

src/serde/de.rs
```diff
@@ -0,0 +1,997 @@
...
+    T: de::Deserialize<'de>,
+{
+    type Item = IonResult<T>;
+
+    fn next(&mut self) -> Option<IonResult<T>> {
+        // Skip any nothing
+        while let StreamItem::Nothing = self.de.reader.current() {
+            match self.de.reader.next() {
+                Ok(_) => (),
+                Err(_) => return None,
```
jnicholls, line 984: I wonder, do the implementations of `StreamReader::next()` actually return an error when at the end of file? Or do they just return `Nothing`? I'm just wondering if this will ever terminate, or do we get stuck in an infinite loop of `Nothing`...

jmt-lab: The next should error out on end of file (hopefully)...

jmt-lab: Cause there is no is_end() indicator or something else so i assumed the behavior (which could be a downfall)

jnicholls:
Maybe it's not for this PR, but I do wonder about Ion annotations and how to handle them in the future. We may need to create an `Annotations` type that people can put into their own data structures, that would capture all the annotations for an Ion element when deserialized, and likewise serialize to a sequence of symbols/strings. Thoughts? /cc @zslayton

src/serde/de.rs
```diff
@@ -0,0 +1,997 @@
...
+                IonType::Struct => {
+                    self.reader.step_in()?;
+                    self.reader.next()?;
+                    let value = visitor.visit_enum(VariantAccess::new(self))?;
+                    self.reader.step_out()?;
+                    self.reader.next()?;
+                    Ok(value)
+                }
+                _ => Err(IonError::DecodingError {
+                    description: "expeted an enumeration".to_string(),
```
zslayton, line 378:
```suggestion
                    description: "expected an enumeration".to_string(),
```

src/serde/de.rs
```diff
@@ -0,0 +1,997 @@
...
+                description: "unexpected end of file".to_string(),
+            })
+        }
+    }
+
+    fn deserialize_identifier<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        let result = visitor.visit_string(self.reader.read_string()?);
```
zslayton, line 392: Would it make more sense to use Ion's `symbol` type instead of `string` for an identifier?

src/serde/de.rs
```diff
@@ -0,0 +1,997 @@
...
+        self.reader.next()?;
+        result
+    }
+
+    fn deserialize_ignored_any<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        self.deserialize_any(visitor)
```
zslayton, line 401: Is `deserialize_ignored_any` supposed to just step over the current item in the stream? If so, why not just call `visitor.reader.next()` instead of doing serialization logic? (I'm just going off the method name.)

src/serde/de.rs
```diff
@@ -0,0 +1,997 @@
...
+    where
+        K: DeserializeSeed<'de>,
+    {
+        if self.reader.ion_type().is_none() {
+            self.reader.step_out()?;
+            Ok(None)
+        } else {
+            let key = self.reader.field_name().map(|x| x.to_string())?;
+            let deserializer = MapKeyDeserializer { key };
+            Ok(Some(seed.deserialize(deserializer)?))
```
zslayton, line 442: I'm not sure what `MapKeyDeserializer` does, but is copying every struct name into its own string necessary?

src/serde/ser.rs
```diff
@@ -0,0 +1,804 @@
+use chrono::{DateTime, Utc};
+use std::io::Cursor;
+
+use crate::{
+    BinaryWriterBuilder, Integer, IonError, IonResult, IonType, StreamItem, StreamReader, Symbol,
+    TextWriterBuilder, Timestamp, Writer,
```
zslayton, line 6: Please address any nits identified by `cargo clippy` (e.g. unused imports).

src/serde/ser.rs
```diff
@@ -0,0 +1,804 @@
...
+    fn serialize_char(self, v: char) -> Result<Self::Ok, Self::Error> {
+        self.writer.write_string(v.to_string())
+    }
+
+    fn serialize_str(self, v: &str) -> Result<Self::Ok, Self::Error> {
+        self.writer.write_string(v.to_string())
+    }
+
+    fn serialize_bytes(self, v: &[u8]) -> Result<Self::Ok, Self::Error> {
+        self.writer.write_blob(v.to_owned())
```
zslayton, line 148:
```suggestion
        self.writer.write_blob(v)
```

src/serde/ser.rs
```diff
@@ -0,0 +1,804 @@
...
+    fn serialize_f64(self, v: f64) -> Result<Self::Ok, Self::Error> {
+        self.writer.write_f64(v)
+    }
+
+    fn serialize_char(self, v: char) -> Result<Self::Ok, Self::Error> {
+        self.writer.write_string(v.to_string())
+    }
+
+    fn serialize_str(self, v: &str) -> Result<Self::Ok, Self::Error> {
+        self.writer.write_string(v.to_string())
```
zslayton, line 144:
```suggestion
        self.writer.write_string(v)
```

src/serde/ser.rs
```diff
@@ -0,0 +1,804 @@
...
+
+    fn serialize_unit(self) -> Result<Self::Ok, Self::Error> {
+        self.serialize_none()
+    }
+
+    fn serialize_unit_struct(self, _name: &'static str) -> Result<Self::Ok, Self::Error> {
+        self.serialize_unit()
+    }
+
+    fn serialize_unit_variant(
```
zslayton, line 170: Both here and in the deserializer code: it would be _really_ helpful to have a comment on each function that shows an example of the Rust data type and the corresponding Ion representation being produced. The serde model has a not of names for things that aren't very familiar to me. (For example: "unit variant" or "newtype variant".)

src/serde/ser.rs
```diff
@@ -0,0 +1,804 @@
...
+    where
+        T: Serialize,
+    {
+        self.writer.step_in(IonType::Struct)?;
+        self.writer.set_field_name(variant);
+        value.serialize(&mut *self)?;
+        self.writer.step_out()
+    }
+
+    fn serialize_seq(self, _len: Option<usize>) -> Result<Self::SerializeSeq, Self::Error> {
```
zslayton, line 206: Is there a way (Rust annotations?) that we could signal to the serializer that we want an s-expression instead of a list for a particular data type?

src/serde/ser.rs
```diff
@@ -0,0 +1,804 @@
...
+
+struct MapKeySerializer {}
+
+fn key_must_be_a_string() -> IonError {
+    IonError::EncodingError {
+        description: "ion does not support non-string keys for maps".to_string(),
+    }
+}
+
+impl<'a> ser::Serializer for MapKeySerializer {
```
zslayton, line 404: This is a clever solution! However, I wonder if applications that rely on honest-to-goodness map types would prefer to retain the key's data types.

Some options:
1. Simply err if the data cannot be represented sanely using an Ion `struct`. (that is: ONLY text keys)
2. Convert all keys to strings on a best-effort basis to produce a `struct`. (This is lossy.)
3. Serialize maps as lists of key/value pairs instead of `struct`s.

For the moment I think I'm inclined to go with the first option until we can decide on a final strategy. If we choose either 2 or 3, we'll be inventing a serialization format that's not-quite-Ion. Other Ion implementations won't recognize them the way that this serde implementation would.

src/serde/ser.rs
```diff
@@ -0,0 +1,804 @@
...
+        _variant_index: u32,
+        _variant: &'static str,
+        _len: usize,
+    ) -> Result<Self::SerializeStructVariant, Self::Error> {
+        Err(key_must_be_a_string())
+    }
+}
+
+pub enum TableSerializer<'a, E> {
+    Table(&'a mut Serializer<E>),
```
zslayton, line 574: What is a `Table` in this context?

zslayton (changes requested):
Overall this looks pretty good! 

We need to be very careful about producing output that cannot be understood the same way by other Ion libraries.

I'd appreciate it if you could go through and add comments. I think I followed the conversions happening in most places, but it'd be nice not to have to think as hard. 😄

src/serde/datetime.rs
```diff
@@ -0,0 +1,162 @@
...
+impl Serialize for Timestamp {
+    #[inline]
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::ser::Serializer,
+    {
+        let datetime: DateTime<FixedOffset> = self.clone().try_into().unwrap();
+        let mut state = serializer.serialize_struct("$datetime", 1)?;
+        state.serialize_field("$datetime", datetime.to_rfc3339().as_str())?;
+        state.end()
```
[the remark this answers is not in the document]
jmt-lab: THe above struct is not actually written to the ion file at all. Until rust implements specialization for traits we have to be clever (learned this from toml-rs crate). We serialize it into a special struct then Ion's deserializer detects this structure and writes an ion timestamp to the ion file. Now does this break multi format serialization (ie serializing same object with timestamp to json) yes unfortunately, but the problem is the same with toml-rs. 

We could potentially hide this behind another feature for which i can add a comment on specifying that using it means you can only use ion for Timestamp.

src/serde/ser.rs
jmt-lab: This is part of the datetime/timestamp serialization trick

src/serde/de.rs
zslayton: `next()` returns `Nothing` if there are no more items at the current level of depth.

zslayton:
The latest commit appears to add comments and a `thread_local` for improved interoperability with non-Ion implementations of `serde`, but I believe many of my comments from the original review still need to be discussed/addressed. In particular, I'm worried about `unwrap()`ping instead of surfacing an error and deciding on the right behavior for structs. Converting Ion values to strings feels hazardous to me.

jnicholls:
I can address the comments. I know @jmt-lab mentioned a limited amount of time to invest. Happy to jump in and take it to completion.

jmt-lab:
Thanks jnicholls, I appreciate the help

jmt-lab:
@zslayton and @jnicholls I was able to fit in removal of the unwraps with proper error handling

zslayton:
Hi @jnicholls, are you still interested in driving this PR?

jnicholls:
Sorry for the ghosting :( Yes, I'll drive this PR starting next week. Thanks for the poke.

progwriter:
Hi folks, any progress here? This is a _really_ useful feature, I'd love to have this merged in/released, as I'm sure many other people would as well.

fnordpig:
<poke>

zslayton:
Hi all, I'm going to close this PR out for the moment as @desaikd is working on a new one that rebases it on the latest code in `main`. Thanks very much to everyone that contributed to this!

PR #406, opened 2022-08-10
Add a ToIonDataSource implementation for std::io::Cursor and references to byte arrays.

jnicholls:
## What's changing?

* Adding an implementation of `ToIonDataSource` for `std::io::Cursor` for completeness.
* Adding an implementation of `ToIonDataSource` for `&[u8; N]` as an additional acceptable type for any method accepting a `ToIonDataSource`, such as the readers.
* Update the `IonDataSource` trait to have default implementations, since it requires BufRead.
* Update the `ToIonDataSource` to put a trait bound on the `DataSource` associated type of `IonDataSource` instead of `BufRead` which is more specific and correct.

### DIFF
--- a/src/data_source.rs
+++ b/src/data_source.rs
@@ -23,38 +23,6 @@ use crate::result::{decoding_error, IonError, IonResult};
 /// byte array.
 pub trait IonDataSource: BufRead {
     /// Ignore the next `number_of_bytes` bytes in the data source.
-    fn skip_bytes(&mut self, number_of_bytes: usize) -> IonResult<()>;
-
-    /// Returns the next byte in the data source if available.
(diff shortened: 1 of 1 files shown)


zslayton:
The change looks good, but the build is currently failing. Could you take a look when you have a chance?

jnicholls:
Yeah, I forgot to add some files and test updates, sorry. I have a few branches going. I'll get back to this in a little bit. Thanks.

jnicholls:
Seems like the builds are stuck pending... 🤷‍♂️

zslayton (approved):

zslayton:
Thanks for the PR!

PR #407, opened 2022-08-10
Encoding oversized binary decimals no longer fails quietly

### LINKED ISSUE #404: Writing large decimal values in binary can fail quietly
The binary Ion writer uses a very modest buffer size for encoding decimal values:

https://github.com/amzn/ion-rust/blob/1e3331162f3dc32e0b7cd3dc3026da031b41b446/src/binary/decimal.rs#L21

There is currently no mechanism to fallback to the heap if that is insufficient, so an `Err` is raised if the encoded decimal would be larger than `16` bytes long. Unfortunately, there is a bug in the `raw_binary_writer`:

https://github.com/amzn/ion-rust/blob/1e3331162f3dc32e0b7cd3dc3026da031b41b446/src/binary/raw_binary_writer.rs#L660

Notice that there is no `?` at the end of the line. An earlier run of `cargo clippy --fix` seems to have addressed the unused `Result` by prepending a `let _ = ...`, causing any `Err` to be silently ignored.

Together, these problems cause streams containing oversized decimals to be corrupted. The writer fails to encode the decimal and adds no bytes to the output, but continues on without raising an error.

As a short term fix, we should:
1. Bump the size of the stack-allocated decimal encoding buffer to 32.
2. Add a `?` to the line shown above so errors are surfaced to the application appropriately.
3. Map the not-enough-space error to a `DecodingError` with a human-friendly error message (e.g. `"encountered a decimal that was larger than the configured maximum"`)

Longer term, we should weigh whether we want to support even larger decimals. If so, we'll need a mechanism to fall back to the heap.

zslayton:
The binary decimal encoder uses a fixed-size, stack-allocated buffer
to encode decimal values. Prior to this commit, encoding a decimal
whose encoded representation was too large for the buffer would
fail silently, emitting no bytes but also not raising an error.
This had the potential to corrupt the output.

1. Increases the size of the decimal encoding buffer from 16 bytes
   to 32.
2. Maps the IOError representing the buffer overrun to an
   EncodingError.
3. Surfaces any errors that occur to the application instead of
   quietly dropping them.

Fixes #404.

### DIFF
--- a/src/binary/decimal.rs
+++ b/src/binary/decimal.rs
@@ -6,19 +6,15 @@ use arrayvec::ArrayVec;
 use bigdecimal::Zero;
 
 use crate::ion_eq::IonEq;
-use crate::{
-    binary::{
-        int::DecodedInt, raw_binary_writer::MAX_INLINE_LENGTH, var_int::VarInt, var_uint::VarUInt,
-    },
-    result::IonResult,
-    types::{
-        coefficient::{Coefficient, Sign},
-        decimal::Decimal,
-        magnitude::Magnitude,
-    },
-};
+use crate::{binary::{
+    int::DecodedInt, raw_binary_writer::MAX_INLINE_LENGTH, var_int::VarInt, var_uint::VarUInt,
+}, IonError, result::IonResult, types::{
+    coefficient::{Coefficient, Sign},
+    decimal::Decimal,
+    magnitude::Magnitude,
+}};
 
-const DECIMAL_BUFFER_SIZE: usize = 16;
+const DECIMAL_BUFFER_SIZE: usize = 32;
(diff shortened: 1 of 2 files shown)


src/binary/decimal.rs
```diff
@@ -108,24 +109,38 @@ where
 
     fn encode_decimal_value(&mut self, decimal: &Decimal) -> IonResult<usize> {
         let mut bytes_written: usize = 0;
-        // First encode the decimal. We need to know the encoded length before
-        // we can compute and write out the type descriptor.
+        // First encode the decimal value to a stack-allocated buffer.
```
popematt, line 112: Just curious—is it possible (and useful) to calculate the size prior to encoding it?

desaikd (approved):

Issue #293, opened 2021-06-17
Chrono timestamps always have precision


marcbowes:
```
2000-01-01T00:00:00+00:00 -> [69, 80, 0f, d0, 81, 81, 80, 80, 80, c9]
```

So the first type is 68 not 69 (because my output has +1 byte) and the last (extra) byte is c9. I think that's because the above example shows up with 0d-9 nanoseconds. So... looks like 2 bugs! If you don't mind, I'll open (and own) and issue for the second one (treat is as orthogonal to this PR).

_Originally posted by @marcbowes in https://github.com/amzn/ion-rust/pull/289#discussion_r653173492_

zslayton:
Support for `chrono` timestamps has been removed.

PR #409, opened 2022-08-19
Use parse_timestamp instead of chrono to parse Timestamp

PlasmaIntec:
*Issue #, if available:*
#293 

*Description of changes:*
Use parse_timestamp instead of chrono to parse Timestamp. The expected bytes written were checked against using `ion dump`.

### DIFF
--- a/src/binary/timestamp.rs
+++ b/src/binary/timestamp.rs
@@ -162,7 +162,9 @@ mod binary_timestamp_tests {
                 assert_eq!(buf.len(), expected);
                 assert_eq!(written, expected);
             }
-            _ => panic!("reader.next() should only return reader::StreamItem::Value(IonType::Timestamp)"),
+            _ => panic!(
+                "reader.next() should only return reader::StreamItem::Value(IonType::Timestamp)"
+            ),
         }
         Ok(())
     }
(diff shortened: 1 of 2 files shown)


jobarr-amzn (approved):
lgtm :)

jobarr-amzn (changes requested):
I've learned that there are reasons to keep the module private, we shouldn't merge this as-is. More details will follow eventually.

src/binary/timestamp.rs
```diff
@@ -134,28 +134,32 @@ where
...
         #[case] input: &str,
         #[case] expected: usize,
     ) -> IonResult<()> {
-        let chrono = DateTime::parse_from_rfc3339(input).unwrap();
-        let timestamp = chrono.into();
-        let mut buf = vec![];
-        let written = buf.encode_timestamp_value(&timestamp)?;
-        assert_eq!(buf.len(), expected);
-        assert_eq!(written, expected);
+        let text_value = parse_timestamp(&(input.to_owned() + " ")).unwrap().1;
```
zslayton, line 153: Rather than changing the visibility of the text parser functions, I'd rather we used a text reader here.

```rust
let mut reader = ReaderBuilder::new().build(input);
match reader.next()? {
  IonType::Timestamp => // do test
  _ => // fail
}
```

zslayton (changes requested):
Thanks for the fix, and sorry for the delayed review! I've been on vacation.

jobarr-amzn (approved):

PR #410, opened 2022-08-30
adds changes for `Ord` implementation of `Timestamp`

desaikd:
*Description of changes:*
This PR works on changes for `Ord` implementation of `Timestamp` to consider both datetime and fractional seconds for comparison.

* considers datetime comparison result as well as fractional seconds
comparison result for `Ord`
* changes `fractional_seconds_equal()` to convert `Option<&Mantissa>` to `&Mantissa`, where default value `Mantissa::Digits(0)` is used for `None`. This guarantees a `fractional_seconds`  value to be present at all times and makes it easier to compare those fractional seconds.

*Test:*
* adds test case for fractional second and year precision timestamps
ordering

### DIFF
--- a/src/types/timestamp.rs
+++ b/src/types/timestamp.rs
@@ -304,7 +304,7 @@ impl Timestamp {
         }
     }
 
-    /// Tests the fractional seconds fields of two timestamps for Ion equality. This function will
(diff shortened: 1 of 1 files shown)


zslayton (approved):
🚀

PR #415, opened 2022-09-09
Add a 'newline' text writer format

jobarr-amzn:
*Description of changes:* 
I needed to generate newline-centric Ion data from binary and didn't have an easy way to do it, so I ended up doing something more involved with regular expressions, a small perl script, and reasoning about indentation. This is the capability that I wish I had then.

I've also added a `space_between_top_values` parameter to the `RawTextWriterBuilder` so that top-level values can be handled differently, and extracted a heap-allocated `WhitespaceConfig` struct as suggested in the comment around raw text writer whitespace configuration. 

Please let me know wherever this can be improved, my Rust is as of yet very unoxidized. 

Representatives of each format, from the documentation:

`default`
```ion
{foo: 1, bar: 2, baz: 3} [1, 2, 3] true "hello"
```
---

`newline`
```ion
{foo: 1, bar: 2, baz: 3}
[1, 2, 3]
true
"hello"
```
---

`pretty`
```ion
{
    foo: 1,
    bar: 2,
    baz: 3
}
[
    1,
    2,
    3
]
true
"hello"
```


jobarr-amzn:
I've already patched ion-cli to support `--format newline` as well, and will fire that PR out after this ships. Happy to take suggestions on the `newline` name as well. `oneline`? Something else?

src/text/raw_text_writer.rs
```diff
@@ -203,17 +238,21 @@ impl<W: Write> RawTextWriter<W> {
...
                 // ...then this is the first value inside a container. We'll write the
                 // `space_after_container_start` so it will (e.g.) appear on its own line.
-                write!(&mut self.output, "{}", self.space_after_container_start)?;
+                write!(&mut self.output, "{}", self.whitespace_config.space_after_container_start)?;
             }
         } else {
             // Otherwise, this is not the first value in this container. Emit the container's
             // delimiter (for example: in a list, write a `,`) before we write the value itself.
             self.write_value_delimiter()?;
-            write!(&mut self.output, "{}", self.space_between_values)?;
```
rmarrowstone, line 212: Did you consider hooking into the `write_value_delimiter` fn that's already switching on `ContainerType` or otherwise combining this with that logic?

rmarrowstone:
Per your note, I don't love the 'newline' name either. I don't have a great suggestion, but would suggest something like 'distinct' or '(per|each)(Item|datum)' something something...

jobarr-amzn:
Oof, this didn't show up in `cargo build` locally but I need to address the new `TextKind` in [`ion_c_writer.rs#L17`](https://github.com/amzn/ion-rust/blob/dabd570daac3072a91391ff2a84cfc3ce7de0a02/src/value/ion_c_writer.rs#L17).

I'll have to go to back to the drawing board a bit.

src/text/raw_text_writer.rs
jobarr-amzn: I did look at it, but I didn't see a satisfactory way. If I had to put this in `write_value_delimiter` then I'd still have to switch there on whether or not to add the `whitespace_config.space_between_values`. It also looks to me like `write_value_delimiter` is focused on the syntactic delimiters, not the presentation/whitespace additions. 

I'd rather leave all the "spacing" concerns in one place, I think.

This also makes me wonder whether the builder ought to validate that all the supplied whitespace values are semantically whitespace.

src/text/raw_text_writer.rs
```diff
@@ -132,15 +139,55 @@ struct EncodingLevel {
...
+    // Top-level values are independent of other values in the stream, we may separate differently
+    space_between_top_level_values: &'static str,
+    // Non-top-level values are within a container
+    space_between_nested_values: &'static str,
+    indentation: &'static str,
+    space_after_field_name: &'static str,
+    space_after_container_start: &'static str,
+}
+
+static DEFAULT_WS_CONFIG: WhitespaceConfig = WhitespaceConfig {
```
zslayton, line 152: It's subjective, but I have a preference for spelling variable names out. Could we make these `WS`s into `WHITESPACE`s?

src/text/raw_text_writer.rs
```diff
@@ -29,14 +26,26 @@ impl RawTextWriterBuilder {
...
+    /// top-level values.
+    ///
+    /// For example:
+    /// ```ignore
+    /// {foo: 1, bar: 2, baz: 3}
+    /// [1, 2, 3]
+    /// true
+    /// "hello"
+    /// ```
+    // This doesn't solve the problem of final newlines. Should find a way to solve that some day.
```
zslayton, line 45: Could you open an issue for this? I think it would be the right behavior for the `lines` and `pretty` formats.

If you'd like to keep this comment, I think we should link to the issue you open for it.

src/text/raw_text_writer.rs
```diff
@@ -59,43 +68,44 @@ impl RawTextWriterBuilder {
...
-            // Field names and values are separated by a single space
-            space_after_field_name: String::from(" "),
-            // The first value in a container appears on a line by itself
-            space_after_container_start: String::from("\n"),
+            whitespace_config: WhitespaceConfig { ..PRETTY_WS_CONFIG },
         }
     }
 
-    pub fn with_space_between_values<S: Into<String>>(
+    pub fn with_space_between_top_values(
```
zslayton, line 75: For consistency with other usages, I'd like to use the wordier `with_space_between_top_level_values`. (Note the **\_level\_**)

src/text/raw_text_writer.rs
```diff
@@ -59,43 +68,44 @@ impl RawTextWriterBuilder {
...
     }
 
-    pub fn with_space_after_container_start<S: Into<String>>(
+    pub fn with_space_after_container_start(
         mut self,
-        space_after_container_start: S,
+        space_after_container_start: &'static str,
     ) -> RawTextWriterBuilder {
-        self.space_after_field_name = space_after_container_start.into();
+        self.whitespace_config.space_after_field_name = space_after_container_start;
```
zslayton, line 108: There's some dissonance here that I probably caused but can't remember why. Should `space_after_container_start` and `space_after_field_name` have two separate setters? Or should we unify their names somehow?

src/text/raw_text_writer.rs
```diff
@@ -107,12 +117,9 @@ impl RawTextWriterBuilder {
...
             field_name: None,
             containers: vec![EncodingLevel::default()],
-            // TODO: We should consider putting these in a single struct (`WhitespaceConfig`?)
-            //       and `Box`ing it. Each `String` is 24 bytes, making the writer much bigger.
-            space_between_values: self.space_between_values,
-            indentation: self.indentation,
-            space_after_field_name: self.space_after_field_name,
-            space_after_container_start: self.space_after_container_start,
+            // Should we validate here that all the strings in `whitespace_config` actually are
+            // semantically whitespace?
```
zslayton, line 121: Good question! Let's open an issue for it and link that here as a `TODO`.

src/text/raw_text_writer.rs
```diff
@@ -132,15 +139,55 @@ struct EncodingLevel {
...
+    space_after_field_name: " ",
+    // The first value in a container appears next to the opening delimiter
+    space_after_container_start: "",
+};
+
+static LINES_WS_CONFIG: WhitespaceConfig = WhitespaceConfig {
+    // Each value appears on its own line
+    space_between_top_level_values: "\n",
+    // Otherwise use the compact/default layout from `DEFAULT_WS_CONFIG`
+    ..DEFAULT_WS_CONFIG
```
zslayton, line 169: 💯

src/text/raw_text_writer.rs
```diff
@@ -132,15 +139,55 @@ struct EncodingLevel {
...
 pub struct RawTextWriter<W: Write> {
     output: BufWriter<W>,
     annotations: Vec<String>,
     field_name: Option<String>,
     containers: Vec<EncodingLevel>,
-    space_between_values: String,
-    indentation: String,
-    space_after_field_name: String,
-    space_after_container_start: String,
+    whitespace_config: Box<WhitespaceConfig>,
```
zslayton, line 190: Nice, this drops 96 bytes of stack space down to 8! 🙌

src/text/text_writer.rs
```diff
@@ -26,6 +26,22 @@ impl TextWriterBuilder {
...
+    /// For example:
+    /// ```ignore
+    /// {foo: 1, bar: 2, baz: 3}
+    /// [1, 2, 3]
+    /// true
+    /// "hello"
+    /// ```
+    pub fn lines() -> TextWriterBuilder {
+        TextWriterBuilder {
+            text_kind: TextKind::Newline,
```
zslayton, line 41: Could we change `TextKind::Newline` to `TextKind::Lines` for consistency?

zslayton (commented):
Nice job! Some superficial comments below.

src/text/raw_text_writer.rs
jobarr-amzn: Done!

src/text/raw_text_writer.rs
jobarr-amzn: Done!

src/text/text_writer.rs
jobarr-amzn: Done!

src/text/raw_text_writer.rs
jobarr-amzn: Done: https://github.com/amzn/ion-rust/issues/437

src/text/raw_text_writer.rs
jobarr-amzn: Done: https://github.com/amzn/ion-rust/issues/438

src/text/raw_text_writer.rs
jobarr-amzn: I think this is just a bug? Here's the relevant snippet of the compact text format configuration: 

```
    // Single space between field names and values
    space_after_field_name: " ",
    // The first value in a container appears next to the opening delimiter
    space_after_container_start: "",
```

and here's the relevant snippet of the pretty text format configuration:

```
    // Field names and values are separated by a single space
    space_after_field_name: " ",
    // The first value in a container appears on a line by itself
    space_after_container_start: "\n",
```

Both of them have different values for these two fields, and those different values will be used. It's only this setter which conflates them, and no tests exercise this setter.

jobarr-amzn: I added some tests that exercise the builder setters.

zslayton (approved):
Excellent! 🙌 

There's a `clippy` suggestion about using `!is_empty` instead of `!= ""`. Could you address that before merging?

PR #416, opened 2022-09-14
Add non-blocking RawTextReader & TextBuffer

nirosys:
*Issue #, if available:* n/a

*Description of changes:*
This PR aims at providing a non-blocking RawTextReader implementation mirroring the non-blocking binary reader implementation. The PR looks huge, but is mostly just copies of the blocking reader & buffer, with some modifications.

> NOTE: I've split up the commits so that the first commit contains the original reader and buffer implementations, as well as any other changes needed, and the second commit includes only the modifications to the reader and buffer. This should make it a little easier to see the changes made to them without the whole new file thing getting in the way.

The non-blocking versions of `RawTextReader` and `TextBuffer`, based on the original blocking versions, were added to a new module `text::non_blocking`. The changes allow for a `RawTextReader` to be created using any `AsRef<[u8]>`, with the ability to append data if the underlying type is a `Vec<u8>`, matching the non-blocking binary reader.

`TextBuffer` was implemented to validate UTF-8 contents on first line load, and when adding more data through the `read_from`, or `append_bytes` function. The validation allows for trailing incomplete sequences (though this is imperfect when considering unicode modifiers..) and will keep invalid partial sequences separate from the validated sequences in anticipation of data being added later.

The validation of the buffer during creation does differ from the existing TextBuffer implementation which would identify invalid UTF-8 sequences during calls to `load_next_n_lines` as the characters are appended to TextBuffer's `line`.  

Returning lines from `TextBuffer` creates references to `str`, leaving the already validated UTF-8 data in the underlying buffer. This does cause an issue when appending data, which triggers a `restack` and shifts the unconsumed data to the beginning of the buffer.  Any references returned prior to this restack become invalid. More effort can be put here if we need references to be smarter, and either adjust for restack, or identify when they are invalid do to consumption, etc. 

Known Issues:

* `restack` will invalidate any string data returned by the text buffer. This requires `&str`s returned by the text buffer (and thus text reader) to be evaluated/copied before appending more data. Or making sure no `&str`s are obtained until after data is appended. A better abstraction around the strings could be created here if this becomes an issue. 

---


src/text/non_blocking/raw_text_reader.rs
```diff
@@ -335,6 +334,12 @@ impl<T: ToIonDataSource> RawTextReader<T> {
                             // Retry the parser on the extended buffer in the next loop iteration
                             continue;
                         }
+                        Err(TextError::Incomplete { line, column }) => {
+                            // If load_next_line() returns TextError::Incomplete, then that means
+                            // it has incomplete UTF-8 data in the buffer, suggesting the user is
+                            // not done providing the ion data.
+                            return RootParseResult::Incomplete(line, column);
+                        }
```
zslayton, line 343: For the purposes of this blocking reader, should this just `continue`? There's incomplete data at the end of the line (but maybe also new bytes before that) so just try parsing again?

src/text/non_blocking/raw_text_reader.rs
```diff
@@ -378,9 +383,12 @@ impl<T: ToIonDataSource> RawTextReader<T> {
         // EOF are ambiguous.
         const SENTINEL_ION_TEXT: &str = "\n0\n";
         // Make a note of the buffer's length; we're about to modify it.
-        let original_length = self.buffer.remaining_text().len();
+        let mut remaining_text = self.buffer.remaining_text().to_owned();
```
zslayton, line 387: Since this only happens at EOF, can we not `append_bytes()` to the buffer rather than allocating a new `String`?

zslayton (commented):
Nice job! 🚀 Some questions and superficial comments below.

src/text/non_blocking/text_buffer.rs
```diff
@@ -1,64 +1,133 @@
...
+
+use thiserror::Error;
+
+/// Represents a span of bytes that have been validated as UTF-8 strings.
+///
+/// The range of bytes is [.0, .1)
+#[derive(Copy, Clone)]
+struct ValidUtf8Span(usize, usize);
+impl ValidUtf8Span {
+    fn incr_start(&self, bytes: usize) -> ValidUtf8Span {
```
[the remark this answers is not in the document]
nirosys: Absolutely.

src/text/non_blocking/text_buffer.rs
```diff
@@ -1,64 +1,133 @@
...
+    /// Returns a &str view of the bytes within `data` specified by `span`.
+    /// This function assumes prior UTF-8 validation has occurred.
+    fn string_view(&self, span: ValidUtf8Span) -> &str {
+        use std::str::from_utf8_unchecked;
+        unsafe { from_utf8_unchecked(&self.data.as_ref()[span.0..span.1]) }
+    }
+
+    /// Returns the amount of data currently available in the buffer that has not been marked as
+    /// read via the [consume] method. This includes data that has not been split into lines yet.
+    pub fn data_remaining(&self) -> usize {
```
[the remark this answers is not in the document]
nirosys: 👍

src/text/non_blocking/text_buffer.rs
```diff
@@ -69,154 +138,315 @@ impl<R: BufRead> TextBuffer<R> {
...
+            Err(error) => {
+                // Errors are OK here, as long as we do not have a length of erroneous data. OK
+                // errors need to be partial multi-byte characters at the end of the buffer, which
+                // imply that the data isn't complete.
+                let upto = error.valid_up_to();
+                if upto > 0 && error.error_len().is_none() {
+                    upto + valid_start
+                } else {
+                    Err(TextError::utf8_error(upto))?
+                }
```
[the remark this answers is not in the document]
nirosys: Good call.

src/text/non_blocking/raw_text_reader.rs
zslayton: Ah, of course! That makes sense, thanks for explaining. I wonder if we can optimize this another way, but solving that puzzle can wait. :)

src/text/non_blocking/raw_text_reader.rs
zslayton: > My understanding is that all multi-byte sequences should be in some sort of container (a string, clob, ''d symbol, etc), is that right?

Yep!

src/text/non_blocking/text_buffer.rs
```diff
@@ -69,154 +138,315 @@ impl<R: BufRead> TextBuffer<R> {
...
+    }
+
+
     #[test]
     #[should_panic]
-    fn test_consume_err_illegal_utf8_offset() {
-        let mut buffer = text_buffer("😎");
-        buffer.load_next_line().unwrap();
-        // This jumps into the middle of the emoji.
-        // If it were to succeed, the next call to remaining_text() would panic.
```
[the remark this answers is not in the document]
nirosys: I missed this comment, and honestly have no idea how or why I would have removed those comments. 😬  Re-adding them

src/text/non_blocking/text_buffer.rs
```diff
@@ -371,47 +376,53 @@ mod tests {
...
+        // Add only part of our atom character, copying all but the last byte of the modifier.
         input
-            .append_bytes(&atom_modified[..5])
+            .append_bytes(&atom_modified[..=4])
             .expect("unable to add partial UTF-8 character");
-        // This should fail since we have a partial UTF-8 sequence in the first line.
+
+        // This should succeed, and provide access to all of the data up to the start of the
+        // incomplete sequence. Unfortunately, the 'invalid' sequence starts with the modifier, so
+        // the atom symbol is returned.
```
zslayton, line 410: From the end user perspective this should be ok, right? The text buffer may return the partial glyph ('atom' instead of 'purple background atom') but the Ion reader will say that whatever larger value it's processing (string, symbol, clob) is `Incomplete`.

zslayton (approved):

src/text/non_blocking/text_buffer.rs
nirosys: Yes, I believe so. Every string-like needs a container except identifiers, but they are limited to ASCII characters.

Issue #411, opened 2022-09-01
Change type of `Timestamp::fractional_seconds` to be `Mantissa`


desaikd:
Current implementation of `Timestamp` has `fractional_seconds` type as `Option<Mantissa>`. 

https://github.com/amzn/ion-rust/blob/944511591310f7e273074eaa252f6cd37b5728bb/src/types/timestamp.rs#L126-L131

The above could be simplified by making type of `fractional_seconds` as `Mantissa`  when creating the `Timestamp`. We can put a default value for `fractional_seconds` which could either be `Mantissa::Digits(0)` or `Mantissa::Decimal(Decimal::new(0,0))`.

zslayton:
This is an implementation detail; I don't believe it needs to block the 1.0 release. Thoughts, @popematt?

popematt:
Now that I've looked a little closer, I think you're right.

PR #417, opened 2022-09-15
Make type of Timestamp::fractional_seconds Mantissa

PlasmaIntec:
*Issue #, if available:*
#411 

*Description of changes:*

Previously, the type of `Timestamp::fractional_seconds` was `Option<Mantissa>`

After the implementation of `Ord` for `Timestamp`, `Timestamp::fractional_seconds` can be defaulted to `Mantissa::Digits(0)`


src/types/timestamp.rs
```diff
@@ -1200,8 +1162,7 @@ impl TryInto<ion_c_sys::timestamp::IonDateTime> for Timestamp {
             (Precision::Month, _) => TSPrecision::Month,
             (Precision::Day, _) => TSPrecision::Day,
             (Precision::HourAndMinute, _) => TSPrecision::Minute,
-            (Precision::Second, None) => TSPrecision::Second,
```
PlasmaIntec, line 1203: Note that we are losing the ability to convert into a `TSPrecision::Second`.

Should we retain that mapping by doing a `is_zero()` check on the `mantissa`?

PlasmaIntec: Ah looks like we have a test for this exact scenario: `types::timestamp::ionc_tests::convert_from_to_ionc::case_6_second_plus0400`

zslayton:
I just realized that I've allowed this PR to languish. Sorry about that! I'll review it tomorrow.

src/types/timestamp.rs
```diff
@@ -311,15 +291,9 @@ impl Timestamp {
 
         // TODO: make Timestamp::fractional_seconds to be Mantissa when creating a Timestamp to get rid of below conversion
         // convert Option<&Mantissa> to &Mantissa
```
zslayton, line 293: Looks like this `TODO` comment can be removed now.

src/types/timestamp.rs
```diff
@@ -311,15 +291,9 @@ impl Timestamp {
...
+        let m1 = &self.fractional_seconds;
 
-        let m2 = match &other.fractional_seconds {
-            None => &Mantissa::Digits(0),
-            Some(m) => m,
-        };
+        let m2 = &other.fractional_seconds;
 
         // compare fractional seconds
         match (m1, m2) {
```
zslayton, line 299: I don't see `m1` and `m2` getting used outside the `match` line below, so maybe we could just simplify it to:

```rust
match (&self.fractional_seconds, &other.fractional_seconds) {
```
?

src/types/timestamp.rs
```diff
@@ -1136,7 +1098,7 @@ impl From<DateTime<FixedOffset>> for Timestamp {
         // Get a copy of the offset to store separately
         let offset = Some(*fixed_offset_date_time.offset());
         let precision = Precision::Second;
-        let fractional_seconds = Some(Mantissa::Digits(9));
+        let fractional_seconds = Mantissa::Digits(0);
```
zslayton, line 1101: Types from the `chrono` crate (`NaiveDateTime`, `DateTime<FixedOffset>`) always use nanosecond precision, which is why this (and the one above on what is now line 1088) were using `Digits(9)`. Was this an intentional change?

src/types/timestamp.rs
```diff
@@ -1164,12 +1126,12 @@ impl From<ion_c_sys::timestamp::IonDateTime> for Timestamp {
...
         let fractional_seconds = if let TSPrecision::Fractional(mantissa) = ionc_dt.precision() {
-            Some(match mantissa {
+            match mantissa {
                 IonCMantissa::Digits(digits) => Mantissa::Digits(*digits),
                 IonCMantissa::Fraction(fraction) => Mantissa::Arbitrary(fraction.clone().into()),
-            })
+            }
         } else {
-            None
+            Mantissa::Digits(0)
```
zslayton, line 1134: I'm seeing `Mantissa::Digits(0)` in enough places that I think it would be worth implementing [the `Default` trait](https://doc.rust-lang.org/std/default/trait.Default.html) for `Mantissa`. What do you think?

src/types/timestamp.rs
zslayton: The `ion-c` bindings have been removed since the last release, which I think obviates this issue. Could you try merging from `main` when you have a chance?

src/types/timestamp.rs
```diff
@@ -1676,7 +1635,7 @@ mod timestamp_tests {
...
     #[case::timestamp_with_different_offset(Timestamp::with_ymd_hms(2021, 4, 6, 10, 15, 0).build_at_offset(5 * 60).unwrap(), Timestamp::with_ymd_hms(2021, 4, 6, 10, 15, 0).build_at_offset(-5 * 60).unwrap(), Ordering::Less)]
     #[case::timestamp_with_unknown_offset(Timestamp::with_ymd_hms(2021, 4, 6, 10, 15, 0).build_at_unknown_offset().unwrap(), Timestamp::with_ymd_hms(2021, 4, 6, 10, 15, 0).build_at_offset(-5 * 60).unwrap(), Ordering::Less)]
-    #[case::timestamp_with_second_precison_and_year_precision(Timestamp::with_ymd(2001, 1, 1).build().unwrap(), Timestamp::with_ymd_hms(2001, 1, 1, 0, 0, 0).with_fractional_seconds(Decimal::new(00000000000000000000u128, -20)).build_at_unknown_offset().unwrap(), Ordering::Equal)]
+    #[case::timestamp_with_second_precison_and_year_precision(Timestamp::with_ymd(2001, 1, 1).build().unwrap(), Timestamp::with_ymd_hms(2001, 1, 1, 0, 0, 0).with_fractional_seconds(Decimal::new(00000000000000000000u128, -20)).build_at_unknown_offset().unwrap(), Ordering::Less)]
```
zslayton, line 1638: You've encountered one of the least intuitive corners of the Ion spec here. This ordering should not have changed--the two timestamps represent the same point in time, but with different resolution.

When testing for _equivalence_ (that is: whether two Ion timestamps represent the exact same data), we take precision into account. However, when comparing two timestamps for _ordering_, we only look at the point in time to which the `timestamp` refers. If one timestamp is `2021T` (year precision) and another is `2021-01-01T00:00.000-00:00` (millisecond precision), they both refer to the earliest possible moment in the year 2021 and so are considered `Ordering::Equal`.

src/binary/timestamp.rs
```diff
@@ -75,25 +75,23 @@ where
...
-                                Mantissa::Arbitrary(decimal) => {
-                                    bytes_written += self.encode_decimal(decimal)?;
-                                }
-                            };
-                        }
+                        // TODO: Both branches encode directly due to one
+                        // branch owning vs borrowing the decimal
+                        // representation. #286 should provide a fix.
+                        match &timestamp.fractional_seconds {
+                            Mantissa::Digits(precision) => {
```
zslayton, line 82: The now-removed line above:
```rust
if let Some(ref mantissa) = timestamp.fractional_seconds {
```
had a side effect of letting us skip the encoding logic altogether if the fractional seconds value was going to result in no bytes being output. I think we can reproduce that optimization here by doing:

```suggestion
                        match &timestamp.fractional_seconds {
                            Mantissa::Digits(0) => {}, // Do nothing.
                            Mantissa::Digits(precision) => {
```

src/types/timestamp.rs
```diff
@@ -75,7 +75,11 @@ impl Mantissa {
             Ordering::Equal
         } else if d1.coefficient.is_zero() && d2.coefficient.is_zero() {
             // Coefficient zeros' signs don't have to be compared for fractional seconds.
-            d1.exponent.cmp(&d2.exponent)
+            if d1.coefficient.sign() == Sign::Positive {
+                d2.exponent.cmp(&d1.exponent)
+            } else {
+                d1.exponent.cmp(&d2.exponent)
```
zslayton, line 81: Could you explain this change a bit? I'm unsure why the ordering logic would have changed.

src/binary/timestamp.rs
```diff
@@ -75,25 +75,23 @@ where
...
+                                // Consider the following case: `2000-01-01T00:00:00.123Z`.
+                                // That's 123 millis, or 123,000,000 nanos.
+                                // Our mantissa is 0.123, or 123d-3.
+                                let scaled = utc.nanosecond() / 10u32.pow(9 - *precision); // 123,000,000 -> 123
+                                let exponent = (*precision as i64).neg(); // -3
+                                let fractional = Decimal::new(scaled, exponent); // 123d-3
+                                bytes_written += self.encode_decimal(&fractional)?;
+                            }
+                            Mantissa::Arbitrary(decimal) => {
+                                bytes_written += self.encode_decimal(&decimal)?;
```
zslayton, line 92: There are a handful of `cargo clippy` comments that were added by our CI build for little things like unnecessary `&`s. They look like this:
image

zslayton (commented):
Thanks for the PR, this looks great! A few questions/comments below.

src/types/timestamp.rs
PlasmaIntec: I see- thanks for the clarification!

src/types/timestamp.rs
PlasmaIntec: I updated this branch to return `Ordering::Equal`, since `Mantissa::decimals_compare` is only used for testing fractional seconds fields for ordering, and we won't care for the precision once we know that both coefficients are zero

src/types/timestamp.rs
PlasmaIntec: Agreed

src/types/timestamp.rs
PlasmaIntec: Yes- basically, the `test_timestamp_fractional_seconds_scale` unit test would break since the builder would use a `DateTime<FixedOffset>` to create the `Timestamp`, and `From<DateTime<FixedOffset>>` would default `fractional_seconds` to nanosecond precision.

Should we revert this to nanosecond precision and construct `timestamp_with_seconds` with `.with_nanoseconds_and_precision(0, 0)` instead?

zslayton: I think this should remain `Digits(9)` since someone converting from `DateTime<FixedOffset>` to `Timestamp` in other contexts would expect the nanoseconds to be preserved.

In the unit test you mention: 

https://github.com/amzn/ion-rust/blob/fb164687c5dca001b144c4fe36f8edd06c5d4556/src/types/timestamp.rs#L1568-L1569

it should be that `with_ymd_hms` sets the `Timestamp`'s precision to `Second` and its mantissa to `Digits(0)` after constructing the `DateTime<FixedOffset>`.

zslayton:
Hi @PlasmaIntec, just checking whether I can provide any clarification on [my final comment](https://github.com/amzn/ion-rust/pull/417#discussion_r983552982). Please let me know if I can clarify anything.

zslayton:
I'm closing this as it's now pretty far behind `main`. If you're interested in continuing, please re-base it on `main` and I'll reopen the PR. Thanks!

PR #418, opened 2022-09-16
Removes dependency on `ion-c-sys`

zslayton:
This PR removes all code that was previously behind the `ion_c` feature gate. Many of the native implementation APIs were developed with an eye towards maintaining compatability with the `ion-c-sys` implementation. However, this limited our ability to offer idiomatic APIs. Now that `ion-rust`'s native implementations are relatively complete, we can remove the `ion-c-sys` dependency to facilitate refactoring.

`ion-hash` continues to have a dependency on `ion-c-sys` for the time being, but it has been modified to depend on to the most recently published version of `ion-rust` that supported it. A follow-on PR will update `ion-hash` to depend on the native implementations instead.


desaikd (approved):
🚢

rmarrowstone (approved):
Huzzah!

Issue #405, opened 2022-08-08
Cannot read value of a Decimal


trashbyte:
I cannot find any way to get the actual underlying value of a `Decimal` object. The `Decimal` type is opaque and the `coefficient` field is not publicly exposed in any way I can find, aside from being printed by the `Display` implementation, which is hardly an efficient way to access the value. Am I missing something? Is this an intentional design choice?

zslayton:
>  Is this an intentional design choice?

No, the library is simply incomplete. 

At present, both `Decimal` and `Timestamp` are difficult to use as standalone types. They have enough functionality to guarantee fidelity during roundtripping, but not to use them on their own.

See #286 and #273.

rmarrowstone:
Another relevant issue is the discussion around using BigDecimal: https://github.com/amzn/ion-rust/issues/302

rmarrowstone:
I've been looking at this today and my current thinking is that we should _not_ export and in fact remove all dependencies on `BigDecimal`. We should prefer to keep our dependencies lean and it is not clear that `bigdecimal` is the obvious choice, so I don't think we want to be opinionated about this. I realize that we are also using the `num` crates, but they are much more commonly used: they have more forks and stars and are used in the aws-sdk-rust, for example.

As of https://github.com/amzn/ion-rust/pull/365 our `Decimal` exposes `scale` and `precision`. I propose that we expose the bare minimum to be usable: the magnitude of the decimal and the sign.

Users will need to write their own arithmetic or convert our `Decimal` to their chosen representation, but they will have one less dependency forced on them. We can also later enrich our `Decimal` model further if needed. Additionally, If there is a clear de facto standard crate in time we can provide transforms to/from that.

PR #428, opened 2022-09-22
Add functions to access Sign and Magnitude of Decimal

rmarrowstone:
This change adds access to the Sign and unscaled Magnitude of a Decimal
in its inherent implementation. This addresses GH Issue #405.

I did not expose Coefficient as it was. That seemed to leak too much of
the implementation and not be very user friendly.

I chose to expose the Sign enum as it was: it seems clearest this way
and doesn't expose any implementation details which are likely to change.

I intentionally decoupled the unscaled magnitude from the stored
representation as I think it's more useful this way.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/types/decimal.rs
+++ b/src/types/decimal.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::cmp::Ordering;
 
 use bigdecimal::{BigDecimal, Signed};
@@ -7,7 +8,7 @@ use crate::ion_eq::IonEq;
 use crate::result::{illegal_operation, IonError};
 use crate::types::coefficient::{Coefficient, Sign};
 use crate::types::magnitude::Magnitude;
-use num_traits::Zero;
+use num_traits::{ToPrimitive, Zero};
 use std::convert::{TryFrom, TryInto};
 use std::fmt::{Display, Formatter};
 use std::ops::Neg;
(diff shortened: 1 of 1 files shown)


src/types/decimal.rs
```diff
@@ -47,6 +50,30 @@ impl Decimal {
...
+    pub fn unscaled_as_u64(&self) -> Option<u64> {
+        match self.coefficient.magnitude() {
+            Magnitude::U64(m) => { Some(*m) }
+            Magnitude::BigUInt(m) => { m.to_u64() }
+        }
+    }
+
+    /// Returns the unscaled magnitude of the Decimal value as a BigUInt.
+    /// This may allocate a BigUint if the magnitude is fitable in a u64.
+    pub fn unscaled_as_biguint(&self) -> Cow<BigUint> {
```
rmarrowstone, line 70: Would really appreciate feedback on the choice to return a `Cow` here.

rmarrowstone:
The code coverage appears to be failing due to prior changes, but also seems to show red for added comments?!? I would appreciate some context and/or level-setting about what to expect here. Thanks!

rmarrowstone:
Closing this after getting some feedback. Am going to put out a refactor to collapse Magnitude with UInteger then expose that.

PR #430, opened 2022-09-23
Removes the separate `no_text_values` collection in Struct

zslayton:
* Introduces the `SymbolRef` type, a borrowed view of a text value for field lookups.
* Introduces the `AsSymbolRef` trait to allow values to be temporarily viewed as a `SymbolRef`.
* Modifies the `IonStruct` trait to use `AsSymbolRef` for lookups instead of `AsRef<str>`. This change enables lookups for unknown text.
* Modifies the `StructRef` type to use `AsSymbolRef` for its lookup methods.
* Modifies the `Struct` type to use `AsSymbolRef` for its lookup methods.
* Modifies the `Struct` type to store its values in a `SmallVec`, improving the performance of the common case. The first field value for any given field name will be stored directly in the map. If the field name is repeated, the additional values will be stored elsewhere on the heap. Prior to this change, *all* field values were stored on the heap in a `Vec`.
* Adds `len()` and `is_empty()` methods to the `Struct` type. We will likely wish to move these into the `IonStruct` trait in the future.
* Tweaks the ion-tests test runner to print the Ion text of the elements that caused test failures instead of using their auto-generated `Debug` format.


src/value/borrowed.rs
```diff
@@ -293,24 +294,29 @@ impl<'val> IonStruct for StructRef<'val> {
...
-    fn get<T: AsRef<str>>(&self, field_name: T) -> Option<&Self::Element> {
-        self.text_fields
-            .get(field_name.as_ref())?
-            .last()
-            .map(|(_s, v)| v)
+    fn get<T: AsSymbolRef>(&self, field_name: T) -> Option<&Self::Element> {
+        if let Some(text) = field_name.as_symbol_ref().text() {
+            self.text_fields.get(text)?.last().map(|(_s, v)| v)
+        } else {
+            self.no_text_fields.last().map(|(_name, value)| value)
```
desaikd, line 301: Should we remove `no_text_fields` from here as well?

src/symbol_ref.rs
```diff
@@ -0,0 +1,87 @@
...
+// All text types can be viewed as a `SymbolRef`.
+impl<'a, A: AsRef<str> + 'a> AsSymbolRef for A {
+    fn as_symbol_ref(&self) -> SymbolRef {
+        SymbolRef {
+            text: Some(self.as_ref()),
+        }
+    }
+}
+
+// Owned `Symbol` values can be viewed as a `SymbolRef`. ue to lifetime conflicts in the
```
desaikd, line 42:
```suggestion
// Owned `Symbol` values can be viewed as a `SymbolRef` due to lifetime conflicts in the
```

src/value/borrowed.rs
```diff
@@ -293,24 +294,29 @@ impl<'val> IonStruct for StructRef<'val> {
...
+                    .get(text)
+                    .into_iter()
+                    .flat_map(|v| v.iter())
+                    .map(|(_s, v)| v),
+            )
+        } else {
+            Box::new(self.no_text_fields.iter().map(|(_name, value)| value))
+        }
     }
 }
```
desaikd, line 322: Should we include `len()` and `is_empty()` here as well? Or was this based on current use case of `Element`?

src/value/owned.rs
```diff
@@ -253,51 +287,59 @@ impl IonStruct for Struct {
...
+        self.len() == other.len()
             // we need to test equality in both directions for both text_fields and no_text_fields
             // A good example for this is annotated vs not annotated values in struct
             //  { a:4, a:4 } vs. { a:4, a:a::4 } // returns true
             //  { a:4, a:a::4 } vs. { a:4, a:4 } // returns false
-            && self.eq_text_fields(other) && other.eq_text_fields(self)
-            && self.eq_no_text_fields(other) && other.eq_no_text_fields(self)
+            && self.fields_eq(other) && other.fields_eq(self)
     }
 }
```
desaikd, line 345: We should probably add some tests for `len` and `is_empty` functions.

src/value/owned.rs
```diff
@@ -185,32 +187,67 @@ impl PartialEq for Sequence {
...
 
-/// An owned implementation of [`Struct`]
+/// An owned implementation of [`IonStruct`]
 #[derive(Debug, Clone)]
 pub struct Struct {
-    text_fields: LinkedHashMap<Rc<str>, Vec<(Symbol, Element)>>,
-    no_text_fields: Vec<(Symbol, Element)>,
+    // A mapping of field name to any values associated with that name.
+    // If a field name is repeated, each value will be in the associated SmallVec.
+    // Since repeated field names are not common, we store the values in a SmallVec;
```
desaikd, line 195: 👍🏻

src/value/borrowed.rs
zslayton: I intentionally avoided overhauling `StructRef` in this PR (but then forgot to make that explicit! 😬) I plan to make the same change in a follow-on if we're ok with this.

src/symbol_ref.rs
zslayton, line 42: Good catch! Tweaking the rewrite as that word was intended to start a new sentence.

```suggestion
// Owned `Symbol` values can be viewed as a `SymbolRef`. Due to lifetime conflicts in the
```

src/value/owned.rs
zslayton: Agreed!

src/value/borrowed.rs
desaikd: That works. We can do it in a separate PR.

src/value/owned.rs
desaikd: Added some unit tests for this.

src/value/borrowed.rs
desaikd: As discussed elsewhere in this PR, this will be handled in a separate PR

src/symbol_ref.rs
desaikd: Added a fix for this

desaikd (approved):

PR #431, opened 2022-09-24
Collapse Magnitude with UInteger

rmarrowstone:
This change removes the Magnitude struct and all impl bits, in favor
of collapsing the unique functionality there with the UInteger type.

Both were an enum over u64 and BigUint, with an overlapping set of
inherent and trait impls.


src/types/decimal.rs
```diff
@@ -47,6 +49,12 @@ impl Decimal {
         self.coefficient.number_of_decimal_digits()
     }
 
+    /// Returns the Sign of the Decimal value.
+    /// Unscaled zero values may have either a Positive or Negative Sign.
```
zslayton, line 53: Couple of thoughts here:
* I believe the phrase "unscaled value" means the same thing as "coefficient" -- is that correct? If so, I'd rather stick to "coefficient" for consistency with the rest of the code base.
* A coefficient of zero will _only_ have a `Negative` sign if it's truly `-0`. I think you may be referring to what will happen if we have to return the coefficient as a `u64`, but that won't impact the `Sign` value which is stored and evaluated separately. (Or have I misunderstood something?)

src/types/integer.rs
```diff
@@ -55,9 +56,8 @@ pub trait IntAccess {
     fn as_big_int(&self) -> Option<&BigInt>;
 }
 
-/// Represents a UInt of any size. Used for reading binary integers and symbol IDs.
-///
-/// Equality tests do NOT compare across storage types; U64(1) is not the same as BigUInt(1).
+/// Represents a UInt of any size. Used for reading binary integers and symbol Ids.
+/// Use to represent the unsigned magnitude of Decimal values and fractional seconds.
```
zslayton, line 60:
```suggestion
/// Used to represent the unsigned magnitude of Decimal values and fractional seconds.
```

src/types/integer.rs
```diff
@@ -136,6 +161,90 @@ impl From<UInteger> for Integer {
...
+
+impl_uinteger_from_small_unsigned_int_types!(u8, u16, u32, u64, usize);
+
+macro_rules! impl_uinteger_from_small_signed_int_types {
+    ($($t:ty),*) => ($(
+        impl From<$t> for UInteger {
+            fn from(value: $t) -> UInteger {
+                // Discard the sign and convert the value to a u64.
+                let magnitude: u64 = value.checked_abs()
+                        .and_then(|v| Some(v.abs() as u64))
```
zslayton, line 212: Long ago when I wrote this, Rust didn't have the [`unsigned_abs`](https://doc.rust-lang.org/std/primitive.i64.html#method.unsigned_abs) method to handle this conversion.

rmarrowstone: ack. changed to use that.

zslayton (approved):

zslayton:
image

Sweet!

PR #436, opened 2022-09-30
changes borrowed module to use `SymbolRef` and simplifies `StructRef`

desaikd:
*Description of changes:*
This PR works on changes borrowed module to use SymbolRef and simplifies `StructRef`. 

*List of changes:*
* adds implementation of `Hash`, `From<&str>`, `Borrow<str>` for `SymbolRef`
* makes `SymbolRef` publicly visible in crate
* removes usage of `SymboltokenRef` from borrowed module and instead uses `SymbolRef`
* removes the separate `no_text_values` collection in `StructRef`
* adds `len()` and `is_empty()` for `StructRef`
* adds `IonEq` implmentation for `ElementRef`
* some minor doc comment changes for owned module

*Test:*
- modifies doc test use `SymbolRef` for borrowed module tests
- adds tests for `len` and `is_empty` methods of `StructRef`

### DIFF
--- a/src/value/borrowed.rs
+++ b/src/value/borrowed.rs
@@ -8,94 +8,66 @@
 //! backed by octets or string data, `&[u8]` and `&str` are used.
 
 use super::{IonElement, IonSequence, IonStruct, IonSymbolToken};
-use crate::symbol_ref::AsSymbolRef;
+use crate::ion_eq::IonEq;
+use crate::symbol_ref::{AsSymbolRef, SymbolRef};
 use crate::types::decimal::Decimal;
 use crate::types::integer::Integer;
 use crate::types::timestamp::Timestamp;
-use crate::types::SymbolId;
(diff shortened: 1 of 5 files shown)


src/value/borrowed.rs
```diff
@@ -580,8 +623,100 @@ mod borrowed_value_tests {
...
+        }
+    }
+
+    #[rstest(
+    container, is_empty,
+    case::struct_(
+        ElementRef::new_struct(
+            vec![
+                ("greetings", ElementRef::from(ValueRef::String("hello".into()))),
+                ("name", ElementRef::from(ValueRef::String("Ion".into())))
```
zslayton, line 679: `ValueRef::String(&str)` wraps a `&str`, so I don't think we need the `.into()`s.

zslayton (approved):
🚀

PR #439, opened 2022-10-06
adds implementaion of `Display` for `UInteger` and `Integer`

desaikd:
*Description of changes:*
This PR works on adding `Display` implementation for `UInteger` and ``Integer`.

*Test:*
Adds unit tests for `Display` of `UInteger` and `Integer`.

### DIFF
--- a/src/text/text_formatter.rs
+++ b/src/text/text_formatter.rs
@@ -323,10 +323,7 @@ impl<'a, W: std::fmt::Write> IonValueFormatter<'a, W> {
     }
 
     pub fn format_integer(&mut self, value: &Integer) -> IonResult<()> {
-        match value {
-            Integer::I64(i) => write!(self.output, "{}", i)?,
-            Integer::BigInt(i) => write!(self.output, "{}", i)?,
-        }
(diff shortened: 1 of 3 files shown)


src/text/text_formatter.rs
```diff
@@ -323,10 +323,7 @@ impl<'a, W: std::fmt::Write> IonValueFormatter<'a, W> {
     }
 
     pub fn format_integer(&mut self, value: &Integer) -> IonResult<()> {
-        match value {
-            Integer::I64(i) => write!(self.output, "{}", i)?,
-            Integer::BigInt(i) => write!(self.output, "{}", i)?,
-        }
+        write!(self.output, "{}", value)?;
```
rmarrowstone, line 326: Is there an equivalent change in `raw_text_writer`? I know there was for `Timestamp`, but I don't have my IDE open so just checking.

rmarrowstone (approved):

src/text/text_formatter.rs
desaikd: That's correct. I have added the change for `raw_text_writer` as well. Thanks!

PR #440, opened 2022-10-10
Implement blocking RawTextReader wrapper

nirosys:
*Issue #, if available:* n/a

*Description of changes:*
This PR adapts the blocking `RawTextReader` to act as a blocking wrapper around the non-blocking `RawTextReader`.

The blocking reader uses a non-blocking reader with a `Vec<u8>` backed text buffer, starting with a default size of 4kb. Data is read from the IonDataSource into the text buffer maintaining the 4kb size unless more data is needed in order to parse a larger value, such as a blob.

With each data consuming call, the reader will perform a check if the underlying source has more data, and if the current text buffer has less than 25% data remaining. If so, then the reader will fill the text buffer before calling the consuming function on the non-blocking reader.

If an incomplete error is returned during a consuming call, the blocking reader will attempt to fill more data into the text buffer. The reader will continue to feed data into the text buffer, until there is no more data, or the consuming function no longer returns an incomplete error. The incomplete error is bubbled up to the caller if the source data is exhausted. 

#### Potential Addition
Right now the blocking reader will use a 4kb buffer unless more space is needed to parse larger values. This is something the user should have control over, and probably warrants having a `with_capacity`-style creation method. I can follow up with a PR to add that too.

#### Discovered Issue
While extending the unit tests in order to pass the reported codecov %s, I ran into an issue where parsing a struct with 4k spaces between the last value and the closing struct delimiter failed. Adding a comma to end the last value resulted in a correct parse. Removing the spaces, also resulted in a correct parse. I updated the code to force newly read characters to be added to the current loaded lines (to see if it was a buffer reading issue), which ensured the closing delimiter was in the buffer when parsing and even saw the delimiter in the error message, but the failure to parse continued.

I'll dig more into this to figure out what the underlying issue is. Adding the `with_capacity`-style method above will help to determine if this is due to this PR's buffer reading, or some other issue. 

#### Update
The most recent commit addresses feedback on this PR, as well as a couple bugs that were discovered while implementing the end of stream signal.

Part of the fixes include adding a "carried state", which is a mechanism to allow tracking of state for more complex `next` logic. The first place this is used is parsing of struct fields where the field name is parsed, then the value, which can fail when an incomplete text error is reached while parsing the value.

The second place is when stepping out of a container. After popping the container from the parent list, if the reader is within another container it needs to find the delimiter or container end. During this, an incomplete text error can be reached, so state needs to be carried for the next call to `next`, after the reader gets more data.

In addition to the bug fixes, the stream end marker was added so that the reader can rely on Incomplete Text errors, rather than trying to read more data when the buffer starts getting low.

#### Update 2 
This round of updates addresses the issues brought up in review. Most of note is the handling of errors in the non-blocking `step_out` method. Details about the fix are left as comments within the `step_out` function.

Another note for this update is that I've included line, and column, numbers for errors that previously had constant placeholders. The implementation for column numbers is currently not accurate and due to recent changes in restack, has the potential to be altered when restacking, causing non-zero column numbers to be shifted by previously consumed lengths. I plan on following up soon with an update to handle lines and columns correctly.

---

### DIFF
--- a/src/raw_reader.rs
+++ b/src/raw_reader.rs
@@ -7,13 +7,15 @@ use std::fmt::{Display, Formatter};
 /// `RawReader` is a shorthand for a [Reader](crate::Reader) implementation that returns [RawStreamItem]s and
 /// uses [RawSymbolToken] to represent its field names, annotations, and symbol values.
 pub trait RawReader: IonReader<Item = RawStreamItem, Symbol = RawSymbolToken> {
-    // Defines no additional functionality
+    // Mark the stream as complete. This allows the reader to understand when partial parses on
+    // data boundaries are not possible.
+    fn stream_done(&mut self);
 }
 impl<T> RawReader for T
 where
     T: IonReader<Item = RawStreamItem, Symbol = RawSymbolToken>,
 {
-    // No additional implementations are necessary
+    fn stream_done(&mut self) {}
 }
 
 /// Allows a Box<dyn RawReader> to be used as a RawReader.
--- a/src/raw_reader.rs
+++ b/src/raw_reader.rs
@@ -9,13 +9,13 @@ use std::fmt::{Display, Formatter};
 pub trait RawReader: IonReader<Item = RawStreamItem, Symbol = RawSymbolToken> {
     // Mark the stream as complete. This allows the reader to understand when partial parses on
     // data boundaries are not possible.
-    fn stream_done(&mut self);
+    fn stream_complete(&mut self);
 }
 impl<T> RawReader for T
 where
     T: IonReader<Item = RawStreamItem, Symbol = RawSymbolToken>,
 {
-    fn stream_done(&mut self) {}
+    fn stream_complete(&mut self) {}
 }
 
 /// Allows a Box<dyn RawReader> to be used as a RawReader.
--- a/src/text/non_blocking/raw_text_reader.rs
+++ b/src/text/non_blocking/raw_text_reader.rs
@@ -27,11 +27,25 @@ const INITIAL_PARENTS_CAPACITY: usize = 16;
 
 // Represents reader state that can be carried over between calls to `next`.
 // Currently this is used to communicate state when `next` needs to return due to lack of data.
+#[derive(PartialEq, Debug)]
 enum ReaderState {
+    // Any previous `next` calls completed successfully and we have no carried state.
+    Ready,
     // A previous `next` call was able to parse the struct field's name, but not the value.
     StructFieldValue,
-    // A previous `next` call popped the container, but was unable to find an outter container's delimeter or end.
-    ContainerDelimeter,
+    // A previous `next` call popped the container, but was unable to find an outer container's delimiter or end.
+    ContainerDelimiter,
+}
+
+impl ReaderState {
+    // Returns true if the current ReaderState is Ready.
+    fn is_ready(&self) -> bool {
+        *self == Self::Ready
+    }
+
+    fn reset(&mut self) {
+        *self = Self::Ready;
+    }
 }
 
 pub struct RawTextReader<A: AsRef<[u8]>> {
@@ -52,7 +66,11 @@ pub struct RawTextReader<A: AsRef<[u8]>> {
     is_eos: bool,
     parents: Vec<ParentContainer>,
     // Optional state from previous read attempts.
-    carried_state: Option<ReaderState>,
+    carried_state: ReaderState,
+    // The depth at which step_out is called.
+    step_out_depth: Option<usize>,
+    // The "call depth" of the current nested step_out.
+    step_out_nest: usize,
 }
 
 /// Represents the final outcome of a [RawTextReader]'s attempt to parse the next value in the stream.
@@ -82,7 +100,9 @@ impl<A: AsRef<[u8]>> RawTextReader<A> {
             is_eof: false,
             is_eos: false,
             parents: Vec::with_capacity(INITIAL_PARENTS_CAPACITY),
-            carried_state: None,
+            carried_state: ReaderState::Ready,
+            step_out_depth: None,
+            step_out_nest: 0,
         }
     }
 
@@ -92,12 +112,12 @@ impl<A: AsRef<[u8]>> RawTextReader<A> {
 
     // Mark the data stream as being complete. This tells the reader that all data has been read
     // into the reader.
-    pub fn stream_done(&mut self) {
+    pub fn stream_complete(&mut self) {
         self.is_eos = true;
     }
 
-    // Returns true if the stream has been marked as completely loaded via `stream_done`.
-    pub fn is_stream_done(&self) -> bool {
+    // Returns true if the stream has been marked as completely loaded via `stream_complete`.
+    pub fn is_stream_complete(&self) -> bool {
         self.is_eos
     }
 
@@ -119,7 +139,7 @@ impl<A: AsRef<[u8]>> RawTextReader<A> {
 
         // Unset variables holding onto information about the previous position.
         self.current_ivm = None;
-        if self.carried_state.is_none() {
+        if self.carried_state.is_ready() {
             self.current_value = None;
             self.current_field_name = None;
         }
@@ -155,7 +175,7 @@ impl<A: AsRef<[u8]>> RawTextReader<A> {
             IonType::Struct => {
                 // If we haven't carried over a state indicating that we have read our field name,
                 // then we need to read it.
-                let have_name = matches!(self.carried_state, Some(ReaderState::StructFieldValue));
+                let have_name = self.carried_state == ReaderState::StructFieldValue;
                 let result = (!have_name).then(|| self.next_struct_field_name());
 
                 if let Some(Ok(None)) = result {
@@ -173,11 +193,11 @@ impl<A: AsRef<[u8]>> RawTextReader<A> {
                             // Read field name, we need to store it.
                             self.current_field_name = field_name;
                             // Note our state so we can start here if reading the value fails.
-                            self.carried_state = Some(ReaderState::StructFieldValue);
+                            self.carried_state = ReaderState::StructFieldValue;
                         }
                     }
                     let field_value = self.next_struct_field_value()?;
-                    self.carried_state = None; // All done, reset our state.
+                    self.carried_state.reset(); // All done, reset our state.
                     Ok(Some(field_value))
                 }
             }
@@ -203,6 +223,67 @@ impl<A: AsRef<[u8]>> RawTextReader<A> {
             }
             Err(e) => return Err(e),
         };
+
+        Ok(())
+    }
+
+    // This function implements the logic of `step_out`, but is extracted in order to make tracking
+    // of the step_out depth, which is handled by the actual `step_out` function, easier.
+    //
+    // See `step_out` for more info about how we handle errors when stepping out.
+    fn step_out_impl(&mut self) -> IonResult<()> {
+        // If we do not yet have a target depth, establish it.
+        if self.step_out_depth.is_none() {
+            self.step_out_depth = Some(self.depth() - 1);
+        }
+
+        if self.carried_state != ReaderState::ContainerDelimiter {
+            if self.parents.is_empty() {
+                return illegal_operation(
+                    "Cannot call `step_out()` when the reader is at the top level.",
+                );
+            }
+
+            // The container we're stepping out of.
+            let parent = self.parents.last().unwrap();
+
+            // If we're not at the end of the current container, advance the cursor until we are.
+            // Unlike the binary reader, which can skip-scan, the text reader must visit every value
+            // between its current position and the end of the container.
+            if !parent.is_exhausted() {
+                while let RawStreamItem::Value(_) | RawStreamItem::Null(_) = self.next()? {}
+            }
+
+            // Remove the parent container from the stack and clear the current value.
+            let _ = self.parents.pop();
+            self.current_value = None;
+
+            if self.parents.is_empty() {
+                // We're at the top level; nothing left to do.
+                return Ok(());
+            }
+            // We have stepped out and need to track that we have in just in case we error due to
+            // incomplete text while looking.
+            self.carried_state = ReaderState::ContainerDelimiter;
+        }
+
+        // We've stepped out, but the reader isn't at the top level. We're still inside another
+        // container. Make sure the container was followed by either the appropriate delimiter
+        // or the end of its parent.
+        let container_type = self.parents.last().unwrap().ion_type();
+        match container_type {
+            IonType::List => {
+                self.parse_expected("list delimiter or end", list_delimiter)?;
+            }
+            IonType::SExpression => {
+                self.parse_expected("s-expression delimiter or end", s_expression_delimiter)?;
+            }
+            IonType::Struct => {
+                self.parse_expected("struct delimiter or end", struct_delimiter)?;
+            }
+            scalar => unreachable!("Stepping out of a scalar type: {:?}", scalar),
+        };
+        self.carried_state = ReaderState::Ready;
 
         Ok(())
     }
@@ -246,7 +327,7 @@ impl<A: AsRef<[u8]>> RawTextReader<A> {
                     self.process_stream_item(item)
                 } else {
                     // If we are not at the end of the stream, we need to get more data.
-                    incomplete_text_error(0, 0)
+                    incomplete_text_error(self.buffer.lines_loaded(), self.buffer.line_offset())
                 }
             }
             RootParseResult::NoMatch => {
@@ -308,7 +389,7 @@ impl<A: AsRef<[u8]>> RawTextReader<A> {
             Ok(Some(value)) => Ok(value),
             Ok(None) => {
                 if !self.is_eos {
-                    incomplete_text_error(self.buffer.lines_loaded(), 0)
+                    incomplete_text_error(self.buffer.lines_loaded(), self.buffer.line_offset())
                 } else {
                     decoding_error(format!(
                         "Unexpected end of input while reading {} on line {}: '{}'",
@@ -340,7 +421,7 @@ impl<A: AsRef<[u8]>> RawTextReader<A> {
             RootParseResult::NoMatch => {
                 // If we are not at the end of the stream we could be missing a match due to partial
                 // data.
-                if !self.is_eos {
+                if self.is_eos {
                     // Return an error that contains the text currently in the buffer (i.e. what we
                     // were attempting to parse.)
                     let error_message = format!(
@@ -350,7 +431,7 @@ impl<A: AsRef<[u8]>> RawTextReader<A> {
                     );
                     decoding_error(error_message)
                 } else {
-                    incomplete_text_error(0, 0)
+                    incomplete_text_error(self.buffer.lines_loaded(), self.buffer.line_offset())
                 }
             }
             RootParseResult::Failure(error_message) => decoding_error(error_message),
@@ -456,7 +537,7 @@ impl<A: AsRef<[u8]>> RawTextReader<A> {
         // We unfortunately need to copy here in order to append the SENTINEL_ION_TEXT, since we
         // aren't guaranteed a vector-backed reader.
         let mut remaining_text = self.buffer.remaining_text().to_owned();
-        // Make a note of the buffer's length; we're about to modify it.
+        // Make a note of the text's length; we're about to modify it.
         let original_length = remaining_text.len();
 
         // Append our sentinel value to the end of the input buffer.
@@ -498,7 +579,9 @@ impl<A: AsRef<[u8]>> RawTextReader<A> {
                 // again. Return the value we found.
                 RootParseResult::Ok(value)
             }
-            Err(Incomplete(_needed)) => RootParseResult::Incomplete(self.buffer.lines_loaded(), 0),
+            Err(Incomplete(_needed)) => {
+                RootParseResult::Incomplete(self.buffer.lines_loaded(), self.buffer.line_offset())
+            }
             Err(Error(ion_parse_error)) => {
                 RootParseResult::Failure(format!(
                     "Parsing error occurred near line {}: '{}': '{:?}'",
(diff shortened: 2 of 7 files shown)


nirosys:
Moved this out of draft and into review. I cannot reproduce the coverage %s that codecov is generating. Running the github actions locally gives a 99.06% coverage for the blocking `raw_text_reader.rs` (which is most of this PR). I'm not entirely sure how codecov.io is producing the patch %, but the lines it is marking in the PR source as not covered include comments and blank lines.

src/text/raw_text_reader.rs
```diff
@@ -480,170 +99,106 @@ impl<T: ToIonDataSource> IonReader for RawTextReader<T> {
...
 
     fn next(&mut self) -> IonResult<RawStreamItem> {
-        // Parse the next value from the stream, storing it in `self.current_value`.
-        self.load_next_value()?;
-
-        // If we're positioned on an IVM, return the (major, minor) version tuple
-        if let Some((major, minor)) = self.current_ivm {
-            return Ok(RawStreamItem::VersionMarker(major, minor));
-        }
+        self.feed_if_needed()?;
```
zslayton, line 102: Rather than using the (reasonable!) heuristic you've implemented on each call to `next()`, I propose:
1. Doing an initial read at reader creation time. Our [builder APIs return an `IonResult<ReaderType>`](https://github.com/amzn/ion-rust/blob/94c83bf336f0c3902ac81b594f5d38774c097782/src/reader.rs#L37) to accommodate I/O during construction. (In that particular example, we peek at the first four bytes of the file to detect whether we're dealing with text or binary.)
2. Only reading from the source when we encounter an `Incomplete`.

That way most calls to `next()` will be free in terms of I/O checking.

src/text/raw_text_reader.rs
```diff
@@ -480,170 +99,106 @@ impl<T: ToIonDataSource> IonReader for RawTextReader<T> {
...
+
+        match self.reader.next() {
+            Err(err @ IonError::IncompleteText { .. }) => {
+                // We received an incomplete.. and need more data..
+                loop {
+                    // If there is no more data to read, we need to bubble the incomplete up to the
+                    // caller.
+                    if 0 == self.read_source(1024)? {
```
zslayton, line 110: Why 1024 instead of whatever the buffer has room for?

src/text/raw_text_reader.rs
```diff
@@ -480,170 +99,106 @@ impl<T: ToIonDataSource> IonReader for RawTextReader<T> {
...
-        // Parse the next value from the stream, storing it in `self.current_value`.
-        self.load_next_value()?;
-
-        // If we're positioned on an IVM, return the (major, minor) version tuple
-        if let Some((major, minor)) = self.current_ivm {
-            return Ok(RawStreamItem::VersionMarker(major, minor));
-        }
+        self.feed_if_needed()?;
+
+        match self.reader.next() {
```
zslayton, line 104: Here's a slightly more concise version if you like:
```rust
fn next(&mut self) -> IonResult<RawStreamItem> {
    self.feed_if_needed()?;
    loop {
        let mut result = self.reader.next();
        if let IonError::IncompleteText { .. } = result {
            if 0 == self.read_source(1024)? {
                return Err(err);
            }
       } else {
            return result;
        }
}
```

src/text/raw_text_reader.rs
```diff
@@ -670,103 +222,56 @@ impl<T: ToIonDataSource> IonReader for RawTextReader<T> {
...
-
-        // Remove the parent container from the stack and clear the current value.
-        let _ = self.parents.pop();
-        self.current_value = None;
-
-        if self.parents.is_empty() {
-            // We're at the top level; nothing left to do.
-            return Ok(());
-        }
+        self.feed_if_needed()?;
```
zslayton, line 237: I suggest making a `retry_on_incomplete` helper method that takes a closure to loop over as long as it's returning `Incomplete`. Then we can use it in `step_in()`, `step_out()` and (when this wrapper is format-agnostic) all of the `read_*` calls that can trigger I/O in the binary reader.

zslayton (commented):
I love a good mostly-deletions PR! 🎉 

image

This blocking wrapper is _tantalizingly close_ to being able to work for any format of raw reader, which is where I'd like to get to. In a follow-on PR, could you make it generic over another `R: RawReader`? Then we can delete the blocking raw binary reader implementation too! (There are probably prerequisite changes that need to happen for this; in particular, I think we'll need to consolidate `Incomplete` and `IncompleteText` into a single error type. This in turn means putting some more thought into how we report reader position in a format-agnostic way.)

Some minor comments below.

src/text/non_blocking/text_buffer.rs
```diff
@@ -284,6 +284,7 @@ impl TextBuffer<Vec<u8>> {
 
         let read_buffer = &mut self.data.as_mut_slice()[self.data_end..];
         let bytes_read = source.read(read_buffer)?;
+        self.data.resize(self.data_end + bytes_read, 0);
```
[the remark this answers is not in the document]
nirosys: oof.. Thank you for calling that out. I had run into an issue where parsing was failing due to unexpected characters in the text (large number of '\0's), that was ultimately due to the `read_from` reserving the needed space for the requested bytes. I quickly fixed it with this resize, to truncate the buffer down to only what was read.  At the time I was focused on another area of the code, and never went back to investigate why the issue was happening.

I've dug deeper now, and I'll post a new commit that removes this resize in favor of fixing the underlying bug in the text buffer. 

When UTF-8 validation is done, my original code validated `&self.data.as_ref()[valid_end..]`. Once a `read_from` happens that pre-extends the buffer beyond the actual data size the UTF-8 validation starts validating the end of the buffer without read data. Since the data is initialized with '\0's, the unread bytes do not get evaluated by `from_utf8`, however the `Ok(_)` match assumed the entire buffer was valid and would return the buffer's length as the validated end.

src/text/raw_text_reader.rs
nirosys: I think we need a way for the user to flag the non-blocking reader as 'done' in terms of data being provided. The reason I started reading ahead of consuming functions was to reduce the chances of the parser returning back values and advancing when partial values were present.

For example, in the unit tests that used the builder APIs, the chain of first 4 bytes followed by the rest of the buffer, resulted in a first `read` of 4 bytes. For text encoded timestamps, that lined up perfectly with the 4 digit year, which would parse as an integer. Then the next read would generate an error. Those boundaries seem to be possible anywhere, and there seems to be ample opportunities for partial values to be parsed as valid values.

What do you think of having the user signal when data is complete, and if that signal has not been made and a parse would consume the remainder of the buffer, an incomplete error is returned and the data is left as is?

With that I think we can rely on the `Incomplete` to signal when to read.

zslayton: Oh! Thank you for explaining the problem. Take a look at the old `RawTextReader`'s [`parse_value_at_eof()` method](https://github.com/amzn/ion-rust/blob/main/src/text/raw_text_reader.rs#L376). What we need (and I think this is similar to what you described) is a way for the blocking reader to say:
```
if self.reader.next() == Incomplete and self.input().is_eof() {
  return self.reader.parse_value_at_eof();
}
```
What I don't know (and we don't necessarily need to solve this here/now) is how to do this in a way that's format agnostic. The binary reader doesn't have this problem; it can always tell where the boundaries of a value are.

We may end up needing to add some methods to the `RawReader` trait.

src/text/raw_text_reader.rs
nirosys: Arbitrarily chose 1024, but it isn't tied to the remaining space in the buffer in order to allow the buffer to grow to support reading blobs that are larger than the buffer.

zslayton: I wonder if we should double it on each retry so it ramps up faster for large values?

src/text/raw_text_reader.rs
nirosys: With the recent commit, the retry logic is only contained in the `next`, and `step_out` methods. I do want to pull it into a helper method, but ran into issues due to borrows of self. I need to spend a little more time with it, but didn't want to hold up the PR any more than it already has been.

src/text/non_blocking/raw_text_reader.rs
```diff
@@ -721,29 +795,37 @@ impl<A: AsRef<[u8]>> IonReader for RawTextReader<A> {
...
     fn step_out(&mut self) -> IonResult<()> {
-        if self.parents.is_empty() {
-            return illegal_operation(
-                "Cannot call `step_out()` when the reader is at the top level.",
-            );
-        }
+        // It is possible to fail partway through stepping out, after we've popped the parent off
+        // and started looking for an outer container's delimiter. So we first need to check our
+        // carried state and see if we're re-attempting step_out.
+        if !matches!(self.carried_state, Some(ReaderState::ContainerDelimeter)) {
```
zslayton, line 801: A bit simpler:
```suggestion
        if self.carried_state != Some(ReaderState::ContainerDelimiter) {
```

src/text/non_blocking/raw_text_reader.rs
```diff
@@ -721,29 +795,37 @@ impl<A: AsRef<[u8]>> IonReader for RawTextReader<A> {
...
-        if self.parents.is_empty() {
-            // We're at the top level; nothing left to do.
-            return Ok(());
+            if self.parents.is_empty() {
+                // We're at the top level; nothing left to do.
+                return Ok(());
+            }
+            // We have stepped out and need to track that we have in just in case we error due to
+            // incomplete text while looking.
+            self.carried_state = Some(ReaderState::ContainerDelimeter);
```
zslayton, line 828: This strikes me as late in the method to set `carried_state`; what if the loop on line 815 encounters incomplete data while it's exhausting all of the child elements of the container we're stepping out of?

src/text/raw_text_reader.rs
```diff
@@ -11,45 +9,43 @@ use crate::{Decimal, Integer, IonError, IonType};
...
+        };
+        reader.read_source(size)?;
+        Ok(reader)
     }
 
     pub fn bytes_read(&self) -> usize {
         self.reader.bytes_read()
     }
 
+    pub fn stream_done(&mut self) {
```
zslayton, line 45: Is this intentionally `pub`? I expected the blocking implementation to handle this automatically.

src/text/raw_text_reader.rs
```diff
@@ -99,25 +70,30 @@ impl<T: ToIonDataSource> IonReader for RawTextReader<T> {
...
+                        return result;
+                    } else {
+                        self.reader.stream_done();
                     }
                 }
+                // The assumption here is that most buffer sizes will start at a magnitude the user
+                // is comfortable with in terms of memory usage. So if we're reading more in order
+                // to reach a parsable point we do not want to start consuming more than an order of
+                // magnitude more memory just to get there.
+                read_size = std::cmp::min(read_size * 2, self.max_read_size * 10);
```
zslayton, line 93: We're willing to read by 10x the `max_read_size`, so I think we should reconsider that variable name. `original_buffer_size`, maybe?

src/text/raw_text_reader.rs
```diff
@@ -760,4 +727,221 @@ mod reader_tests {
...
+
+    impl crate::data_source::ToIonDataSource for TestIonSource {
+        type DataSource = Self;
+
+        fn to_ion_data_source(self) -> Self::DataSource {
+            self
+        }
+    }
+
+    impl std::io::BufRead for TestIonSource {
```
zslayton, line 769: Could `TestIonSource` be replaced by [`io::Cursor`](https://doc.rust-lang.org/std/io/struct.Cursor.html) from the stdlib?

src/text/raw_text_reader.rs
```diff
@@ -760,4 +727,221 @@ mod reader_tests {
...
+        assert_eq!(SOURCE.len(), metrics.borrow().bytes_read);
+        assert!(result.is_ok());
+        assert_eq!(IonType::Symbol, reader.ion_type().unwrap());
+
+        Ok(())
+    }
+
+    #[test]
+    #[ignore]
+    fn repro_multi_container() -> IonResult<()> {
```
zslayton, line 856: Is this `repro` test a debugging leftover?

src/text/non_blocking/raw_text_reader.rs
```diff
@@ -25,6 +25,15 @@ use crate::IonType;
 
 const INITIAL_PARENTS_CAPACITY: usize = 16;
 
+// Represents reader state that can be carried over between calls to `next`.
+// Currently this is used to communicate state when `next` needs to return due to lack of data.
+enum ReaderState {
+    // A previous `next` call was able to parse the struct field's name, but not the value.
+    StructFieldValue,
+    // A previous `next` call popped the container, but was unable to find an outter container's delimeter or end.
+    ContainerDelimeter,
```
zslayton, line 34:
```suggestion
    // A previous `next` call popped the container, but was unable to find an outer container's delimiter or end.
    ContainerDelimiter,
```

src/text/non_blocking/raw_text_reader.rs
```diff
@@ -25,6 +25,15 @@ use crate::IonType;
 
 const INITIAL_PARENTS_CAPACITY: usize = 16;
 
+// Represents reader state that can be carried over between calls to `next`.
+// Currently this is used to communicate state when `next` needs to return due to lack of data.
+enum ReaderState {
+    // A previous `next` call was able to parse the struct field's name, but not the value.
+    StructFieldValue,
```
zslayton, line 32: My initial thought here was that we should just make parsing a field name and value transactional; if either fails we just bubble up `Incomplete` and retry both on the next attempt.

We still need to track state for stepping out of a container, so we could keep this implementation even though it's a bit more complex. Interested to hear your thoughts, though.

src/text/non_blocking/raw_text_reader.rs
```diff
@@ -37,8 +46,13 @@ pub struct RawTextReader<A: AsRef<[u8]>> {
     // Otherwise, it is None.
     current_ivm: Option<(u8, u8)>,
     bytes_read: usize,
+    // True if the current text buffer is exhausted.
     is_eof: bool,
+    // True if the caller has indicated that all data has been read for the 'stream'.
+    is_eos: bool,
     parents: Vec<ParentContainer>,
+    // Optional state from previous read attempts.
+    carried_state: Option<ReaderState>,
```
zslayton, line 55: I think we can merge a couple of these member fields; rather than holding a `carried_state: Option<ReaderState>`, I think it can have a `state: ReaderState` with a new variant that represents "ready to read the next value".

See the non-blocking raw binary reader's [`ReaderState` type](https://github.dev/amzn/ion-rust/blob/b01a603216edfe83d51aa57647ea912665f5334b/src/binary/non_blocking/raw_binary_reader.rs#L233-L261) for comparison.

src/text/non_blocking/raw_text_reader.rs
```diff
@@ -37,8 +46,13 @@ pub struct RawTextReader<A: AsRef<[u8]>> {
     // Otherwise, it is None.
     current_ivm: Option<(u8, u8)>,
     bytes_read: usize,
+    // True if the current text buffer is exhausted.
     is_eof: bool,
```
zslayton, line 50: I realize this isn't part of your PR, but I wonder if this `is_eof` needs to be tracked separately from just checking `self.buffer.is_empty()` anywhere we currently read it, perhaps via a helper method with a friendly name.

src/raw_reader.rs
```diff
@@ -7,13 +7,15 @@ use std::fmt::{Display, Formatter};
 /// `RawReader` is a shorthand for a [Reader](crate::Reader) implementation that returns [RawStreamItem]s and
 /// uses [RawSymbolToken] to represent its field names, annotations, and symbol values.
 pub trait RawReader: IonReader<Item = RawStreamItem, Symbol = RawSymbolToken> {
-    // Defines no additional functionality
+    // Mark the stream as complete. This allows the reader to understand when partial parses on
+    // data boundaries are not possible.
+    fn stream_done(&mut self);
```
zslayton, line 12: I think I prefer the phrasing in your comment. How about `stream_complete`?
```suggestion
    fn stream_complete(&mut self);
```

src/text/non_blocking/raw_text_reader.rs
zslayton: > What is the benefit of throwing the field name away? Is it just to keep the code simple?

Yep!

> Even if it is only during read boundaries, the data, and read/server response/etc size play into how impactful that is.

Intuitively, I'd expect the work wasted at read boundaries to be negligible in practice. I'd prefer to start simple and add additional state machinery if/when it shows up in profiling.

src/text/non_blocking/raw_text_reader.rs
```diff
@@ -415,17 +568,19 @@ impl<A: AsRef<[u8]>> RawTextReader<A> {
                 RootParseResult::Eof
             }
             Ok((_remaining_text, value)) => {
+                // If we match, and try to consume the remaining buffer in its entirety before we
+                // know the data has been fully loaded we need to treat it as an incomplete error
+                // so that we do not inadvertently succeed on a partial parse.
+                if original_length == self.buffer.remaining_text().len() && !self.is_eos {
+                    return RootParseResult::Incomplete(0, 0);
```
zslayton, line 575: Is the `(0, 0)` a placeholder?

src/text/non_blocking/raw_text_reader.rs
```diff
@@ -723,47 +878,36 @@ impl<A: AsRef<[u8]>> IonReader for RawTextReader<A> {
...
-            IonType::SExpression => {
-                self.parse_expected("s-expression delimiter or end", s_expression_delimiter)?;
-            }
-            IonType::Struct => {
-                self.parse_expected("struct delimiter or end", struct_delimiter)?;
+        // If an incomplete error occurs during a step_out, we will re-enter step_out and:
+        //    If we have any error other than ContainerDelimiter, we need to finish exhausting
+        //    the current container and complete the step_out that failed.
+        //
+        //    Once the failed step_out is done, we need to "bubble up" and step_out of any of
```
zslayton, line 885: I don't think I understand this bit -- let's talk through it offline.

src/text/non_blocking/raw_text_reader.rs
```diff
@@ -1261,4 +1408,91 @@ mod reader_tests {
...
     }
+
+    // Test stepping out at the top level with incomplete error during step_out call.
+    #[test]
+    fn resume_step_out() -> IonResult<()> {
+        // The spacing on this is important, since we need to know where (approximately) the
+        // incomplete error occurs.
+        let source = r#"{field:{another_field:{foo:"We should fail somewhere in this string.."}}}"#;
+
+        let mut reader = RawTextReader::new(source[..47].as_bytes().to_owned());
```
zslayton, line 1419: Could you add a comment showing where 47 is in the input? I see it's somewhere in the string, but knowing exactly where would help me mentally run the unit test.

src/text/non_blocking/raw_text_reader.rs
```diff
@@ -1261,4 +1408,91 @@ mod reader_tests {
...
+        assert_eq!(reader.depth(), 2);
+
+        match reader.step_out() {
+            Err(IonError::IncompleteText { .. }) => {
+                assert_eq!(reader.depth(), 3); // we should fail at depth 3, within the foo field.
+                reader.read_from(&mut source[47..].as_bytes(), 512)?;
+
+                // Resume our initial step_out, this should bring us back to depth 1.
+                reader.step_out()?;
+                assert_eq!(reader.depth(), 1);
```
zslayton, line 1438: If I'm at depth 3 and `step_out` fails, then the next time I call `step_out` successfully I should be at depth 2. If I'm understanding this correctly, the current implementation allows the reader to make _some_ progress before failing and the user can then resume using it from that unknown interstitial position. I think there's a lot of room for confusion here (how far did the reader get? where is it now?), and it's not consistent with our other `Reader` implementations' contracts.

src/text/non_blocking/text_buffer.rs
```diff
@@ -292,8 +307,9 @@ impl TextBuffer<Vec<u8>> {
             })?
         }
 
-        self.data_end += bytes_read;
-        self.data_exhausted = false;
+        if self.data_end > 0 {
+            self.data_exhausted = false;
+        }
```
zslayton, line 312: If you prefer:
```suggestion
        self.data_exhausted = self.data_end > 0;
```

src/text/raw_text_reader.rs
```diff
@@ -1,476 +1,59 @@
...
-                } else {
-                    // Otherwise, this is the end of the struct.
-                    Ok(None)
-                }
-            }
-            other => unreachable!(
-                "The reader's `parents` stack contained a scalar value: {:?}",
-                other
-            ),
+    pub fn new_with_size(input: T, size: usize) -> IonResult<RawTextReader<T>> {
```
zslayton, line 30: Can you open an issue to track exposing this to the `Builder` API?

src/text/raw_text_reader.rs
```diff
@@ -1,476 +1,59 @@
...
-    /// Constructs an [IonError::IllegalOperation] which explains that the reader was asked to
-    /// perform an action that is only allowed when it is positioned over the item type described
-    /// by the parameter `expected`.
-    fn expected<E: Display>(&self, expected: E) -> IonError {
-        illegal_operation_raw(format!(
-            "type mismatch: expected a(n) {} but positioned over a(n) {}",
-            expected,
-            self.current()
-        ))
+        self.bytes_fed += bytes_read;
```
zslayton, line 52: I can't remember, is this information not already tracked in the raw reader or text buffer?

src/text/raw_text_reader.rs
zslayton: You can create a reader that uses a `&mut Cursor` and then get the cursor back when you're done. Personally, I'd also be in favor of having a `Reader::input_source()` or something that lent out a reference to the underlying data source. We don't have to solve that in this PR, though.

src/text/non_blocking/raw_text_reader.rs
```diff
@@ -207,6 +227,67 @@ impl<A: AsRef<[u8]>> RawTextReader<A> {
...
+            let _ = self.parents.pop();
+            self.current_value = None;
+
+            if self.parents.is_empty() {
+                // We're at the top level; nothing left to do.
+                return Ok(());
+            }
+            // We have stepped out and need to track that we have in just in case we error due to
+            // incomplete text while looking.
+            self.carried_state = ReaderState::ContainerDelimiter;
```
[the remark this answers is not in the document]
nirosys: The `ContainerDelimiter` state is specifically for when we have finished reading all of the container's items and have stepped out of the container, but need to find an inner delimiter, or closing delimiter, for a parent container. That is why it is set after exhausting the contents, and after we test for an empty `parents` list.

If an error occurs while reading the container's values, like your example, the carried state *needs* to be the state from where the error occurred. Since the calls to `next` in the exhaustion loop can step into other containers, or fail to parse any value, we can't set a carried state until we're done exhausting the contents and (from the perspective of the step_out) are back in control of what's being parsed.

In the case of `ContainerDelimiter`, if we set it before reading, and an error occurs while reading a value, the next step into `step_out` would cause the reader to start looking for a parent container's delimiter. If we set it to any other state, then another instance of `step_out` could clobber that state with whatever it was doing at the time.

I've added a unit test for your example and will add some more, and push them up when they're done.

src/text/non_blocking/raw_text_reader.rs
nirosys: 👍

src/text/non_blocking/raw_text_reader.rs
```diff
@@ -207,6 +227,67 @@ impl<A: AsRef<[u8]>> RawTextReader<A> {
...
     }
 
+    // This function implements the logic of `step_out`, but is extracted in order to make tracking
+    // of the step_out depth, which is handled by the actual `step_out` function, easier.
+    //
+    // See `step_out` for more info about how we handle errors when stepping out.
+    fn step_out_impl(&mut self) -> IonResult<()> {
+        // If we do not yet have a target depth, establish it.
+        if self.step_out_depth.is_none() {
+            self.step_out_depth = Some(self.depth() - 1);
```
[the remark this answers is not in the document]
nirosys: You're right. I had originally set it to the depth step_out was called at and did not update the comment when I changed to the target depth.

src/text/non_blocking/raw_text_reader.rs
nirosys: Sorry for the sporadic responses, I let this one marinate for a bit. Your comment about the reader implementations' contracts is very surprising.

Confusion for whom? The user? Or maintainers?
What reader contract? I may have missed documentation describing this.

If the intent truly results in requiring the entire remaining content of a container to be in memory before a `step_out` can actually make progress, and re-parsing the same data until it does, then I have concerns. Let's discuss.

nirosys: Update after our discussion:

I misunderstood Zach's comment. We do not need to roll back to where the reader was when `step_out`/`next` was called, we need to ensure that if we cannot complete the `step_out` or `next` then we do not allow the user to advance the reader in any other way until the initial failed call is completed. This should match the behavior of the binary reader.

src/text/non_blocking/raw_text_reader.rs
```diff
@@ -124,15 +173,32 @@ impl<A: AsRef<[u8]>> RawTextReader<A> {
...
+                        None => (), // We have a field name from a previous call to this function..
+                        Some(err @ Err(_)) => {
+                            // error occurred, bubble up.
+                            err?;
+                        }
+                        Some(Ok(field_name)) => {
+                            // Read field name, we need to store it.
+                            self.current_field_name = field_name;
+                            // Note our state so we can start here if reading the value fails.
+                            self.carried_state = ReaderState::StructFieldValue;
```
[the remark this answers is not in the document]
nirosys: I will get this implemented with the next update.

src/text/raw_text_reader.rs
nirosys: Yea, this was missed in the last update. I had removed the original use of `bytes_fed`, and the function that exposed it, but missed the actual field definition and update to it. Deleted.

src/text/raw_text_reader.rs
nirosys: Absolutely

src/text/non_blocking/raw_text_reader.rs
```diff
@@ -25,16 +25,19 @@ use crate::IonType;
...
-    // A previous `next` call was able to parse the struct field's name, but not the value.
-    StructFieldValue,
-    // A previous `next` call popped the container, but was unable to find an outer container's delimiter or end.
-    ContainerDelimiter,
+    // A previous `step_out` call failed to complete, but may have progressed through some values.
+    // We need to continue the step_out before doing anything else.
+    SteppingOut {
+        // The depth the reader is attempting to step-out to.
+        target_depth: usize,
+        // Whether the step_out process is looking for the parent container or not.
```
zslayton, line 38: Hm, I'm unsure what "looking for the parent container" means here. It sounds distinct from expecting the closing delimiter. Maybe you could add an example?

src/text/non_blocking/raw_text_reader.rs
```diff
@@ -59,18 +62,19 @@ pub struct RawTextReader<A: AsRef<[u8]>> {
...
     is_eos: bool,
     parents: Vec<ParentContainer>,
-    // Optional state from previous read attempts.
-    carried_state: ReaderState,
-    // The depth at which step_out is called.
-    step_out_depth: Option<usize>,
+    // Current state of the reader. This state, combined with `need_continue` will signal if an
+    // action needs to be completed (need_continue = true) or if we are attempting for the first
+    // time (need_continue = false).
+    state: ReaderState,
```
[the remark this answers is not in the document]
zslayton: I don't think I've grasped why this couldn't be stored in the `SteppingOut` variant, but we don't need to hold up the PR for it.

zslayton (approved):

PR #442, opened 2022-10-12
Enables GATs iteration in the Element API

zslayton:
This PR replaces all uses of `Box<dyn Iterator<Item=_> + 'a>` in the `Element` and `ElementRef` APIs with generic associated types. This means that heap allocations are no longer required to iterate over annotations, sequence elements, or struct fields.

Because this change uses generic associated types, it also configures the CI tests to use the `beta` channel of `cargo`. GATs will be available in the next release of Rust (v1.65, out in November); once that happens, we can and should switch back to `stable`. (#443)

This change also removes the recently added dependency on the `linkhash` crate. That dependency was originally pulled in so structs could remember the order in which their fields were inserted for later iteration. However, it only worked if the struct did not contain duplicate field names. This PR changes the field storage for `Struct` and `StructRef` to store their fields in a `Vec` and also maintain a `HashMap<Symbol, IndexVec>` that remembers all of the indexes at which values for a given field can be found. 

Further, it moves `Sequence` elements and `Struct` fields into an `Rc` to make `clone()` operations cheaper. A future PR will modify iterators over collection types to clone the `Rc` and thus remove one of their lifetime constraints, making it much easier to write a recursive iterator over an element tree without constant cloning.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -35,6 +35,7 @@ jobs:
         with:
           profile: minimal
           toolchain: beta
+          components: rustfmt, clippy
           override: true
       - name: Cargo Build 
         uses: actions-rs/cargo@v1
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -34,7 +34,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: stable
+          toolchain: beta
           override: true
       - name: Cargo Build 
         uses: actions-rs/cargo@v1
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -34,6 +34,8 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
+          # We're temporarily testing against `beta` to have access to GATs. When Rust v1.65 is released in
+          # 2022-11, we should change this back to `stable`.
           toolchain: beta
           components: rustfmt, clippy
           override: true
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -36,6 +36,7 @@ jobs:
           profile: minimal
           # We're temporarily testing against `beta` to have access to GATs. When Rust v1.65 is released in
           # 2022-11, we should change this back to `stable`.
+          # See: https://github.com/amzn/ion-rust/issues/443
           toolchain: beta
           components: rustfmt, clippy
           override: true
--- a/examples/read_all_values.rs
+++ b/examples/read_all_values.rs
@@ -62,7 +62,7 @@ fn read_all_values<R: RawReader>(reader: &mut R) -> IonResult<usize> {
                 match ion_type {
                     Struct | List | SExpression => reader.step_in()?,
                     String => {
-                        let _text = reader.map_string(|_s| ())?;
+                        reader.map_string(|_s| ())?;
                     }
                     Symbol => {
                         let _symbol_id = reader.read_symbol()?;
@@ -83,10 +83,10 @@ fn read_all_values<R: RawReader>(reader: &mut R) -> IonResult<usize> {
                         let _boolean = reader.read_bool()?;
                     }
                     Blob => {
-                        let _blob = reader.map_blob(|_b| ())?;
+                        reader.map_blob(|_b| ())?;
                     }
                     Clob => {
-                        let _clob = reader.map_clob(|_c| ())?;
+                        reader.map_clob(|_c| ())?;
                     }
                     Null => {}
                 }
--- a/src/value/borrowed.rs
+++ b/src/value/borrowed.rs
@@ -170,7 +170,7 @@ impl<'val> IonSequence for SequenceRef<'val> {
     type Element = ElementRef<'val>;
     type ElementsIterator<'a> = ElementRefIterator<'a, 'val> where 'val: 'a;
 
-    fn iter<'a>(&'a self) -> Self::ElementsIterator<'a> {
+    fn iter(&self) -> Self::ElementsIterator<'_> {
         ElementRefIterator::new(&self.children)
     }
 
@@ -258,7 +258,7 @@ where
 
             by_name
                 .entry(field_name.clone())
-                .or_insert_with(|| IndexVec::new())
+                .or_insert_with(IndexVec::new)
                 .push(by_index.len());
             by_index.push((field_name, field_value));
         }
--- a/src/value/borrowed.rs
+++ b/src/value/borrowed.rs
@@ -13,7 +13,7 @@ use crate::symbol_ref::{AsSymbolRef, SymbolRef};
 use crate::types::decimal::Decimal;
 use crate::types::integer::Integer;
 use crate::types::timestamp::Timestamp;
-use crate::value::iterators::{ElementRefIterator, FieldRefIterator};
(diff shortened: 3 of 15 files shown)


zslayton:
I can't make any sense of the code coverage results; I don't think I trust `codecov` anymore. The `ion-tests` integration exercises the DOM thoroughly, however.

src/value/owned.rs
```diff
@@ -187,52 +197,113 @@ impl PartialEq for Sequence {
...
+
+    /// Gets the last value in the Struct that is associated with the specified field name.
+    ///
+    /// Note that the Ion data model views a struct as a bag of (name, value) pairs and does not
+    /// have a notion of field ordering. In most use cases, field names are distinct and the last
+    /// appearance of a field in the struct's serialized form will have been the _only_ appearance.
+    /// If a field name appears more than once, this method makes the arbitrary decision to return
+    /// the value associated with the last appearance. If your application uses structs that repeat
+    /// field names, you are encouraged to use [get_all] instead.
+    fn get_last<A: AsSymbolRef>(&self, field_name: A) -> Option<&Element> {
```
rmarrowstone, line 244: Curious why this is `get_last` instead of `get_first` (just `get_any`). My only concern is that `get_last` forces the implementation to read and map the entire struct. That may be unproblematic now, but I don't see a good reason to choose last over first or any.

src/value/borrowed.rs
```diff
@@ -249,84 +250,126 @@ where
...
-            number_of_fields += 1;
+impl<'data> FieldRefs<'data> {
+    /// Gets all of the indexes that contain a value associated with the given field name.
+    fn get_indexes<A: AsSymbolRef>(&self, field_name: A) -> Option<&IndexVec> {
+        match field_name.as_symbol_ref().text() {
+            // If the provided field name symbol has undefined text...
+            None => {
+                // ...then build a cheap, stack-allocated `Symbol` that represents unknown text
+                let symbol = SymbolRef::with_unknown_text();
+                // ...and use the unknown text symbol to look up matching field values
```
rmarrowstone, line 288: I'm trying to reason about what the behavior of this would be. What fields would match the `unknown text symbol`?

src/value/owned.rs
rmarrowstone: I guess my assumption is that we would want to build the map (and the indexing to it) lazily. That doesn't seem to be happening now (and I don't think it needs to now). I confess that reviewing rust code in github is still relatively new to me, so please let me know what I'm missing.

src/value/owned.rs
```diff
@@ -187,52 +197,113 @@ impl PartialEq for Sequence {
...
+                // ...then build a cheap, stack-allocated `Symbol` that represents unknown text
+                let symbol = Symbol::unknown_text();
+                // ...and use the unknown text symbol to look up matching field values
+                self.by_name.get(&symbol)
+            }
+            Some(text) => {
+                // Otherwise, look it up by text
+                self.by_name.get(text)
+            }
+        }
```
jobarr-amzn, line 224:
```suggestion
        field_name.as_symbol_ref().text()
            .and_then(|text| self.by_name.get(text))
            .or_else(|| self.by_name.get(&(Symbol::unknown_text())))
```

It looks like something like this will work here. I think it's clearer, this is the kind of logic these combinators were written for.

My suggestion here doesn't preserve the comments or the first unknown then known ordering, which is not ideal. Take or leave as little as you like :)

src/value/owned.rs
zslayton: > My only concern is that get_last forces the implementation to read and map the entire struct. ... I guess my assumption is that we would want to build the map (and the indexing to it) lazily.

This is a good instinct, and a `LazyElement` is something we're planning to implement in the future. As you mention, this implementation (`Element`/`Struct`) is fully materialized in memory, so the choice of first or last is truly arbitrary. 

This particular method (`get_last`) lives on a private type (`Fields`) that isn't exposed to the end user. It's invoked by a public method called `get` that's defined on the `IonStruct` trait. If we were going to change the contract, that's the method we'd want to change.

At present, [`get` promises to return the last field](https://github.com/amzn/ion-rust/blob/4c7fc62e452e285862c181494267d86209a947a0/src/value/mod.rs#L483-L484) it found in the serialized struct. From an Ion data model perspective, this isn't really correct; struct fields don't have an order. The most correct thing to do would be to only offer a `get_all()` method that returned an iterator whose iteration order was not guaranteed. However, nearly all `struct` use cases don't use repeated fields and users expect to be able to access the one-and-only value for a given field trivially, so `get` is available as a convenience. If it turns out someone wants the _first_ appearance of a given field, they can call `self.get_all(name).first()` and (now that GATs are enabled) that will be basically the same cost as `get`.

src/value/borrowed.rs
zslayton: Fields whose symbol ID intentionally maps to unknown text. This is the case for `$0` as well as any symbol whose entry in the symbol table was non-text (`null`, `5`, `false`, etc).

.github/workflows/rust.yml
```diff
@@ -34,7 +34,10 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: stable
+          # We're temporarily testing against `beta` to have access to GATs. When Rust v1.65 is released in
+          # 2022-11, we should change this back to `stable`.
+          toolchain: beta
+          components: rustfmt, clippy
```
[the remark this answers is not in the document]
zslayton: Created #443 and added a link to this comment, thanks.

src/value/borrowed.rs
```diff
@@ -249,84 +250,126 @@ where
...
+    by_name: HashMap<SymbolRef<'a>, IndexVec>,
+}
 
-            number_of_fields += 1;
+impl<'data> FieldRefs<'data> {
+    /// Gets all of the indexes that contain a value associated with the given field name.
+    fn get_indexes<A: AsSymbolRef>(&self, field_name: A) -> Option<&IndexVec> {
+        match field_name.as_symbol_ref().text() {
+            // If the provided field name symbol has undefined text...
+            None => {
```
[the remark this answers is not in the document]
zslayton: Maybe slightly? Done.

src/value/iterators.rs
```diff
@@ -0,0 +1,193 @@
+use crate::value::borrowed::ElementRef;
+use crate::value::owned::Element;
+use crate::{Symbol, SymbolRef};
+use smallvec::SmallVec;
+// This macro defines a new iterator type for a given `Iterator => Item` type name pair.
+//
+// The implementation produced can be used to iterate over any `Vec<Item>` or `&[Item]`.
+macro_rules! create_new_slice_iterator_type {
+    ($($iterator_name:ident => $item_name:ty),*) => ($(
```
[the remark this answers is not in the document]
zslayton: Done

jobarr-amzn (approved):

examples/read_all_values.rs
```diff
@@ -62,7 +62,7 @@ fn read_all_values<R: RawReader>(reader: &mut R) -> IonResult<usize> {
                 match ion_type {
                     Struct | List | SExpression => reader.step_in()?,
                     String => {
-                        let _text = reader.map_string(|_s| ())?;
+                        reader.map_string(|_s| ())?;
```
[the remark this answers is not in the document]
zslayton: This example program (our only one!) is sorely in need of updating in general. 😞 I'll leave this for a future PR.

PR #448, opened 2022-11-07
adds implementation of `Catalog`

desaikd:
*Description of changes:*
This PR works on adding implementation of `Catalog`. 

*List of changes:*
- adds implementation of `SharedSymbolTable`
- adds implementation of `Catalog`
  - adds trait `Catalog`, `MutableCatalog`
  - adds struct `SimpleCatalog`

*Test:*
- adds unit tests for `Catalog` implementation.

### DIFF
--- a/src/catalog.rs
+++ b/src/catalog.rs
@@ -3,28 +3,25 @@ use crate::result::{illegal_operation, illegal_operation_raw};
 use crate::shared_symbol_table::SharedSymbolTable;
 use std::collections::{BTreeMap, HashMap};
 
-/// Catalog is a collection of Shared Symbol Tables
-/// for more information on [Catalog]: https://amzn.github.io/ion-docs/docs/symbols.html#the-catalog
+/// A Catalog is a collection of Shared Symbol Tables.
+/// For more information about the concept of a catalog,
+/// see [the `symbols` section of the specification](https://amzn.github.io/ion-docs/docs/symbols.html#the-catalog).
 pub trait Catalog {
     /// Returns the Shared Symbol Table with given table name
     /// If a table with the given name doesn't exists or if the table name is an empty string
     /// then returns an error
+    /// If a table with multiple versions exists for the given name then it will return the latest version of table
     fn get_table(&self, name: &str) -> IonResult<SharedSymbolTable>;
     /// Returns the Shared Symbol Table with given table name and version
     /// If a table with given name and version doesn't exists then it returns an error
     fn get_table_with_version(&self, name: &str, version: usize) -> IonResult<SharedSymbolTable>;
 }
 
-pub trait MutableCatalog {
-    /// Adds a Shared Symbol Table with name into the Catalog
-    fn put_table(&mut self, table: SharedSymbolTable);
-}
-
-struct SimpleCatalog {
+struct MapCatalog {
     tables_by_name: HashMap<String, BTreeMap<usize, SharedSymbolTable>>,
 }
 
-impl SimpleCatalog {
+impl MapCatalog {
     pub fn new() -> Self {
         Self {
             tables_by_name: HashMap::new(),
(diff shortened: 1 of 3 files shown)


src/catalog.rs
```diff
@@ -0,0 +1,126 @@
...
+            })?;
+
+        let table = versions.get(&version).ok_or_else(|| {
+            illegal_operation_raw(format!("Symbol table with name: {} does not exist", name))
+        })?;
+        Ok(table.to_owned())
+    }
+}
+
+impl MutableCatalog for SimpleCatalog {
```
zslayton, line 69: The `MutableCatalog` trait feels a bit off to me. Consider a few possible implementations of `Catalog`:
* `MapCatalog`: entirely in memory.
* `FileSystemCatalog`: looks for SST files with a given name within a given directory.
* `DynamoDbCatalog`: looks for SST documents in a Dynamo table.

`put_table` is a good fit for `MapCatalog` but doesn't fit as well for the other two. They could produce IO errors, and a user might wish to do bulk uploads of SSTs to a DynamoDB table rather than individual PUTs.

I propose turning this `impl MutableCatalog for SimpleCatalog` into just `impl SimplCatalog`; we can standardize a trait for reading from the `Catalog` while allowing each implementation to decide when and how to build the catalog up.

src/shared_symbol_table.rs
```diff
@@ -0,0 +1,54 @@
...
+pub struct SharedSymbolTable {
+    name: String,
+    version: usize,
+    imports: Vec<Option<String>>,
+}
+
+impl SharedSymbolTable {
+    pub fn new(
+        name: String,
+        version: Option<usize>,
```
zslayton, line 16: I think this can just be `version: usize`. Finding an `import` struct that doesn't specify a `version` field happens at a lower layer than creating a `SharedSymbolTable` instance in memory; the reader can handle defaulting the `version` to 1 if it's missing.

src/shared_symbol_table.rs
```diff
@@ -0,0 +1,54 @@
...
+        })
+    }
+
+    /// Returns the version of this [SharedSymbolTable]
+    pub fn version(&self) -> usize {
+        self.version
+    }
+
+    /// Returns the name of this [SharedSymbolTable]
+    pub fn name(&self) -> &String {
```
zslayton, line 51: Consider returning a `&str` instead. Returning a `&String` is effectively the same thing, but leaks an implementation detail: it reveals that the `SharedSymbolTable` stores its name as a `String` and not a `Rc<str>`, `SmolString`, or whatever else.

src/catalog.rs
```diff
@@ -0,0 +1,126 @@
...
+        Self {
+            tables_by_name: HashMap::new(),
+        }
+    }
+}
+
+impl Catalog for SimpleCatalog {
+    fn get_table(&self, name: &str) -> IonResult<SharedSymbolTable> {
+        if name.is_empty() {
+            return illegal_operation("Can not get a symbol table where the name is empty");
```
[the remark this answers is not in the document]
desaikd: Sure. I will change this for all the errors.

src/catalog.rs
```diff
@@ -0,0 +1,126 @@
...
+    /// If a table with given name and version doesn't exists then it returns an error
+    fn get_table_with_version(&self, name: &str, version: usize) -> IonResult<SharedSymbolTable>;
+}
+
+pub trait MutableCatalog {
+    /// Adds a Shared Symbol Table with name into the Catalog
+    fn put_table(&mut self, table: SharedSymbolTable);
+}
+
+struct SimpleCatalog {
```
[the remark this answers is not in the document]
desaikd: sure. I tried to keep it close to `ion-java` but I can definitely change this.

src/catalog.rs
desaikd: Makes sense. I  will merge the `MutableCatalog` functionality into `SimpleCatalog` for this.

src/shared_symbol_table.rs
desaikd: Understood. I will change this with next commit.

src/catalog.rs
```diff
@@ -0,0 +1,126 @@
...
+/// Catalog is a collection of Shared Symbol Tables
+/// for more information on [Catalog]: https://amzn.github.io/ion-docs/docs/symbols.html#the-catalog
+pub trait Catalog {
+    /// Returns the Shared Symbol Table with given table name
+    /// If a table with the given name doesn't exists or if the table name is an empty string
+    /// then returns an error
+    fn get_table(&self, name: &str) -> IonResult<SharedSymbolTable>;
+    /// Returns the Shared Symbol Table with given table name and version
+    /// If a table with given name and version doesn't exists then it returns an error
+    fn get_table_with_version(&self, name: &str, version: usize) -> IonResult<SharedSymbolTable>;
```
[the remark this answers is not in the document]
desaikd: `get_table_with_name` returns the latest version of the table based on given table name.

src/catalog.rs
```diff
@@ -3,77 +3,75 @@ use crate::result::{illegal_operation, illegal_operation_raw};
...
-/// Catalog is a collection of Shared Symbol Tables
-/// for more information on [Catalog]: https://amzn.github.io/ion-docs/docs/symbols.html#the-catalog
+/// A Catalog is a collection of Shared Symbol Tables.
+/// For more information about the concept of a catalog,
+/// see [the `symbols` section of the specification](https://amzn.github.io/ion-docs/docs/symbols.html#the-catalog).
 pub trait Catalog {
     /// Returns the Shared Symbol Table with given table name
     /// If a table with the given name doesn't exists or if the table name is an empty string
     /// then returns an error
+    /// If a table with multiple versions exists for the given name then it will return the latest version of table
```
zslayton, line 13: I think the spec says that if no version is specified, `1` is assumed. No `TODO` here, just noting that this behavior is different from the use case in the spec.

I think that for our initial implementation, we can have `ion-rust` require an exact version match. The world of partial-matching on imports is pretty complex.

PR #451, opened 2022-12-07
Stop impl Display for Element from panicking on typed nulls

### LINKED ISSUE #450: `impl Display for Element` panics for typed nulls
For example, this test:
```rs
#[test]
fn print_a_typed_null() {
    let value: Element = element_reader().read_one("null.bool".as_bytes()).unwrap();
    println!("{}", value)
}
```
results in this output:
```
thread 'print_a_typed_null' panicked at 'called `Option::unwrap()` on a `None` value'
```

It's caused by this line:
https://github.com/amzn/ion-rust/blob/aa33f0fb8b5e4965d56e48fd337b813dded32278/src/value/owned.rs#L466

It's probably a good idea to also check why it wasn't caught by any of our existing tests (particularly https://github.com/amzn/ion-rust/blob/main/tests/element_test_vectors.rs).

popematt:
*Issue #, if available:*

Fixes #450 

*Description of changes:*

* Added a test based on the "good" test vectors in `ion-tests` to ensure that `Display` is consistent with the default `TextWriter`.
* Fixes the implementation of `Display` for `Element` so that it doesn't try to unwrap `None`.
* Updates `format_struct()`, `format_list()`, and `format_sexp()` to be consistent with the output of a `TextWriter` (I removed some extra whitespace inside the containers), and updates the unit tests accordingly.

### DIFF
--- a/src/text/text_formatter.rs
+++ b/src/text/text_formatter.rs
@@ -405,7 +405,7 @@ impl<'a, W: std::fmt::Write> IonValueFormatter<'a, W> {
     }
 
     pub(crate) fn format_struct(&mut self, value: &Struct) -> IonResult<()> {
-        write!(self.output, "{{ ")?;
+        write!(self.output, "{{")?;
         let mut peekable_itr = value.iter().peekable();
         while let Some((field_name, field_value)) = peekable_itr.next() {
             self.format_symbol(field_name.text().unwrap())?;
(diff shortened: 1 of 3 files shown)


desaikd (approved):
Nice 🚀

zslayton (approved):

jobarr-amzn (approved):

Issue #287, opened 2021-06-14
IonHash completness


marcbowes:
Tracking issue for finishing an initial implementation of IonHash.

Maintainer note: where possible, create and ref. subissues.

At the time of writing, we have a mostly complete implementation of the "bones" of ion-hash. This means that, for all types, we can compute T, Q and representation. So, what's left?

First, bugs! We have several of these:

- [ ] Timestamp fractional seconds always use nanoseconds
- [ ] Decimals don't support negative zero

Symbols and annotations aren't implemented correctly:

- [ ] Symbols are read as strings and don't resolve correctly
- [ ] SID0 completely fails to load

Then, some code cleanup issues:

- [ ] Lots of `&mut sink` output variables are begging for a better abstraction
- [ ] Use the ion-hash-test tests directly instead of a curated subset

Out of scope: making it work with reader & writer traits. I think an Element-based hasher is a reasonable MVP.

PR #453, opened 2022-12-20
Completes impl of Ion Hash algorithm and bumps to latest ion-rust version

### LINKED ISSUE #263: Use ion-hash-tests verbatim
Due to #219, the current tests in ion-hash use a subset of the full test suite with the unloadable elements removed. Once this bug is fixed, we should go back and use the tests from the submodule directly.

### LINKED ISSUE #420: Remove `ion-hash`'s dependency on ion-c-sys
`ion-hash` depends on the `ion-c-sys` binary writer implementation to encode its values for hashing. Following PR #418, `ion-rust` no longer offers an optional dependency on `ion-c-sys`. As part of that PR, `ion-hash` was modified to depend on the last published version of `ion-rust` that offered that optional dependency.

We need to update `ion-hash` to use the native Rust writer/encoders in `ion-rust`. Then we can point its `Cargo.toml` dependency on `ion-rust` back to `path=".."`.

Having tried this briefly, there is at least one encoding difference between the ion-c and Rust encoders that will need to be investigated. Per [the `timestamp` section of the ion-hash spec](https://amzn.github.io/ion-hash/docs/spec.html#6-timestamp):

> The fraction_exponent and fraction_coefficient denote the fractional seconds of the timestamp as a decimal value. The fractional seconds value is coefficient * 10 ^ exponent. It must be greater than or equal to zero and less than 1. A coefficient of zero shall not be encoded, as zero is the default. Fractions whose coefficient is zero and exponent is greater than -1 are illegal and shall not be encoded. The following hex sequences are the representation field of valid Ion binary timestamps, and are all equivalent:
> 
> ```
> 80 0F D0 81 81 80 80 80       // 2000-01-01T00:00:00Z with no fractional seconds
> 80 0F D0 81 81 80 80 80 80    // The same instant with 0d0 fractional seconds and implicit zero coefficient
> 80 0F D0 81 81 80 80 80 80 00 // The same instant with 0d0 fractional seconds and explicit zero coefficient
> 80 0F D0 81 81 80 80 80 C0    // The same instant with 0d-0 fractional seconds
> 80 0F D0 81 81 80 80 80 81    // The same instant with 0d1 fractional seconds
> ```
> 
> For hashing purposes, minimal/compact form is required, so the representation of all of the above timestamps shall be 80 0F D0 81 81 80 80 80.

The binary `timestamp` encoder in Rust delegates to the `decimal` encoder and does not always meet this ion-hash requirement. 

The `decimal` encoder preserves the difference between `0d5` and `0d7`, which are different decimals, while the `timestamp` encoder doesn't care about the difference between them and must define its own encoding logic to compact them. The current implementation produces valid `timestamp`s, but invalid hashes.

We should modify the `timestamp` encoder to check for these 'compact form' cases and handle them separately rather than delegating them.

There may be other issues, but that is the one that currently presents itself.

popematt:
**Issue #, if available:**

Fixes #263 
Fixes #420 
Progress on #287 (maybe complete?)

**Description of changes:**

* Updates to use the refactored `IonElement` API that was introduced in `ion-rs=0.14.0`.
* Removes the dependency on `ion-c-sys`. (I think that might mean we can remove `ion-c-sys` entirely from the repo, but I'm not sure.)
* Changes usages of `strukt` to `struct_` to match the convention used elsewhere in this GitHub repo
* Adds support for hashing annotated values
* Adds support for symbols with unknown text
* Updates ion_hash_tests.rs
  * Use `ion-hash-test` test cases directly
  * Collect test failures and report multiple instead of panicking on the first failed test case
  * Handle test names properly
* Adds hard coded test cases for the ion-hash-test cases that cannot be loaded correctly by ion-rust yet.

To be clear, I've finished implementing the algorithm, but that doesn't mean this library is finished. What's left to do? Well, here are a few things that I think are required:
* I think the overall architecture of this implementation is unnecessarily complex and can be simplified.
* Also might need to add support for MD5 and other common hashing functions.
* May need to revisit whether this should be a separate crate or just a feature of the `ion-rs` crate (so that they don't get out of sync again in the future).
* We should remove the dependency on `sha2`, or at least gate it behind a feature flag.

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.*


ion-hash/tests/ion_hash_tests.rs
```diff
@@ -257,3 +289,76 @@ fn seq_to_bytes<E: Element>(elem: &E) -> Vec<u8> {
...
+    fn test_unknown_field_name_text() {
+        let fields_itr = vec![(Symbol::unknown_text(), Element::from(1))];
+        let struct_ = owned::Struct::from_iter(fields_itr);
+        let element = Element::from(struct_);
+
+        let digest = IdentityDigest::hash_element(&element).unwrap();
+
+        let expected_string = format!(
+            "{:02x?}",
+            &[0x0b, 0xd0, 0x0c, 0x0b, 0x71, 0x0c, 0x0e, 0x0c, 0x0b, 0x20, 0x01, 0x0c, 0x0e, 0x0e]
```
jobarr-amzn, line 345: I wrote this out to understand it and double-check:
```rust
// s(struct) -> B || TQ || escape(concat(sort(H(field1), H(field2), ..., H(fieldn)))) || E
// s(field)  -> h(s(fieldname) || s(fieldvalue))
// s(scalar) -> B || TQ || escape(representation) || E

&[
  0x0b, 0xd0, // B(struct) || TQ(struct) || escape(
  0x0c, 0x0b, //   B(fieldname1/symbol)
  0x71,       //     TQ(symbol) empty symbol representation
  0x0c, 0x0e, //   E(fieldname1/symbol)
  0x0c, 0x0b, //   B(fieldvalue1/int)
  0x20, 0x01, //     TQ(int) || representation(1)
  0x0c, 0x0e, //   E(fieldvalue1/int) )
  0x0e        // E(struct)
]
```

jobarr-amzn (approved):
Thanks for this :)

ion-hash/tests/ion_hash_tests.rs
popematt: Yeah, I think that's it. I just copied these expected hashes directly from the corresponding test case in `ion-hash-test`, so it didn't occur to me to add an explanatory note like that.

jobarr-amzn: I wonder if it's possible/helpful to make an expression (macros?) that would look something like:
```rust
block(tq(STRUCT), escape(block(tq(EMPTY_SYMBOL)), block(tq(INTEGER), int_repr(1))))
```

PR #458, opened 2023-01-09
Format Agnostic RawReader Pass

nirosys:
*Issue #, if available:* N/A

*Description of changes:*
This PR adds a format agnostic Position for error reporting of locations within an Ion document. For both text and binary formats the Position holds a byte offset. Positions within text documents can also add an optional Line and Column to the Position for better error reporting in the text format.

The addition of Position allowed the text only error `IncompleteText` to be removed and the RawTextReader can now use the `Incomplete` error like RawBinaryReader.

This PR also adjusts the handling of incomplete data in the RawBinaryReader's `step_out`. Prior to this PR if `step_out` was called without enough data in the buffer, the function would return `Ok(())` and set the reader's state to `Skipping` with the remaining bytes that could not be skipped. This PR continues to consume, and set the `Skipping` state, but returns an `Incomplete` error rather than `Ok(())` in order to match the behavior of the RawTextReader.

---

### DIFF
--- a/src/result.rs
+++ b/src/result.rs
@@ -5,64 +5,51 @@ use std::{fmt, io};
 use thiserror::Error;
 
 /// Represents the position, as line and column, within a text format ion document.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub enum TextPosition {
     /// No text position available; implies binary format is being read.
     None,
     /// The position represented as line and column.
-    LineColumn(usize, usize),
+    LineAndColumn(usize, usize),
 }
 
-#[derive(Clone, Debug)]
+/// Position represents the location within an ion document where an error has been
+/// identified. For all formats `byte_offset` will contain the number of bytes into the document
+/// that have been processed prior to encountering the error. When working with the text format,
+/// `line_column` will be updated to contain the line and column as well.
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Position {
     pub byte_offset: usize,
     pub line_column: TextPosition,
 }
 
 impl Position {
-    fn with_offset(offset: usize) -> Self {
+    pub fn with_offset(offset: usize) -> Self {
         Position {
             byte_offset: offset,
             line_column: TextPosition::None,
         }
     }
 
-    fn with_line_info(offset: usize, line: usize, column: usize) -> Self {
+    pub fn with_text_position(&self, line: usize, column: usize) -> Self {
         Position {
-            byte_offset: offset,
-            line_column: TextPosition::LineColumn(line, column),
-        }
-    }
-}
-
-impl PartialEq for Position {
-    fn eq(&self, other: &Self) -> bool {
-        if other.byte_offset != self.byte_offset {
-            false
-        } else {
-            use TextPosition::*;
-            match (&other.line_column, &self.line_column) {
-                (None, None) => true,
-                (None, LineColumn(_, _)) | (LineColumn(_, _), None) => false,
-                (LineColumn(ol, oc), LineColumn(sl, sc)) => ol == sl && oc == sc,
-            }
+            line_column: TextPosition::LineAndColumn(line, column),
+            ..*self
         }
     }
 
-    fn ne(&self, other: &Self) -> bool {
-        !self.eq(other)
+    fn has_text_position(&self) -> bool {
+        self.line_column != TextPosition::None
     }
 }
-
-impl Eq for Position {}
 
 impl Display for Position {
-    // Formats the position based on whether we have a LineColumn or not.
+    // Formats the position based on whether we have a LineAndColumn or not.
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), Error> {
         use TextPosition::*;
         match &self.line_column {
             None => write!(f, "{}", self.byte_offset),
-            LineColumn(line, column) => {
+            LineAndColumn(line, column) => {
                 write!(f, "{} ({}:{})", self.byte_offset, line, column)
             }
         }
@@ -182,25 +169,12 @@ pub fn incomplete_data_error_raw(label: &'static str, offset: usize) -> IonError
     }
 }
 
-pub fn incomplete_text_error<T>(
-    label: &'static str,
-    byte_offset: usize,
-    line: usize,
-    column: usize,
-) -> IonResult<T> {
-    Err(incomplete_text_error_raw(label, byte_offset, line, column))
+pub fn incomplete_text_error<T>(label: &'static str, position: Position) -> IonResult<T> {
+    Err(incomplete_text_error_raw(label, position))
 }
 
-pub fn incomplete_text_error_raw(
-    label: &'static str,
-    byte_offset: usize,
-    line: usize,
-    column: usize,
-) -> IonError {
-    IonError::Incomplete {
-        label,
-        position: Position::with_line_info(byte_offset, line, column),
-    }
+pub fn incomplete_text_error_raw(label: &'static str, position: Position) -> IonError {
+    IonError::Incomplete { label, position }
 }
 
 /// A convenience method for creating an IonResult containing an IonError::DecodingError with the
--- a/src/result.rs
+++ b/src/result.rs
@@ -4,52 +4,66 @@ use std::{fmt, io};
 
 use thiserror::Error;
 
-/// Represents the position, as line and column, within a text format ion document.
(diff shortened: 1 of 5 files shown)


src/result.rs
```diff
@@ -1,9 +1,74 @@
...
+
+    fn with_line_info(offset: usize, line: usize, column: usize) -> Self {
+        Position {
+            byte_offset: offset,
+            line_column: TextPosition::LineColumn(line, column),
+        }
+    }
+}
+
+impl PartialEq for Position {
```
popematt, line 38: Is there a reason you can't derive `PartialEq` (and `Eq` and `Hash`) for `Position` (and `TextPosition`)?

src/result.rs
```diff
@@ -1,9 +1,74 @@
...
+            }
+        }
+    }
+
+    fn ne(&self, other: &Self) -> bool {
+        !self.eq(other)
+    }
+}
+
+impl Eq for Position {}
```
popematt, line 57: In Java, it's a best practice to implement `equals()` and `hashCode()`, and it seems like a principle that would be language-independent. Should there also be an `impl Hash for Position` here?

src/result.rs
```diff
@@ -1,9 +1,74 @@
...
-use std::fmt::{Debug, Error};
+use std::fmt::{Debug, Display, Error};
 use std::{fmt, io};
 
 use thiserror::Error;
 
+/// Represents the position, as line and column, within a text format ion document.
+#[derive(Clone, Debug)]
+pub enum TextPosition {
+    /// No text position available; implies binary format is being read.
```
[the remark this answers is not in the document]
nirosys: There is no true enforcement mechanism, it is entirely up to the site where the Position is created. Right now it is only used in errors and the behavior is maintained through the use of format specific error creation functions. Nothing is stopping code from being added that deviates from that behavior however.

src/result.rs
```diff
@@ -1,9 +1,74 @@
...
+    None,
+    /// The position represented as line and column.
+    LineColumn(usize, usize),
```
zslayton, line 13: Subjective: I have a mild preference for `LineAndColumn` over `LineColumn`.

src/result.rs
```diff
@@ -1,9 +1,74 @@
...
+
+impl Position {
+    fn with_offset(offset: usize) -> Self {
+        Position {
+            byte_offset: offset,
+            line_column: TextPosition::None,
+        }
+    }
+
+    fn with_line_info(offset: usize, line: usize, column: usize) -> Self {
```
zslayton, line 30: This is the only place the phrase `line_info` appears, which makes me wonder if this should be the slightly wordier `with_line_and_column` or perhaps `with_text_position`?

src/text/non_blocking/raw_text_reader.rs
```diff
@@ -343,7 +345,12 @@ impl<A: AsRef<[u8]>> RawTextReader<A> {
                     self.process_stream_item(item)
                 } else {
                     // If we are not at the end of the stream, we need to get more data.
-                    incomplete_text_error(self.buffer.lines_loaded(), self.buffer.line_offset())
+                    incomplete_text_error(
+                        "text",
+                        self.buffer.bytes_consumed(),
+                        self.buffer.lines_loaded(),
+                        self.buffer.line_offset(),
```
zslayton, line 352: A thought you can take or leave: maybe it makes sense to give `TextBuffer` a convenience method since it's the owner of all of these bits of information?

```rust
text_buffer.incomplete_text_error("text") // or something?
```

src/result.rs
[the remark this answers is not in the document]
popematt: I agree with Zack, but I also think we could just reduce the visibility of this to `pub(crate)` given that we can always increase the visibility later without it being a breaking change.

src/result.rs
```diff
@@ -1,9 +1,74 @@
 use std::convert::From;
-use std::fmt::{Debug, Error};
+use std::fmt::{Debug, Display, Error};
 use std::{fmt, io};
 
 use thiserror::Error;
 
+/// Represents the position, as line and column, within a text format ion document.
+#[derive(Clone, Debug)]
+pub enum TextPosition {
```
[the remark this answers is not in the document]
nirosys: I didn't make any changes here with the latest commit. I'm trying to consider what type of information could be added. I've had debugability on the brain so I considered more diagnostic information such as depth, or other data related to positioning but couldn't come up with anything text specific.

Are there any ideas for future functionality that you can think of?

Maybe taking `TextPosition` away, and adding an `Option<(usize, usize)>` field, on the `Position` struct for line and column, and then just extend Position if/when we need to?

src/result.rs
nirosys: Me too, updated in latest.

src/result.rs
nirosys: Added `with_text_position` as a builder-style method in the latest commit.

src/result.rs
nirosys: In the latest commit I got rid of the manual implementations and added `fn has_text_position(&self) -> bool` to help out with ease of access.

After thinking more about Zack's comment about converting TextPosition to a struct, I'm kind of leaning towards wanting to eliminate TextPosition entirely and adding line and column info as an `Option<(usize, usize)>` which would then provide the niceties of Option.

src/text/non_blocking/raw_text_reader.rs
nirosys: I had considered something similar but didn't feel like the text buffer was the place to generate the error, since the reader was the one who really identified it.

In the latest commit I added a `get_position` function to `TextBuffer` and have it return a `Position` with all of that data. I also updated `incomplete_text_error*` to take a Position instead of individual values, which simplifies this a little more and encapsulates that positioning info a little better.

nirosys (commented):
I had a few comments from last week that I completely forgot to hit the actual "Finish your review" button on. 😶‍🌫️ 

I've submitted a new commit which should address most things discussed. I have comments inline with more info. Sorry that took so long, thank you both for the feedback.

src/result.rs
```diff
@@ -5,64 +5,51 @@ use std::{fmt, io};
...
 
-#[derive(Clone, Debug)]
+/// Position represents the location within an ion document where an error has been
+/// identified. For all formats `byte_offset` will contain the number of bytes into the document
+/// that have been processed prior to encountering the error. When working with the text format,
+/// `line_column` will be updated to contain the line and column as well.
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Position {
     pub byte_offset: usize,
     pub line_column: TextPosition,
```
zslayton, line 23: I'm inclined to provide accessors instead of `pub` fields to give us:
1. The ability to prevent users from changing field values after we hand them a `Position`.
2. The ability to rename fields or change their data type down the road.

I'd be good with `pub(crate)` for dev convenience, though. What do you think?

src/result.rs
```diff
@@ -5,64 +5,51 @@ use std::{fmt, io};
...
         }
     }
 
-    fn with_line_info(offset: usize, line: usize, column: usize) -> Self {
+    pub fn with_text_position(&self, line: usize, column: usize) -> Self {
         Position {
-            byte_offset: offset,
-            line_column: TextPosition::LineColumn(line, column),
+            line_column: TextPosition::LineAndColumn(line, column),
+            ..*self
```
zslayton, line 37: Nice

src/result.rs
popematt: Can we make this whole struct `pub(crate)` for now, not just the fields? I don't recall seeing any usages that would force it to be `pub`.

zslayton: With the `IonError` type offering accessors for byte offset, line, and column? If so, I'd be ok with that.

nirosys: Thoughts on converting IonError to a struct and moving the individual errors (Incomplete, IoError, etc) into an ErrorKind, similar to the way `std::io::Error` works?

It's a pretty big change, imo, but it would let us move things like position out of the enum variants and instead require position info for any error creation, rather than relying on the variants to include it. That would make accessors make more sense to me, since the information would be expected for each error variant.

Thoughts?

zslayton:
Sounds good to me! Let's open an issue for it and tackle it in a separate
PR.

src/result.rs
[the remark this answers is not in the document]
nirosys: Cool, I'll make the changes to Position, and remove TextPosition shortly.

zslayton (approved):
Looks good! There's one `clippy` suggestion to incorporate before merging.

popematt (approved):

PR #463, opened 2023-01-25
Fix permissions for clippy-check

nirosys:
*Issue #, if available:* N/A

*Description of changes:*
The clippy-check github action has been failing with the move to the amazon-ion org. This seems to be due to default permissions that changed between the amzn org, and our new org. Documentation seems to say that we should be able to allow write permissions via the workflow config, so this PR is an attempt at doing that.

---


desaikd (approved):
Workflow check have passed for the PR 🚀

Issue #468, opened 2023-02-21
Some generated unit tests ignore failure case


zslayton:
cc @popematt 

A set of unit tests were introduced in PR #451 with [this method signature](https://github.com/amazon-ion/ion-rust/pull/451/files#diff-f4d869788fbfef40174e562fa77827feb9d63b4fc212752344afcf780be9e1f7R470-R472):

```rust
    #[test_resources("ion-tests/iontestdata/good/**/*.ion")]
    #[test_resources("ion-tests/iontestdata/good/**/*.10n")]
    fn test_to_string(file_name: &str) -> IonResult<()> {
      // ...
    }
```

Unfortunately, due to a limitation/bug in the `test_generator` crate that provides the `test_resources` macro, the expanded form of those tests looks like:

```rust
#[allow(non_snake_case)]
fn test_to_string_ion_tests_iontestdata_good_testfile30_ion() {
    test_to_string("ion-tests/iontestdata/good/testfile30.ion".into());
}
```

Notice that `test_to_string` is called but the `Result` it returns is ignored. `rustc` emitted a warning about this, but it was buried in the test output making it easy to miss:

![image](https://user-images.githubusercontent.com/611616/220382462-5defd126-264f-4666-a137-24218eff2a74.png)

Build/test output can be seen [here](https://github.com/amazon-ion/ion-rust/pull/451/checks#step:4:386).

Wrapping the test with an assertion like so:

```rust
    #[test_resources("ion-tests/iontestdata/good/**/*.ion")]
    #[test_resources("ion-tests/iontestdata/good/**/*.10n")]
    fn test_to_string(file_name: &str) {
        match test_to_string_result(file_name) {
            Ok(_) => {}
            Err(e) => panic!("{:?}", e),
        }
    }

    fn test_to_string_result(file_name: &str) -> IonResult<()> {
        // ... original method impl ...
    }
```

causes several errors/failures to be surfaced. I have not investigated the individual errors thoroughly; at least a few will be addressed by @desaikd's work on loading shared symbol tables.

PR #471, opened 2023-02-21
Removes uses of deprecated methods from `chrono`

zslayton:
Starting in v0.4.23, [`chrono` deprecated a number of methods](https://github.com/chronotope/chrono/pull/827) that would panic if the parameters were out of bounds (e.g. `month=400`) in favor of new versions that would return a `None` if the inputs were bad.

This PR replaces usages of the deprecated methods with their safe counterparts. Unfortunately, the safe counterparts are rather more verbose. However, most of the impacted code was in unit tests.

### DIFF
--- a/src/types/timestamp.rs
+++ b/src/types/timestamp.rs
@@ -134,6 +134,20 @@ fn first_n_digits_of(num_digits: u32, value: u32) -> u32 {
     value / 10u32.pow(total_digits - num_digits)
 }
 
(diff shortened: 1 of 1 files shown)


desaikd (approved):
Looks good! There are still some `clippy` warnings though.

zslayton:
> There are still some clippy warnings though.

Most of them are being tracked in #468, one is a bug in `clippy` itself being tracked in #472. Thanks for the review!

PR #473, opened 2023-02-27
Fix silent test failures caused by unimplemented features

### LINKED ISSUE #468: Some generated unit tests ignore failure case

popematt:
**Issue #, if available:**

Fixes #468 

**Description of changes:**

Updates `impl_display_for_element_tests::test_to_string` so that it doesn't return a `Result`. All intermediate results must now be dealt with in this function.

Adds a skip list for failing test cases. All currently skipped tests cases are due to unimplemented functionality.

### DIFF
--- a/tests/element_test_vectors.rs
+++ b/tests/element_test_vectors.rs
@@ -393,7 +393,7 @@ mod impl_display_for_element_tests {
     use ion_rs::TextWriterBuilder;
     use std::fs::read;
 
-    const TO_STRING_SKIP_LIST: [&'static str; 12] = [
+    const TO_STRING_SKIP_LIST: &[&'static str] = &[
         // These tests have shared symbol table imports in them, which the Reader does not
         // yet support.
         "ion-tests/iontestdata/good/subfieldInt.ion",
--- a/tests/element_test_vectors.rs
+++ b/tests/element_test_vectors.rs
@@ -390,26 +390,51 @@ mod impl_display_for_element_tests {
     use super::*;
     use ion_rs::value::native_writer::NativeElementWriter;
     use ion_rs::value::reader::element_reader;
-    use ion_rs::{IonResult, TextWriterBuilder};
+    use ion_rs::TextWriterBuilder;
     use std::fs::read;
+
+    const TO_STRING_SKIP_LIST: [&'static str; 12] = [
+        // These tests have shared symbol table imports in them, which the Reader does not
+        // yet support.
+        "ion-tests/iontestdata/good/subfieldInt.ion",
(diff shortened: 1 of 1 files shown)


tests/element_test_vectors.rs
```diff
@@ -390,26 +390,51 @@ mod impl_display_for_element_tests {
     use super::*;
     use ion_rs::value::native_writer::NativeElementWriter;
     use ion_rs::value::reader::element_reader;
-    use ion_rs::{IonResult, TextWriterBuilder};
+    use ion_rs::TextWriterBuilder;
     use std::fs::read;
 
+    const TO_STRING_SKIP_LIST: [&'static str; 12] = [
```
zslayton, line 396: You can get around having to state the length of the array (and updating it with each change) by using a reference:

```suggestion
    const TO_STRING_SKIP_LIST: &[&'static str] = &[
```

zslayton (approved):

Issue #472, opened 2023-02-21
Clippy warns about use of a redundant closure


zslayton:
`clippy` currently emits a warning about a redundant closure:

![image](https://user-images.githubusercontent.com/611616/220436637-92a731b1-aa40-482d-a24e-0cc77eff69dd.png)

However, if you accept its suggestion and remove the closure, the resulting code causes a compiler error about lifetime inference. This appears to be [a bug in `clippy`](https://github.com/rust-lang/rust-clippy/issues/10378).

PR #475, opened 2023-03-07
Removes IonElement trait, ElementRef impl

zslayton:
`IonElement` provides a trait abstraction over two flavors of `Element` API:

1. The 'borrowed' API, where each `ElementRef` has a lifetime tied to the lifetime of the source data.

2. The 'owned' API, where each `Element` is self-contained, with no connection to the source data.

While the borrowed API was intended to be lighter weight, the current implementation materializes nearly the same amount of data as the owned API. Collections (`list`, `sexp`, `struct`) are heap-allocated. Of the scalar types, only `string`s and `lob`s avoid any allocation in the borrowed API.

The `IonElement` trait was designed before GATs were stabilized and had some rough edges caused by the inherent difficulty in abstracting over lifetime constraints in traits in the absence of GATs.

In practice, users that need an `Element` API are exclusively using the `owned` variant.

In light of these observations and in anticipation of the upcoming v1.0 release, this PR removes both the `IonElement` trait and the `borrowed` implementation, leaving only the owned `Element` API. These constructs can be reintroduced in the future if and when they are needed.

Follow-on pull requests will:
* Offer builder APIs to construct `Struct`, `List`, and `SExpression` more ergonomically.
* Merge `ElementReader`'s functionality into `IonReader`.
* Merge `ElementWriter's` functionality into `IonWriter`.
* Restore the documentation and examples to reflect the new APIs.


src/value/owned.rs
```diff
@@ -514,6 +491,15 @@ impl From<BigInt> for Element {
     }
 }
 
+impl From<Integer> for Element {
+    fn from(integer_val: Integer) -> Self {
+        match integer_val {
+            Integer::I64(i64_int) => i64_int.into(),
+            Integer::BigInt(big_int) => big_int.into(),
+        }
```
zslayton, line 499: 🗺️ We already had `From<i64>` (e.g. `16.into()`) and `From<BigInt>`, adding `From<Integer>` allows simple conversions in places where we've already constructed an `Integer` enum.

src/value/owned.rs
```diff
@@ -538,6 +524,12 @@ impl From<bool> for Element {
     }
 }
 
+impl From<&str> for Element {
+    fn from(string_val: &str) -> Self {
+        Value::String(string_val.to_owned()).into()
+    }
+}
+
```
zslayton, line 532: 🗺️ Allows `let s: Element = "hello".into();`

src/value/owned.rs
```diff
@@ -583,14 +575,8 @@ impl<'a, T: 'a> Iterator for ElementDataIterator<'a, T> {
...
 
-impl IonElement for Element {
-    type SymbolToken = Symbol;
-    type Sequence = Sequence;
-    type Struct = Struct;
-    type Builder = Element;
-    type AnnotationsIterator<'a> = SymbolsIterator<'a>;
-
-    fn ion_type(&self) -> IonType {
+impl Element {
```
zslayton, line 578: 🗺️ For now I've converted the functionality that was associated with the `IonElement` trait into concrete methods on the `Element` type. I expect to refactor these in a follow-on PR.

src/value/owned.rs
```diff
@@ -610,89 +596,100 @@ impl IonElement for Element {
...
-    fn as_integer(&self) -> Option<&Integer> {
+    pub fn as_integer(&self) -> Option<&Integer> {
         match &self.value {
             Value::Integer(i) => Some(i),
             _ => None,
         }
     }
 
-    fn as_f64(&self) -> Option<f64> {
+    pub fn as_float(&self) -> Option<f64> {
```
zslayton, line 628: 🗺️ I renamed this to reflect the Ion type (`float`) rather than the Rust type (`f64`) as we do in the other methods.

src/value/owned.rs
```diff
@@ -610,89 +596,100 @@ impl IonElement for Element {
...
     }
 
-    fn as_str(&self) -> Option<&str> {
+    pub fn as_text(&self) -> Option<&str> {
         match &self.value {
             Value::String(text) => Some(text),
             Value::Symbol(sym) => sym.text(),
             _ => None,
         }
     }
```
zslayton, line 655: 🗺️ Rather than have a single `as_string()` method that returns text for either `String` or `Symbol` elements, we now have `as_string()` for `String` elements, `as_symbol()` for `Symbol` elements, and `as_text()` for the old behavior that doesn't care which text type the `Element` is.

ion-hash/src/representation.rs
```diff
@@ -68,7 +53,7 @@ where
             Some(Integer::I64(v)) => match v {
                 0 => {}
                 _ => {
-                    let magnitude = v.abs() as u64;
+                    let magnitude = v.unsigned_abs();
```
zslayton, line 56: 🗺️ Incorporating a `clippy` suggestion.

ion-hash/src/representation.rs
```diff
@@ -110,7 +95,7 @@ where
                     return Ok(());
                 }
 
-                self.update_escaping(&v.to_be_bytes());
+                self.update_escaping(v.to_be_bytes());
```
zslayton, line 98: 🗺️ Incorporating a `clippy` suggestion.

ion-hash/src/representation.rs
```diff
@@ -154,7 +136,7 @@ where
 
     fn write_repr_string(&mut self, value: Option<&str>) -> IonResult<()> {
         match value {
-            Some(s) if s.len() > 0 => self.update_escaping(s.as_bytes()),
+            Some(s) if !s.is_empty() => self.update_escaping(s.as_bytes()),
```
zslayton, line 139: 🗺️ Incorporating a `clippy` suggestion.

ion-hash/src/representation.rs
```diff
@@ -163,17 +145,14 @@ where
 
     fn write_repr_blob(&mut self, value: Option<&[u8]>) -> IonResult<()> {
         match value {
-            Some(bytes) if bytes.len() > 0 => self.update_escaping(bytes),
+            Some(bytes) if !bytes.is_empty() => self.update_escaping(bytes),
```
zslayton, line 148: 🗺️ Incorporating a `clippy` suggestion.

src/value/owned.rs
```diff
@@ -610,89 +596,114 @@ impl IonElement for Element {
...
+
+    pub fn as_symbol(&self) -> Option<&Symbol> {
         match &self.value {
             Value::Symbol(sym) => Some(sym),
             _ => None,
         }
     }
 
-    fn as_bool(&self) -> Option<bool> {
+    pub fn as_boolean(&self) -> Option<bool> {
```
zslayton, line 671: 🗺️ Renaming to reflect the Ion type (`boolean`) rather than the Rust type (`bool`).

src/value/owned.rs
```diff
@@ -610,89 +596,114 @@ impl IonElement for Element {
...
         match &self.value {
             Value::Boolean(b) => Some(*b),
             _ => None,
         }
     }
 
-    fn as_bytes(&self) -> Option<&[u8]> {
+    pub fn as_lob(&self) -> Option<&[u8]> {
```
zslayton, line 678: 🗺️ As with the `string`/`symbol`/`text`, I've broken this into `clob`/`blob`/`lob`.

zslayton (commented):
🗺️ PR tour

desaikd (approved):
Looks good 🚀 
There are still some `clippy` warnings for `element_stream_reader` though.

zslayton:
> There are still some `clippy` warnings for `element_stream_reader` though.

Those are related to #472.

PR #477, opened 2023-03-10
Make Display for Decimal more human-friendly

jobarr-amzn:
*Description of changes:*

Before this all Decimals would be displayed as e.g. 37d-2 instead of 0.37. After this decimals are displayed as `ABC.DEF`, `ABCd3` or `0.000ABC` unless the number of padding zeroes grows large, in which case it will fall back to `A.BCd-E` for some exponent.

This also prefers to quantify the scale of larger numbers, e.g. `ABCDEFGd0` will be written as A.BCDEFGd6 instead of `ABCDEFG.`

"large" is henceforth "six" until someone changes it ;)

### DIFF
--- a/src/types/decimal.rs
+++ b/src/types/decimal.rs
@@ -296,8 +296,42 @@ impl TryFrom<f64> for Decimal {
 
 impl Display for Decimal {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        // TODO: This is correct, but not the most human-friendly format.
-        write!(f, "{}d{}", self.coefficient, self.exponent)
+        let digits = &*self.coefficient.magnitude.to_string();
+        let len = digits.len();
+        // The index of the decimal, relative to the magnitude representation
+        //       0123
+        // Given ABCDd-2, the decimal gets inserted at position 2.
+        let i_d = len as i64 + self.exponent;
+
+        if self.coefficient.sign == Sign::Negative {
+            write!(f, "-").unwrap();
+        };
+
+        if self.exponent == 0 && len > 6 {
+            // e.g. A.BCDEFGd6
+            write!(f, "{}.{}d{}", &digits[0..1], &digits[1..len], (i_d - 1))
+        } else if self.exponent == 0 {
+            // e.g. ABC.
+            write!(f, "{}.", &digits)
+        } else if self.exponent >= 0 {
+            // e.g. ABCd1
+            // let zeroes = "0".repeat(self.exponent as usize);
+            write!(f, "{}d{}", &digits, self.exponent)
+        } else {
+            // exponent < 0, there is a fractional component
+            if i_d > 0 {
+                // e.g. A.BC or AB.C
+                let ui_d = i_d as usize;
+                write!(f, "{}.{}", &digits[0..ui_d], &digits[ui_d..len])
+            } else if i_d > -6 {
+                // e.g. 0.ABC or 0.000ABC
+                let zeroes = "0".repeat(i_d.abs() as usize);
+                write!(f, "0.{}{}", &*zeroes, &digits)
+            } else {
+                // e.g. A.BCd-12
+                write!(f, "{}.{}d{}", &digits[0..1], &digits[1..len], (i_d - 1))
+            }
+        }
     }
 }
 
@@ -347,12 +381,15 @@ mod decimal_tests {
     use rstest::*;
 
     #[rstest]
-    #[case(Decimal::new(1, 0), "1d0")]
-    #[case(Decimal::new(123, -2), "123d-2")]
-    #[case(Decimal::new(123, 2), "123d2")]
-    #[case(Decimal::negative_zero_with_exponent(0), "-0d0")]
-    #[case(Decimal::negative_zero_with_exponent(-4), "-0d-4")]
-    #[case(Decimal::negative_zero_with_exponent(4), "-0d4")]
+    #[case(Decimal::new(123, 1), "123d1")]
+    #[case(Decimal::new(123, 0), "123.")]
+    #[case(Decimal::new(-123,  0),"-123.")]
+    #[case(Decimal::new( 123, -1),  "12.3")]
+    #[case(Decimal::new( 123, -3),   "0.123")]
+    #[case(Decimal::new(-123, -5),  "-0.00123")]
+    #[case(Decimal::new( 123, -5),   "0.00123")]
+    #[case(Decimal::new( 123, -10),  "1.23d-8")]
+    #[case(Decimal::new(-123, -10), "-1.23d-8")]
     fn test_display(#[case] decimal: Decimal, #[case] expected: &str) {
         let mut buffer = String::new();
         write!(buffer, "{decimal}").unwrap();
--- a/src/types/decimal.rs
+++ b/src/types/decimal.rs
@@ -298,7 +298,7 @@ impl Display for Decimal {
     #[rustfmt::skip] // https://github.com/rust-lang/rustfmt/issues/3255
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         // Inspired by the formatting conventions of Java's BigDecimal.toString()
-        const WIDE_NUMBER: i64 = 6; // if you think about it, six is a lot 🙃
+        const WIDE_NUMBER: usize = 6; // if you think about it, six is a lot 🙃
 
         let digits = &*self.coefficient.magnitude.to_string();
         let len = digits.len();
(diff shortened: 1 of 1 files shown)


src/types/decimal.rs
```diff
@@ -296,8 +296,37 @@ impl TryFrom<f64> for Decimal {
 
 impl Display for Decimal {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        // TODO: This is correct, but not the most human-friendly format.
-        write!(f, "{}d{}", self.coefficient, self.exponent)
+        let digits = &*self.coefficient.magnitude.to_string();
+        let len = digits.len();
+        // The index of the decimal
+        let i_d = len as i64 + self.exponent;
```
popematt, line 302: What does "index of the decimal" mean?

src/types/decimal.rs
```diff
@@ -296,8 +296,37 @@ impl TryFrom<f64> for Decimal {
...
+        if self.coefficient.sign == Sign::Negative {
+            write!(f, "-").unwrap();
+        };
+
+        if self.exponent > 6 {
+            // e.g. A.BCd12
+            write!(f, "{}.{}d{}", &digits[0..1], &digits[1..len], (i_d - 1))
+        } else if self.exponent > 0 {
+            // e.g. ABC0000. or ABC.
+            let zeroes = "0".repeat(self.exponent as usize);
```
popematt, line 313: This line is _adding_ more digits of precision to the output.

jobarr-amzn: Good callout. I've fixed it and the tests pass now.

src/types/decimal.rs
```diff
@@ -296,8 +296,42 @@ impl TryFrom<f64> for Decimal {
 
 impl Display for Decimal {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        // TODO: This is correct, but not the most human-friendly format.
-        write!(f, "{}d{}", self.coefficient, self.exponent)
+        let digits = &*self.coefficient.magnitude.to_string();
+        let len = digits.len();
+        // The index of the decimal, relative to the magnitude representation
```
popematt, line 301:
```suggestion
        // The index of the decimal point, relative to the magnitude representation
```
Given that we're working with a `Decimal` data type here, I think we should be more explicit that this refers to the decimal _point_. (It does, right?)

src/types/decimal.rs
```diff
@@ -296,8 +296,42 @@ impl TryFrom<f64> for Decimal {
...
+            write!(f, "{}d{}", &digits, self.exponent)
+        } else {
+            // exponent < 0, there is a fractional component
+            if i_d > 0 {
+                // e.g. A.BC or AB.C
+                let ui_d = i_d as usize;
+                write!(f, "{}.{}", &digits[0..ui_d], &digits[ui_d..len])
+            } else if i_d > -6 {
+                // e.g. 0.ABC or 0.000ABC
+                let zeroes = "0".repeat(i_d.abs() as usize);
```
popematt, line 328: Can you please add Clippy's suggestions here?

src/types/decimal.rs
```diff
@@ -296,8 +296,42 @@ impl TryFrom<f64> for Decimal {
...
 impl Display for Decimal {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        // TODO: This is correct, but not the most human-friendly format.
-        write!(f, "{}d{}", self.coefficient, self.exponent)
+        let digits = &*self.coefficient.magnitude.to_string();
+        let len = digits.len();
+        // The index of the decimal, relative to the magnitude representation
+        //       0123
+        // Given ABCDd-2, the decimal gets inserted at position 2.
+        let i_d = len as i64 + self.exponent;
```
zslayton, line 304: How about `d_index`?

src/types/decimal.rs
```diff
@@ -296,8 +296,42 @@ impl TryFrom<f64> for Decimal {
...
+
+        if self.coefficient.sign == Sign::Negative {
+            write!(f, "-").unwrap();
+        };
+
+        if self.exponent == 0 && len > 6 {
```
zslayton, line 310: Could you make `6` a named constant and/or add a comment explaining how you chose it?

src/types/decimal.rs
```diff
@@ -296,8 +296,42 @@ impl TryFrom<f64> for Decimal {
...
+            // e.g. A.BCDEFGd6
+            write!(f, "{}.{}d{}", &digits[0..1], &digits[1..len], (i_d - 1))
+        } else if self.exponent == 0 {
+            // e.g. ABC.
+            write!(f, "{}.", &digits)
+        } else if self.exponent >= 0 {
+            // e.g. ABCd1
+            // let zeroes = "0".repeat(self.exponent as usize);
```
zslayton, line 318: Leftover?
```suggestion
```

src/types/decimal.rs
```diff
@@ -296,8 +296,42 @@ impl TryFrom<f64> for Decimal {
...
+            write!(f, "{}d{}", &digits, self.exponent)
+        } else {
+            // exponent < 0, there is a fractional component
+            if i_d > 0 {
+                // e.g. A.BC or AB.C
+                let ui_d = i_d as usize;
```
zslayton, line 324: If it's helpful, you can shadow the existing name in this scope so you don't have to think of a new one. (IMO, `ui_d` is a bit cryptic, though I gather it stands for `unsigned_index_of_d`?). Shadowing has the potential to be a bit confusing for readers outside of an IDE, but in this case it only lasts one line.

src/types/decimal.rs
zslayton, line 328: This will allocate a new string, but you can achieve the same thing using the format string you pass to `write!`. See [the docs for _Fill/Alignment_ and _Sign/#/0_ for details](https://doc.rust-lang.org/std/fmt/#fillalignment).

Alternatively, create one very long `static ZEROS: &str = "0000000...";`  and just take a slice of the desired length 😛

src/types/decimal.rs
jobarr-amzn: I added a comment:
```rust
        // The index of the decimal, relative to the magnitude representation
        //       0123
        // Given ABCDd-2, the decimal gets inserted at position 2.
        let i_d = len as i64 + self.exponent;
```

src/types/decimal.rs
jobarr-amzn: Originally I used `decimal_index`, but then `rustfmt` forces some of the `write!` cases into multiple lines. I shortened some variable names to keep everything on one line because I found the multiline version so much less appealing.

src/types/decimal.rs
jobarr-amzn: Will do both. Happy to change to some other value, as well.

src/types/decimal.rs
jobarr-amzn: I ran into https://github.com/rust-lang/rust-clippy/issues/7416 and had a very frustrating time before I managed to get clippy running. 

Fixed the [cast_abs_to_unsigned](https://rust-lang.github.io/rust-clippy/master/index.html#cast_abs_to_unsigned) issue.

jobarr-amzn: Using the format string passed to `write!` was a great suggestion, took a little fiddling but I like it :D

src/types/decimal.rs
```diff
@@ -295,41 +295,37 @@ impl TryFrom<f64> for Decimal {
 }
 
 impl Display for Decimal {
+    #[rustfmt::skip] // https://github.com/rust-lang/rustfmt/issues/3255
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        // Inspired by the formatting conventions of Java's BigDecimal.toString()
+        const WIDE_NUMBER: i64 = 6; // if you think about it, six is a lot 🙃
```
zslayton, line 301: This could be a `usize`, which would allow you to skip:
* casting `len as i64` down on line 308
* casting `WIDE_NUMBER as usize` on line 314

zslayton (approved):
🚀 A nice improvement!

src/types/decimal.rs
jobarr-amzn: Yeah `WIDE_NUMBER` should be a `usize`.

* If I don't cast `len as i64` on line 308 then`dot_index` gets inferred as `usize`, but that's not right because `dot_index` should be able to be negative.
* As far as I understand it, I either have to cast `WIDE_NUMBER as usize` on line 314 to agree with `len` or cast `WIDE_NUMBER as i64` on line 324 for easiest comparison with the negative-valued `dot_index` there. The latter is the right call because `WIDE_NUMBER` *should* be a `usize`. 

Unless I'm missing something, which is probable :) I'd also happily take feedback or second opinions on the `WIDE_NUMBER` concept- whether we ought to  have it at all, whether it ought to be applied "evenly" on both whole and fractional numbers, and how large it should be.

PR #478, opened 2023-03-10
Makes `ion-hash` a feature of `ion-rs`

### LINKED ISSUE #454: Consider making `ion-hash` an optional feature in `ion-rs`
As a maintainer, it's mildly weird/annoying to have `ion-rs` and `ion-hash` versioned differently in the same crate. (Can we release _just_ `ion-hash`? Would we create a GitHub release for it?) One solution is to just give them identical version numbers, but Rust crates support "features", and this seems like a great use case for it. Moving `ion-hash` to be a feature of `ion-rs` would mean that:
* There's less chance of having mismatched or unresolvable version requirements for `ion-rs`
* Users of the library only need one dependency instead of two
* We can guarantee that `ion-hash` always works with the latest version of `ion-rs` (unlike when `ion-rs` 0.14.0 was released, and `ion-hash` was not updated)

popematt:
*Issue #, if available:*

Resolves #454 

*Description of changes:*

* Moves `ion-hash` from its own crate into `ion-rs`
* Moves the `ion-hash-tests` submodule to the root of the repo
* Updates the GHA workflow to run cargo tests with no features and with all features.
* Creates a new version of `ion-hash` crate to essentially tombstone the package.


Cargo.toml
```diff
@@ -27,6 +27,10 @@ members = [
   "ion-hash"
 ]
 
+[features]
+default = []
+ion-hash = ["digest"]
```
zslayton, line 32: Does this mean that the optional `digest` dependency should be included when the `ion-hash` feature is enabled? If so, should `sha2` also be in this list?

zslayton (approved):
yessss 🚀

One question about optional dependencies below. If it's straightforward, can you address the `clippy` warnings before merging? (With the exception of the `redundant closure` warning, which is caused by issue #472.)

Cargo.toml
popematt: This means that `ion-hash` will automatically bring in `digest`. I don't think it should bring in `sha2` because `sha2` is not required. Someone might want to use md5 or another hash function instead.

jobarr-amzn (approved):

PR #479, opened 2023-03-13
Adds builder API and macros for Element

zslayton:
This PR aims to make the `Element` API more ergonomic.

Changes:
* Splits `Sequence` into `List` and `Struct` types with their own methods (including `builder()`)
* Adds `ListBuilder`, `SExpBuilder`, and `StructBuilder` types.
* Adds `ion_list![...]`, `ion_sexp!(...)` and `ion_struct!{...}` macros for concise element construction.
* Adds an `ion!(...)` macro for parsing data into a single `Element`.
* Removes `Rc` from all `Element` types except `Symbol` in anticipation of a larger change to make `Element` `Sync`/`Send`.
* Adds `clone_builder()` to the container types, allowing duplicates of a container to be made with elements added/removed.

-----

### Examples

Some of the 'before' cases shown here were more verbose than strictly necessary, but as they were pulled
directly from the `ion-rs` source I think they're good illustrations of the hoops some developers had to
jump through.

#### `list`
**Before**
```rust
let list = Element::new_list(
    vec![
        Element::from("greetings".to_owned()),
        Element::from(5),
        Element::from(true)
    ].into_iter()
)
```
**With builder**
```rust
let list = List::builder()
    .push("greetings")
    .push(5)
    .push(true)
    .build();
```

**With `ion_list!` macro**
```rust
let list = ion_list!["greetings", 5, true]; // Expands to the builder code in the previous example
```

-----

#### `sexp`
**Before**
```rust
let list = Element::new_sexp(
    vec![
        Element::from("greetings".to_owned()),
        Element::from(5),
        Element::from(true)
    ].into_iter()
)
```
**With builder**
```rust
let sexp = SExp::builder()
    .push("greetings")
    .push(5)
    .push(true)
    .build();
```

**With `ion_sexp!` macro**
```rust
let sexp = ion_sexp!("greetings" 5 true); // Expands to the builder code in the previous example
```

-----

#### `struct`
**Before**
```rust
let struct_ = Element::new_struct(
    vec![
        ("foo", Element::from(Value::String("hello".into()))),
        ("bar", Element::from(Value::Integer(Integer::U64(5u64)))),
        ("baz", Element::from(Value::Boolean(true)),
    ].into_iter()
);
```
**With builder**
```rust
let struct_ = Struct::builder()
    .with_field("foo", "hello")
    .with_field("bar", 5)
    .with_field("baz", true)
    .build();
```

**With `ion_struct!` macro**
```rust
let sexp = ion_struct! { // Expands to the builder code in the previous example
    "foo": "hello",
    "bar": 5,
    "baz": true
};
```

-----

#### `ion!`

Especially in the context of unit tests, it's nice to be able to turn Ion text into an `Element` for the sake of comparison.

Previously, this required users to create an `ElementReader`, pass it the data, and ask it for 
an `Option<Result<Element>>`.

```rust
let element = NativeElementReader
    .iterate_over("(foo bar baz)")
    .unwrap() // this never fails; removing it is a TODO
    .next()
    .expect("Data passed to the ion! macro did not contain any values.")
    .expect("Invalid Ion data passed to the `ion!` macro.")
```

This PR adds the `ion!` macro, which achieves the same thing more concisely:

```rust
let element = ion!("(foo bar baz)");
```
The macro will panic if the Ion data is invalid. It returns the first value from the stream. If we think it's useful, 
in the future we could also provide an `ion_stream!(...)` macro that returns a `Vec<Element>` with all of the
values instead of just the first.


ion-hash/src/representation.rs
```diff
@@ -26,7 +26,8 @@ pub(crate) trait RepresentationEncoder {
             IonType::Symbol => self.write_repr_symbol(elem.as_symbol())?,
             IonType::String => self.write_repr_string(elem.as_string())?,
             IonType::Clob | IonType::Blob => self.write_repr_blob(elem.as_lob())?,
-            IonType::List | IonType::SExpression => self.write_repr_seq(elem.as_sequence())?,
+            IonType::List => self.write_repr_list(elem.as_list())?,
+            IonType::SExpression => self.write_repr_sexp(elem.as_sexp())?,
```
zslayton, line 30: 🗺️ The changes in `ion-hash` were made to accommodate splitting the monolithic `Sequence` type into `List` and `SExp` types.

src/symbol.rs
```diff
@@ -118,6 +118,12 @@ impl From<&str> for Symbol {
     }
 }
 
+impl From<String> for Symbol {
+    fn from(text: String) -> Self {
+        Symbol::owned(text)
+    }
+}
```
zslayton, line 125: 🗺️ While implementing `StructBuilder` I realized you couldn't provide an owned `String` as a field name because `String` didn't implement `Into<Symbol>`.

src/text/text_formatter.rs
```diff
@@ -550,9 +549,11 @@ mod formatter_test {
     fn test_format_struct() -> IonResult<()> {
         formatter(
             |ivf| {
-                ivf.format_struct(&Struct::from_iter(
-                    vec![("greetings", Element::from(Value::String("hello".into())))].into_iter(),
-                ))
+                ivf.format_struct(
+                    &Struct::builder()
+                        .with_field("greetings", "hello")
+                        .build_struct(),
```
zslayton, line 555: 🗺️ The new `ion_list!`, `ion_sexp!`, and `ion_struct!` macros produce an `Element`. The methods in this file operate on the more specific `List`, `SExp`, and `Struct` types that an `Element` can contain. As a result, we're using the builder APIs rather than the new macros.

src/text/text_value.rs
```diff
@@ -94,7 +94,7 @@ impl TextValue {
     }
 }
 
-/// Converts a given type into a `Vec<OwnedSymbolToken>`.
+/// Converts a given type into a `Vec<RawSymbolToken>`.
```
zslayton, line 97: 🗺️ This was a dangling documentation reference to the long ago removed `OwnedSymbolToken` type.

src/value/builders.rs
```diff
@@ -0,0 +1,453 @@
...
+            // Removes the only 'b' field
+            .remove_field("b")
+            // Removes the first 'd' field, leaves the second
+            .remove_field("d")
+            .build();
+        let expected = ion!(r#"{a: 1, c: "foo", d: baz}"#);
+        assert_eq!(actual, expected);
+        assert_eq!(actual, expected);
+    }
+}
```
zslayton, line 453: 🗺️ I can't comment directly on `element_stream_reader.rs` directly because it wasn't changed in this PR, but the "redundant closure" warning from `clippy` is caused by #472.

src/value/element_stream_reader.rs
```diff
@@ -481,9 +481,9 @@ mod reader_tests {
         next_type(reader, IonType::Struct, false);
         reader.step_in()?;
         next_type(reader, IonType::Integer, false);
-        assert_eq!(reader.field_name()?, text_token("foo"));
+        assert_eq!(reader.field_name()?, Symbol::owned("foo"));
```
zslayton, line 484: 🗺️ The `text_token` method was a vestige of the old `IonSymbolToken` trait and API.

src/value/mod.rs
```diff
@@ -12,6 +12,7 @@
 //! [simd-json-value]: https://docs.rs/simd-json/latest/simd_json/value/index.html
 //! [serde-json-value]: https://docs.serde.rs/serde_json/value/enum.Value.html
 
+pub mod builders;
```
zslayton, line 15: 🗺️ Changes in this file are superficial; I just updated the unit tests to use the new macros where possible.

src/value/owned.rs
```diff
@@ -1,186 +1,183 @@
...
-impl Symbol {
-    pub fn symbol_id(&self) -> Option<SymbolId> {
-        match self.text() {
-            Some(_) => None,
-            None => Some(0),
-        }
+impl Element {
+    pub fn null(null_type: IonType) -> Element {
+        null_type.into()
     }
```
zslayton, line 23: 🗺️ I renamed all of the `new_TYPE` methods to simply `TYPE` (e.g. `Element::new_blob(...)` to `Element::blob(...)` with the exception of the container types. `struct` is a Rust keyword, so the container types' methods' names now indicate that they return the appropriate builder. (e.g. `new_struct()` -> `struct_builder()`)

src/value/owned.rs
```diff
@@ -1,186 +1,183 @@
...
-impl Element {
-    fn new_null(e_type: IonType) -> Element {
-        Value::Null(e_type).into()
+    pub fn timestamp(timestamp: Timestamp) -> Element {
+        timestamp.into()
     }
 
-    fn new_bool(bool: bool) -> Element {
-        Value::Boolean(bool).into()
+    pub fn float(float: f64) -> Element {
```
zslayton, line 52: 🗺️ I renamed the `new_f64` method to `new_float` for consistent use of Ion type names.

src/value/owned.rs
```diff
@@ -467,111 +466,111 @@ impl PartialEq for Element {
 
 impl Eq for Element {}
 
-impl From<Value> for Element {
-    fn from(val: Value) -> Self {
-        Self::new(vec![], val)
```
zslayton, line 472: 🗺️ Changes in this file were made to guarantee that Rust values with an obvious mapping to Ion `Element`s could be passed into APIs without needing explicit boilerplate conversion code.

src/value/reader.rs
```diff
@@ -257,30 +253,16 @@ mod reader_tests {
...
-                Struct(
-                    fields.into_iter().map(|(tok, val)| {
-                        (tok, Element::new(vec![Symbol::owned("a")], val))
-                    }).collect()).into()
-            })
-            .collect(),
+            ion_struct! {
+                "string_field": (["a"], "oink!"),
+                "string_field": (["a"], "moo!"),
+                "bool_field": (["a"], true)
```
zslayton, line 259: 🗺️ `Into<Element>` is implemented for tuples of `([...annotations...], value)`, allowing for this syntax. I'm interested in getting feedback on this; alternative ideas welcome.

I considered an `annotated!` method (`"bool_field": annotated!("a", true)`) but found it to be too verbose.

zslayton (commented):
🗺️ PR tour

ion-hash/src/representation.rs
JustinDaleGray: Should we consider similar ergonomics for `Blob` and `Clob`?

```suggestion
            IonType::Clob | IonType::Blob => self.write_repr_blob(elem.as_lob())?,
```

src/text/text_formatter.rs
JustinDaleGray: Is this because the macros are not (or should not) be generic enough to be used recursively to return additional nested lists, structs, or sexp? Should we consider the macros returning an Option of a struct or element?

src/value/owned.rs
JustinDaleGray: Can a float be case into a specific float type, e.g. `Element.as_f64`?

src/text/text_formatter.rs
desaikd: Similar thought should e use `ion_struct!`, `ion_list!` return `Struct`, `List` instead of `Element` and use `ion!` for returning an `Element`? This seems more intuitive to me and may also align with how builder uses `list_builder()`, `sexp_builder()` for building `List` or `SExp`?

ion-hash/src/representation.rs
zslayton: > Should we consider similar ergonomics for `Blob` and `Clob`?
> 
> ```
>             IonType::Clob | IonType::Blob => self.write_repr_blob(elem.as_lob())?,
> ```

`Element`  offers `as_lob`, `as_clob`, and `as_blob` methods:
https://github.com/amazon-ion/ion-rust/blob/5c88f8eb712aeee185bc03cbbd4be1171bed4b0a/src/value/owned.rs#L677-L696

Similarly, it also has `as_sequence`, `as_list`, and `as_sexp` methods:
https://github.com/amazon-ion/ion-rust/blob/5c88f8eb712aeee185bc03cbbd4be1171bed4b0a/src/value/owned.rs#L698-L718

src/text/text_formatter.rs
zslayton: > Is this because the macros are not (or should not) be generic enough to be used recursively to return additional nested lists, structs, or sexp? [....]

No, `build()` will always successfully return an `Element`. Take a look at the `Value` enum and `Element` struct:

https://github.com/amazon-ion/ion-rust/blob/5c88f8eb712aeee185bc03cbbd4be1171bed4b0a/src/value/owned.rs#L401-L424

`List`, `SExp`, and `Struct` are each variants of the `Value` enum, and an `Element` is a `(Value, annotations)` pair.

In most cases, users will want an `Element` since that's what most APIs expect/return. That's why the macros and the `build()` methods both return an `Element`. However, in some cases (like the code this comment chain is on), you actually want the more specific `List`, `SExp`, or `Struct` types, not the more general `Element` type. In those cases, you can configure your (e.g.) `ListBuilder` and then call `fn build_list(self) -> List` instead of `fn build(self) -> Element`. Does that make sense?

src/value/owned.rs
zslayton: In the Ion data model, a `float` is always 64 bits. Once it's read from the stream into memory, it's basically always an `f64`. However, on the writing side we make an effort to encode it as an `f32` if it can be done losslessly.

src/text/text_formatter.rs
zslayton: > Similar thought should we have ion_struct!, ion_list! return Struct, List instead of Element and use ion! for returning an Element? This seems more intuitive to me and may also align with how builder uses list_builder(), sexp_builder() for building List or SExp?

My aim is to make the most common use cases the most streamlined. Users almost never need/want a `List`/`SExp`/`Struct` after constructing one, and having those types be the default return values for the macros would mean you'd need to do something like:
```rust
let e = Element::from(ion_list![1, 2, 3]);
```
or
```rust
let e: Element = ion_list![1, 2, 3].into();
```
instead of just
```rust
let e = ion_list![1, 2, 3];
```

The macros are named `ion_list!`, `ion_sexp!` and `ion_struct!` because each offers syntax that mirrors the Ion text equivalent.

| Ion | Rust |
|----|------|
|`[1, 2, 3]` | `ion_list! [1, 2, 3]` |
|`(1 2 3)` | `ion_sexp! (1 2 3)` |
|`{foo: "bar"}` | `ion_struct! {"foo": "bar"}`|

Notice that the container macros are intended to make it easy to turn Rust variables into Ion data:
```rust
let first_name = "Lisa";
let last_name = "Jones";
let e = ion_struct! {"first": first_name, "last": last_name};
```
while the `ion!` macro expects an already-valid Ion data stream.  Inserting existing values requires injecting them into Ion data like this:
```rust
let first_name = "Lisa";
let last_name = "Jones";
let ion_data = format!(r#"{{"first": "{}", "last": "{}"}}"#, first_name, last_name);
let e = ion!(ion_data);
```
which is error prone and expensive.

src/value/reader.rs
popematt: Alternately, we could create a trait that extends `Into<Element>` with a function `with_annotations()`.

E.g.:
```
ion_struct! {
  a: 1.with_annotations(["foo"])
}
```

src/value/builders.rs
```diff
@@ -0,0 +1,453 @@
...
+        field_value: E,
+    ) -> Self {
+        self.fields.push((field_name.into(), field_value.into()));
+        self
+    }
+
+    /// Removes the first field with the specified name from the [`Struct`] being constructed.
+    pub fn remove_field<A: AsRef<str>>(mut self, field_to_remove: A) -> Self {
+        // TODO: This removes the first field with a matching name.
+        //       Do we need other versions for remove_all or remove_last?
```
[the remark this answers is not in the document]
zslayton: The `remove` functions exist to we can offer immutable changes to a given `Element` via the `clone_builder()` API:
```rust
let e1 = ion_list! [1, 2, 3];
let e2 = e1.clone_builder().remove(1).build(); // [1, 2]
```
Currently this is a deep clone, but I plan to optimize it to something much lighter in a future PR.

src/value/builders.rs
```diff
@@ -0,0 +1,453 @@
...
+        SExp::new(self.values)
+    }
+
+    /// Builds an s-expression [`Element`] with the previously specified child elements.
+    pub fn build(self) -> Element {
+        self.build_sexp().into()
+    }
+}
+
+/// Constructs [Struct] values incrementally.
```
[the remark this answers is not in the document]
zslayton: > We should have an example showing how to use this to merge two structs (unless there's a different way that is better).

Sure, I'll add that.

> Is it worth adding a with_fields() or some other function to make it easier to add all the fields from one struct to this builder?

For the moment, this is achieved with `clone_builder()`:

```rust
let e1 = ion_struct! {"a": 1, "b": 2};
let e2 = e1.clone_builder().with_field("c", 3); // {a: 1, b: 2, c: 3}
```

We can/should add more sophisticated methods. (Currently there's no method to explicitly overwrite a given field; you'd have to remove the old one and then add the new one. Not ideal.) If it's ok, I'd like to do this in another PR since this one's already pretty big.

src/value/builders.rs
```diff
@@ -0,0 +1,453 @@
...
+/// Constructs an struct [`Element`] with the specified fields.
+///
+/// For each field, the name must implement `Into<Symbol>` and the value must implement
+/// `Into<Element>`.
+///
+/// ```
+/// use ion_rs::{ion, ion_struct};
+/// let field_name_2 = "x";
+/// // Construct an s-expression Element from Rust values
+/// let actual = ion_struct!{
```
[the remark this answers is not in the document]
zslayton: Sure, will do!

src/value/owned.rs
```diff
@@ -1,186 +1,183 @@
...
-        let mut children: Vec<Element> = Vec::new();
-        for elem in iter {
-            children.push(elem);
-        }
-        Self {
-            children: Rc::new(children),
-        }
+/// An in-memory representation of an Ion s-expression
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SExp {
```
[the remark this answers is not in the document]
zslayton: I've been pushing for `SExpression` everywhere and this was the PR that made me swing the other way. X_X I started appreciating how short `SExp` is and, given that it's a common abbreviation, began to prefer it.

How would you feel about me renaming the enum variant `SExp` instead of renaming the `struct`?

src/value/builders.rs
```diff
@@ -0,0 +1,453 @@
...
+        use $crate::value::reader::ElementReader;
+        let bytes: &[u8] = $source.as_ref();
+        NativeElementReader
+            .iterate_over(bytes)
+            .unwrap()
+            .next()
+            .expect("Data passed to the ion! macro did not contain any values.")
+            .expect("Invalid Ion data passed to the `ion!` macro.")
+    }};
+}
```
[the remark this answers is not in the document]
zslayton: Yeah, I wondered about that. The only reason I could think is that macros are more powerful and we could probably augment `ion!` in a backwards compatible way if it was a macro. At the moment though, I don't have a design that requires it. Thoughts?

src/value/reader.rs
zslayton: > Alternately, we could create a trait that extends `Into<Element>` with a function `with_annotations()`.
> 
> E.g.:
> 
> ```
> ion_struct! {
>   a: 1.with_annotations(["foo"])
> }
> ```

_Whoa,_ now THAT's thinking with portals! I'm going to give this one a try, I really like it.

src/value/builders.rs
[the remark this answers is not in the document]
popematt: >> Is it worth adding a with_fields() or some other function to make it easier to add all the fields from one struct to this builder?
>
>For the moment, this is achieved with clone_builder():

I should have been more clear. Should we have a better way to write something like this? (I don't actually know if this code will compile, but I think you can understand my meaning.)
```
let struct_a = ion_struct! { a: 1, b: 2 };
let struct_b = ion_struct! { c: 3, d: 4 };
let merged_struct: Element = {
    let builder = struct_a.clone_builder();
    for (fname, value) in struct_b.fields() {
        builder.with_field(fname, value);
    }
    builder.build()
}
println!("{merged_struct}"); // prints { a:1, b:2, c:3, d:4 }
```

src/value/owned.rs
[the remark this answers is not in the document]
popematt: Sure, as long as they're consistent.

src/value/builders.rs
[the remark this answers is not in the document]
popematt: If we want to augment it to have any compile-time validation of the input (which is the most likely enhancement we would want to make, IMO) we would have to convert it to a proc macro. (Really, though, any change that would require it to be a proc macro will do.) Proc macros must be defined in a separate crate, so we would also have to move the macro to a new crate, which would not be a backwards compatible change.

src/value/builders.rs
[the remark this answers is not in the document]
zslayton: We should have a better way to, yes. Is it ok if I add it in a future PR? If so, I'll open an issue to track it.

src/value/builders.rs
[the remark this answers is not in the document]
zslayton: > If we want to augment it to have any compile-time validation of the input

I'd love to have this. I think the `ion!` macro should only accept literal data (making compile-time guarantees feasible), while we could create something else like an `ion_fmt!` to inject Rust values into a text stream (with runtime error handling) if we decide that's useful.

I'm going to leave it as a macro for now in the hopes that we can get compile-time validation in before 1.0.

src/text/text_formatter.rs
zslayton: > My aim is to make the most common use cases the most streamlined. Users almost never need/want a List/SExp/Struct after constructing one, and having those types be the default return values for the macros would mean you'd need to do something like:

After some more consideration, I ended up swinging the other way on this.

Instead of returning a fully constructed `Element`, the macros now return their corresponding bare value type:
* `ion_list!` returns `List`
* `ion_sexp!` returns `SExp`
* `ion_struct!` returns `Struct`

I did this for a few reasons:
* When working with any other value type (`Decimal`, `Integer`, `bool`, `f64`, etc) it's understood that there's a required conversion step that needs to happen before you can use it as an `Element`. This change makes the bare collection value types consistent with the scalar types; when you build a `List`, `SExp`, or `Struct`, there's now a required conversion step that needs to happen before you can use it as an `Element`.
* Because the bare value types implement `Into<Element>`, you only end up paying the syntactic cost of `.into()` at the top level. For example:
```rust
let list: Element = ion_list! [
  1,
  true,
  ion_struct! {"foo": "bar"}, // nested; like all other bare values in this list, no .into() required
  7.5f64
].into(); // Top level, .into() required
```
* Functions that take an `Into<Element>` instead of an `Element` can accept the bare value types without explicit conversion.
```rust
fn do_something<E: Into<Element>>(value: E) {
  // ...something...
}

do_something(ion_list! [1, 2, 3]); // No explicit conversion required
```
* Thanks to Matt's idea to offer `with_annotations(...)` on values that haven't become `Element`s yet, we can also construct a full `Element` that way:
```rust
let list_element = ion_list! [1, 2, 3].with_annotations(["foo", "bar"]);
```
* There are some methods in the builder APIs where it's nice to be able to easily construct and pass in the bare value types `List`, `SExp`, and `Struct`. While the builder APIs are simple/limited at the moment, I have a sense that we may use that more in the future.

src/value/builders.rs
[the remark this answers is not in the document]
zslayton: I went ahead and added a `StructBuilder::with_fields` method.
```rust
let s1 = ion_struct! {foo: 1};
let s2 = ion_struct! {bar: 2};
let merged = Struct::builder()
  .with_fields(s1.fields())
  .with_fields(s2.fields())
  .build();
```
I have one more trait impl I need to do for Struct (`IntoIterator`) and then it'll look like this:
```rust
let s1 = ion_struct! {foo: 1};
let s2 = ion_struct! {bar: 2};
let merged = Struct::builder()
  .with_fields(&s1)
  .with_fields(&s2)
  .build();
```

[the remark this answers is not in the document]
popematt: > Is it ok if I add it in a future PR? If so, I'll open an issue to track it.

Yes, that's fine of course. Sorry that I didn't see this sooner. If you want to leave any out _any_ suggestions to avoid scope creep in this PR, that's fine by me.

src/value/builders.rs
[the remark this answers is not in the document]
zslayton: Ah, I understand. Ok, I'll make it a function for now and keep it `pub(crate)` so we can change it later.

popematt (approved):

PR #481, opened 2023-03-14
Update `rstest` syntax

jobarr-amzn:
*Issue #, if available:* #155

*Description of changes:* I was looking at old issues here and bumped into #155- but I saw only three examples of the old `rstest`. Might as well clean them up?

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #482, opened 2023-03-14
adds accessor  methods for `Timestamp`

desaikd:
*Issue https://github.com/amazon-ion/ion-rust/issues/489*

*Description of changes:*
This PR works on adding accessor methods for `Timestamp`.

*List of changes:*
* adds `year`, `month`, `day`, `hour`, `minute` and `second` methods for given timestamp in local time.
* adds `utc_hour`, `utc_minute` and `utc_second` methods for given timestamp in UTC time.

*Test:*
adds unit tests for different accessor methods defined in `Timestamp`.


src/types/timestamp.rs
```diff
@@ -572,6 +572,66 @@ impl Timestamp {
...
+
+    /// Returns the month that has been specified in the [Timestamp].
+    pub fn month(&self) -> u32 {
+        self.date_time.month()
+    }
+
+    /// Returns the day that has been specified in the [Timestamp].
+    pub fn day(&self) -> u32 {
+        self.date_time.day()
+    }
```
zslayton, line 589: For `year`, `month`, and `day`, I believe the current implementation always returns the UTC date. We'll need to offer local and `utc_` versions of each of those as well. (Consider 11PM on New Year's Eve in NYC, when it's already New Year's day in London.)

src/types/timestamp.rs
```diff
@@ -1584,6 +1644,114 @@ mod timestamp_tests {
...
+
+    #[test]
+    fn test_timestamp_second() -> IonResult<()> {
+        let timestamp = Timestamp::with_ymd_hms(2021, 4, 6, 10, 15, 30).build_at_offset(-5 * 60)?;
+        assert_eq!(timestamp.second(), 30);
+        Ok(())
+    }
+
+    #[test]
+    fn test_timestamp_utc_hour() -> IonResult<()> {
```
zslayton, line 1716: For each unit test, please include examples that test the unit's boundary across timezone offsets. (For example, in this `hour` test we see `5`/`10`/`15` -- we also need to verify the behavior when the result goes backwards or forwards across the day boundaries at `0`/`24`.)

src/types/timestamp.rs
```diff
@@ -572,6 +572,66 @@ impl Timestamp {
...
+    }
+
+    /// Returns the UTC minute(s) that has been specified in the [Timestamp].
+    pub fn utc_minute(&self) -> u32 {
+        self.date_time.minute()
+    }
+
+    /// Returns the UTC second(s) that has been specified in the [Timestamp].
+    // this will be similar to `second` method because `Timestamp` creation takes minutes offset
+    // i.e. we can not put an offset with second precision and hence no conversion for offset would be needed
```
zslayton, line 631: Huh, good observation!

src/types/timestamp.rs
```diff
@@ -1584,6 +1644,114 @@ mod timestamp_tests {
...
+            .with_month(2)
+            .with_day(4)
+            .build()?;
+
+        assert_eq!(timestamp_2.day(), 4);
+        Ok(())
+    }
+
+    #[test]
+    fn test_timestamp_hour() -> IonResult<()> {
```
zslayton, line 1676: Could you add a unit test showing that `hour()` works as expected when the offset is a `15` or `30` minute interval? (e.g. Chennai)

src/types/timestamp.rs
```diff
@@ -572,6 +572,66 @@ impl Timestamp {
...
+        self.date_time.day()
+    }
+
+    /// Returns the hour(s) that has been specified in the [Timestamp].
+    pub fn hour(&self) -> u32 {
+        // verify if the timestamp has an offset
+        if let Some(offset) = self.offset {
+            // `NativeDateTime#hours()` returns hours normalized as per UTC
+            // for local time we need to +/- the difference
+            return (self.date_time.hour() as i32 + (offset.local_minus_utc() / 3600)) as u32;
```
zslayton, line 597: This looks ok to me, but if any of the unit tests we add show that something more complex is required, we can get the `DateTime<FixedOffset>` version of `self.date_time` using [`DateTime::from_local`](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.from_local) and then ask it for the corresponding field.

src/types/timestamp.rs
```diff
@@ -572,6 +572,66 @@ impl Timestamp {
...
+        self.date_time.minute()
+    }
+
+    /// Returns the second(s) that has been specified in the [Timestamp].
+    pub fn second(&self) -> u32 {
+        self.date_time.second()
+    }
+
+    /// Returns the UTC hour(s) that has been specified in the [Timestamp].
+    pub fn utc_hour(&self) -> u32 {
```
popematt, line 620: Idea—we could get rid of the `xyz()`/`utc_xyz()` dichotomy entirely by making all of the functions return the value in the "local"/offset time, and then adding one function to convert to a UTC timestamp.

I'm not quite sure myself whether this is a good idea, which is why I've labelled it an "idea" rather than a "suggestion", but I hope that by presenting the idea, others might chime in.

src/types/timestamp.rs
```diff
@@ -572,6 +572,66 @@ impl Timestamp {
...
+        self.date_time.month()
+    }
+
+    /// Returns the day that has been specified in the [Timestamp].
+    pub fn day(&self) -> u32 {
+        self.date_time.day()
+    }
+
+    /// Returns the hour(s) that has been specified in the [Timestamp].
+    pub fn hour(&self) -> u32 {
```
popematt, line 592: Have you thought about returning an `Option` from these functions so that they can reflect the precision of the `timestamp`?
E.g.
```rust
/// Returns the hour(s) that has been specified in the [Timestamp] if the timestamp's precision 
/// includes hours. If you need hour, regardless of precision, use `hour().unwrap_or_default()`,
/// which will correctly return an hour of `0` when the precision does not include hours.
pub fn hour(&self) -> Option<u32> {
    // ...
}
```

I think it makes it match the data model a little better. It's a little more verbose when you always want some value, but it's a little more ergonomic when you only want to deal with the subfields that are included by the timestamp's precision. What do you think?

zslayton: This is a good observation. I'm unsure which way to go on it; Ion `Timestamp`s _always_ have a hour/minutes/second in that they represent a fixed point in time and those fields are `0` when the precision is too low. However, this does represent an opportunity to remind users that the `0` might have been implicit rather than explicitly part of the value.

I suspect that as a generalized datetime value, providing field defaults is probably what users would prefer. @jpschorr, what's your take on this? Would you rather check the timestamp's `precision`, or `hour().is_none()`?

src/types/timestamp.rs
zslayton: It's probably pretty cheap to do this. Users would likely appreciate having a `fn to_utc(&self) -> Timestamp` method anyway.

desaikd: Reference issue: https://github.com/amazon-ion/ion-rust/issues/489

src/types/timestamp.rs
desaikd: + 1. I think it would rather simple to check `precision` instead of `hour().is_none()`.

src/types/timestamp.rs
desaikd: added `to_utc` for `Timestamp`, removed all `utc_*` methods.

src/types/timestamp.rs
```diff
@@ -1584,6 +1634,117 @@ mod timestamp_tests {
...
+        case(Timestamp::with_ymd_hms(2021, 4, 6, 10, 59, 30).build_at_offset(5 * 60), 59)
+    )]
+    fn test_timestamp_minute(
+        timestamp: IonResult<Timestamp>,
+        expected_minutes: u32,
+    ) -> IonResult<()> {
+        assert_eq!(timestamp?.minute(), expected_minutes);
+
+        Ok(())
+    }
```
jobarr-amzn, line 1699: We got rid of the last vestige of `#[rstest(...)]` tests earlier this week: https://github.com/amazon-ion/ion-rust/pull/481

```suggestion
    #[rstest]
    #[case(Timestamp::with_ymd_hms(2021, 4, 6, 10, 15, 30).build_at_offset(-90), 10)]
    #[case(Timestamp::with_ymd_hms(2021, 4, 6, 10, 15, 30).build_at_offset(-5 * 60), 10)]
    #[case(Timestamp::with_ymd_hms(2021, 4, 6, 10, 15, 30).build_at_offset(5 * 60), 10)]
    #[case(Timestamp::with_ymd_hms(2021, 4, 6, 10, 15, 30).build_at_offset(15), 10)]
    #[case(Timestamp::with_ymd_hms(2021, 4, 6, 10, 15, 30).build_at_offset(30), 10)]
    #[case(Timestamp::with_ymd_hms(2021, 4, 6, 10, 15, 30).build_at_offset(0), 10)]
    #[case(Timestamp::with_ymd_hms(2021, 4, 6, 0, 15, 30).build_at_offset(5 * 60), 0)]
    #[case(Timestamp::with_ymd_hms(2021, 4, 6, 23, 15, 30).build_at_offset(-5 * 60), 23)]
    #[case(Timestamp::with_ymd_hms(2021, 4, 6, 0, 15, 30).build_at_offset(23 * 60), 0)]
    fn test_timestamp_hour(
        #[case] timestamp: IonResult<Timestamp>,
        #[case] expected_hours: u32
    ) -> IonResult<()> {
        assert_eq!(timestamp?.hour(), expected_hours);

        Ok(())
    }

    #[rstest]
    #[case(Timestamp::with_ymd_hms(2021, 4, 6, 10, 15, 30).build_at_offset(-90), 15)]
    #[case(Timestamp::with_ymd_hms(2021, 4, 6, 10, 15, 30).build_at_offset(-5 * 60), 15)]
    #[case(Timestamp::with_ymd_hms(2021, 4, 6, 10, 15, 30).build_at_offset(5 * 60), 15)]
    #[case(Timestamp::with_ymd_hms(2021, 4, 6, 10, 15, 30).build_at_offset(0), 15)]
    #[case(Timestamp::with_ymd_hms(2021, 4, 6, 10, 0, 30).build_at_offset(5 * 60), 0)]
    #[case(Timestamp::with_ymd_hms(2021, 4, 6, 10, 59, 30).build_at_offset(5 * 60), 59)]
    fn test_timestamp_minute(
        #[case] timestamp: IonResult<Timestamp>,
        #[case] expected_minutes: u32,
    ) -> IonResult<()> {
        assert_eq!(timestamp?.minute(), expected_minutes);

        Ok(())
    }
```

jobarr-amzn (changes requested):

popematt (commented):
FYI—there's a `TODO` you can get rid of on lines 162-164.

src/types/timestamp.rs
zslayton: `self.date_time.day()` is the UTC `day`. In order to return the local `day`, you have to localize the `year`/`month`/`day` fields of `self.date_time` to accommodate the new year's eve case I described above.

If the `Timestamp` doesn't have an offset because its precision is too low, you can use the current values.

src/types/timestamp.rs
```diff
@@ -234,7 +234,7 @@ impl Timestamp {
     /// NOTE: This is a potentially lossy operation. A Timestamp with picoseconds would return a
     /// number of nanoseconds, losing precision. Similarly, a Timestamp with milliseconds would
     /// also return a number of nanoseconds, erroneously gaining precision.
-    fn fractional_seconds_as_nanoseconds(&self) -> Option<u32> {
+    pub fn fractional_seconds_as_nanoseconds(&self) -> Option<u32> {
```
zslayton, line 237: This method name is kind of onerous (I'm the one who wrote it!) and it returns `Option<u32>`, so I think we should keep it private. The public methods should be something like `nanoseconds(&self) -> u32` and `milliseconds(&self) -> u32`. They can use this method under the hood.

src/types/timestamp.rs
```diff
@@ -572,6 +572,56 @@ impl Timestamp {
...
+
+    /// Returns the second(s) that has been specified in the [Timestamp].
+    pub fn second(&self) -> u32 {
+        self.date_time.second()
+    }
+
+    /// Return a UTC timestamp for this [Timestamp]
+    pub fn to_utc(&self) -> Timestamp {
+        let utc_date_time = DateTime::<Utc>::from_utc(self.date_time, Utc);
+        utc_date_time.naive_utc().into()
```
zslayton, line 623: Is this different from doing:
```suggestion
        self.date_time.into()
```
?

src/types/timestamp.rs
```diff
@@ -1584,6 +1634,117 @@ mod timestamp_tests {
...
+
+    #[test]
+    fn test_timestamp_month() -> IonResult<()> {
+        let timestamp = Timestamp::with_year(2021).with_month(2).build()?;
+        assert_eq!(timestamp.month(), 2);
+        Ok(())
+    }
+
+    #[test]
+    fn test_timestamp_day() -> IonResult<()> {
```
zslayton, line 1652: We need some tests for the `year`/`month`/`day` accessors where the `Timestamp` has high precision and an offset.

src/types/timestamp.rs
zslayton: I'm unresolving this thread because I don't see tests that cross the `day` boundary. If the hour is `10`, what happens when the offset is `-11:00`? Or if it's `15` and the offset is `10:00`?

zslayton (changes requested):

src/types/timestamp.rs
desaikd: Yes, that is true both are same.

zslayton (approved):
🚀 

Couple of small things left before merging.

jobarr-amzn (approved):

PR #485, opened 2023-03-15
Element reader

zslayton:
This PR introduces an updated `ElementReader` trait that provides methods for materializing values from the reader into `Element`s, and is available for all user-level reader implementations regardless of format or blocking strategy.

Other changes:
* Adds static convenience methods to the `Element` type that mirror the method
   names in the `ElementReader` trait, simplifying one-off parsing.
  * `Element::read_one()`, `Element::read_next()`, `Element::read_all()`
* Removes `NativeElementReader`, which used to be part of the `native`/`ion-c` dichotomy.
* Streamlines some of the integration test logic in `element-test-vectors`.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #447, opened 2022-11-05
Implement `Display` for `value::owned::Struct`, `value::owned::Sequence`, etc.


jpschorr:
`value::owned::Element` implements `Display`: https://github.com/amzn/ion-rust/blob/main/src/value/owned.rs#L456-L483

But e.g., `Struct`, `Sequence` to not implement `Display`

So if I have a `Struct` (perhaps I've checked type and called `as_struct` on the `Element` and passed to another function), I can't easily display without something like `Element::from(struct_.clone()).to_string()`

zslayton:
Fixed in #486, this will be included in the next release.

PR #486, opened 2023-03-16
Mass renaming, container Display impls

zslayton:
Builds on the changes in PR #485. After that's merged, I'll update this to target `main`.

-----

**The diff contains no logic changes.**

This PR performs several superficial cleanup tasks, including:
* Renames the `crate::value` module to `crate::element`
* Renames several types (and corresponding methods) to better align with [the official Ion type names](https://amazon-ion.github.io/ion-docs/docs/spec.html#the-ion-data-model)
  * `Boolean` -> `Bool`
  * `Integer` -> `Int`
  * `UInteger` -> `UInt`
  * `SExpression` -> `SExp` (this rename was started in an earlier PR, but additional references to the type name were found)
* Adds `Display` impls for `List`, `SExp`, and `Struct` (addressing #447)
* Adds `Display` impl for `Value` so it can eventually be used independent of `Element`.
* Modifies `Display` for `IonType` to use official type names


popematt (approved):
To be completely honest, I did not read _all_ of the changes. I agree with the overview of the renaming changes, but I did not read through those commits because of their sheer size. I did, however, look at the `Display`-related commits and the clippy commits. 👍

src/element/owned.rs
```diff
@@ -867,4 +898,40 @@ mod value_tests {
...
+    fn sexp_display_roundtrip() {
+        let sexp = ion_sexp! (1 2 3 true false);
+
+        // Use the Display impl to serialize the sexp to text
+        let text_sexp = format!("{sexp}");
+        // Parse the result and make sure it represents the same data
+        let expected_element: Element = sexp.into();
+        let actual_element = Element::read_one(text_sexp).unwrap();
+        assert!(expected_element.ion_eq(&actual_element));
+    }
```
jobarr-amzn, line 924: I just can't get over how neat these macros are. Great work, Zack and Matt!

src/element/mod.rs
```diff
@@ -309,7 +309,7 @@ mod tests {
             ops: vec![AsBool],
             op_assert: Box::new(|e: &Element| {
                 let expected = Element::from(true);
-                assert_eq!(Some(true), e.as_boolean());
```
jobarr-amzn, line 312: This change is lost too, you belong in the `bool` rename commit.

src/element/mod.rs
```diff
@@ -501,11 +501,11 @@ mod tests {
         // table of negative assertions for each operation
         let neg_table: Vec<(ElemOp, ElemAssertFn)> = vec![
             (IsNull, Box::new(|e| assert!(!e.is_null()))),
-            (AsBool, Box::new(|e| assert_eq!(None, e.as_boolean()))),
+            (AsBool, Box::new(|e| assert_eq!(None, e.as_bool()))),
```
jobarr-amzn, line 504: Lost

src/element/owned.rs
```diff
@@ -724,7 +724,7 @@ impl Element {
         }
     }
 
-    pub fn as_boolean(&self) -> Option<bool> {
+    pub fn as_bool(&self) -> Option<bool> {
```
jobarr-amzn, line 727: Ah, looks like we picked up an additional rename here.

jobarr-amzn (commented):
A few lost patch fragments in the wrong commits, but I stopped noting them after this shipped :P

src/element/reader.rs
```diff
@@ -235,7 +235,7 @@ mod reader_tests {
             Null(IonType::Clob),
             Null(IonType::Blob),
             Null(IonType::List),
-            Null(IonType::SExpression),
+            Null(IonType::SExp),
```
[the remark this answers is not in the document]
zslayton: 😅 Yeah, there were a few of these. I really need to set up a pre-commit git hook for `cargo test --all-features` to prevent this sort of thing.

PR #487, opened 2023-03-17
Modernizes the `ElementWriter` trait

### LINKED ISSUE #202: Element should be writeable to a writer
Currently it's not possible to roundtrip through ion-rs, i.e.

bytes -> reader -> loader -> element -> bytes

Maybe the Element API should include a `write_to(writer)` method or something like that. Beyond just the ability to write an Element to a writer, I think we should have the property that the output bytes are as unmodified as possible (I don't know how to define this). So, modulo things like spacing/comments/etc. it should be possible to read in Ion text, change some value using the Element API and then write the Ion text out again.. the final output should not do weird things like re-order fields, etc.

zslayton:
This PR has a weird commit history that was caused by needing to be rebased a few times in the wake of other (sweeping) changes getting merged. The diff is correct, however, so it can be safely ignored for the purposes of review. It'll get squashed at merge time anyway.

-----

Changes:
* `ElementWriter` is now automatically implemented by all `IonWriter` implementations, allowing `IonWriter`s to write an `Element` to output without creating an instance of an intermediate type (e.g. `NativeElementWriter`).
* Removed `NativeElementWriter`.
* Adds the `Element::value()` method to get a reference to its `Value`. This allows matching on the `Value` directly instead of matching on its `IonType` and then calling the appropriate `as_TYPE().unwrap()` method.
* Added `output()` and `output_mut()` to the `IonWriter` trait, giving users a way to see/manipulate their (eg) output buffer while the writer still exists. Prior to this, calling `drop(writer)` to unfreeze the output buffer was common.
* Fixes a bug in `RawBinaryWriter::write_f32` that caused negative zero to be serialized as positive zero.
* Renames `IonSequence::iter()` to `IonSequence::elements()`.
* Several doc comments included code fences (triple backticks) containing example output (ie not Rust). Previously, those code fences were marked with `ignore` to keep them from being treated as doc tests; this caused them to be listed in the `cargo test` output as `IGNORED`. Now they are marked `text`, indicating that they are not test code at all.

Fixes #202.


src/element/writer.rs
```diff
@@ -54,3 +92,30 @@ pub enum Format {
...
+            null true 0 1e0 2.0 2022T foo "bar" (foo bar baz) [foo, bar, baz] {foo: true, bar: false}
+        "#;
+        let expected_elements = Element::read_all(ion.as_bytes())?;
+        writer.write_elements(&expected_elements)?;
+        writer.flush()?;
+        let actual_elements = Element::read_all(writer.output().as_bytes())?;
+        assert!(expected_elements.ion_eq(&actual_elements));
+        Ok(())
+    }
+}
```
zslayton, line 121: 🗺️ This unit test used to live in `NativeElementWriter` and was moved here when that file was deleted.

src/text/parse_result.rs
```diff
@@ -51,7 +51,7 @@ pub(crate) type IonParseResult<'a, O> = IResult<&'a str, O, IonParseError<'a>>;
 /// with other parsers that return `IonParseResult`, you can call `upgrade()` on your parser's
 /// output before returning it.
 ///
-/// ```ignore
+/// ```ignore // Doc tests cannot use internal APIs
```
zslayton, line 54: 🗺️ This doc comment's code fence _does_ contain Rust code, so `ignore` is appropriate, but it also uses crate-visible methods and so will not compile.

src/text/parse_result.rs
```diff
@@ -70,9 +70,9 @@ pub(crate) type IonParseResult<'a, O> = IResult<&'a str, O, IonParseError<'a>>;
 /// // The same function as above, but spelled out a bit more explicitly
 /// fn verbose_recognize_boolean(input: &str) -> IonParseResult<&str> {
 ///     // Combine several core `nom` parsers to make a new parser with the desired logic
-///     let parser = terminated(          <-- Takes two parsers; if both succeed, returns the output of the first
-///         tag("true").or(tag("false")), <-- Parser 1 matches the text `true` or the text `false`
-///         tag(" ")                      <-- Parser 2 matches a single space
+///     let parser = terminated(          // <-- Takes two parsers; if both succeed, returns the output of the first
+///         tag("true").or(tag("false")), // <-- Parser 1 matches the text `true` or the text `false`
+///         tag(" ")                      // <-- Parser 2 matches a single space
```
zslayton, line 75: 🗺️ Hazard of using `ignore`: the Rust code was not being compiled, so no one noticed that these lines should've been comments.

zslayton (commented):
🗺️ PR tour

src/element/element_stream_reader.rs
```diff
@@ -110,7 +110,7 @@ impl ElementStreamReader {
                 value
                     .as_sequence()
                     .unwrap()
-                    .iter()
+                    .elements()
```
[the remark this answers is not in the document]
popematt: Does the loss of `iter()` in `IonSequence` make any difference w.r.t. being able to use it in a `for` loop?

src/text/parse_result.rs
popematt: It seems like a bad idea to have doc comment code that uses private/pub(crate) functions. Aren't you basically publishing an example that the reader can never use?

src/element/writer.rs
```diff
@@ -54,3 +92,30 @@ pub enum Format {
...
+
+    #[test]
+    fn element_roundtrip() -> IonResult<()> {
+        let mut buffer = Vec::new();
+        let mut writer = TextWriterBuilder::new().build(&mut buffer)?;
+
+        let ion = r#"
+            null true 0 1e0 2.0 2022T foo "bar" (foo bar baz) [foo, bar, baz] {foo: true, bar: false}
+        "#;
+        let expected_elements = Element::read_all(ion.as_bytes())?;
```
jobarr-amzn, line 114: Would the semantics of the test be changed by using a builder macro for `expected_elements`? I feel like that would be an improvement but perhaps I'm missing something.

Separately, this is read text, write text, read text. Is there a version of this that round trips binary or round trips across formats, e.g. read text, write binary, then read binary?

src/text/raw_text_writer.rs
```diff
@@ -737,6 +727,14 @@ impl<W: Write> IonWriter for RawTextWriter<W> {
...
         Ok(())
     }
+
+    fn output(&self) -> &W {
+        self.output.get_ref()
+    }
+
+    fn output_mut(&mut self) -> &mut W {
+        self.output.get_mut()
+    }
```
jobarr-amzn, line 737: This looks like a move- I see the doc comments in the deleted code, are they preserved elsewhere? 

**EDIT**: Nevermind I see much more detailed commentary added in `writer.rs`.

tests/element_test_vectors.rs
```diff
@@ -104,7 +101,7 @@ trait ElementApi {
         for (index, (element1, element2)) in e1.iter().zip(e2.iter()).enumerate() {
             if !element1.ion_eq(element2) {
                 return format!(
-                    "The values at position #{index} were not IonEq.\ne1: {element1:?}\ne2: {element2:?}"
+                    "The values at position #{index} were not IonEq.\ne1: {element1}\ne2: {element2}"
```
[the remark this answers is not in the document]
jobarr-amzn: Thanks!

tests/ion_hash_tests.rs
```diff
@@ -123,7 +123,7 @@ fn ion_hash_tests() -> IonHashTestResult<()> {
 
 fn test_file(file_name: &str) -> IonHashTestResult<()> {
     let data = read(file_name).map_err(|source| ion_rs::IonError::IoError { source })?;
-    let elems = Element::read_all(&data)?;
+    let elems = Element::read_all(data)?;
```
jobarr-amzn, line 126: What's going on here?

jobarr-amzn (approved):

src/element/element_stream_reader.rs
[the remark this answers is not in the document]
zslayton: Good question. The short answer is no, what was previously:
```rust
for element in sequence.iter() {...}
```
is now
```rust
for element in sequence.elements() {...}
```
I still intend to `impl IntoIterator for &List` and `for &SExp`, which will enable the more desirable:
```rust
for element in &sequence {...}
```

src/text/parse_result.rs
zslayton: This is a private module, so it isn't exposed in the user-facing documentation.

Any doc comment that includes a code fence is treated as code that should compile using the public API; I'm unsure how to convince `cargo` to compile it with pub(crate) APIs. If you know of a way, let me know!

src/element/writer.rs
```diff
@@ -5,38 +5,76 @@
...
+            Value::String(s) => self.write_string(s),
+            Value::Symbol(s) => self.write_symbol(s),
+            Value::Bool(b) => self.write_bool(*b),
+            Value::Blob(b) => self.write_blob(b),
+            Value::Clob(c) => self.write_clob(c),
+            Value::SExp(s) => {
+                self.step_in(IonType::SExp)?;
+                self.write_elements(s.elements())?;
+                self.step_out()
+            }
```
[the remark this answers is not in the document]
zslayton: Could you elaborate on what you mean by "closed form?"

I considered extending `ElementWriter` to also have `write_struct(&Struct)`, `write_list(&List)`, and `write_sext(&SExp)`. Would that address it?

tests/ion_hash_tests.rs
zslayton: `clippy` noticed that `data` (which was already a reference) was being turned into a doubly-indirect reference unnecessarily. This change was unrelated to the core purpose of the PR.

PR #488, opened 2023-03-17
Address boundary handling issues in non-blocking raw text reader, and its blocking wrapper

nirosys:
*Issue #, if available:* n/a

*Description of changes:*
Previous implementation of the non-blocking reader, and its blocking wrapper, had two bugs when handling incomplete errors, and continuing after an incomplete error.

The non-blocking reader did not take into account the `step_out` nesting when unsetting the `need_continue`. When stepping out of a container that contained a container that had not yet been read, this would result in the reader stepping out beyond the target depth.

The blocking reader did not properly mark the stream as complete when stepping out of a container. If a `step_out` caused the last remaining data from the stream to be read (completing the stream) root level values at the end of the stream could be ambiguous and continue to cause incomplete data errors, until the stream had been marked complete.

---

### DIFF
--- a/src/text/non_blocking/raw_text_reader.rs
+++ b/src/text/non_blocking/raw_text_reader.rs
@@ -291,7 +291,7 @@ impl<A: AsRef<[u8]>> RawTextReader<A> {
 
     /// Continues any previously incomplete parsing attempt.
     fn continue_state(&mut self) -> IonResult<()> {
-        if self.need_continue {
+        if self.need_continue && self.step_out_nest == 0 {
             self.need_continue = false;
 
             match self.state {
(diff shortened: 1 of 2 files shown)


nirosys:
Absolutely, porting them over from my other branch now.

nirosys:
Added a unit test which currently validates the `step_out` nesting fix when continuing from an incomplete. I do not think it validates the end of stream on `step_out` anymore (It has been modified since I found the issue). It will take me a bit to repro, in order to get the data, and read sizes, to line up so it exhausts on `step_out`.

I'll make sure to include that in the format agnostic PR if no one minds its absence here.

zslayton (approved):
Nice! Can you address the `clippy` warnings before merging?

nirosys:
tyvm! Added commit for clippy issues.

PR #490, opened 2023-03-17
Adds basic crate-level documentation

zslayton:
Provides examples for creating, reading, and writing `Element`s.

image

image

image

### DIFF
--- a/src/binary/non_blocking/raw_binary_reader.rs
+++ b/src/binary/non_blocking/raw_binary_reader.rs
@@ -531,20 +531,20 @@ impl<A: AsRef<[u8]>> RawBinaryBufferReader<A> {
     }
 }
 
-/// If the RawBinaryBufferReader's data source is a Vec<u8>, it gains the ability to add data to
+/// If the RawBinaryBufferReader's data source is a `Vec<u8>`, it gains the ability to add data to
 /// the buffer between read operations. This is useful when reading from a streaming data source
 /// like a file or TCP socket; the reader can read the contents of its buffer, add more bytes as
 /// they arrive, and then continue reading.
 impl RawBinaryBufferReader<Vec<u8>> {
     /// Copies the provided bytes to end of the reader's input buffer.
-    fn append_bytes(&mut self, bytes: &[u8]) {
+    pub fn append_bytes(&mut self, bytes: &[u8]) {
         self.buffer.append_bytes(bytes);
     }
 
-    /// Tries to read `length` bytes from `source`. Unlike [append_bytes], this method does not do
+    /// Tries to read `length` bytes from `source`. Unlike [Self::append_bytes], this method does not do
     /// any copying. A slice of the reader's buffer is handed to `source` so it can be populated
     /// directly.
-    fn read_from<R: Read>(&mut self, source: R, length: usize) -> IonResult<usize> {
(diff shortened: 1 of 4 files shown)


src/binary/non_blocking/raw_binary_reader.rs
zslayton, line 1: 🗺️ `cargo doc` pointed out some nits in this file and the next.

src/element/owned.rs
```diff
@@ -598,6 +598,18 @@ impl From<Symbol> for Value {
...
+    fn from(value: &[u8]) -> Self {
+        Value::Blob(value.into())
+    }
+}
+
+impl From<Vec<u8>> for Value {
+    fn from(value: Vec<u8>) -> Self {
+        Value::Blob(value)
+    }
+}
```
zslayton, line 611: 🗺️ While writing the doc tests, I realized we didn't have an `Into<Value>` impl for the `blob`-ish Rust types.

src/lib.rs
```diff
@@ -29,8 +187,11 @@ mod symbol_table;
 mod system_reader;
 mod writer;
 
+#[doc(inline)]
 pub use data_source::IonDataSource;
+#[doc(inline)]
 pub use raw_symbol_token::RawSymbolToken;
+#[doc(inline)]
```
zslayton, line 194: 🗺️ `#[doc(inline)]` prevents these from showing up in a big "Re-exports" list in the docs. We'll probably add this to more types before 1.0.

zslayton (commented):
🗺️ PR tour

src/lib.rs
```diff
@@ -1,6 +1,164 @@
...
+//! writer.flush()?;
+//! assert_eq!(
+//!   "{foo: \"hello\", bar: true, baz: [4, 5, 6]}".as_bytes(),
+//!   writer.output().as_slice()
+//! );
+//! # Ok(())
+//! # }
+//! ```
 
+use element::owned::Element;
```
desaikd, line 161: clippy suggests this unused import

src/lib.rs
```diff
@@ -1,6 +1,164 @@
...
+//! # fn main() -> IonResult<()> {
+//! use std::fs::File;
+//! use ion_rs::element::owned::Element;
+//! use ion_rs::element::reader::ElementReader;
+//! use ion_rs::ReaderBuilder;
+//! let ion_file = File::open("/foo/bar/baz.ion").unwrap();
+//! let mut reader = ReaderBuilder::default().build(ion_file)?;
+//! // A simple pretty-printer
+//! for element in reader.elements() {
+//!     println!("{}", element?)
```
desaikd, line 102: Should we add a sample file content and output for this example?

desaikd (approved):
Few minor comments otherwise looks good!
Really nice to have those macro examples 🚀

src/lib.rs
zslayton: It does stand out as the only example without an `assert_eq!`, but I figured seeing how to access the elements in a file would be enough for folks to use the other tests' examples. I'm going to leave this as-is for the moment to get v0.16.0 out, but I'm open to changing it in the future.

src/lib.rs
zslayton: Fixed!

PR #491, opened 2023-03-18
Remove `crate::element::owned` module

zslayton:
This PR removes the `crate::element::owned` module, moving its contents (`Element`, `Value`, etc) into the parent `element` module.

`owned` was originally created to stand in contrast with the now-removed `borrowed` implementation of `Element`. The name is a bit peculiar when it stands on its own.

The diff contains no logic changes, only the results of an automated refactor.


zslayton (commented):
As this PR contains no logic changes, I'm going to go ahead and merge it. We can always back it out if needed.

desaikd (approved):

Issue #494, opened 2023-03-27
Fatal error generated when insufficient data reached while parsing escaped sequence


nirosys:
When insufficient data is encountered while parsing an escaped string sequence (appears to be all types of sequences, but I have not verified), an error is returned that is not correctly identified as Insufficient data. This is causing the raw text reader to fail rather than requesting more data.

nirosys:
After some digging, just re-read the error and realized this is not a UTF-8 boundary issue, but rather the parser not correctly identifying an incomplete error while parsing escaped sequences.

Trivial repro:
```rust
let source = "\"123456\\u269b\"";
//                       ^-- First read stops here.
let mut reader = RawTextReader::new(source.as_bytes()[..10].to_owned());
reader.next()?;
```

Running that returns an Err of:
```
unexpected error after partial escaped sequence: unrecognized input near line 0: could not parse an escaped character (Unicode, hex, or literal): Parsing requires more data: '"123456\u2''
```

nirosys:
A fix for this was merged a while ago, closing.

PR #495, opened 2023-03-28
Handle incomplete errors when parsing escaped string sequences

nirosys:
*Issue #, if available:* #494

*Description of changes:*
This PR adds two changes:
1) Updates error handling while parsing escaped string sequences in order to bubble up nom's Incomplete errors to the reader, so that they can be communicated further.
2) Updates `read_from` to ensure that the amount of data read is limited to the length provided by the caller. Prior to this PR the entire rest of the buffer would be used for reading data. While this could be beneficial (the memory is allocated, using it makes sense) the caller expects to read no more than `length`, so as a matter of correctness we shouldn't read more than `length`.

A regression test was also added to mimic the behavior seen in #494.

---

### DIFF
--- a/src/text/non_blocking/raw_text_reader.rs
+++ b/src/text/non_blocking/raw_text_reader.rs
@@ -1795,4 +1795,32 @@ mod reader_tests {
 
         Ok(())
     }
+
+    #[test]
+    fn generate_incomplete_on_truncated_escape() -> IonResult<()> {
+        let source = "\"123456\\u269b\"";
+        //                       ^-- First read stops here. (offset 9)
+        let mut reader = RawTextReader::new(source.as_bytes()[..10].to_owned());
+
+        match reader.next() {
+            Err(IonError::Incomplete {
+                position:
+                    Position {
+                        line_column: Some((line, column)),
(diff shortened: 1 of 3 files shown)


src/text/non_blocking/raw_text_reader.rs
```diff
@@ -1795,4 +1795,32 @@ mod reader_tests {
...
+                assert_eq!(line, 0); // Line is still 0 since we haven't actually seen a '\n' yet.
+                assert_eq!(column, 0); // start of the string; the value being parsed.
+            }
+            Err(e) => panic!("unexpected error after partial escaped sequence: {e}"),
+            Ok(item) => {
+                panic!("unexpected successful parsing of partial escaped sequence data: {item:?}")
+            }
+        }
+        reader.append_bytes(source[10..].as_bytes())?;
+        next_type(&mut reader, IonType::String, false);
```
popematt, line 1823: Am I understanding this correctly? These lines are testing that you can append more data and have the parsing continue successfully, right? If so, it's probably a good idea to assert that we are getting the expected value, not just the expected type.

Issue #496, opened 2023-03-29
Optimize `RawTextWriter::add_annotation`


zslayton:
The `RawTextWriter::add_annotation` method has the following `TODO`:

https://github.com/amazon-ion/ion-rust/blob/cffa7452bffaa8b8af3d702b00281186c42238d2/src/text/raw_text_writer.rs#L400-L409

This could be addressed by adding a `Shared` variant to the `RawSymbolTokenRef` enum:

https://github.com/amazon-ion/ion-rust/blob/024cb2e8f15f349c3ecb109e0ae629ca8acc0377/src/raw_symbol_token_ref.rs#L7-L10

The `Shared` variant could just `clone()` the `Arc<str>` inside the provided `Symbol::Shared` instead of allocating a new `String` via `to_owned`.

This would just be an optimization / implementation detail, and so can be made after a 1.0 release.

PR #497, opened 2023-03-29
Make `Element` implement `Send`

zslayton:
This PR modifies the `Symbol` type to use `Arc<str>` instead of `Rc<str>` when sharing references to text, eliminating the last remaining use of a non-`Send` type (`Rc`) in the `Element` API.

Following this change, `Element`s can be sent between threads or shared using native mechanisms like `Arc<Element>`.

Changes:
* Modifies the `Symbol::Shared` enum variant to use `Arc<str>` instead of `Rc<str>`.
* Adds a unit test demonstrating that `Element` implements `Send` and safeguarding against future changes that accidentally violate that property.
* Addresses a `clippy` warning added in Rust 1.70 nightly about using struct builder syntax unnecessarily.

### DIFF
--- a/src/symbol.rs
+++ b/src/symbol.rs
@@ -4,13 +4,13 @@ use std::borrow::Borrow;
 use std::cmp::Ordering;
 use std::fmt::{Display, Formatter};
 use std::hash::{Hash, Hasher};
-use std::rc::Rc;
+use std::sync::Arc;
 
 /// Stores or points to the text of a given [Symbol].
 #[derive(Debug, Eq)]
 enum SymbolText {
     // This Symbol refers to a string in the symbol table
-    Shared(Rc<str>),
+    Shared(Arc<str>),
     // This Symbol owns its own text
     Owned(String),
     // This Symbol is equivalent to SID zero (`$0`)
@@ -42,7 +42,7 @@ impl Clone for SymbolText {
     fn clone(&self) -> Self {
         match self {
             SymbolText::Owned(text) => SymbolText::Owned(text.to_owned()),
-            SymbolText::Shared(text) => SymbolText::Shared(Rc::clone(text)),
+            SymbolText::Shared(text) => SymbolText::Shared(Arc::clone(text)),
             SymbolText::Unknown => SymbolText::Unknown,
         }
     }
@@ -88,7 +88,7 @@ impl Symbol {
         }
     }
 
-    pub fn shared(text: Rc<str>) -> Symbol {
+    pub fn shared(text: Arc<str>) -> Symbol {
         Symbol {
             text: SymbolText::Shared(text),
         }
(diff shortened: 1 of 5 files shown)


src/text/raw_text_writer.rs
zslayton, line 1: 🗺️ The changes in this file address [a `clippy` complaint](https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_struct_initialization) about using struct builder initialization syntax without overriding any of the base struct's fields.

src/text/raw_text_writer.rs
```diff
@@ -400,7 +395,8 @@ impl<W: Write> RawTextWriter<W> {
     pub fn add_annotation<A: AsRawSymbolTokenRef>(&mut self, annotation: A) {
         // TODO: This function currently allocates a new string for each annotation.
         //       It will be common for this text to come from the symbol table; we should
-        //       make it possible to pass an Rc<str> or similar when applicable.
+        //       make it possible to pass an Arc<str> or similar when applicable.
+        //       See: https://github.com/amazon-ion/ion-rust/issues/496
```
zslayton, line 399: 🗺️ I noticed this `TODO` and opened an issue (#496) to track eventually implementing this optimization.

src/symbol.rs
```diff
@@ -4,13 +4,13 @@ use std::borrow::Borrow;
...
 use std::hash::{Hash, Hasher};
-use std::rc::Rc;
+use std::sync::Arc;
 
 /// Stores or points to the text of a given [Symbol].
 #[derive(Debug, Eq)]
 enum SymbolText {
     // This Symbol refers to a string in the symbol table
-    Shared(Rc<str>),
+    Shared(Arc<str>),
```
zslayton, line 13: 🗺️ Multiple users have expressed a desire to be able to store an `Element` in an `Arc` or send an `Element` across threads, which is currently impossible because `Rc` does not implement `Send`.

Using `Arc` imposes a higher overhead cost when creating or dropping `Symbol`s, but makes `Element`s much more flexible and eliminates a source of surprise from the API.

zslayton (commented):
🗺️ PR tour

popematt (approved):
Just curious—is it possible to use Rc and implement Send ourselves for Symbol such that when the Symbol is being moved across threads, it just copies the underlying String? That might be a way to stay thread-safe without having to accept the overhead of Arc for single-threaded use.

zslayton:
> Just curious—is it possible to use Rc and implement Send ourselves for Symbol such that when the Symbol is being moved across threads, it just copies the underlying String? That might be a way to stay thread-safe without having to accept the overhead of Arc for single-threaded use.

That's a good thought, but [`Send` is a marker trait](https://doc.rust-lang.org/std/marker/trait.Send.html); it doesn't offer any methods we can implement to customize that behavior. I also suspect (but don't have data to prove) that allocating a new `String` for the copy would be more expensive than `Arc::clone()`, which is just incrementing an atomic integer.

jobarr-amzn (approved):

PR #499, opened 2023-03-29
Implements `IntoIterator` for container Elements

zslayton:
This PR adds implementations of `IntoIterator` for each of the container `Element` types, allowing the expected looping syntax:

```rust
for element in &list {
    // ...
}

for element in &sexp {
    // ...
}

for (field_name, value) in &struct_ {
    // ...
}
```

It also moves the container types to their own modules for easier navigation.

Each commit is self-contained. Only the final commit (`Implements IntoIterator...`) contains any new logic. The rest are mechanical refactorings. I recommend reviewing them in order.

### DIFF
--- a/src/element/struct.rs
+++ b/src/element/struct.rs
@@ -1,0 +1,203 @@
+use crate::element::builders::StructBuilder;
+use crate::element::iterators::{FieldIterator, FieldValuesIterator, IndexVec};
+use crate::element::Element;
+use crate::ion_eq::IonEq;
+use crate::symbol_ref::AsSymbolRef;
+use crate::text::text_formatter::IonValueFormatter;
+use crate::Symbol;
+use std::collections::HashMap;
+use std::fmt::{Display, Formatter};
+
+// This collection is broken out into its own type to allow instances of it to be shared with Arc/Rc.
+#[derive(Debug, Clone)]
+struct Fields {
+    // Key/value pairs in the order they were inserted
+    by_index: Vec<(Symbol, Element)>,
+    // Maps symbols to a list of indexes where values may be found in `by_index` above
+    by_name: HashMap<Symbol, IndexVec>,
+}
+
+impl Fields {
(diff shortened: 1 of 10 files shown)


src/element/struct.rs
```diff
@@ -0,0 +1,203 @@
...
+        }
+    }
+
+    /// Iterates over all of the (field name, field value) pairs in the struct.
+    fn iter(&self) -> impl Iterator<Item = &(Symbol, Element)> {
+        self.by_index.iter()
+    }
+}
+
+/// An in-memory representation of an Ion Struct
```
desaikd, line 78: Should we add macro examples here as well?

src/element/struct.rs
```diff
@@ -201,3 +209,22 @@ impl PartialEq for Struct {
...
 impl Eq for Struct {}
+
+#[cfg(test)]
+mod tests {
+    use crate::element::Element;
+    use crate::ion_struct;
+
+    #[test]
+    fn for_field_in_struct() {
+        // Simple example to exercise List's implementation of IntoIterator
```
desaikd, line 220:
```suggestion
        // Simple example to exercise Struct's implementation of IntoIterator
```

desaikd (approved):
🚀

src/element/struct.rs
```diff
@@ -0,0 +1,230 @@
...
+
+// Allows `for (name, value) in &my_struct {...}` syntax
+impl<'a> IntoIterator for &'a Struct {
+    type Item = (&'a Symbol, &'a Element);
+    type IntoIter = FieldIterator<'a>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.iter()
+    }
+}
```
zslayton, line 172: 🗺️ This impl (and its unit test) is the only new code in this module. Everything else was moved here from `mod`.

zslayton (commented):
🗺️ PR tour

PR #501, opened 2023-03-30
Custom string type

zslayton:
This PR introduces a new `Str` type that replaces all usages of
Rust's owned `String` type in the `Reader` and `Element` APIs.

For the time being, the `Str` type is a simple opaque wrapper
around `String` with trait implementations that allow it to be
used in much the same way. Having an opaque wrapper will allow
us to swap out the implementation later without a breaking
change, enabling us to provide stack-allocated small strings,
shared strings (interning), etc as desired.

The name `Str` was selected to:
1. Avoid collisions with `String` and `&str`.
2. Use (an abbreviation of) the same type name as Ion itself (vs, say, `Text`)

We can add more trait implementations and method pass-throughs
as needed.

### DIFF
--- a/src/types/string.rs
+++ b/src/types/string.rs
@@ -10,7 +10,7 @@ use std::fmt::{Display, Formatter};
 /// ```
 #[derive(Eq, Ord, PartialOrd, Debug, Clone, Hash)]
 pub struct Str {
-    // For the time being, the `Text` type is an opaque wrapper around the standard Rust `String`
+    // For the time being, the `Str` type is an opaque wrapper around the standard Rust `String`
     // type. Having this opaque wrapper means that we can swap out its implementation without a
     // breaking change, allowing us to offer string interning, stack-allocated small strings, or
     // other optimizations as needed.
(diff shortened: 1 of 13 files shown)


src/types/string.rs
zslayton, line 1: 🗺️ This file/type is the only new logic in the PR. The rest of the files are mechanical refactorings to use `Str` instead of `std::string::String`.

zslayton (commented):
🗺️ PR tour

src/types/string.rs
```diff
@@ -0,0 +1,114 @@
...
+    pub fn len(&self) -> usize {
+        self.text.len()
+    }
+
+    /// Returns `true` if this is the empty string (`""`); otherwise, returns `false`.
+    ///
+    /// ```
+    /// use ion_rs::Str;
+    /// let s: Str = "hello!".into();
+    /// assert_eq!(s.len(), 6);
```
desaikd, line 40: This should be `is_empty` examples.

desaikd (approved):

jobarr-amzn:
I also like that we're not using `Text` as a name here because we use that elsewhere to describe the commonality of Symbols *and* Strings, or to contrast with the binary Ion encoding. One more use as name of the custom String type would be too much.

jobarr-amzn (approved):

Issue #500, opened 2023-03-30
Value/Content APIs Factoring to Capture Structurally Similar Content Types


almann:
### Problem Statement

Ion contains types that are structurally identical, but differ only in their _type_.  In our current factoring of `Value`, we end up creating variants that are basically the same content, but differ only in the type of `struct` used in the variant.  This leads to some code duplication and traits, that require generic code to be used if one wants to operate across these structurally identical types.  I believe we could "push down" the value variants as a way to factor the API to avoid this boilerplate.

This is a distillation of my thinking after an offline thread with @zslayton.

### Background

In the original `Element`/`Value` APIs the `Value` was the `enum` representing _type_ and its _contents_ of the value were modeled directly in the `enum`.  The following is the original factoring abbreviated for illustration purposes:

```rs
struct Bytes {
    bytes: Vec<u8>,
}

struct Sequence {
    children: Vec<Element>,
}

enum Value {
    Int(Int),
    Float(f64),
    Decimal(Decimal),
    Timestamp(Timestamp),
    String(String),
    Symbol(Symbol),
    Bool(bool),
    Blob(Bytes),
    Clob(Bytes),
    SExp(Sequence),
    List(Sequence),
    Struct(Struct),
}
```

The latest refactoring of these APIs have captured the type within the _content_ for the variants that are useful as intermediaries for the various APIs that build up these types and need to capture the intent of the user (e.g., I would like to build this sequence that is to become a `Value::List(...)`)  this need is captured for example in the [builder APIs and macros PR][1].  This factoring looks something like this (again, abbreviated/digested for illustration purposes):

```rs
struct Blob {
    bytes: Vec<u8>,
}

struct Clob {
    bytes: Vec<u8>,
}

// trait to allow us to generically operate across Blob/Clob
trait Lob { ... }

struct SExp {
    children: Vec<Element>,
}

struct List {
    children: Vec<Element>,
}

// trait to allow us to generically operate across SExp/List
// called IonSequence in the real APIs
trait Sequence { ... }

enum Value {
    Int(Int),
    Float(f64),
    Decimal(Decimal),
    Timestamp(Timestamp),
    String(String),
    Symbol(Symbol),
    Bool(bool),
    Blob(Blob),
    Clob(Clob),
    SExp(SExp),
    List(List),
    Struct(Struct),
}
```

[...]

PR #502, opened 2023-03-31
Makes `List` and `SExp` thin wrappers around `Sequence`

zslayton:
See issue #500 for discussion.

This PR removes the `IonSequence` trait and reintroduces a concrete `Sequence` type. It also converts `List` and `SExp` into tuple structs that wrap `Sequence`.

`ListBuilder` and `SExpBuilder` have also been removed in favor of a single `SequenceBuilder` implementation with `build_list()` and `build_sexp()` methods.

`List` and `SExp` each implement `Deref<Target=Sequence>`, allowing them both to expose the functionality of their underlying `Sequence` of elements transparently.

This eliminates a fair amount of boilerplate code that was needed to provide functionality for both types. It also enables unified handling for `List` and `SExp` in match expressions in situations where the author does not care about the Ion type distinction. Finally, users no longer have to import `IonSequence` in order to access the basic functionality of `List` and `SExp`.

### DIFF
--- a/src/element/builders.rs
+++ b/src/element/builders.rs
@@ -1,30 +1,64 @@
-use crate::element::{Element, Struct};
+use crate::element::{Element, Sequence, Struct};
 use crate::Symbol;
 
-/// Constructs [List] values incrementally.
+/// Constructs [Sequence], [List], and [SExp] values incrementally.
 ///
+/// Building a [Sequence]:
 /// ```
-/// use ion_rs::element::Element;
-/// let actual: Element = Element::list_builder()
+/// use ion_rs::element::{Element, Sequence};
+/// let actual: Sequence = Sequence::builder()
 ///     .push(1)
 ///     .push(true)
 ///     .push("foo")
-///     .build()
-///     .into();
-/// let expected = Element::read_one(r#"[1, true, "foo"]"#).unwrap();
+///     .build();
+/// let expected: Sequence = Sequence::new([
+///   Element::integer(1),
+///   Element::boolean(true),
+///   Element::string("foo"),
+/// ]);
 /// assert_eq!(actual, expected);
 /// ```
-pub struct ListBuilder {
+/// Building a [List]:
+/// ```
+/// use ion_rs::element::{Element, List, Sequence};
+/// let actual: List = Sequence::builder()
+///     .push(1)
+///     .push(true)
+///     .push("foo")
+///     .build_list();
+/// let expected: List = List(Sequence::new([
+///   Element::integer(1),
+///   Element::boolean(true),
+///   Element::string("foo"),
+/// ]));
+/// assert_eq!(actual, expected);
+/// ```
+/// Building a [SExp]:
+/// ```
+/// use ion_rs::element::{Element, SExp, Sequence};
+/// let actual: SExp = Sequence::builder()
+///     .push(1)
+///     .push(true)
+///     .push("foo")
+///     .build_sexp();
+/// let expected: SExp = SExp(Sequence::new([
+///   Element::integer(1),
+///   Element::boolean(true),
+///   Element::string("foo"),
+/// ]));
+/// assert_eq!(actual, expected);
+/// ```
+pub struct SequenceBuilder {
     values: Vec<Element>,
 }
 
-impl ListBuilder {
-    /// Crate visible; users should call [`List::builder()`] or [`Element::list_builder()`] instead.
+impl SequenceBuilder {
+    /// Crate visible; users should call [`Sequence::builder()`] instead.
     pub(crate) fn new() -> Self {
         Self { values: Vec::new() }
     }
 
-    /// Helper method for [`List::clone_builder()`].
+    /// Helper method for [`Sequence::clone_builder()`].
     pub(crate) fn with_initial_elements(elements: &[Element]) -> Self {
         let mut new_elements = Vec::with_capacity(elements.len());
         new_elements.extend_from_slice(elements);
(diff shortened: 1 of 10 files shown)


src/element/builders.rs
```diff
@@ -101,9 +82,19 @@ impl SExpBuilder {
...
+    /// Builds a [`List`] with the previously specified elements.
+    pub fn build_list(self) -> List {
+        List(self.build())
+    }
+
     /// Builds a [`SExp`] with the previously specified elements.
-    pub fn build(self) -> SExp {
-        SExp::new(self.values)
+    pub fn build_sexp(self) -> SExp {
+        SExp(self.build())
```
zslayton, line 97: 🗺️ The unified `SequenceBuilder` type now offers `build()`, `build_list()`, and `build_sexp()` methods that construct the expected types.

src/element/list.rs
```diff
@@ -19,21 +17,25 @@ use std::fmt::{Display, Formatter};
...
 
-    pub fn builder() -> ListBuilder {
-        ListBuilder::new()
+impl Deref for List {
+    type Target = Sequence;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
     }
+}
```
zslayton, line 34: 🗺️ This impl allows users to transparently access the underlying `Sequence`'s methods.

zslayton (commented):
🗺️ PR tour

src/element/list.rs
popematt: [`Deref` polymorphism](https://rust-unofficial.github.io/patterns/anti_patterns/deref.html) is considered an anti-pattern. Can you explain why you went with this approach instead of any of the alternatives?

zslayton: I discussed this [in our design thread on issue 500](https://github.com/amazon-ion/ion-rust/issues/500#issuecomment-1491885057).

popematt: Oh, I missed that. I'll go take a look.

zslayton: I just pushed a commit to replace `Deref` with `delegate!`. The result is slightly more verbose, but it's probably worth it to sidestep potential controversy.

src/element/list.rs
```diff
@@ -19,21 +19,48 @@ use std::fmt::{Display, Formatter};
...
+    }
+
+    delegate! {
+        to self.0 {
+            pub fn clone_builder(&self) -> SequenceBuilder;
+            pub fn elements(&self) -> ElementsIterator<'_>;
+            pub fn get(&self, index: usize) -> Option<&Element>;
+            pub fn len(&self) -> usize;
+            pub fn is_empty(&self) -> bool;
+        }
```
popematt, line 50: Nice! I actually quite like this.

src/element/list.rs
```diff
@@ -19,21 +19,48 @@ use std::fmt::{Display, Formatter};
...
+//         &self.0
+//     }
+// }
+
 impl List {
-    pub(crate) fn new(children: Vec<Element>) -> Self {
-        Self { children }
+    pub fn builder() -> SequenceBuilder {
+        Sequence::builder()
+    }
```
popematt, line 41: It is a little unfortunate that you can write this code now:
```rust
List::builder().build_sexp();
```
Would it be worth putting a type parameter into `SequenceBuilder` to prevent this from happening? I'm putting a comment over by SequenceBuilder with an example of what I mean.

src/element/builders.rs
```diff
@@ -1,83 +1,64 @@
...
 /// assert_eq!(actual, expected);
 /// ```
-pub struct SExpBuilder {
+pub struct SequenceBuilder {
     values: Vec<Element>,
 }
 
-impl SExpBuilder {
-    /// Crate visible; users should call [`SExp::builder()`] or [`Element::sexp_builder()`] instead.
+impl SequenceBuilder {
```
popematt, line 55: If we parameterize `SequenceBuilder`, the return type of `build()` can be set when the SequenceBuilder is created so that you can't write `List::builder().build_sexp()`.
```rust
struct SequenceBuilder<T> {
    _target_type: Phantom<T>,
    values: Vec<Element>,
}

impl <T> SequenceBuilder<T> {
    /// Crate visible; users should call [`Sequence::builder()`] instead.
    pub(crate) fn new() -> Self {
        Self { values: Vec::new() }
    }

    /// Helper method for [`Sequence::clone_builder()`].
    pub(crate) fn with_initial_elements(elements: &[Element]) -> Self {
        let mut new_elements = Vec::with_capacity(elements.len());
        new_elements.extend_from_slice(elements);
        Self {
            values: new_elements,
        }
    }

    // etc.
}

// Repeate for Sequence and SExp
impl SequenceBuilder<List> {
    pub fn build(self) -> List {
        List(self.values.into())
    }
}
```

src/element/list.rs
zslayton: Oof, yeah, that's unfortunate. Good catch.

src/element/builders.rs
zslayton: This would work, though I think my current leaning would be to simply eliminate the `List::builder() -> SequenceBuilder` and `SExp::builder() -> SequenceBuilder` methods in favor of the (already available) `Sequence::builder()`.

The doc comment examples for `List` and `SExp` can demonstrate using `Sequence::builder()` to construct them, and most users will reach for the macros anyway.

An interesting (minor) benefit of using the same builder type for all sequences is that I can do:
```rust
let sexp = ion_sexp!(1 2 3);
// I can make a List using the SExp's elements as a starting point
let list = sexp.clone_builder().build_list();
```

It's possible that we'll want to introduce a `Stream` sequence type--values with no container, which are used at the top level and (in Ion 1.1) template expansion. We would also be able to use `SequenceBuilder` for that.

popematt: > An interesting (minor) benefit of using the same builder type for all sequences is that I can do:
> ```rust
> let list = sexp.clone_builder().build_list();
> ```

We could use the generics so that `List::builder()` only lets you build a list, but `list.clone_builder()` will let you build a `List`, `SExp`, or `Sequence`. E.g.:
```rust
impl SequenceBuilder<Cloned> {
    pub fn build_list(self) -> List {
        List(self.values.into())
    }
    pub fn build_sexp(self) -> SExp {
        List(self.values.into())
    }
    // Etc.
}
```

That being said, I think the solution you've proposed is fine (and certainly simpler).

zslayton:
@almann I think the change in this PR gets us closer to what you're after. I'm going to go ahead and merge this, but I won't close #500; we can continue to discuss it.

PR #505, opened 2023-04-04
Makes `Value::List` and `Value::SExp` wrap `Sequence`

zslayton:
This PR partially addresses issue #500.

It modifies the `Value` enum's `List` and `SExp` variants from:

```rust
enum Value {
  // ...
  List(List),
  SExp(SExp),
  // ...
}
```

to

```rust
enum Value {
  // ...
  List(Sequence),
  SExp(Sequence),
}
```

while also preserving the standalone `element::List(Sequence)` and `element::SExp(Sequence)` types, which are used primarily by the SequenceBuilder, `ion_list!`, and `ion_sexp!`.

Following this change, matching sequence types is flattened/simplified:
```rust
if let List(seq) | SExp(seq) = value {
  // iterate over `seq`
}
```
but we retain the ability to construct strongly typed standalone values when necessary, such as in the `ion_list!`
and `ion_sexp!` macros.

```rust
// Constructs an element::List(Sequence)
let list: List = ion_list![1, 2, 3];
// Discards the outer wrapper element::List, moving the Sequence into a `Value::List(...)`.
let value: Value = list.into(); // Discards the outer `List`
```

### DIFF
--- a/src/element/list.rs
+++ b/src/element/list.rs
@@ -77,6 +77,12 @@ impl From<Sequence> for List {
     }
 }
 
+impl From<List> for Sequence {
+    fn from(value: List) -> Self {
+        value.0
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use crate::ion_list;
--- a/src/element/mod.rs
+++ b/src/element/mod.rs
@@ -66,8 +66,8 @@ pub enum Value {
     Bool(bool),
     Blob(Vec<u8>),
     Clob(Vec<u8>),
-    SExp(SExp),
-    List(List),
+    SExp(Sequence),
+    List(Sequence),
     Struct(Struct),
 }
 
@@ -86,8 +86,8 @@ impl Display for Value {
             Value::Clob(clob) => ivf.format_clob(clob),
             Value::Blob(blob) => ivf.format_blob(blob),
             Value::Struct(struct_) => ivf.format_struct(struct_),
-            Value::SExp(sexp) => ivf.format_sexp(sexp),
-            Value::List(list) => ivf.format_list(list),
+            Value::SExp(sequence) => ivf.format_sexp(sequence),
+            Value::List(sequence) => ivf.format_list(sequence),
         }
         .map_err(|_| std::fmt::Error)?;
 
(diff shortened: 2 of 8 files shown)


src/element/list.rs
```diff
@@ -77,6 +77,12 @@ impl From<Sequence> for List {
     }
 }
 
+impl From<List> for Sequence {
+    fn from(value: List) -> Self {
+        value.0
+    }
+}
```
zslayton, line 84: 🗺️ This `impl` allows a `List(Sequence)` to discard the type wrapper and become a `Sequence`. For now, it's used to convert `List(Sequence)` to `Value::List(Sequence)`.

src/element/mod.rs
```diff
@@ -436,21 +436,7 @@ impl Element {
...
-    }
-
-    pub fn as_sexp(&self) -> Option<&SExp> {
-        match &self.value {
-            Value::SExp(sexp) => Some(sexp),
-            _ => None,
-        }
-    }
-
-    pub fn as_list(&self) -> Option<&List> {
```
zslayton, line 451: 🗺️ Now that a `Value` doesn't contain a `List` or `SExp`, we can only lend out a `&Sequence`.

We could retain these methods and have them continue to return `&Sequence` when the type matches. Open to thoughts there.

src/element/sequence.rs
```diff
@@ -39,6 +39,12 @@ impl Sequence {
     }
 }
 
+impl AsRef<Sequence> for Sequence {
```
zslayton, line 42: 🗺️ Allows methods that take an `AsRef<Sequence>` to accept a `List`, `SExp`, or `Sequence`.

src/text/text_formatter.rs
```diff
@@ -417,31 +417,33 @@ impl<'a, W: std::fmt::Write> IonValueFormatter<'a, W> {
...
-    pub(crate) fn format_list(&mut self, value: &List) -> IonResult<()> {
+    pub(crate) fn format_list<S: AsRef<Sequence>>(&mut self, sequence: S) -> IonResult<()> {
         write!(self.output, "[")?;
-        let mut peekable_itr = value.elements().peekable();
+        self.format_sequence_elements(sequence, ", ")?;
+        write!(self.output, "]")?;
+        Ok(())
+    }
+
+    fn format_sequence_elements<S: AsRef<Sequence>>(
```
zslayton, line 434: 🗺️ While I was here, I broke out the common sequence formatting functionality into a helper method.

zslayton (commented):
🗺️ PR tour

src/element/mod.rs
popematt: I think we should leave them out in favor of narrowing conversions. Let's not add too many ways to do the same thing unless it's absolutely necessary to do so.

popematt (approved):

PR #506, opened 2023-04-05
Introduces `Bytes`, `Blob`, and `Clob` wrapper types

### LINKED ISSUE #500: Value/Content APIs Factoring to Capture Structurally Similar Content Types

zslayton:
This PR introduces new opaque wrappers for `Vec<u8>`, which was used throughout the APIs to represent owned blobs and clobs.

`Bytes` captures the core functionality of an owned, immutable byte array, but does not include Ion type information.

`Blob` and `Clob` are both trivial tuple struct wrappers around `Bytes`, adding an Ion type to the data.

The `Value::Blob` and `Value::Clob` enum variants (which already provide an Ion type) now wrap an instance of `Bytes`.

This PR fixes #500.


src/binary/non_blocking/raw_binary_reader.rs
```diff
@@ -769,8 +770,8 @@ impl<A: AsRef<[u8]>> IonReader for RawBinaryBufferReader<A> {
         self.read_symbol_id().map(RawSymbolToken::SymbolId)
     }
 
-    fn read_blob(&mut self) -> IonResult<Vec<u8>> {
-        self.read_blob_bytes().map(Vec::from)
+    fn read_blob(&mut self) -> IonResult<Blob> {
+        self.read_blob_bytes().map(Vec::from).map(Blob::from)
```
zslayton, line 774: 🗺️ The `IonReader` trait's blob- and clob- related methods had to be updated to return `Blob` and `Clob`.

Before:
```rust
fn read_blob(&mut self) -> IonResult<Vec<u8>>;
fn read_clob(&mut self) -> IonResult<Vec<u8>>;
```

After:
```rust
fn read_blob(&mut self) -> IonResult<Blob>;
fn read_clob(&mut self) -> IonResult<Clob>;
```

This change had to be made in each implementation of `IonReader`, of which there are several.

src/element/lob.rs
```diff
@@ -0,0 +1,116 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Clob(pub Bytes);
+
+impl Clob {
+    pub fn as_slice(&self) -> &[u8] {
+        self.as_ref()
+    }
+}
+
+impl AsRef<[u8]> for Blob {
```
zslayton, line 44: 🗺️ There are a lot of trait impls here that feel repetitive. I'm confident that we can represent them more concisely if needed (perhaps via a macro or a `Lob` trait with blanket implementations), but I want to leave that for a future PR.

src/element/mod.rs
```diff
@@ -64,8 +69,8 @@ pub enum Value {
     String(Str),
     Symbol(Symbol),
     Bool(bool),
-    Blob(Vec<u8>),
-    Clob(Vec<u8>),
+    Blob(Bytes),
+    Clob(Bytes),
```
zslayton, line 73: 🗺️ Notice that we can still handle blob and clob values in a single `match`/`if let` arm:
```rust
if let Blob(b) | Clob(b) = value {
  // ...do stuff with `b`...
}
```

src/element/mod.rs
```diff
@@ -176,7 +181,21 @@ impl From<&[u8]> for Value {
...
+        let bytes: Bytes = blob.into();
+        Value::Blob(bytes)
+    }
+}
+
+impl From<Clob> for Value {
+    fn from(clob: Clob) -> Self {
+        let bytes: Bytes = clob.into();
+        Value::Clob(bytes)
     }
```
zslayton, line 199: 🗺️ The standalone `Blob` and `Clob` types can be trivially widened to `Value`.

src/element/reader.rs
```diff
@@ -147,8 +147,8 @@ impl<'a, R: IonReader<Item = StreamItem, Symbol = Symbol>> ElementLoader<'a, R>
                     Timestamp => Value::Timestamp(self.reader.read_timestamp()?),
                     Symbol => Value::Symbol(self.reader.read_symbol()?),
                     String => Value::String(self.reader.read_string()?),
-                    Clob => Value::Clob(self.reader.read_clob()?),
-                    Blob => Value::Blob(self.reader.read_blob()?),
+                    Clob => Value::Clob(self.reader.read_clob()?.into()),
+                    Blob => Value::Blob(self.reader.read_blob()?.into()),
```
zslayton, line 151: 🗺️ The `.into()` here is converting the lobs into their inner `Bytes` representations. From a memory layout perspective, this is a no-op as `Bytes` is the only field in `Blob`/`Clob`.

zslayton (commented):
🗺️ PR tour

popematt (approved):

PR #508, opened 2023-04-06
Introduces an `Annotations` type

zslayton:
This PR replaces Element's `Vec<Symbol>` representation of an annotations sequence with a new type: `Annotations`. This layer of indirection will allow us to experiment with alternative data structures in the future.


src/binary/non_blocking/raw_binary_reader.rs
```diff
@@ -1283,7 +1283,7 @@ impl<'a, A: AsRef<[u8]>> TxReader<'a, A> {
 #[cfg(test)]
 mod tests {
     use crate::binary::non_blocking::raw_binary_reader::RawBinaryBufferReader;
-    use crate::text::text_value::IntoAnnotations;
+    use crate::text::text_value::IntoRawAnnotations;
```
zslayton, line 1286: 🗺️ The `IntoAnnotations` trait name was a bit misleading once I introduced an `Annotations` type. I have renamed it `IntoRawAnnotations`, reflecting the fact that it accepts raw symbol tokens.

src/element/annotations.rs
```diff
@@ -0,0 +1,110 @@
...
+/// use ion_rs::element::Annotations;
+/// let annotations: Annotations = ["foo", "bar", "baz"].into();
+/// for annotation in &annotations {
+///     assert_eq!(annotation.text().map(|s| s.len()), Some(3));
+/// }
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Annotations {
+    symbols: Vec<Symbol>,
+}
```
zslayton, line 16: 🗺️ Our new `Annotations` type.

src/element/annotations.rs
```diff
@@ -0,0 +1,110 @@
...
+    /// assert!(annotations.contains("bar"));
+    /// assert!(annotations.contains("baz"));
+    ///
+    /// assert!(!annotations.contains("quux"));
+    /// assert!(!annotations.contains("quuz"));
+    /// ```
+    pub fn contains<S: AsRef<str>>(&self, query: S) -> bool {
+        let query: &str = query.as_ref();
+        self.iter().any(|symbol| symbol.text() == Some(query))
+    }
```
zslayton, line 73: 🗺️ While the implementations of `contains()` and `starts_with()` are not complex, they are needed commonly enough that providing convenience functions for them seems reasonable.

src/element/element_stream_reader.rs
```diff
@@ -179,7 +179,7 @@ impl IonReader for ElementStreamReader {
         let iterator = self
             .current_value
             .as_ref()
-            .map(|value| value.annotations())
+            .map(|value| value.annotations().into_iter())
```
zslayton, line 182: 🗺️ `value.annotations()` now returns an `&Annotations` instead of an iterator.

src/element/mod.rs
```diff
@@ -256,13 +256,16 @@ impl IonEq for Element {
...
 pub struct Element {
-    annotations: Vec<Symbol>,
+    annotations: Annotations,
     value: Value,
 }
 
 impl Element {
-    pub fn new(annotations: Vec<Symbol>, value: Value) -> Self {
-        Self { annotations, value }
+    pub(crate) fn new(annotations: Annotations, value: impl Into<Value>) -> Self {
```
zslayton, line 264: 🗺️ I've restricted access to the constructors for `Element` and `Annotations` to encourage users to reach for the more flexible builders and constructor methods. We can reintroduce it later if needed.

zslayton (commented):
🗺️ PR tour

src/element/mod.rs
```diff
@@ -256,13 +256,16 @@ impl IonEq for Element {
 /// An `(annotations, value)` pair representing an Ion value.
 #[derive(Debug, Clone)]
 pub struct Element {
-    annotations: Vec<Symbol>,
+    annotations: Annotations,
```
desaikd, line 259: (Question) Should we rather have `Annotation` type defined which will be a `Symbol` and then we can return a `Vec<Annotation>` for `Element::annotations()` and one can perform all the `Vec` related operations on it?
I guess my question is which part are we trying to make abstract, do we want to make the entire `Vec` opaque or just provide something like `Annotation`?

zslayton: The goal is to abstract away the `Vec` itself. On the stack, a `Vec` is 24 bytes on a 64-bit machine:
* 8 bytes for its current length (`usize`)
* 8 bytes for its capacity (`usize`)
* 8 bytes for a pointer to its array on the heap.

An `Element` is a `(Value, Annotations)` pair, which means that its stack size is 24 bytes + the size of a `Value`. Trimming down `Annotations` would also trim down `Element`, making it faster to construct/move. (See the [_Type Sizes_ section](https://nnethercote.github.io/perf-book/type-sizes.html) of the Rust performance book for more details.)

Most Ion values don't have any annotations; I believe that replacing `Vec` with something like `ThinVec` would be beneficial. Of course, we would need to write some benchmarks to be sure. For now, I'm just wrapping `Vec` in an opaque type so we can swap out its implementation later without causing problems.

> Should we rather have Annotation type defined which will be a Symbol

Having an `Annotation` type is an interesting idea, though I imagine its meaning will always be very closely aligned with the `Symbol` type. (`Symbol` is a custom type of ours anyway, so we can already modify its implementation if needed.)

src/element/annotations.rs
```diff
@@ -0,0 +1,110 @@
...
+    /// values.
+    /// ```
+    /// use ion_rs::element::Annotations;
+    /// let annotations: Annotations = ["foo", "bar", "baz"].into();
+    /// assert!(annotations.starts_with("foo"));
+    /// assert!(!annotations.starts_with("quux"));
+    /// ```
+    pub fn starts_with(&self, text: impl AsRef<str>) -> bool {
+        self.iter().next().and_then(|a| a.text()) == Some(text.as_ref())
+    }
```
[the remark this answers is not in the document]
zslayton: > I agree that inspecting the first annotation is probably the most common thing to do with the annotations. However, I think people might want to be able to match on the first annotation. It is possible, but very unwieldy to match using a boolean function [....]

That's a good thought! I can get behind `first()`. 👍 

> By forcing the result to be a bool, I think that there is the slight possibility of users introducing subtle bugs when there is unknown symbol text. I say "slight" because I think unknown symbol text is uncommon in real world data, but if you do have unknown symbol text, it would be very easy to accidentally do the wrong thing with this function. Really, there are three possible outcomes—true, false, and indeterminate (in the case of unknown symbol text).

`ion-rust` makes a conscious decision to (at the user level) treat symbols whose text is unknown due to missing context as a hard error. This is different approach from `ion-java`, where it's possible to forge ahead in the absence of a shared symbol table if you do so with great caution.

`ion-rust` _cannot_, however, treat `$0` as an error as it is explicitly part of the domain of the `symbol` type. (i.e. Its text is _known to be unknown_, it's not unknown due to circumstance.)

By accepting an `AsRef<str>`, these convenience APIs are choosing not to support the `$0` case. (We could add support for it later with some trait work if we wanted to.) If the user asks "Is the first annotation `dog`?", the answer is `true` or `false`. If the first annotation is `$0`, it's not equivalent to the symbol 'dog'.

src/element/mod.rs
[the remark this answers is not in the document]
zslayton: Great question. For now it isn't because it's not because there isn't an efficient implementation of `From<Vec<Symbol>> for Annotations`. This is due to use of a blanket trait impl for an `IntoIterator<S: Into<Symbol>>` that allows us to accept slices, arrays, vecs, and other containers of `&str`, `String`, `Symbol`, etc. I have a detailed comment on the existing trait impl.

Requiring an `Annotations` means that callers will need to build it themselves using `pub(crate) fn Annotations::new()`, which takes a `Vec<Symbol>`.

src/element/annotations.rs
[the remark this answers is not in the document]
popematt: > ion-rust cannot, however, treat $0 as an error as it is explicitly part of the domain of the symbol type. (i.e. Its text is known to be unknown, it's not unknown due to circumstance.)

This is one of the little corners of Ion that still makes my head spin sometimes, but you've helped me understand why `starts_with` and `contains` are not problematic.

src/element/annotations.rs
```diff
@@ -0,0 +1,110 @@
...
+
+// NB: This blanket implementation covers a LOT of type combinations:
+//
+//     A {slice, array, Vec...} of {&str, String, Symbol, ...}
+//
+// Unfortunately, it also prevents us from having special cases for Vec<Symbol>
+// (which would otherwise be able to be converted directly to an `Annotations` without cloning)
+// or for a single value that could be turned into a `Symbol`. If we need that ability in the
+// future, consider defining a custom `IntoAnnotations` trait.
+impl<S: Into<Symbol>, I: IntoIterator<Item = S>> From<I> for Annotations {
```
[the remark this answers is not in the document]
zslayton: Hm, possibly! I'd like to leave that for another PR, however.

src/element/annotations.rs
```diff
@@ -17,13 +17,14 @@ pub struct Annotations {
...
+    // else (`thinvec`?) in the future. Users are unlikely to need to construct an `Annotations`
+    // themselves, but can use a `From` impl if necessary.
     pub(crate) fn new(symbols: Vec<Symbol>) -> Self {
         Annotations { symbols }
     }
 
-    // Constructs an Annotations object representing an empty symbol sequence
-    pub(crate) fn empty() -> Self {
+    /// Constructs an Annotations object representing an empty symbol sequence
+    pub fn empty() -> Self {
```
zslayton, line 27: I made this public for two reasons:
1. It's already possible to construct an `Annotations`, empty or otherwise, using a `From` implementation.
2. I needed an empty `Annotations` for the doc tests and making one via `From` required being annoyingly explicit.

popematt: 3. Constructing an empty `Annotations` does not (cannot?) expose any of the implementation details, so really, the question should be "why not?".

desaikd (approved):
LGTM!

src/element/annotations.rs
```diff
@@ -0,0 +1,124 @@
...
+    /// assert_eq!(annotations.first(), Some("foo"));
+    ///
+    /// let empty_sequence: Vec<&str> = vec![];
+    /// let annotations: Annotations = empty_sequence.into();
+    /// assert_eq!(annotations.first(), None);
+    ///
+    /// let annotations: Annotations = [Symbol::unknown_text()].into();
+    /// assert_eq!(annotations.first(), None)
+    /// ```
+    pub fn first(&self) -> Option<&str> {
```
popematt, line 97: I'm not thrilled with the fact that this treats "no annotations" and "the first annotation has unknown text" as equivalent, but I'm okay with it because it is explicitly called out in the documentation.

Would it be terrible to use `Result` instead of `Option` for this? Then we could distinguish them.
```rust
enum GetAnnotationError {
    NoSuchElement,
    UnknownSymbolText,
}

pub fn first(&self) -> Result<&str, AnnotationTextError> {
    // ...
}
```
Let me know what you think. I think my suggestion is an improvement (otherwise I wouldn't suggest it), but I'm willing to approve the PR as is if you disagree.

zslayton: I agree that it's more correct, but I think it only benefits those of us writing spec compliance tests. (i.e. Those of us using `$0`.) I think I'd prefer to keep it as-is and introduce a `first_symbol` or similar that returns `&Symbol` if we ever have demand for it.

popematt (approved):

PR #509, opened 2023-04-10
Remove map_<type> APIs and replace them with read_<type>

nirosys:
*Issue #, if available:* #335

*Description of changes:*
This PR removes the `map_`-style functions from the reader trait and replaces their use with `read_`-style functions.

`read_str` was added to the IonReader trait. The function already existed in the binary reader, so it was moved into the trait impl and the implemented for the other readers.

---


zslayton (approved):
👍

PR #515, opened 2023-04-21
Fixes incorrect handling of SID 0 as symbol text

popematt:
*Issue #, if available:*

None.

*Description of changes:*

* Changes the system symbol constants to be `Option<&str>` and changes SID `$0` to `None` rather than `"$0"`
* Propagate SID $0 through the text writers as a `RawSymbolToken` or `RawSymbolTokenRef` rather than forcing all symbols to be converted to text.
* Updates `round_trip_skip_list()` in `element_test_vector.rs` to no longer skip SID 0 test cases.


src/text/text_writer.rs
```diff
@@ -129,42 +129,42 @@ impl<W: Write> IonWriter for TextWriter<W> {
...
         };
-        self.raw_writer.write_symbol(text)
+        self.raw_writer.write_symbol(raw_symbol_token_ref)
     }
 
     fn set_field_name<A: AsRawSymbolTokenRef>(&mut self, name: A) {
-        let text = match name.as_raw_symbol_token_ref() {
-            RawSymbolTokenRef::SymbolId(symbol_id) => {
+        let raw_symbol_token_ref = match name.as_raw_symbol_token_ref() {
+            RawSymbolTokenRef::SymbolId(symbol_id) if symbol_id != 0 => {
```
zslayton, line 157: Same thought here RE: allowing `$0` and removing the `panic!`.

zslayton (commented):
Excellent! I see a couple of other things that would be nice to address while you're making changes to this code if you don't mind.

zslayton (approved):
🚀

PR #517, opened 2023-04-21
Adds IonData wrapper that uses Ion equivalence for equality

popematt:
**Issue #, if available:**

None

**Description of changes:**

* Adds `IonData`, a struct that can wrap anything that implements `IonEq`.
  * For `IonData<Element>`, I added some delegations to `Element` I thought would be required for it to be useful. There's the `read_*()` functions and `annotations()`, and `value()`.
* Adds `IonOrd`, a trait for a total ordering that is consistent with `IonEq`.
* Adds implementations of `IonOrd` for `Element`, `Value`, and others.
* Moves `IonEq` from `src/` to `src/ion_data/` so that all of the Ion-data-model-semantics stuff is in one place.
* Small, incidental changes:
  * Adds derive `PartialOrd` and `Ord` to `Bytes`
  * Moves `Element.ion_type()` implementation to `Value`; `Element.ion_type()` now delegates to `Value.ion_type()`
  * Adds `IntoIterator` impl for `&'a Sequence`

I have no illusions that this is perfect, but I want to get something started because we need this for handling things like distinct elements efficiently in `ion-schema-rust`.


src/element/mod.rs
```diff
@@ -77,6 +123,28 @@ pub enum Value {
     Struct(Struct),
 }
 
+impl Value {
+    pub fn ion_type(&self) -> IonType {
+        use Value::*;
```
almann, line 128: I am shocked this didn't already exist 😄

src/element/mod.rs
```diff
@@ -253,6 +321,31 @@ impl IonEq for Element {
     }
 }
 
+// Ordering is done as follows:
+// 1. Ion type -- because it is the cheapest comparison
```
almann, line 325: I have a stronger rationale than just expense--actually ordering the Ion types together makes logical sense for the natural ordering we're defining here.

src/ion_data/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+///
+/// [IonData] itself does not implement [IonEq] so that it is not possible to create, for example,
+/// `IonData<IonData<Element>>`.
+///
+/// [Hash] and [Ord] are not guaranteed to be implemented for all [IonData], but when they are, they
+/// are required to be consistent with [IonEq] (and [Eq]).
+///
+/// __WARNING__—The Ion specification does _not_ define a total ordering over all Ion values. Do
+/// not depend on getting any particular result from [Ord]. Use it only as an opaque total ordering
+/// over all [IonData]. _The algorithm used for [Ord] may change between versions._
```
almann, line 25: I am persnickety about liking doc comments around types (especially linked ones) to be in typewriter font: 

```suggestion
/// A wrapper for lifting "data" to "Ion data". Anything that implements [`IonEq`] can be converted
/// into [`IonData`]. As [`IonData`], the semantics of [`Eq`] are the same as [`IonEq`]—that is to say, [`Eq`]
/// follows the rules for Ion Equivalence in the Ion specification.
///
/// [`IonData`] itself does not implement [`IonEq`] so that it is not possible to create, for example,
/// `IonData<IonData<Element>>`.
///
/// [Hash] and [Ord] are not guaranteed to be implemented for all [`IonData`], but when they are, they
/// are required to be consistent with [`IonEq`] (and [`Eq`]).
///
/// __WARNING__—The Ion specification does _not_ define a total ordering over all Ion values. Do
/// not depend on getting any particular result from [`Ord`]. Use it only as an opaque total ordering
/// over all [`IonData`]. _The algorithm used for [`Ord`] may change between versions._
```

src/ion_data/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+use crate::{Element, IonResult};
+use std::cmp::Ordering;
+use std::fmt::{Display, Formatter};
+
+use crate::element::{Annotations, Value};
+use delegate::delegate;
+pub use ion_eq::IonEq;
+pub(crate) use ion_ord::IonOrd;
+
+/// A wrapper for lifting "data" to "Ion data". Anything that implements [IonEq] can be converted
```
almann, line 13: The first sentence doesn't really indicate the point of the class IMNSHO.  Maybe something like:

> A wrapper for lifting Ion compatible data into using Ion oriented comparisons (versus the Rust value semantics).  This enables the default semantics to be what a Rust user expects for native values, but allows a user to opt-in to Ion's structural equivalence/order.

src/ion_data/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+            pub fn annotations(&self) -> &Annotations;
+        }
+    }
+}
+
+impl From<IonData<Element>> for Element {
+    fn from(value: IonData<Element>) -> Self {
+        value.0
+    }
+}
```
almann, line 110: I'm not sure these are worth having TBH.  Why not just provide the trivial converters and avoid the special purpose code:

```rust
impl<T> From<T> for IonData<T> where T: IonEq {
    fn from(value: T) -> Self {
        IonData<T>(value)
    }
}

impl<T> From<IonData<T>> for T where T: IonEq {
    fn from(value: IonData<T>) -> Self {
        value.0
    }
}
```

Assuming no problems with blanket conversions here, the call sites should look very reasonable:

```rust
let v: IonData<_> = Element::read_one("500").into();
```

This also works similarly for anything else that happens to not be an `Element`.

src/lib.rs
```diff
@@ -179,10 +179,10 @@ mod blocking_reader;
 mod catalog;
 // Public as a workaround for: https://github.com/amazon-ion/ion-rust/issues/484
 pub mod constants;
-pub mod ion_eq;
 mod raw_symbol_token;
 mod raw_symbol_token_ref;
 // Public as a workaround for: https://github.com/amazon-ion/ion-rust/issues/484
+mod ion_data;
 pub mod reader;
```
almann, line 186: I think this comment is not aligned properly (`// Public as ...`)

src/lib.rs
```diff
@@ -209,6 +209,11 @@ pub use types::string::Str;
 pub use types::timestamp::Timestamp;
 pub use types::IonType;
 
+pub use ion_data::IonData;
+pub mod ion_eq {
+    pub use crate::ion_data::IonEq;
+}
```
almann, line 215: Why do we want a top-level trivial module for `ion_eq`?  Is this just to maintain backwards compatibility?  I would get rid of it if we can.

src/ion_data/ion_ord.rs
```diff
@@ -0,0 +1,262 @@
...
+    fn ion_cmp(&self, other: &Self) -> Ordering {
+        self.iter().cmp(other.iter())
+    }
+}
+
+impl IonOrd for f64 {
+    fn ion_cmp(&self, other: &Self) -> Ordering {
+        self.total_cmp(other)
+    }
+}
```
almann, line 30: It surprised me that `IonOrd` was not implemented (even trivially) for all the underlying value types.  I understand why now looking at `element::Value`, but I think `IonOrd` should be built on top of `IonOrd` and not a combination of `IonOrd` and `PartialOrd`.  If we don't like this, I think this really should be documented because what is unusual about this is that some of the underlying value types work with `IonData` and some don't.  E.g., I *think* I can  `IonData<f64>` but not `IonData<i64>` if I read this all correctly (and I may not be).

I was re-looking at `IonEq` and I think this is the source of the asymmetry.  Ideally we'd have blanket implementations and specializations for when `IonEq` doesn't apply (but specialization is not stable AFAIK), so I think it best to just comment the intention for `IonOrd`/`IonEq` **or** implement it for all of the constituent "value" types underlying the value enum.

almann (changes requested):

src/ion_data/mod.rs
popematt: I was hoping that you would have a better way of describing the purpose of this type. Thanks! 😉

src/lib.rs
popematt: Good catch.

src/lib.rs
popematt: This was just for backwards compatibility. I can get rid of it.

src/ion_data/ion_ord.rs
popematt: I can add trivial `IonEq` and `IonOrd` for all of the remaining value types, if you think that is also an appropriate solution.

src/ion_data/mod.rs
popematt: We already have `impl<T> From<T> for IonData<T> where T: IonEq` (see ~L45).

I can't do `impl<T> From<IonData<T>> for T where T: IonEq` because:
```
error[E0210]: type parameter `T` must be covered by another type when it appears before the first local type (`IonData<T>`)
  --> src/ion_data/mod.rs:50:6
   |
50 | impl<T> From<IonData<T>> for T where T: IonEq {
   |      ^ type parameter `T` must be covered by another type when it appears before the first local type (`IonData<T>`)
   |
   = note: implementing a foreign trait is only possible if at least one of the types for which it is implemented is local, and no uncovered type parameters appear before that first local type
   = note: in this case, 'before' refers to the following order: `impl<..> ForeignTrait<T1, ..., Tn> for T0`, where `T0` is the first and `Tn` is the last
```

However, I did take a page out of `Cell`'s book, and provide an `into_inner()` method as part of the generic implementation of `IonData`. This implementation of `From<IonData<Element>> for Element` is redundant and could be removed.

src/ion_data/mod.rs
popematt: Will do!

src/element/mod.rs
popematt: One could also make a case for annotations being the first thing to sort by. (I'm not going to argue that here, but I understand that as a point of view.) I can update the rationale.

src/ion_data/mod.rs
popematt: As for the `IonData<Element>::read_*()` functions, I think they do save you from having to repeat a lot of boilerplate elsewhere.
```rust
let elements = IonData::read_all(ion)?
// vs
let elements = Element::read_all(ion)?
            .into_iter()
            .map(IonData::from)
            .collect::<Vec<_>>()
```

...unless you can help me get a nice conversion trait impl going. I can't use `impl From<Vec<Element>> for Vec<IonData<Element>>` because of the "Only traits defined in the current crate can be implemented for arbitrary types" rule.

popematt:
@almann or @zslayton, can you recommend any good idioms for making this also work nicely with borrowed values (i.e. `&Element`)?

src/ion_data/mod.rs
```diff
@@ -1,24 +1,21 @@
...
 mod ion_ord;
 
-use crate::{Element, IonResult};
 use std::cmp::Ordering;
 use std::fmt::{Display, Formatter};
 
-use crate::element::{Annotations, Value};
-use delegate::delegate;
 pub use ion_eq::IonEq;
 pub(crate) use ion_ord::IonOrd;
```
almann: Something I think we should consider (and I am okay just opening an issue for this) is whether we actually need a public `IonEq` trait if `IonData` solves this.  The use case of `IonEq` is that users can apply Ion equality without conflation, `IonData` formalizes that relationship with order as well (and maybe eventually hashing).

I also think this has another benefit--end users cannot implement `IonEq` outside the crate--I see this as an effective reduction in API surface.

almann (approved):
:shipit: A couple questions below, but overall I am good.

src/ion_data/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+    }
+}
+
+impl<T: IonEq> Eq for IonData<T> {}
+
+impl<T: IonEq> From<T> for IonData<T> {
+    fn from(value: T) -> Self {
+        IonData(value)
+    }
+}
```
popematt, line 45: @almann, it's here.

almann (approved):
🚢

src/ion_data/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+/// not depend on getting any particular result from [Ord]. Use it only as an opaque total ordering
+/// over all [IonData]. _The algorithm used for [Ord] may change between versions._
+#[derive(Debug, Clone)]
+pub struct IonData<T: IonEq>(T);
+
+impl<T: IonEq> IonData<T> {
+    // Checks if two values are equal according to Ion's structural equivalence.
+    pub fn eq(a: &T, b: &T) -> bool {
+        T::ion_eq(a, b)
+    }
```
almann, line 30: It just occurred to me that something like this might be a bit more flexible (works on anything that `AsRef<T>`)

```suggestion
    /// Checks if two values are equal according to Ion's structural equivalence.
    pub fn eq<R>(a: R, b: R) -> bool where R: AsRef<T> {
        T::ion_eq(a.as_ref(), b.as_ref())
    }
```

popematt: Any thoughts about using this vs. putting `AsRef` into the function signature?
```rust
impl <T: IonEq, R: Deref<Target = T>> IonEq for R {
    fn ion_eq(&self, other: &Self) -> bool {
        // ...
    }
}
```

Using `AsRef` in the function signature means that in some cases, we need to start adding type annotations (turbofish, not ion) when there's more than one `AsRef` implementation for a given type.

Do you think that implementing `IonEq` for smart pointers would capture your intent, though in a different way?

almann: Does this work for the case where we don't want to expose `IonEq`, would external users have to use `Into<IonData>` to get at this?

popematt: > would external users have to use `Into<IonData>` to get at this?

No, they would not. As long as something implements `IonEq`, external users can (without knowing about `IonEq`) use `IonData::eq(foo, bar)` if all they want to do is check for Ion equivalence.

almann: I think I would be good with this formulation--the net effect is basically the same I think.

popematt: Oh shoot. I was wrong. I still had `IonEq` as `pub` instead of `pub(crate)`. I think `IonEq` might need to be `pub`, but we can prevent other implementors by using the [sealed traits](https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed) pattern. I'm going to play around with it a little more to see if I can figure out something that works nicely.

src/ion_data/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+
+use std::cmp::Ordering;
+use std::fmt::{Display, Formatter};
+
+pub(crate) use ion_eq::IonEq;
+pub(crate) use ion_ord::IonOrd;
+
+/// A wrapper for lifting Ion compatible data into using Ion oriented comparisons (versus the Rust
+/// value semantics). This enables the default semantics to be what a Rust user expects for native
+/// values, but allows a user to opt-in to Ion's structural equivalence/order.
```
zslayton, line 12: I'd like this doc comment to be a bit crisper about the motivation. Why would one want this behavior enough to wrap one's values? I happen to know your motivation from offline discussion, but I think someone encountering this construct for the first time in the wild would need to read the code to understand it. Consider adding a couple of example use cases that show why Rust value semantics can be a poor fit for working with Ion.

zslayton (approved):
LGTM, but one request regarding doc comments below.

src/types/bool.rs
```diff
@@ -0,0 +1,45 @@
+use crate::ion_data::{IonEq, IonOrd};
+use std::cmp::Ordering;
+use std::fmt::{Display, Formatter};
+
+/// Represents an Ion `bool` value, for the purpose of implementing some Ion-related traits.
+///
+/// Most of the time, you can use [bool] instead of this type.
+#[derive(Debug, Clone, PartialOrd, Ord, PartialEq, Eq)]
+pub struct Bool(bool);
```
almann, line 9: Probably deriving `Copy` is fine, but I am curious why we need this because we only need to define `IonEq`/`IonOrd` for `bool` (which we should be able to do in the module that defines those traits).  Whatever I am missing from why this is needed probably needs to go in the doc comment.

src/types/float.rs
```diff
@@ -0,0 +1,65 @@
+use crate::ion_data::{IonEq, IonOrd};
+use num_traits::Zero;
+use std::cmp::Ordering;
+use std::fmt::{Display, Formatter};
+
+/// Represents an Ion `float` value, for the purpose of implementing some Ion-related traits.
+///
+/// Most of the time, you can use [f64] instead of this type.
+#[derive(Debug, Clone, PartialOrd, PartialEq)]
+pub struct Float(f64);
```
almann, line 10: Same here as for `Bool...`

almann (changes requested):
Overall, I am happy with where this is going, however I am a bit puzzled by the added `Bool`/`Float` wrappers.  I think if this is needed to hide `IonEq` we may want to reconsider doing this.

src/ion_data/ion_eq.rs
```diff
@@ -22,39 +22,36 @@ use num_traits::Zero;
...
+// impl<T: IonEq> IonEq for Vec<T> {
+//     fn ion_eq(&self, other: &Self) -> bool {
+//         self.as_slice().ion_eq(other.as_slice())
+//     }
+// }
 
-impl IonEq for bool {
+impl<R: Deref> IonEq for R
+where
+    <R as Deref>::Target: IonEq,
```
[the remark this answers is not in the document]
popematt: Actually, there is a difference here, and it causes an error when I change it. Here's a minimum reproducible case:
```rust
#[test]
fn foo() {
    let element_vec: Vec<Element> = Element::read_all("foo 1 true").unwrap();
    bar(element_vec.into());
}

fn bar<T: IonOrd>(ion: IonData<T>) {
    // Do something
}
```
Fails with:
```
error[E0277]: the size for values of type `[element::Element]` cannot be known at compilation time
```
I'm assuming that this is because the trait impl is monomorphized differently depending on how the signature is formulated. (While rearranging a `where` clause shouldn't do something like this, in this case we have added a second type parameter _and_ we have changed how `Deref::Target` is specified.)

Do you have any insight or theory on the details of why this occurs?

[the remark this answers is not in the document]
popematt: Aha... it needs `?Sized`:
```rust
impl<E, R> IonEq for R where
  E: IonEq + ?Sized,
  R: Deref<Target = E>,
```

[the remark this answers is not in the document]
popematt: I ended up sticking with my original factoring because it minimizes the number of type parameters, though I cleaned it up by getting rid of some redundant parts.
```rust
impl<R: Deref> IonEq for R
where
    R::Target: IonEq,
```

popematt:
Hopefully this is good to go now. Latest commit includes:
* Remove `Float` and `Bool` in favor of just using `ion_eq_f64()`, etc. functions with a detailed note about why the functions exist.
* Remove some`AsRef` impls that this PR added that weren't actually needed.
* Minor clarifications, cleanups.

almann (approved):
:shipit: LGTM

PR #518, opened 2023-04-22
Fix rustdoc

### LINKED ISSUE #516: Documenting Private Items Fails
When trying to generate all of our API documentation (useful for doing low-level work), the following fails with numerous warnings and errors.

```
$ cargo doc --open --document-private-items
```

While not critical to our public documentation, we probably should be strict and make sure this doesn't break.  I suggest fixing the errors (and probably the warnings) and put this as a step in CI to make sure we keep ourselves honest.

almann:
Cleans up doc comments such that `cargo doc --document-private-items --all-features` works and is in CI.

Fixes #516

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,6 @@
 #![allow(dead_code)]
 #![deny(rustdoc::broken_intra_doc_links)]
+#![deny(rustdoc::bare_urls)]
 //! # Reading and writing `Element`s
 //!
 //! The [Element] API offers a convenient way to read and write Ion data when its exact shape is
--- a/src/text/non_blocking/raw_text_reader.rs
+++ b/src/text/non_blocking/raw_text_reader.rs
@@ -404,11 +404,10 @@ impl<A: AsRef<[u8]> + Expandable> RawTextReader<A> {
     }
 
     /// Assumes that the reader is inside a struct AND that a field has already been successfully
-    /// parsed from input using [`next_struct_field_name`][1] and attempts to parse the next value.
+    /// parsed from input using [`next_struct_field_name`](Self::next_struct_field_name) and
+    /// attempts to parse the next value.
     /// In this input position, only a value (or whitespace/comments) are legal. Anything else
     /// (including EOF) will result in a decoding error.
-    ///
-    /// [1]: Self::next_struct_field_name
     fn next_struct_field_value(&mut self) -> IonResult<AnnotatedTextValue> {
         // Only called after a call to [next_struct_field_name] that returns Some(field_name).
         // It is not legal for a field name to be followed by a '}' or EOF.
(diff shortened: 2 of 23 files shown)


src/text/non_blocking/raw_text_reader.rs
```diff
@@ -404,9 +404,11 @@ impl<A: AsRef<[u8]> + Expandable> RawTextReader<A> {
     }
 
     /// Assumes that the reader is inside a struct AND that a field has already been successfully
-    /// parsed from input using [next_struct_field_name] and attempts to parse the next value.
+    /// parsed from input using [`next_struct_field_name`][1] and attempts to parse the next value.
     /// In this input position, only a value (or whitespace/comments) are legal. Anything else
     /// (including EOF) will result in a decoding error.
+    ///
+    /// [1]: Self::next_struct_field_name
```
jobarr-amzn, line 411: Only two examples in this patch of a reference-style link, and none that I can find outside it. Might as well stick with inline links, and if we find a more complicated case use named reference-style links. 
```suggestion
    /// parsed from input using [`next_struct_field_name`](Self::next_struct_field_name) and
    /// attempts to parse the next value.
    /// In this input position, only a value (or whitespace/comments) are legal. Anything else
    /// (including EOF) will result in a decoding error.
```

src/text/parsers/text_support.rs
```diff
@@ -177,10 +177,12 @@ pub(crate) fn escaped_char_unicode(input: &str) -> IonParseResult<char> {
     Ok((remaining_input, character))
 }
 
-/// Rust's [char] type represents any Unicode code point EXCEPT a surrogate. If we encounter a high
+/// Rust's [`char`][1] type represents any Unicode code point EXCEPT a surrogate. If we encounter a high
 /// surrogate in the stream, we cannot convert it to a `char` yet. We must find the (mandatory)
 /// low surrogate that follows it in the stream; then we can combine the high and low surrogates
```
jobarr-amzn, line 182:
```suggestion
/// Rust's [`char`](prim@char) type represents any Unicode code point EXCEPT a surrogate. If we
/// encounter a high surrogate in the stream, we cannot convert it to a `char` yet. We must find the
/// (mandatory) low surrogate that follows it in the stream; then we can combine the high and low
/// surrogates into a complete code point. This code point can then be returned as a Rust
```

src/text/parsers/text_support.rs
```diff
@@ -177,10 +177,12 @@ pub(crate) fn escaped_char_unicode(input: &str) -> IonParseResult<char> {
...
-/// into a complete code point. This code point can then be returned as a Rust [char].
+/// into a complete code point. This code point can then be returned as a Rust [`char`][1].
+///
+/// [1]: prim@char
```
jobarr-amzn, line 185:
```suggestion
/// [`char`](prim@char).
```

src/lib.rs
```diff
@@ -1,5 +1,6 @@
 #![allow(dead_code)]
 #![deny(rustdoc::broken_intra_doc_links)]
+#![deny(rustdoc::bare_urls)]
```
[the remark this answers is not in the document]
almann: This is saying Rustdoc will _error_ when we make this mistake.

jobarr-amzn (approved):

Issue #519, opened 2023-04-22
Add Experimental Feature to Cargo Configuration


almann:
As I am currently steel-threading Ion 1.1 macros, I am developing a bunch of lower-level APIs that may in the future be generally useful.  In addition, it may be nice to not keep all of this capability in a fork or branch and maintain it with our core code base so that it is easier to evolve.

I propose adding an [feature][1] to our `cargo` configuration that allows use to put experimental APIs and capabilities behind as a non-default conditional compilation flag.  We may choose to still maintain some of this stuff behind non-`main` branches, but this gives us flexibility to push things into `main` that are useful, but needs to be tested by applications on a very strict opt-in.

Cargo.toml [change](https://github.com/almann/ion-rust/commit/71f7d05dcd1ae4211feb6d61cb9366e52bd63ef9):
```toml
# Feature for indicating particularly bleeding edge APIs or functionality in the library.
# These are not guaranteed any sort of API stability and may also have non-standard
# Ion behavior (e.g., draft Ion 1.1 capabilities).
experimental = []
```

I've modeled this for the macro steel thread here to see how it would look in practice:

https://github.com/almann/ion-rust/commit/c25531f59667cbf17420470f9a2e8963db96fc9d

[1]: https://doc.rust-lang.org/cargo/reference/features.html

Issue #521, opened 2023-04-22
Thunk API for Deferred/Materialized Values


almann:
#520 (Stream API) requires the modeling of a lazy value.  This issue is for tracking the feature required for this.

I have created draft API and implementation here:

> The `thunk` that underlies this can be found here:
>
> https://github.com/almann/ion-rust/blob/c25531f59667cbf17420470f9a2e8963db96fc9d/src/thunk.rs

I considered some 3P crates for this, but decided to roll my own because my steel thread implementation has requirements that require us to control and mutate in place for certain implementation details of the stream.  Not taking a 3P dependency is itself of value, but not my primary motivation.  The crates I looked at:

* [`lazy`](https://crates.io/crates/lazy) - old crate that hasn't been updated in many years and may has reports of not working in modern Rust.
* [`lazy-st`](https://crates.io/crates/lazy-st) - less old fork of the above breaking out single-threaded and multi-threaded capability in separate crates.
* [`thunk`](https://crates.io/crates/thunk) - old crate that requires unsafe, untagged unions and Rust nightly (not sure if it still would require nightly, but the unsafe code is not appealing).

For implementing our lazy values, we do want to _sometimes_ in-place materialize (e.g., deferred field names), but in other places we may not want to (only support one-time evaluation of `read_xxx` methods and not store the result.  Our lazy value should allow for this level of control.

That said, the crates above could be useful for looking at implementation strategies that we can employ to get the flexibility we need plus performance if needed.

Issue #520, opened 2023-04-22
API for Efficient Stream of Ion Values


almann:
We have an `IonReader` trait that captures both the *"iteration"* over the ***stream*** and the ***value*** the stream is currently on.  This is a fine API, but there are cases (especially when we start to think about macros in Ion 1.1), that this API becomes difficult to compose on.  I propose we create an API that de-couples the ***stream*** from the ***value*** without while deferring computation on the value it is on.

At a high-level, the API might look something like this:

```rs
pub enum Instruction {
  /// Advance to the next token.
  Next,
  /// Skip to the end of the current container
  NextContainerEnd,
}

struct AnnotatedToken<'a>{ ... }

trait TokenStream {
  fn next_token(&mut self, instruction: Instruction) -> IonResult<AnnotatedToken>;
}
```

The above is quite a simplification, but it demonstrates the basic structure.  `AnnotatedToken` would represent the next token deferring computation on a field name, annotations, or the underlying scalar value, the beginning of a container, end of the container, or the end of a stream.

It would be straightforward to adapt this `TokenStream` wrap any `IonReader` (though some care would have to be done to model this on both user-level/system-level/raw-level variants to do it generically), and the other direction should be relatively straightforward as well (make any `TokenStream` into an `IonReader`.   The latter gives us `TokenStream` to `Element`.

almann:
A draft version of a `tokens` module can be found here:

https://github.com/almann/ion-rust/tree/c25531f59667cbf17420470f9a2e8963db96fc9d/src/tokens

The `thunk` that underlies this can be found here:

https://github.com/almann/ion-rust/blob/c25531f59667cbf17420470f9a2e8963db96fc9d/src/thunk.rs

PR #522, opened 2023-04-22
Adds Lazy Evaluation Support

almann:
Adds `thunk::Thunk` as an experimental API for modeling lazy/eager values that are fallible (and thus evaluate to an `IonResult`. The API supports controlling how the value is materialized from a closure and supports moving values into and out of it.

The `Thunk` models three distinct states:

* ***Deferred***: value is backed by a closure.
* ***Materialized***: value is stored directly in the `Thunk`.
* ***Error***: there was some error evaluating the *deferred* value. This state can be thought of as a deferred value that can *never* be evaluated.

This also backs this API with a Cargo feature called "experimental" that we can use for other such APIs.

resolves: #519, #521

Prerequisite for #520.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,10 @@ ion-hash = ["digest"]
 # Feature for indicating particularly bleeding edge APIs or functionality in the library.
 # These are not guaranteed any sort of API stability and may also have non-standard
 # Ion behavior (e.g., draft Ion 1.1 capabilities).
-experimental = []
+experimental = ["experimental-streaming"]
+
+# Experimental streaming APIs
+experimental-streaming = []
 
 [dependencies]
 base64 = "0.12"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,11 @@ members = [
 default = []
 ion-hash = ["digest"]
 
+# Feature for indicating particularly bleeding edge APIs or functionality in the library.
+# These are not guaranteed any sort of API stability and may also have non-standard
+# Ion behavior (e.g., draft Ion 1.1 capabilities).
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -31,6 +31,14 @@ members = [
 default = []
 ion-hash = ["digest"]
 
+# Feature for indicating particularly bleeding edge APIs or functionality in the library.
+# These are not guaranteed any sort of API stability and may also have non-standard
+# Ion behavior (e.g., draft Ion 1.1 capabilities).
+experimental = ["experimental-streaming"]
+
+# Experimental streaming APIs
+experimental-streaming = []
```
almann, line 40: As per discussion with @popematt, added `experimental-streaming` to capture the experimental feature of the API that this supports  `experimental` now depends on `experimental-streaming`.

src/thunk.rs
```diff
@@ -0,0 +1,286 @@
...
+        match value_res {
+            Ok(val) => self.0 = Ok(Materialized(val)),
+            Err(e) => self.0 = Err(e),
+        };
+        // at this point we can now return a reference to our internal materialized value
+        // or the error
+        match &self.0 {
+            Ok(Deferred(_)) => {
+                // XXX should not be possible
+                panic!("Thunk memoization is still deferred")
```
zslayton, line 108: I like [`unreachable!`](https://doc.rust-lang.org/std/macro.unreachable.html) for this use case.

zslayton (approved):

src/thunk.rs
almann: Updated as per suggestion.

PR #523, opened 2023-04-24
Reduce 'pretty' indentation 4=>2

jobarr-amzn:
*Issue #, if available:* None. 

*Description of changes:* This is the same indentation that `jq` uses by default for prettifying JSON, which makes like-for-like comparisons easiest. An indentation level of 4 also gets deep twice as fast as 2.

Please feel free to discard this pull request if it's unwanted, consider it opening the conversation of "what kind of indentation should we use?"

I intend to add an `--indent` option such as the one that [`jq`](https://stedolan.github.io/jq/manual/) supports to [`ion-cli`](https://github.com/amazon-ion/ion-cli/), but I am here moving the default for `pretty`. 


popematt (approved):
I am in favor, especially since you're planning to add `--indent`. Is it difficult to add the indentation config for `ion-rust` in this PR?

almann (approved):

PR #525, opened 2023-04-26
Refactors Value, TextValue, and IonType enum variants to have consistent order

popematt:
**Issue #, if available:**

None

**Description of changes:**

Updates `Value` and `TextValue` to have variant order that is consistent with `IonType`. Also updates any `match` expressions I could find that match all variants of any of those three enums.


almann (approved):
:shipit:

PR #528, opened 2023-04-26
Refactors Annotations conversion trait implementations

### LINKED ISSUE #526: Annotations::into_iter Inconsistent with Vector
I ran into `Annotations::into_iter` not returning an iterator of values (i.e., consuming the source):

```rust
let a: Vec<Symbol> = vec!["a"].into_iter().map(|x| x.into()).collect();
let b: Annotations = a.clone().into();
fn print_type<T>(_: &T) {
    println!("{}", type_name::<T>());
}
for s in a.into_iter() {
    print_type(&s);
}
for s in b.into_iter() {
    print_type(&s);
}
```

which outputs:

```
ion_rs::symbol::Symbol
&ion_rs::symbol::Symbol
```

My expectation would be that `Vec<Symbol>::into_iter` should behave the same as `Annotations::into_iter`, in that it should consume the source into the iterator and have value semantics, but the latter returns references.  This would make it consistent with the intent of being a wrapper around `Vec<Symbol>` (or any `Vec`-like type).

### LINKED ISSUE #527: Annotations Blanket From Trait Prevents IntoIterator Implementation
When looking at fixing #526, I could not add a consuming implementation for `IntoIter<Item = Symbol, ...>` for `Annotations` because it immediately conflicts with the blanked `From` implementation:

```rust
impl<S: Into<Symbol>, I: IntoIterator<Item = S>> From<I> for Annotations {
    fn from(annotations: I) -> Self {
        let symbols: Vec<Symbol> = annotations.into_iter().map(|a| a.into()).collect();
        Annotations::new(symbols)
    }
}
``` 

The above API while allows for certain `into()` conversions to be very convenient, is problematic if we cannot implement a proper consuming `IntoIter`.

The surrounding comment indicates we might consider "consider defining a custom `IntoAnnotations` trait," and I think I agree with that comment as the solution here.

almann:
Adds `IntoAnnotations` trait for generic conversions.  Moves blanked implementation of `From<...>` for `Annotations` to this trait because it conflicts with `Annotations` implementing `IntoIter`.

Adds `IntoIter` for Annotations as a consuming iterator.

Adds `From<Vec<Symbol>>` as a special case conversion for avoiding generic conversion traits.  Adds `FromIterator` to support `collect()` on any iterator transform.

fixes #526, fixes #527 

### DIFF
--- a/src/element/annotations.rs
+++ b/src/element/annotations.rs
@@ -1,4 +1,4 @@
-use crate::element::iterators::SymbolsIterator;
+use crate::element::iterators::{AnnotationsIntoIter, SymbolsIterator};
 use crate::ion_data::IonOrd;
 use crate::Symbol;
 use std::cmp::Ordering;
@@ -35,7 +35,7 @@ impl Annotations {
     /// Returns an [`Iterator`] that yields each of the [`Symbol`]s in this annotations
     /// sequence in order.
     pub fn iter(&self) -> SymbolsIterator {
-        self.into_iter()
+        SymbolsIterator::new(self.symbols.as_slice())
     }
 
     /// Returns the number of annotations in this sequence.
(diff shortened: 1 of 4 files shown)


zslayton (approved):
LGTM modulo the CI errors. `Element`s are immutable (providing clone-on-write methods for containers), so we need to make sure we don't make it possible for someone to destruct an `Element`'s annotations. I think at the moment a user can only get access to an `&Annotations`, so this is fine to add.

PR #533, opened 2023-04-26
Adds CI matrix for crate features

almann:
This change makes it so we build our crate with all the various interesting permutations:

* *default* - build with no features set.
* `--all-features` - build with all the bells and whistles.
* `--features X` where `X` is:
  * `ion-hash`
  * `experimental`
  * `ion-hash,experimental`

When building/testing with these permutations, it uncovered a bug in the `ion-hash` doc tests where the test code depends on an optional feature. This has been conditionally compiled out, preserving the text of the doc comment (only runs when the appropriate feature flag is enabled).

Also updates the build of API docs to error out on leaking internal references.  This change cleans up such documentation comments.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -84,8 +84,8 @@ jobs:
         # only running on the Linux test. (The choice of OS was arbitrary.)
         if: matrix.os == 'ubuntu-latest' && matrix.features == 'all'
         run: rustup component add clippy
-      - name: Run Clippy && matrix.features == 'all'
-        if: matrix.os == 'ubuntu-latest'
+      - name: Run Clippy 
+        if: matrix.os == 'ubuntu-latest' && matrix.features == 'all'
         uses: actions-rs/clippy-check@v1
         with:
           # Adding comments to the PR requires the GITHUB_TOKEN secret.
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -41,13 +41,6 @@ jobs:
           toolchain: stable
           components: rustfmt, clippy
           override: true
-      - name: Cargo Build
-        # cargo test implies build, so only run this explicitly on default
-        if: matrix.features == 'default'
-        uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: --verbose --workspace
       - name: Cargo Test (default/no features)
         if: matrix.features == 'default'
         uses: actions-rs/cargo@v1
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -17,6 +17,8 @@ jobs:
         # meantime, the build only tests Windows Server 2019.
         # See https://github.com/amazon-ion/ion-rust/issues/353
         os: [ubuntu-latest, windows-2019, macos-latest]
+        # build and test for different and interesting crate features
+        features: ['default', 'all', 'ion-hash', 'experimental', 'ion-hash,experimental']
     permissions:
       checks: write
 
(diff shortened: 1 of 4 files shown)


.github/workflows/rust.yml
```diff
@@ -40,21 +42,33 @@ jobs:
           components: rustfmt, clippy
           override: true
       - name: Cargo Build
+        # cargo test implies build, so only run this explicitly on default
+        if: matrix.features == 'default'
```
popematt, line 46: Why run it explicitly at all?

.github/workflows/rust.yml
```diff
@@ -17,6 +17,8 @@ jobs:
         # meantime, the build only tests Windows Server 2019.
         # See https://github.com/amazon-ion/ion-rust/issues/353
         os: [ubuntu-latest, windows-2019, macos-latest]
+        # build and test for different and interesting crate features
+        features: ['default', 'all', 'ion-hash', 'experimental', 'ion-hash,experimental']
```
popematt, line 21: Should we move `ion-hash` under the `experimental` umbrella?

.github/workflows/rust.yml
```diff
@@ -68,16 +82,18 @@ jobs:
         # The clippy check depends on setup steps defined above, but we don't want it to run
         # for every OS because it posts its comments to the PR. These `if` checks limit clippy to
         # only running on the Linux test. (The choice of OS was arbitrary.)
-        if: matrix.os == 'ubuntu-latest'
+        if: matrix.os == 'ubuntu-latest' && matrix.features == 'all'
         run: rustup component add clippy
-      - name: Run Clippy
+      - name: Run Clippy && matrix.features == 'all'
         if: matrix.os == 'ubuntu-latest'
```
popematt, line 88:
```suggestion
      - name: Run Clippy 
        if: matrix.os == 'ubuntu-latest' && matrix.features == 'all'
```

popematt (commented):
Should we check combinations of features for the `coverage.yml` workflow as well?

.github/workflows/rust.yml
almann: Oops, thanks!

.github/workflows/rust.yml
almann: 🤷‍♀️ - Do we consider it experimental?  Cut #534 to deal with this.

almann:
> Should we check combinations of features for the `coverage.yml` workflow as well?

@popematt - I don't think this is worth it.  For code coverage, I think we just want to see the aggregate sum of coverage via `--all-features` especially when the denominator is "all the code."

almann:
Closing and re-opening to try to unstuck _Waiting for status to be reported_

There is something odd going on here:
https://stackoverflow.com/questions/52200096/github-pull-request-waiting-for-status-to-be-reported

popematt:
> Closing and re-opening to try to unstuck _Waiting for status to be reported_
> 
> There is something odd going on here: https://stackoverflow.com/questions/52200096/github-pull-request-waiting-for-status-to-be-reported

Actually, it's waiting for those because of the branch protection rule. I just updated it, and as you can see, now it's waiting for e.g. `Build and Test (macos-latest, all)` instead of `Build and Test (macos-latest)`. (Yes, it uses string matching on the workflow name.)

almann:
> Actually, it's waiting for those because of the branch protection rule. I just updated it, and as you can see, now it's waiting for e.g. `Build and Test (macos-latest, all)` instead of `Build and Test (macos-latest)`. (Yes, it uses string matching on the workflow name.)

Oh thanks! I was quite puzzled as to what I was missing...

popematt (approved):

PR #536, opened 2023-04-29
Introduces an experimental lazy binary reader

zslayton:
The existing `IonReader` trait mimics `ion-java`'s `IonReader` interface, offering a single stateful type that can traverse an Ion stream.

Consider this example data:
```js
customer::{
    name: {
        first: "John",
        middle: "Jacob",
        last: "Jingleheimer-Schmidt",
    },
    address: {
        street: "123 Fake Street",
        city: "Toronto",
        country: "Canada",
    },
    status: legacy_rewards::gold,
}
```

In the existing streaming API, the reader can only see the value on which it is positioned. If it advances beyond that value, it cannot reference it further without materializing it. As such, it needs to fully process each field as it is encountered. Because struct fields can appear in any order, the reader must be able to handle them as they come. This requires the user to write mechanical, repetitive, error-prone dispatch logic similar to the following:

```rust
let mut reader = ...;

struct Address {...};

fn read_top_level(reader: &mut Reader) -> IonResult<Address> {
    if let StreamItem::Value(ion_type) = reader.next()? {
         if ion_type != IonType::Struct {
             return decoding_error(format!("expected a struct, found a(n) {}", ion_type);
         }
    } else {
        return decoding_error("expected a struct, found nothing");
    }
    // Ok, it's a struct. Is it a customer?
    if reaader.annotations().next()? != Some(Ok("customer")) {
        return decoding_error("expected struct to be annotated with `customer::`");
    }
    
    // Ok, it's a `customer::`-annotated struct.
    reader.step_in()?;
    handle_customer(reader)
}

fn handle_customer(reader: &mut Reader) -> IonResult {
    // We're inside the `customer::` struct.
    while let Some(ion_type) = reader.next()? {
        let field_name = reader.field_name()?;
        if field_name == "name" {
            if ion_type != IonType::Struct {
                return decoding_error("`name` field must be a struct");
            }
            handle_address(reader)
        } else if field_name == "address" {
            if ion_type != IonType::Struct {
                return decoding_error("`address` field must be a struct");
            }
            handle_address(reader)
        } else if field_name == "status" {
             if ion_type != IonType::Symbol {
                  return decoding_error("`status` field must be a symbol");
             }
        } else {
             // Ignore other fields or error out as desired
             continue;   
        }
    }
}

fn handle_name(reader: &mut Reader) -> IonResult<Name> {
    // We're inside the `name` struct
    let mut first: Option<String> = None;
    let mut middle: Option<String> = None;
    let mut last: Option<String> = None;
    while let Some(ion_type) = reader.next()? {
        if ion_type != IonType::String {
            return decoding_error("name fields must be strings");
        }
        if field_name == "first" {
            first = Some(reader.read_string()?);
        } else if field_name == "middle" {
            middle = Some(reader.read_string()?);
        } else if field_name == "last" {
            last = Some(reader.read_string()?);
        } else {
            continue;
        }
    }

    if let (Some(first), Some(middle), Some(last)) = (first, middle, last) {
        return Ok(Name::new(first, middle, last));
    }
    decoding_error("missing one or more fields from the `name` struct")
}

fn handle_address(reader: &mut Reader) -> IonResult<Address> {
    todo!("yeesh, I don't have the patience to continue this example")
}

fn handle_status(reader: &mut Reader) -> IonResult<Status> {
    todo!("thanks but no thanks")
}
```

The experimental API introduced in this PR takes advantage of the fact that Ion's encoding context (that is: the symbol table) can only change between top-level values.

Each time that the reader advances, it hands out a "lazy" value--an object containing an immutable slice of the input buffer and some calculated offsets from the encoded value it recognized in the stream. This lazy value serves two purposes:
1. The body of the lazy value can be read upon request, producing other lazy values representing nested data.
2. It immutably borrows the input buffer, effectively "freezing" it so long as any lazy value exists.

In this way, the reader can move to a top-level value and produce any number of lazy references to data within that value. What's more, those values are all immutable and can be read as many times as desired. Further, the immutability means that all of this can be achieved without reference counting, and multithreaded processing within a top-level value can be done without synchronization mechanisms.

In all, this makes it possible to expose an API for traversing top-level values that has more of a "Lazy DOM" feel to it.

Compare the above code with this version that uses the `LazyReader`:

```rust
let binary_ion: &[u8] = ...;
let mut lazy_reader = LazyReader::new(&binary_ion)?;

// Get the first value, confirm that it's a struct.
let customer = lazy_reader.expect_next()?
    .read()?
    .expect_struct()?;
// Make sure it has the `customer` annotation
customer.annotations().expect(["customer"])?;

let name_field = customer.get_expected("name")?.expect_struct()?;
let address_field = customer.get_expected("address")?.expect_struct()?;
let status = customer.get_expected("status")?.expect_symbol()?;

let full_name = Name::new(
    name_field.get_expected("first")?.expect_string()?,
    name_field.get_expected("middle")?.expect_string()?,
    name_field.get_expected("last")?.expect_string()?,
);

let address = ...;
```

Critically, this API has significantly fewer sharp edges. There are no illegal operations--only methods that are legal are exposed to the user in any context.

Perhaps surprisingly, this arrangement is 5-10% faster than the current binary reader (using static dispatch in both) despite being much more flexible. This performance improvement likely comes from 2 sources:
1. Due to everything being immutable, field names and text can live as long as everything else without materialization.
2. Modeling the different contexts the reader can be in (top level, sequence, struct) as their own types means that some sanity checking can be elided. For example, only `reader.next()` needs to check for system values; no such check has to happen in sequences or structs. Additionally, the reader does not need to maintain a stack of parent contexts to support `step_in()` and `step_out`; accessing nested data is done via methods on the lazy container values, not on the reader itself.

This initial PR contains:
* A binary-only `LazyRawReader` that reads a fixed buffer and produces `LazyRawValue`s.
* A binary-only `LazySystemReader` that reads a fixed buffer and produces `LazyValue`s.
* A binary-only `LazyReader` that reads a fixed buffer and produces `LazyValue`s.

Follow-on work includes:
* Writing a blocking wrapper for the above that recognizes `IonError::Incomplete` and shifts the buffer to a new region of input.
* Text parsing using this API
* Traits to abstract over text vs binary 


zslayton (commented):
🗺️ PR tour part 1.

This PR looks huge, but large parts of it were copy/pasted from the existing binary reader for the sake of experimentation. I'll call out which files contain novel logic and which ones are more or less the same as the previously reviewed code it copied.

zslayton:
Closing this draft PR in favor of a newer one with a cleaner commit history.

PR #541, opened 2023-05-01
Adds ReaderTokenStream

almann:
`ReaderTokenStream` adapts any, effectively user-level, `IonReader` into a `TokenStream`.  Ideally, this would work over any abstraction of the `IonReader` (raw/system), but for now we only support `StreamItem` and `Symbol` based readers.

Prerequisite for #520.

----

Note that the this is dependent on #540 and currently based off of that--when/if that is merged **and** this is approved I will change to base to main.


src/tokens/reader_stream.rs
```diff
@@ -0,0 +1,560 @@
...
+    reader_cell: Rc<RefCell<UnderlyingReader<R>>>,
+    pending_token: PendingToken,
+    position: ReaderPosition,
+    // XXX this allows us to explicitly capture a lifetime for the reader we operate on
+    phantom: PhantomData<&'a R>,
+}
+
+/// These generate the methods that produce lazy content for our stream against the underlying
+/// reader.  There is a lot of boilerplate and variants so this avoids copy/paste errors.
+macro_rules! scalar_content_method {
```
almann, line 107: 🗺️ It makes me sad to do this, but it does eliminate *a lot* of boilerplate for generating the `Content` instances against the `Rc<RefCell<...>>`.

src/tokens/reader_stream.rs
```diff
@@ -0,0 +1,560 @@
...
+    Container(ContainerType),
+}
+
+/// Adapter for a [`TokenStream`] over an arbitrary [`IonReader`]
+pub struct ReaderTokenStream<'a, R>
+where
+    R: IonReader<Item = StreamItem, Symbol = Symbol> + 'a,
+{
+    // XXX this is so we can have multiple closures to lazily evaluate tokens
+    reader_cell: Rc<RefCell<UnderlyingReader<R>>>,
```
almann, line 98: 🗺️ This is necessary because we need to maintain three closures (annotations/field names/content) and statically borrow checker only allows one universal borrow for closures.  We use this shared pointer to give each closure what they need to defer the calls to the underlying reader.

src/tokens/reader_stream.rs
```diff
@@ -0,0 +1,560 @@
...
+
+// TODO make this more generic with respect to other readers--the problem is Item/Symbol
+// TODO this has to abstract over potentially system reader to implement macros
+
+/// Internal tracking state of the returned token.  This is necessary because the lifetime of the
+/// token returned from the stream is bound to the stream itself.
+///
+/// This is essentially strongly typed [`Option`] with convenience to operate within the
+/// context of a [`ReaderTokenStream`]
+enum UnderlyingReader<R>
```
almann, line 25: 🗺️ The lifetime of `next_token` is not bound to any static lifetime we can keep track of, this is because if we want to adapt `IonReader`, `next`/`read_XXX` are not associated in any lexical context that it may be valid for as one lifetime that we can track for our `Token`.  This lets us make sure the `Reader` we "gave" to a `Token` can be invalidated when the `Reader` "moves" one while the caller may have held on to that `Token`.

src/tokens/reader_stream.rs
```diff
@@ -0,0 +1,560 @@
...
+    fn invalidate_token(&mut self) {
+        // only invalidate if we have something deferred
+        if self.pending_token.is_deferred() {
+            let underlying = {
+                let mut original = self.reader_cell.borrow_mut();
+                original.invalidate()
+            };
+            self.reader_cell = Rc::new(RefCell::new(underlying));
+        }
+    }
```
almann, line 168: 🗺️ This lets us "garbage collect" outstanding tokens, such tokens will just error when trying to evaluate any deferred value within them.

src/tokens/reader_stream.rs
```diff
@@ -0,0 +1,560 @@
...
+        let reader = underlying.as_ref();
+        reader.parent_type()
+    }
+}
+
+impl<'a, R> TokenStream<'a> for ReaderTokenStream<'a, R>
+where
+    R: IonReader<Item = StreamItem, Symbol = Symbol> + 'a,
+{
+    fn next_token(&mut self, instruction: Instruction) -> IonResult<Token<'a>> {
```
almann, line 231: 🗺️ Originally this was tied to the borrow of `&mut self`.  The problem with this is that you _must_ materialize the token if you want to _implement_ `IonReader` over `IonStream` (see #542).  This is a problematic design since having an `IonReader` that has to materialize the token on every call to `next` is not efficient.

almann (commented):
🗺️ PR tour below.

src/tokens/reader_stream.rs
```diff
@@ -0,0 +1,560 @@
...
+    }
+}
+
+/// Condensed version of the item of the [`IonReader`].
+#[derive(Debug, Eq, PartialEq, Copy, Clone)]
+enum ReaderPosition {
+    /// We are positioned on nothing.
+    Nothing,
+    /// We are positioned on some value.
+    NonContainer,
```
zslayton, line 87: Is this "non-container" (vs "scalar") because it includes null containers, which are arguably scalar values even though their Ion type is a container? If so, that'd be helpful to have in the comments.

src/tokens/reader_stream.rs
```diff
@@ -0,0 +1,560 @@
...
+
+impl<'a, R> TokenStream<'a> for ReaderTokenStream<'a, R>
+where
+    R: IonReader<Item = StreamItem, Symbol = Symbol> + 'a,
+{
+    fn next_token(&mut self, instruction: Instruction) -> IonResult<Token<'a>> {
+        use self::Instruction::*;
+
+        // once we enter this method--we must invalidate any outstanding token references
+        // this has to do with the lifetime of the returned token which cannot be statically
```
zslayton, line 235:
```suggestion
        // once we enter this method--we must invalidate any outstanding token references.
        // this has to do with the lifetime of the returned token which cannot be statically modeled
```

zslayton (approved):

PR #542, opened 2023-05-01
Adds TokenStreamReader

### LINKED ISSUE #520: API for Efficient Stream of Ion Values

almann:
`TokenStreamReader` adapts `TokenStream` into a user-level `Reader`.

This also changes the support types for `ElementReader` to work with `IonReader` trait objects (adding `?Sized` to the type constraints).

The `Element` round-trip tests are also adapted to work with _any_ `IonReader` as a boxed trait object, which allows us to slot in `TokenStreamReader` over `ReaderTokenStream` over a user `Reader` instance, exercising both adapters in this framework.

This also consolidates the round-trip skip-lists into constants so there is less copy/paste between the now three instances of the `ElementApi` trait.

resolves #520

----

Note that the this is dependent on #541 and currently based off of that--when/if that is merged **and** this is approved I will change to base to main.


src/tokens/stream_reader.rs
```diff
@@ -0,0 +1,539 @@
...
+/// Generates the read methods against the underlying token/item state.
+/// This eliminates the boilerplate around each definition avoiding a lot of copy/paste.
+///
+/// * `me` - the identifier for `self` to make it visible to `scalar_exp`.
+/// * `method` - the name of the method.
+/// * `scalar_type` - the return type for the method.
+/// * `variant` - the unqualified variant of [`ScalarValue`] to match over the current token.
+/// * `scalar` - the matched identifier of the contents of [`ScalarValue`].
+/// * `scalar_exp` - the expression to generate the return value.
+macro_rules! read_method_self {
```
almann, line 93: 🗺️ It makes me sad to do this, but this macro really does avoid a lot of boilerplate around implementing the `read_XXX` methods especially given our `Option<RefCell<Token>>` modalities.

src/tokens/stream_reader.rs
```diff
@@ -0,0 +1,539 @@
...
+    T: TokenStream<'a>,
+{
+    type Item = StreamItem;
+    type Symbol = Symbol;
+
+    fn ion_version(&self) -> (u8, u8) {
+        // XXX A `TokenStream` doesn't know its underlying Ion version, we just say 1.0
+        //     because all Ion 1.x versions have the same data model.
+        (1, 0)
+    }
```
almann, line 133: 🗺️ `ElementStreamReader` has a similar weirdness--I honestly don't think this API should be here, it probably belongs at a lower-level or some other trait.

src/tokens/stream_reader.rs
```diff
@@ -0,0 +1,539 @@
...
+        };
+        Box::new(iter.map(Ok))
+    }
+
+    fn field_name(&self) -> IonResult<Self::Symbol> {
+        match &self.curr_token_cell {
+            None => illegal_operation(NO_FIELD_NAME_ERROR_TEXT),
+            Some(token_cell) => {
+                let mut token = token_cell.borrow_mut();
+                match token.share_field_name() {
```
almann, line 214: 🗺️ We could be more aggressive about _not_ interning, only intern when this is called after we materialized it once, but I don't think that is worth it, especially while we're re-thinking the `IonReader` interface in #536.

tests/element_test_vectors.rs
```diff
@@ -419,83 +422,90 @@ mod impl_display_for_element_tests {
     }
 }
 
+const ELEMENT_GLOBAL_SKIP_LIST: SkipList = &[
```
almann, line 425: 🗺️ It appears that these are identical across all `ElementApi` implementations--I'd bet this was a vestige from when we had Ion C lurking.

tests/element_test_vectors.rs
```diff
@@ -67,14 +67,17 @@ fn serialize(format: Format, elements: &[Element]) -> IonResult<Vec<u8>> {
     Ok(buffer)
 }
 
+/// Allow for anything that implements a user-like reader
+type DynamicReader<'a> = Box<dyn IonReader<Item = StreamItem, Symbol = Symbol> + 'a>;
```
almann, line 71: 🗺️ This lets us plug in the `ElementStreamReader` if we want to which is probably not a bad idea for the future (more coverage).

tests/element_test_vectors.rs
```diff
@@ -665,3 +628,77 @@ mod non_blocking_native_element_tests {
...
+        }
+
+        fn non_equivs_skip_list() -> SkipList {
+            &[]
+        }
+
+        fn make_reader<'a>(data: &'a [u8]) -> IonResult<DynamicReader<'a>> {
+            let reader = ReaderBuilder::default().build(data)?;
+            let tokens: ReaderTokenStream<'a, _> = reader.into();
+            let token_reader: TokenStreamReader<'a, _> = tokens.into();
```
almann, line 664: 🗺️ I think this gives us a lot of mileage for testing both adapters.

almann (commented):
🗺️ PR tour below.

src/tokens/stream_reader.rs
```diff
@@ -0,0 +1,539 @@
...
+        //     because all Ion 1.x versions have the same data model.
+        (1, 0)
+    }
+
+    fn next(&mut self) -> IonResult<Self::Item> {
+        match &self.curr_token_cell {
+            None => self.next_token(Instruction::Next),
+            Some(token_cell) => {
+                let token = token_cell.borrow();
+                let insn_opt: Option<_> = match token.content() {
```
zslayton, line 140: ~What does `insn` mean?~

EDIT: I see, it's "instruction". I think this variable name only appears twice in the method, it'd be worth spelling it out.

zslayton (approved):

PR #543, opened 2023-05-01
Move the Ion value types to the types module

popematt:
**Issue #, if available:**

None.

**Description of changes:**

This moves the "types" to the `types` module, as per:

```rust
//! This module provides an implementation of the data types described by the
//! [Ion Data Model](https://amazon-ion.github.io/ion-docs/docs/spec.html#the-ion-data-model)
//! section of the Ion 1.0 spec.
```

It also makes it so that all of the types in the `types` module are exposed as part of `types` rather than as part of a submodule of `types`.

Note that this is just fixing the _internal_ organization of this code. How these types are exposed publicly is another matter that should also be fixed. In some cases, we are exporting the same type at multiple locations. (e.g. `ion_rs::Symbol`, `ion_rs::types::Symbol`, `ion_rs::element::Symbol`) and that's just asking for confusion.


src/element/mod.rs
```diff
@@ -25,26 +25,19 @@ use std::fmt::{Display, Formatter};
...
-pub use lob::{Blob, Clob};
 
-pub use list::List;
-pub use r#struct::Struct;
-pub use sequence::Sequence;
-pub use sexp::SExp;
+pub use crate::types::List;
+pub use crate::types::SExp;
+pub use crate::types::Sequence;
+pub use crate::types::Struct;
```
almann, line 40: I think I would at least put these together so they `rustfmt` together.

src/ion_hash/representation.rs
```diff
@@ -10,9 +10,9 @@ use crate::binary::{self, decimal::DecimalBinaryEncoder, timestamp::TimestampBin
 use crate::element::{Element, Sequence, Struct};
 use crate::ion_hash::element_hasher::ElementHasher;
 use crate::ion_hash::type_qualifier::type_qualifier_symbol;
-use crate::types::decimal::Decimal;
-use crate::types::integer::Int;
-use crate::{result::IonResult, types::timestamp::Timestamp, IonType, Symbol};
+use crate::types::Decimal;
+use crate::types::Int;
+use crate::{result::IonResult, types::Timestamp, IonType, Symbol};
```
almann, line 15: I would move `types::Timestamp` with the other `types` imports.

src/reader.rs
```diff
@@ -12,15 +12,15 @@ use crate::raw_reader::{RawReader, RawStreamItem};
...
 use crate::stream_reader::IonReader;
-use crate::symbol::Symbol;
 use crate::symbol_table::SymbolTable;
-use crate::types::decimal::Decimal;
-use crate::types::integer::Int;
-use crate::types::timestamp::Timestamp;
+use crate::types::Decimal;
+use crate::types::Int;
+use crate::types::Symbol;
+use crate::types::Timestamp;
```
almann, line 19: As I look through a bunch of these, how much should be be consistent with `types::{...}` style imports?

src/stream_reader.rs
```diff
@@ -85,7 +85,7 @@ pub trait IonReader {
     /// error is encountered while reading, returns [crate::IonError].
     fn read_i64(&mut self) -> IonResult<i64>;
 
-    /// Attempts to read the current item as an Ion integer and return it as an [crate::Int]. If the
+    /// Attempts to read the current item as an Ion integer and return it as an [crate::types::Int]. If the
```
almann, line 88:
```suggestion
    /// Attempts to read the current item as an Ion integer and return it as an [`Int`](crate::types::Int). If the
```

src/text/parsers/decimal.rs
```diff
@@ -13,9 +13,9 @@ use crate::text::parsers::numeric_support::{
 };
 use crate::text::parsers::stop_character;
 use crate::text::text_value::TextValue;
-use crate::types::coefficient::{Coefficient, Sign};
-use crate::types::decimal::Decimal;
-use crate::types::integer::UInt;
+use crate::types::Decimal;
+use crate::types::UInt;
+use crate::types::{Coefficient, Sign};
```
almann, line 18: Probably should make these all `types::{...}` or each on their own line.

src/text/parsers/timestamp.rs
```diff
@@ -14,8 +14,8 @@ use crate::text::parse_result::{
 };
 use crate::text::parsers::{stop_character, trim_zeros_and_parse_i32, trim_zeros_and_parse_u32};
 use crate::text::text_value::TextValue;
-use crate::types::decimal::Decimal;
-use crate::types::timestamp::{FractionalSecondSetter, Timestamp};
+use crate::types::Decimal;
+use crate::types::{FractionalSecondSetter, Timestamp};
```
almann, line 18: Consider consolidating these imports/making them consistent.

src/types/decimal.rs
```diff
@@ -6,8 +6,8 @@ use num_bigint::{BigInt, BigUint, ToBigUint};
 use crate::ion_data::IonEq;
 use crate::ion_data::IonOrd;
 use crate::result::{illegal_operation, IonError};
-use crate::types::coefficient::{Coefficient, Sign};
-use crate::types::integer::UInt;
+use crate::types::UInt;
+use crate::types::{Coefficient, Sign};
```
almann, line 10: Consider consolidating these imports/making them consistent.

src/types/decimal.rs
```diff
@@ -387,8 +387,8 @@ impl TryFrom<Decimal> for BigDecimal {
 #[cfg(test)]
 mod decimal_tests {
     use crate::result::IonResult;
-    use crate::types::coefficient::{Coefficient, Sign};
-    use crate::types::decimal::Decimal;
+    use crate::types::Decimal;
+    use crate::types::{Coefficient, Sign};
```
almann, line 391: Consider consolidating these imports/making them consistent.

src/types/iterators.rs
```diff
@@ -99,9 +99,9 @@ impl<'a> Iterator for FieldIterator<'a> {
 
 /// Iterates over the values associated with a given field name in a Struct.
 pub struct FieldValuesIterator<'a> {
-    pub(super) current: usize,
-    pub(super) indexes: Option<&'a IndexVec>,
-    pub(super) by_index: &'a Vec<(Symbol, Element)>,
+    pub(crate) current: usize,
+    pub(crate) indexes: Option<&'a IndexVec>,
+    pub(crate) by_index: &'a Vec<(Symbol, Element)>,
```
almann, line 104: See my top-level comment, this seems to indicate it is not in the right place.

almann (changes requested):
Looks mostly good to me.  Minor cleanup asks below.

I do wonder if `types::iterators` having stuff that is germane to `Element`/`ElementsIterator` and `Annotations` iterators are in the right place.  Generally iterator implementations should _live_ near what they iterate over.  So we either break out those things that are relevant to the `type` module or refactor it as a separate PR.  Thoughts?

popematt:
@almann I have moved the Element and Symbol iterators back to the `elements` mod. If you don't mind, I'd prefer to handle all of the style/formatting clean up in a followup PR.

almann (approved):
:shipit: LGTM

PR #562, opened 2023-06-07
Fix conversion to Decimal from f64 for values greater than +/- i64::MAX and smaller than +/- 1e-15

jpschorr:
Currently, `TryFrom<f64> for Decimal` unconditionally casts the `f64` coefficient to an `i64`, effectively clamping values larger than `i64::MAX` to `i64::MAX`.

This introduces an alternative code path for values larger than `i64::MAX` to create the `Coefficient` through a `BigInt`.

---

### DIFF
--- a/src/types/coefficient.rs
+++ b/src/types/coefficient.rs
@@ -128,7 +128,7 @@ impl TryFrom<Coefficient> for BigInt {
     /// converted is a negative zero, which BigInt cannot represent. Returns Ok otherwise.
     fn try_from(value: Coefficient) -> Result<Self, Self::Error> {
         if value.is_negative_zero() {
-            illegal_operation("Cannot convert negative zero Decimal to BigDecimal")?;
+            illegal_operation("Cannot convert negative zero Decimal to BigInt")?;
         }
         let mut big_int: BigInt = match value.magnitude {
             UInt::U64(m) => m.into(),
@@ -138,6 +138,22 @@ impl TryFrom<Coefficient> for BigInt {
             big_int.mul_assign(-1);
         }
         Ok(big_int)
+    }
+}
+
+impl TryFrom<BigInt> for Coefficient {
+    type Error = IonError;
+
+    fn try_from(value: BigInt) -> Result<Self, Self::Error> {
+        let (sign, magnitude) = value.into_parts();
+        let sign = match sign {
+            num_bigint::Sign::Minus => Sign::Negative,
+            num_bigint::Sign::Plus => Sign::Positive,
+            num_bigint::Sign::NoSign => {
+                return illegal_operation("Cannot convert sign-less BigInt to Decimal.")
+            }
+        };
+        Ok(Coefficient::new(sign, magnitude))
     }
 }
 
--- a/src/types/coefficient.rs
+++ b/src/types/coefficient.rs
@@ -150,7 +150,13 @@ impl TryFrom<BigInt> for Coefficient {
             num_bigint::Sign::Minus => Sign::Negative,
             num_bigint::Sign::Plus => Sign::Positive,
             num_bigint::Sign::NoSign => {
-                return illegal_operation("Cannot convert sign-less BigInt to Decimal.")
+                if magnitude.is_zero() {
+                    Sign::Positive
+                } else {
+                    return illegal_operation(
+                        "Cannot convert sign-less non-zero BigInt to Decimal.",
+                    );
+                }
             }
         };
         Ok(Coefficient::new(sign, magnitude))
--- a/src/types/decimal.rs
+++ b/src/types/decimal.rs
@@ -169,59 +169,76 @@ impl Decimal {
         UInt::BigUInt(scaled_coefficient).cmp(d2.coefficient.magnitude())
     }
 
-    /// Extract the integer and fractional parts of this as a pair of [`num_bigint::BigInt`]s
+    /// Extract the integer and fractional parts and the exponents of this as a `(BigInt, (BigInt, i64))`
     ///
     /// ```ignore
     /// # use num_bigint::BigInt;
     /// # use ion_rs::Decimal;
     /// let d1: Decimal = Decimal::new(123456789, -2);
-    /// let (int_part, frac_part) = d1.into_parts();
+    /// let (int_part, (frac_part, exponent)) = d1.into_parts();
     /// assert_eq!(int_part, BigInt::from(1234567u64));
     /// assert_eq!(frac_part, BigInt::from(89));
+    /// assert_eq!(exponent, -2);
     /// ```
-    pub(crate) fn into_parts(self) -> (BigInt, BigInt) {
+    pub(crate) fn into_parts(self) -> (BigInt, (BigInt, i64)) {
         let magnitude: BigInt = self.coefficient.try_into().unwrap();
         if self.exponent.is_zero() {
-            (magnitude, BigInt::zero())
+            (magnitude, (BigInt::zero(), 0))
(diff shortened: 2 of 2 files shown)


src/types/coefficient.rs
```diff
@@ -141,6 +141,22 @@ impl TryFrom<Coefficient> for BigInt {
...
+impl TryFrom<BigInt> for Coefficient {
+    type Error = IonError;
+
+    fn try_from(value: BigInt) -> Result<Self, Self::Error> {
+        let (sign, magnitude) = value.into_parts();
+        let sign = match sign {
+            num_bigint::Sign::Minus => Sign::Negative,
+            num_bigint::Sign::Plus => Sign::Positive,
+            num_bigint::Sign::NoSign => {
+                return illegal_operation("Cannot convert sign-less BigInt to Decimal.")
```
zslayton, line 153: I was really hoping [the documentation for `NoSign`](https://docs.rs/num-bigint/latest/num_bigint/enum.Sign.html#variants) would shed some light on its intended meaning, but no dice. :/

What does it mean for an integer to be signless?

src/types/decimal.rs
```diff
@@ -311,7 +311,18 @@ impl TryFrom<f64> for Decimal {
...
 
-        Ok(Decimal::new(coefficient as i64, exponent))
+        // prefer a compact representation for the coefficient; fallback to bigint
+        let coefficient: Coefficient = if coefficient <= i64::MAX as f64 {
+            (coefficient as i64).into()
+        } else {
+            coefficient
+                .trunc()
+                .to_bigint()
+                .ok_or_else(|| illegal_operation_raw("Cannot convert large f64 to Decimal."))?
```
zslayton, line 321: Could you add a unit test that exercises this code path? Glancing at [`num_bigint`'s implementation of `BigInt::from_f64`](https://docs.rs/num-traits/0.2.15/src/num_traits/cast.rs.html#516-522), it will succeed for values up to `u64::MAX` (not `i64::MAX`).

zslayton (commented):
Thanks for the fix!

src/types/coefficient.rs
jpschorr: I had hoped the same thing. 

Actually, looking through the source, the implication is that `NoSign` should only be valid for `0` ?

Cf.
- https://github.com/rust-num/num-bigint/blob/6f2b8e0fc218dbd0f49bebb8db2d1a771fe6bafa/src/bigint/division.rs#L51
- https://github.com/rust-num/num-bigint/blob/6f2b8e0fc218dbd0f49bebb8db2d1a771fe6bafa/src/bigrand.rs#L93-L96
- https://github.com/rust-num/num-bigint/blob/6f2b8e0fc218dbd0f49bebb8db2d1a771fe6bafa/src/bigint/shift.rs#L86C1-L88

jpschorr: Cf. https://github.com/rust-num/num-bigint/blob/6f2b8e0fc218dbd0f49bebb8db2d1a771fe6bafa/src/bigint.rs#L211-L230

src/types/decimal.rs
jpschorr: I think that's the wrong crate.

The `to_bigint` runs through [this](https://docs.rs/num-bigint/latest/src/num_bigint/bigint/convert.rs.html#170-177), which in turn runs through [this](https://docs.rs/num-bigint/latest/src/num_bigint/biguint/convert.rs.html#444-471)

src/types/coefficient.rs
```diff
@@ -141,6 +141,28 @@ impl TryFrom<Coefficient> for BigInt {
...
+
+    fn try_from(value: BigInt) -> Result<Self, Self::Error> {
+        let (sign, magnitude) = value.into_parts();
+        let sign = match sign {
+            num_bigint::Sign::Minus => Sign::Negative,
+            num_bigint::Sign::Plus => Sign::Positive,
+            num_bigint::Sign::NoSign => {
+                if magnitude.is_zero() {
+                    Sign::Positive
+                } else {
```
zslayton, line 155: 👍

src/types/decimal.rs
```diff
@@ -167,6 +168,63 @@ impl Decimal {
...
+        } else if self.exponent.is_negative() {
+            let divisor = BigInt::from(10u64).pow((-self.exponent) as u32);
+            magnitude.div_rem(&divisor)
+        } else {
+            let multiplicand = BigInt::from(10u64).pow(self.exponent as u32);
+            (magnitude * multiplicand, BigInt::zero())
+        }
+    }
+
+    /// Returns ([`num_bigint::BigInt`], [`f64`]) representing the differences
```
zslayton, line 195: I think this was previously `(BigInt, BigInt)` -- why the switch to `f64`?

src/types/decimal.rs
```diff
@@ -167,6 +168,63 @@ impl Decimal {
...
+        let (d2_int, d2_frac) = d2.into_parts();
+
+        // turn a fractional part (e.g., `123`) into an f64 less than `0` (e.g., `0.123`)
+        fn to_fract(frac: BigInt) -> f64 {
+            if frac.is_zero() {
+                0.0
+            } else {
+                let frac_oom = max(1, 1 + frac.abs().to_u128().unwrap().ilog10());
+                frac.to_f64().unwrap() / 10f64.powi(frac_oom as i32)
+            }
```
zslayton, line 220: Could you add a comment here explaining the logic? I'm also unsure what `_oom` means here, my brain can't get away from `OutOfMemory`.

src/types/decimal.rs
```diff
@@ -514,6 +593,100 @@ mod decimal_tests {
...
     }
 
+    #[test]
+    fn test_difference_by_parts() {
+        let d1: Decimal = Decimal::new(123456789, -2);
+        let d10: Decimal = Decimal::new(123456789, -1);
+        let d100: Decimal = Decimal::new(123456789, 0);
+
+        // diff d1 with d1
+        let (diff_integer, diff_fractional) = Decimal::difference_by_parts(d1.clone(), d1.clone());
```
zslayton, line 603: Can we make `difference_by_parts` take `&Decimal`s and `clone()` them itself? The cloning seems like something that we might want to change/optimize later.

zslayton (commented):
This looks really good, thanks! A few more questions/comments below.

src/types/decimal.rs
jpschorr: It was.

Because, for example, with a `BigInt`, the fractional part of `1.2` was `2` and the fractional part of `1.000000002` was `2`, which does not allow the comparison that was useful.

src/types/decimal.rs
jpschorr: will do. Order Of Magnitude.

src/types/decimal.rs
```diff
@@ -627,22 +648,36 @@ mod decimal_tests {
...
+            &Decimal::new(12, -1),
+        );
+        // 1 - 1 -> 0
+        assert_eq!(diff_integer, 0.into());
+        // 0.0000002 - 0.2 -> -0.1999998
+        let expected = -0.1999998;
+        // ideally, this would be based on ULPs, not epsilon
+        //  Cf. https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/
+        let relative_error = ((diff_fractional - expected) / expected).abs();
+        assert!(relative_error <= 10.0 * f64::EPSILON);
```
zslayton, line 664: Learning opportunity for me: was `10.0` selected arbitrarily, or is that a best practice?

zslayton (approved):

Issue #547, opened 2023-05-10
Make `ElementStreamReader` publicly accessible


jpschorr:
`ElementStreamReader` seems useful to allow writing one implementation of walking an Ion value, but it is not currently publicly accessible (https://github.com/amazon-ion/ion-rust/blob/main/src/element/mod.rs#L26).

It would be even nicer if it were plumbed through to  `ReaderBuilder` as well.

This would allow something like:

```rust
let mut reader = ion_rs::ReaderBuilder::new()
        .build(element)
        .expect("element to be readable");
```

PR #563, opened 2023-06-07
Add `ElementStreamWriter`, an `IonWriter` that outputs `Element`s

jpschorr:
Related to #547 

- Makes `ElementStreamReader` public
- Adds an `ElementStreamWriter`
  - this is an `IonWriter` that outputs each top-level value to an `Extend<Element>` 

Currently, this is targeted against the branch in #562  as it requires the decimal fixes for the unit tests to succeed, but I'll switch it to compare to main if/when that is merged.

---

### DIFF
--- a/src/element/element_stream_writer.rs
+++ b/src/element/element_stream_writer.rs
@@ -4,7 +4,8 @@ use crate::raw_symbol_token_ref::AsRawSymbolTokenRef;
 use crate::result::illegal_operation;
 use crate::types::Bytes;
 use crate::{
-    Decimal, Int, IonResult, IonType, IonWriter, RawSymbolTokenRef, Str, Symbol, Timestamp,
+    Decimal, Int, IonResult, IonType, IonWriter, RawSymbolTokenRef, Str, Symbol, SymbolTable,
+    Timestamp,
 };
 
 // Represents a level into which the writer has stepped.
@@ -16,7 +17,7 @@ struct ContainerContext {
     container: Container,
 }
 
-// A container for values that writer will writ into.
+// A container for values that the writer will write into.
 // A writer that has not yet called step_in() is at the top level.
 #[derive(Debug, PartialEq)]
 enum Container {
@@ -36,6 +37,7 @@ where
     containers: Vec<ContainerContext>,
     annotations: Annotations,
     field_name: Option<Symbol>,
+    symbol_table: SymbolTable,
 }
 
 impl<'e, E> ElementStreamWriter<'e, E>
@@ -52,6 +54,7 @@ where
             }],
             annotations: Annotations::empty(),
             field_name: None,
+            symbol_table: SymbolTable::default(),
         }
     }
 
@@ -120,9 +123,16 @@ where
         A: AsRawSymbolTokenRef,
     {
         match sym_ref.as_raw_symbol_token_ref() {
-            RawSymbolTokenRef::SymbolId(_symbol_id) => {
-                // TODO what to do with symbol ids?
-                todo!()
+            RawSymbolTokenRef::SymbolId(symbol_id) => {
+                if self.symbol_table.sid_is_valid(symbol_id) {
+                    Ok(self
+                        .symbol_table
+                        .symbol_for(symbol_id)
+                        .cloned()
+                        .unwrap_or(Symbol::unknown_text()))
+                } else {
+                    illegal_operation(format!("Symbol ID ${symbol_id} is undefined."))
+                }
             }
             RawSymbolTokenRef::Text(txt) => Ok(Symbol::owned(txt)),
         }
@@ -140,8 +150,8 @@ where
     }
 
     fn write_ion_version_marker(&mut self, _major: u8, _minor: u8) -> IonResult<()> {
-        // TODO what does this mean for an element writer?
-        todo!()
+        // Nothing to do here
+        Ok(())
     }
 
     fn supports_text_symbol_tokens(&self) -> bool {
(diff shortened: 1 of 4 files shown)


src/element/element_stream_writer.rs
```diff
@@ -0,0 +1,697 @@
...
+    where
+        A: AsRawSymbolTokenRef,
+    {
+        match sym_ref.as_raw_symbol_token_ref() {
+            RawSymbolTokenRef::SymbolId(_symbol_id) => {
+                // TODO what to do with symbol ids?
+                todo!()
+            }
+            RawSymbolTokenRef::Text(txt) => Ok(Symbol::owned(txt)),
+        }
```
jpschorr, line 128: I'm unsure what cases could lead to a symbol id being used here, and thus unsure how to handle this case. Error?

src/element/element_stream_writer.rs
```diff
@@ -0,0 +1,697 @@
...
+    type Output = &'e mut E;
+
+    fn ion_version(&self) -> (u8, u8) {
+        (1, 0)
+    }
+
+    fn write_ion_version_marker(&mut self, _major: u8, _minor: u8) -> IonResult<()> {
+        // TODO what does this mean for an element writer?
+        todo!()
+    }
```
jpschorr, line 145: What should this do? Is it a no-op?

src/element/element_stream_writer.rs
```diff
@@ -0,0 +1,697 @@
...
+    fn set_annotations<I, A>(&mut self, annotations: I)
+    where
+        A: AsRawSymbolTokenRef,
+        I: IntoIterator<Item = A>,
+    {
+        let annotations: IonResult<_> =
+            annotations.into_iter().map(|s| self.to_symbol(s)).collect();
+        // TODO what to do with the expect?
+        self.set_current_annotations(Annotations::new(annotations.expect("symbols")));
+    }
```
jpschorr, line 160: This is one place that symbol ids could be passed to `to_symbol`, but this interface method is not fallible.

src/element/element_stream_writer.rs
```diff
@@ -0,0 +1,697 @@
...
+        self.push_container(container)
+    }
+
+    fn set_field_name<A: AsRawSymbolTokenRef>(&mut self, name: A) {
+        // TODO what to do with the panic?
+        match self.to_symbol(name) {
+            Ok(s) => self.field_name = Some(s),
+            Err(_e) => panic!("set field name invalid symbol"),
+        }
+    }
```
jpschorr, line 227: This is one place that symbol ids could be passed to `to_symbol`, but this interface method is not fallible.

src/element/element_stream_writer.rs
```diff
@@ -0,0 +1,697 @@
...
+            },
+            Value::Int(Int::I64(7)).with_annotations(["foo", "bar", "baz quux"]),
+        );
+    }
+
+    #[test]
+    fn write_decimal() {
+        // returns (BigInt, BigInt) representing the differences between integer and fractional
+        // components of d1 & d2 respectively
+        fn parts_diff(d1: Decimal, d2: Decimal) -> (BigInt, BigInt) {
```
jpschorr, line 399: This is copied verbatim from #562 . Perhaps it could be extracted somewhere appropriate?

jpschorr (commented):
PR Tour of the places that I could use some guidance to bring this in accordance with the larger codebase.

src/element/element_stream_writer.rs
zslayton: Try to resolve the SID using the symbol table. In this case, that's probably just the [system symbol table](https://amazon-ion.github.io/ion-docs/docs/symbols.html#system-symbols). If you instantiate a `SymbolTable`, it'll contain the system symbols by default.

If the symbol ID's out of bounds (ie `>=10`), that's an `illegal_operation` error. If it's `0`, you can construct a `Symbol::unknown_text()`.

src/element/element_stream_writer.rs
```diff
@@ -0,0 +1,697 @@
...
+// Represents a level into which the writer has stepped.
+// A writer that has not yet called step_in() is at the top level.
+#[derive(Debug, PartialEq)]
+struct ContainerContext {
+    annotations: Annotations,
+    field_name: Option<Symbol>,
+    container: Container,
+}
+
+// A container for values that writer will writ into.
```
zslayton, line 19:
```suggestion
// A container for values that the writer will write into.
```

src/element/element_stream_writer.rs
zslayton: Yeah, I think it's a no-op since you're not managing/materializing/emitting a symbol table. @almann and I have been talking about how this (and a few other methods in `IonWriter`) aren't really a good fit for the trait.

src/element/element_stream_writer.rs
zslayton: Great catch. I went to see how the `BinaryWriter` handles this constraint and found an embarrassing leftover:

https://github.com/amazon-ion/ion-rust/blob/fd09e870e5206be16a1c535da0abdd450c6ea6b5/src/binary/binary_writer.rs#L124-L130

🤦

I've opened #564 to track addressing this down the line. For now, have it panic and link to the GH issue.

src/element/element_stream_writer.rs
zslayton: Same thought as above--attempt to resolve; otherwise panic and link to #564.

src/element/element_stream_writer.rs
zslayton: Yeah, I think this can live in `Decimal`. Could you add a couple of example input/output pairs to the doc comment to illustrate its behavior?

jpschorr: Moved in #562, will rebase this onto that and use it here.

src/element/element_stream_writer.rs
jpschorr: Done

src/element/element_stream_writer.rs
jpschorr: Done

src/element/element_stream_writer.rs
jpschorr: Done

jpschorr:
> Thanks! I notice this PR still marked `Draft`, did you have other work you want to incorporate, or can I merge this after #562 lands?

Nothing else for this, just waiting for #562

PR #567, opened 2023-06-14
[User]Reader wraps SystemReader, not RawReader

zslayton:
The `UserReader` (alias: `Reader`) is implemented as a `(RawReader, SymbolTable)` pair. If/when the `RawReader` encounters a symbol table, the `Reader` consumes it and adds its symbols to the `SymbolTable`.

The `SystemReader` is a very similar; it is also a `(RawReader, SymbolTable)` pair. However, when the `RawReader` encounters a symbol table, the `SystemReader` cannot consume it. Instead, it needs to process _only_ the value over which the `RawReader` is currently positioned. This requires some complex state tracking, but it allows the `SystemReader` to expose/return system-level values while also processing them.

This PR modifies the `UserReader` to be a thin wrapper around the `SystemReader`. Whenever the `SystemReader` reports one or more system-level values, the `UserReader` skips over them until it finds the next user-level value or exhausts the input stream. This change is in commit https://github.com/amazon-ion/ion-rust/commit/727417f3548b4f73cf1f40ada9b36a8c4199a75d.

When the `SystemReader` was originally written, implementations of the `RawReader` were not required to have idempotent `read_*` operations; they tended to consume the value when it was read, meaning applications had to materialize/store the value if they wished to refer to it multiple times.  Given this constraint, when the `SystemReader` encountered part of a symbol table, it had to materialize it so it could be processed once by the `SystemReader` and (potentially) a second time by the user.

This constraint no longer exists; all `RawReader` implementations now allow a user to call `read_*` any number of times, always returning the same value. This PR also removes the machinery that was needed to store a materialized copy of system values; it does this in commit e6c3b87af398d0fcf47f85cdf865b5b1df59bfe9.

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -188,7 +188,7 @@ impl<R: RawReader> UserReader<R> {
     }
 
     pub fn symbol_table(&self) -> &SymbolTable {
-        &self.system_reader.symbol_table()
+        self.system_reader.symbol_table()
     }
 }
 
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -5,16 +5,18 @@ use std::ops::Range;
 use delegate::delegate;
 
 use crate::binary::constants::v1_0::IVM;
-use crate::constants::v1_0::system_symbol_ids;
+use crate::binary::non_blocking::raw_binary_reader::RawBinaryReader;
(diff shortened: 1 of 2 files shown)


desaikd (approved):
Looks good!

Issue #507, opened 2023-04-05
Mitigation of 3P BigInt in Public API


almann:
Currently, we expose a 3P `num_bigint::BigInt` (and friends) in our public API.  This is a problem as our stability is a moving target and `num_bigint` is also pre-1.0.  We have a few options here:

1. Wrap `BigInt` in our APIs (similar to what we do in `Decimal` and `Timestamp`), let's call this `LargeInt` for the purposes of discussion.
1. Eliminate `BigInt` entirely and replace with `i128` as a _good enough_:tm: solution.  @zslayton suggested this and may prefer it.
1. Re-export `num_bigint` into `ion_rs::external::num_bigint` and make all of our type references line up to this.  This is effectively aliasing the type in our crate.

I am of the opinion that we should do (3) and provide a `TryInto<i128>` for the `Int` enum and `TryInto<u128>` for the `UInt` enum.  If, and only if, we have a compelling reason to create `LargeInt` should we do so.

(1) and (3) have the problem that a user consuming our wrapper with a local dependency of `BigInt` still requires conversion code (even for `LargeInt`), but is tractable (we cannot auto generate this code, but we could maybe provide a macro to generate the conversion assuming structure of the API--TBH, I don't love this).

(2) has the problem that integers over 16-bytes cannot be represented.  In practice this is probably okay as we're talking integers in the range`-170,141,183,000,000,002,920,328,542,032,742,055,937` to `170,141,183,000,000,002,920,328,542,032,742,055,936`.  I find this disappointing because it would be a regression of functionality only because of the 3P dependency (we wouldn't naturally do this if `BigInt` was in `std`).

popematt:
**Option 4**
Fundamentally, the problem is that we have a transparent `Int` (and `UInt`) enum.

If we make `Int` opaque to hide the underlying implementation, then we can do whatever we want behind the scenes.
We can provide:
* `impl TryFrom<Int> for i128` -- *good enough*™️ for most cases
* `impl From<Int> for Vec<u8>` -- where Vec<u8> is the 2's complement representation of the integer value; this is a lowest-common-denominator format that consumers of the library can use to convert from our opaque int to their own biggish-int type.

In the future, if we deem this soution insufficient, it would be backwards compatible to introduce a `LargeInt` and add `From<Int> for LargeInt` or to re-export `BigInt` and add `impl From<Int> for BigInt`.

almann:
[...]

PR #568, opened 2023-06-15
Makes `UInt` an opaque struct type

zslayton:
Prior to this patch, `UInt` was an enum offering two storage
options: `u64` and `BigUint`.

This patch wraps that enum (now a private type called `UIntData`) in
an opaque struct, giving us the ability to change that representation
in the future without a breaking change.

This also provides a `TryFrom<&UInt>` implementation for all primitive
integer types.

Related to #507.

A follow-on PR will make the same change for the signed `Int` type.

### DIFF
--- a/src/binary/decimal.rs
+++ b/src/binary/decimal.rs
@@ -11,6 +11,7 @@ use crate::binary::var_int::VarInt;
 use crate::binary::var_uint::VarUInt;
 use crate::ion_data::IonEq;
 use crate::result::IonResult;
+use crate::types::integer::UIntData;
 use crate::types::{Coefficient, Decimal, Sign, UInt};
 use crate::IonError;
 
@@ -18,7 +19,9 @@ const DECIMAL_BUFFER_SIZE: usize = 32;
 const DECIMAL_POSITIVE_ZERO: Decimal = Decimal {
     coefficient: Coefficient {
         sign: Sign::Positive,
-        magnitude: UInt::U64(0),
+        magnitude: UInt {
+            data: UIntData::U64(0),
+        },
     },
     exponent: 0,
 };
@@ -69,9 +72,9 @@ where
             }
         } else {
             // Otherwise, allocate a Vec<u8> with the necessary representation.
-            let mut coefficient_bytes = match decimal.coefficient.magnitude() {
-                UInt::U64(unsigned) => unsigned.to_be_bytes().into(),
-                UInt::BigUInt(big) => big.to_bytes_be(),
+            let mut coefficient_bytes = match &decimal.coefficient.magnitude().data {
+                UIntData::U64(unsigned) => unsigned.to_be_bytes().into(),
+                UIntData::BigUInt(big) => big.to_bytes_be(),
             };
 
             let first_byte: &mut u8 = &mut coefficient_bytes[0];
--- a/src/binary/non_blocking/binary_buffer.rs
+++ b/src/binary/non_blocking/binary_buffer.rs
@@ -7,7 +7,7 @@ use crate::binary::uint::DecodedUInt;
 use crate::binary::var_int::VarInt;
 use crate::binary::var_uint::VarUInt;
 use crate::result::{decoding_error, incomplete_data_error, incomplete_data_error_raw};
-use crate::types::{Int, UInt};
+use crate::types::Int;
 use crate::{IonResult, IonType};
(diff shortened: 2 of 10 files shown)


src/binary/decimal.rs
```diff
@@ -11,14 +11,17 @@ use crate::binary::var_int::VarInt;
...
 use crate::IonError;
 
 const DECIMAL_BUFFER_SIZE: usize = 32;
 const DECIMAL_POSITIVE_ZERO: Decimal = Decimal {
     coefficient: Coefficient {
         sign: Sign::Positive,
-        magnitude: UInt::U64(0),
+        magnitude: UInt {
+            data: UIntData::U64(0),
+        },
```
zslayton, line 24: 🗺️ `magnitude: 064.into()` would be more concise, but the trait method `into()` is not `const`.

src/binary/decimal.rs
```diff
@@ -69,9 +72,9 @@ where
             }
         } else {
             // Otherwise, allocate a Vec<u8> with the necessary representation.
-            let mut coefficient_bytes = match decimal.coefficient.magnitude() {
-                UInt::U64(unsigned) => unsigned.to_be_bytes().into(),
-                UInt::BigUInt(big) => big.to_bytes_be(),
+            let mut coefficient_bytes = match &decimal.coefficient.magnitude().data {
+                UIntData::U64(unsigned) => unsigned.to_be_bytes().into(),
+                UIntData::BigUInt(big) => big.to_bytes_be(),
```
zslayton, line 77: 🗺️ This diff contains several instances of this change. Instead of matching directly on the `UInt`, we now match on its private inner `UIntData` field.

src/binary/non_blocking/binary_buffer.rs
```diff
@@ -689,7 +690,7 @@ mod tests {
         let mut buffer = BinaryBuffer::new(&[0b1000_0000]);
         let var_int = buffer.read_uint(buffer.remaining())?;
         assert_eq!(var_int.size_in_bytes(), 1);
-        assert_eq!(var_int.value(), &UInt::U64(128));
+        assert_eq!(var_int.value(), &UInt::from(128));
```
zslayton, line 693: 🗺️ I used `TypeName::from` instead of `value.into()` in several places because that's the only way to specify the expected type inline. You can't use turbofish syntax with `Into`.

src/types/integer.rs
```diff
@@ -64,7 +63,12 @@ pub trait IntAccess {
...
-pub enum UInt {
+pub struct UInt {
+    pub(crate) data: UIntData,
+}
+
+#[derive(Debug, Clone)]
+pub(crate) enum UIntData {
     U64(u64),
     BigUInt(BigUint),
 }
```
zslayton, line 74: 🗺️ This is the core change; the rest of the diff is comprised of ripples.

src/types/integer.rs
```diff
@@ -220,13 +230,34 @@ impl_uint_from_small_signed_int_types!(i8, i16, i32, i64, isize);
...
+                match &value.data {
+                    UIntData::U64(value) => <$t>::try_from(*value).map_err(|_|()),
+                    UIntData::BigUInt(value) => <$t>::try_from(value).map_err(|_|()),
+                }
+            }
+        }
+    )*)
+}
+
+impl_int_types_from_uint!(i8, i16, i32, i64, i128, u8, u16, u32, u64, u128);
```
zslayton, line 256: 🗺️ Previously, we only provided a `TryFrom<&UInt>` implementation for `usize` and `i64`. That was more or less acceptable because `UInt` wasn't part of the public API. However, I think it needs to be in a couple of places. For example, a `Decimal::coefficient` method.

src/types/integer.rs
```diff
@@ -252,44 +283,6 @@ fn big_integer_from_big_uint(value: BigUint) -> Int {
...
-            }),
-            UInt::BigUInt(big_uint) => usize::try_from(big_uint).or_else(|_| {
-                decoding_error(format!(
-                    "Unsigned integer {big_uint:?} was too large to be represented as an usize."
-                ))
-            }),
-        }
-    }
-}
-
```
zslayton, line 292: 🗺️ These manual implementations of `TryFrom` have been supplanted by the macro we just saw above.

zslayton (commented):
🗺️ PR tour

popematt (approved):

PR #570, opened 2023-06-15
Signed type impls of Into<UInt> are now fallible

### LINKED ISSUE #569: Converting from a signed integer to a `UInt` should be fallible
There are a few methods related to decimals that expect a parameter that can be converted into a `UInt`--that is, a `I: Into<UInt>`. For example:

https://github.com/amazon-ion/ion-rust/blob/82f17fc771cb7ea7db63ac05aa4bca2d73803e05/src/types/coefficient.rs#L29-L33

Because `UInt` was an internal-only type, we allowed signed integers to be converted to a `UInt` by taking their absolute value. This allowed bare integers (assumed to be `i32` if no other hints are present) to be passed in directly in several dozen places.

However, now:
1. We no longer use that implicit `abs()` functionality.
2. We need `UInt` to be publicly visible.

We need to convert those `From<SignedInt> for UInt` to `TryFrom<SignedInt> to UInt`.

zslayton:
Fixes #569.

* Signed integer primitives now implement `TryInto<UInt>` and fail if they are negative.
* `Coefficient::new` now takes `I: Into<Magnitude>` where `Magnitude` is a wrapper that signed integers infallibly convert to via absolute value.

### DIFF
--- a/src/binary/int.rs
+++ b/src/binary/int.rs
@@ -197,7 +197,7 @@ impl From<DecodedInt> for Coefficient {
         } = int;
         use types::Sign::{Negative, Positive};
         let sign = if is_negative { Negative } else { Positive };
-        Coefficient::new(sign, value)
+        Coefficient::new(sign, value.unsigned_abs())
     }
 }
 
--- a/src/binary/non_blocking/binary_buffer.rs
+++ b/src/binary/non_blocking/binary_buffer.rs
@@ -690,7 +690,7 @@ mod tests {
         let mut buffer = BinaryBuffer::new(&[0b1000_0000]);
         let var_int = buffer.read_uint(buffer.remaining())?;
         assert_eq!(var_int.size_in_bytes(), 1);
-        assert_eq!(var_int.value(), &UInt::from(128));
+        assert_eq!(var_int.value(), &UInt::from(128u64));
         Ok(())
     }
 
@@ -699,7 +699,7 @@ mod tests {
         let mut buffer = BinaryBuffer::new(&[0b0111_1111, 0b1111_1111]);
         let var_int = buffer.read_uint(buffer.remaining())?;
         assert_eq!(var_int.size_in_bytes(), 2);
-        assert_eq!(var_int.value(), &UInt::from(32_767));
+        assert_eq!(var_int.value(), &UInt::from(32_767u64));
         Ok(())
     }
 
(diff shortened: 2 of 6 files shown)


src/binary/int.rs
```diff
@@ -197,7 +197,7 @@ impl From<DecodedInt> for Coefficient {
         } = int;
         use types::Sign::{Negative, Positive};
         let sign = if is_negative { Negative } else { Positive };
-        Coefficient::new(sign, value)
+        Coefficient::new(sign, value.unsigned_abs())
```
zslayton, line 200: 🗺️ All signed int primitives have an `unsigned_abs()` method that returns their absolute value as an unsigned integer of the closest-sized type (`i8`->`u8`, `i16`->`u16`, etc). Now the `Int` type has an `unsigned_abs()` method that returns a `UInt`.

src/binary/non_blocking/binary_buffer.rs
```diff
@@ -690,7 +690,7 @@ mod tests {
         let mut buffer = BinaryBuffer::new(&[0b1000_0000]);
         let var_int = buffer.read_uint(buffer.remaining())?;
         assert_eq!(var_int.size_in_bytes(), 1);
-        assert_eq!(var_int.value(), &UInt::from(128));
+        assert_eq!(var_int.value(), &UInt::from(128u64));
```
zslayton, line 693: 🗺️ Signed integers can no longer be infallibly converted to a `UInt`. In the absence of other type information, Rust treats integers as `i32`s. This means that to call `UInt::from(...)`, we need to mark our integers as an unsigned type.

src/types/coefficient.rs
```diff
@@ -16,6 +16,67 @@ pub enum Sign {
...
 
+/// A simple wrapper type around a [`UInt`]. Users are not expected to construct
+/// instances of `Magnitude` directly; it is provided for conversion ergonomics.
+///
+/// Signed integer types cannot implement `Into<UInt>`. Instead, they implement
+/// `TryInto<UInt>` and report an error if the input integer is negative.
+///
+/// Signed integers can infallibly implement `Into<Magnitude>` by using their
+/// absolute value.
+pub struct Magnitude(UInt);
```
zslayton, line 27: 🗺️ Previously, as described in #569, signed integers being converted to `UInt` would use their absolute value. This wrapper type offers the same behavior, but is narrowly used in APIs where that wouldn't be surprising.

zslayton (commented):
🗺️ PR tour

popematt (approved):

PR #572, opened 2023-06-16
Replaces `Vec<Element>` with `Sequence`

zslayton:
APIs that returned `Vec<Element>` now return a `Sequence` instead. This allows us to change the storage type without a breaking change. It also means that you can trivially construct a `List` or `SExp` from the output of `Element::read_all`.

This also renames `ElementsIterator` to `SequenceIterator` to avoid confusion with the similarly named `ElementIterator`.

### DIFF
--- a/src/element/builders.rs
+++ b/src/element/builders.rs
@@ -255,7 +255,7 @@ impl StructBuilder {
 /// ```
 #[macro_export]
 macro_rules! ion_list {
-    ($($element:expr),*) => {{
+    ($($element:expr),* $(,)?) => {{
         use $crate::element::Sequence;
         Sequence::builder()$(.push($element))*.build_list()
     }};
@@ -339,7 +339,7 @@ macro_rules! ion_sexp {
 /// ```
 #[macro_export]
 macro_rules! ion_struct {
-    ($($field_name:tt : $element:expr),*) => {{
+    ($($field_name:tt : $element:expr),* $(,)?) => {{
         use $crate::element::Struct;
         Struct::builder()$(.with_field($field_name, $element))*.build()
     }};
@@ -413,6 +413,9 @@ mod tests {
     fn make_list_with_macro() {
         let actual: Element = ion_list![1, true, "foo", Symbol::owned("bar")].into();
         let expected = Element::read_one(r#"[1, true, "foo", bar]"#).unwrap();
+        assert_eq!(actual, expected);
+        // Trailing commas are allowed
+        let actual: Element = ion_list![1, true, "foo", Symbol::owned("bar"),].into();
         assert_eq!(actual, expected);
     }
 
@@ -515,6 +518,16 @@ mod tests {
         }
         .into();
         let expected = Element::read_one(r#"{a: 1, b: true, c: "foo", d: bar}"#).unwrap();
+        assert_eq!(actual, expected);
+
+        // Trailing commas are allowed
+        let actual: Element = ion_struct! {
+            "a": 1,
+            "b": true,
+            "c": "foo",
+            "d": Symbol::owned("bar"), // <-- trailing comma
+        }
+        .into();
         assert_eq!(actual, expected);
     }
 
--- a/src/element/builders.rs
+++ b/src/element/builders.rs
@@ -345,6 +345,29 @@ macro_rules! ion_struct {
     }};
 }
 
+/// Constructs a [`Sequence`] with the specified child values.
+///
+/// Note that a `Sequence` is NOT a type of `Element`. However, one can convert a `Sequence` into a
+/// `List` or `SExp`.
+///
+/// ```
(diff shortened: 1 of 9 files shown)


src/element/builders.rs
zslayton, line 1: 🗺️ This PR makes the `Sequence` type (currently wrapped by `List` and `SExp`) more front-and-center. It represents _any_ sequence of Ion values, including a sequence of top-level values.

As such, we now provide a macro to easily construct one: `ion_seq!`.

src/element/builders.rs
```diff
@@ -345,6 +345,29 @@ macro_rules! ion_struct {
...
+/// Note that a `Sequence` is NOT a type of `Element`. However, one can convert a `Sequence` into a
+/// `List` or `SExp`.
+///
+/// ```
+/// use ion_rs::element::{Element, Sequence};
+/// use ion_rs::{ion_seq, ion_list};
+/// // Construct a Sequence from Rust values
+/// let actual: Sequence = ion_seq!["foo" 7 false ion_list![1.5f64, -8.25f64]];
+/// // Construct a Sequence from serialized Ion data
+/// let expected: Sequence = Element::read_all(r#" "foo" 7 false [1.5e0, -8.25e0] "#).unwrap();
```
zslayton, line 359: 🗺️ `ion_seq!` doesn't take commas between its values in order to better approximate the syntax for top-level values in an Ion stream.

src/element/iterators.rs
```diff
@@ -41,7 +41,7 @@ create_new_slice_iterator_type!(
     // Used for iterating over an Element's annotations
     SymbolsIterator => Symbol,
     // Used for iterating over a Sequence's Elements
-    ElementsIterator => Element
+    SequenceIterator => Element
```
zslayton, line 44: 🗺️ There's another type (`ElementIterator`) that nearly clashes with `ElementsIterator`. I think `SequenceIterator` is a bit clearer anyway.

src/element/mod.rs
```diff
@@ -578,9 +579,17 @@ impl Element {
...
+        Element::iter(data.as_ref())?.collect()
+    }
+
+    /// Returns an iterator over the Elements in the provided Ion data source.
+    /// If the data source cannot be read or contains invalid Ion data, this method
+    /// will return an `Err`.
+    pub fn iter<'a, I: ToIonDataSource + 'a>(
+        source: I,
+    ) -> IonResult<impl Iterator<Item = IonResult<Element>> + 'a> {
+        Ok(ReaderBuilder::default().build(source)?.into_elements())
```
zslayton, line 592: 🗺️ For the 1.0 stabilization process, we plan to label the streaming APIs as experimental and make them opt-in. This method will allow folks who aren't using `Reader`/`ReaderBuilder` to iterate over the elements in a data source without reading the whole thing into memory first.

src/element/reader.rs
```diff
@@ -94,6 +101,18 @@ impl<'a, R: ElementReader + ?Sized> Iterator for ElementIterator<'a, R> {
     }
 }
 
+pub struct OwnedElementIterator<R: ElementReader> {
+    reader: R,
+}
```
zslayton, line 106: 🗺️ `Reader` already had a non-owning iterator, but we needed a version that would take ownership of the `ElementReader` implementation to extend its lifetime.

zslayton (commented):
🗺️ PR tour

src/element/builders.rs
popematt: Which approach does the `ion_sexp!` macro take? Do you think the similarity to Ion syntax is really that significant vs the familiarity of having a comma-delimited list of values?

popematt (approved):
I think I would prefer comma-delimited values in the `ion_seq!` macro, but it's not something I have a particularly strong opinion about. Overall the change is good.

zslayton:
> I think I would prefer comma-delimited values in the ion_seq! macro, but it's not something I have a particularly strong opinion about. Overall the change is good.

I originally wrote it that way to hew close to the `vec![]` syntax, but then realized as I was writing the doc example that the others macros (`ion_list![]`, `ion_sexp!()`, and `ion_struct!{}`) each mirror the corresponding Ion syntax. `ion_sexp!()` doesn't take commas between its values, for example.

I could go either way for `ion_seq!` since:
* It's not an Ion type
* A `Sequence` can be a top-level sequence or the contents of a `List` or `SExp`
* `vec![]` has taught people to expect commas between values

zslayton:
The latest commits make two changes:
1. `ion_seq!` will now accept arguments with or without delimiting commas, allowing users to choose which makes sense based on how they intend to use it and/or what they're comfortable with.
2. `ion_list!` and `ion_struct!` now accept an optional trailing comma, something that's bitten me a bunch of times.

popematt (approved):

Issue #576, opened 2023-06-20
Move unstable APIs behind `experimental-*` feature flags


zslayton:
For `ion-rust` v1.0, we'd like to stabilize a subset of the library's APIs that we believe is both useful and unlikely to change in the near future--the `Element` API. (Note that while `Element` is useful, it is not _complete_; we intend to APIs for accessing deeply nested values and providing stronger typing.)

To this end, we plan to begin migrating APIs that _are_ likely to change soon behind feature flags to communicate to users that they do not yet come with any backwards compatibility guarantees between versions. At a minimum, this includes `IonReader`, `IonWriter`, and their respective implementations. Users can continue to use them as they have been by opting into their experimental nature explicitly.

PR #577, opened 2023-06-20
Adds `"experimental-reader"` feature

zslayton:
This PR is the first of a short series that will move the streaming types behind an experimental feature. It is part of issue #576.

In this PR, we move the public re-exports of:
* `IonReader`
* `Reader`/`StreamItem`
* `SystemReader`/`SystemStreamItem`
* `RawReader`/`RawStreamItem`
* `BlockingRawReader`
behind the feature flag `"experimental-reader"`.

No behavior is changed, and internal modules can continue to access these types as desired.


desaikd (approved):

examples/read_all_values.rs
zslayton, line 1: 🗺️ I changed about 5 lines in this source file and it caused cascading formatting changes that produced a scary-looking diff. x_x

All of the existing code was moved into a module annotated with `#[cfg(feature = "experimental-reader")]` since it heavily leverages those APIs. I also added a fallback `fn main()` that prints an error if you run it without the `experimental-reader` feature enabled. There are no other logic changes.

src/binary/binary_writer.rs
```diff
@@ -201,10 +201,10 @@ impl<W: Write> IonWriter for BinaryWriter<W> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::ion_reader::IonReader;
     use crate::reader::ReaderBuilder;
-    use crate::stream_reader::IonReader;
 
-    use crate::StreamItem::Value;
+    use crate::reader::StreamItem::Value;
```
zslayton, line 207: 🗺️ Now that the root-level public re-exports of the reader-related types are only available conditionally, internal modules need to refer to the internally visible path of each type.

In this example, `crate::StreamItem::Value` became `crate::reader::StreamItem::Value` because `StreamItem` is no longer guaranteed to be publicly visible at the crate root.

src/binary/binary_writer.rs
```diff
@@ -201,10 +201,10 @@ impl<W: Write> IonWriter for BinaryWriter<W> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::ion_reader::IonReader;
     use crate::reader::ReaderBuilder;
-    use crate::stream_reader::IonReader;
```
zslayton, line 205: 🗺️ I renamed the file `stream_reader.rs` to `ion_reader.rs`, a long overdue change. The only thing in that file is the trait `IonReader`.

tests/good_test_vectors.rs
zslayton, line 1: 🗺️ This test suite pre-dates our `ion-tests` integration tests. It would read all of the binary Ion files (`*.10n`) in `ion-tests/good` and assert that there were no errors. The `ion-tests` suite does that and much more, so this is no longer necessary.

zslayton (commented):
🗺️ PR tour

PR #579, opened 2023-06-28
Adds write_value to ElementWriter

popematt:
**Issue #, if available:**

None

**Description of changes:**

Adds a `write_value(&self, value: &Value)` function to `ElementWriter` so that if you have a `&Value` you don't need to take ownership and convert to `Element` in order to easily write the value.

E.g.:
```rust
fn foo<W: IonWriter>(value: &Value, writer: &mut W) {
    // Instead of this:
    writer.write_element(&value.to_owned().into());

    // Now we can write this:
    writer.write_value(value);
}
```

While I was at it, I also updated the documentation for `write_element` and `write_elements`, which can be used at any level, not just for top-level values.

### DIFF
--- a/src/element/writer.rs
+++ b/src/element/writer.rs
@@ -12,10 +12,17 @@ pub use TextKind::*;
 
 /// Serializes [`Element`] instances into some kind of output sink.
 pub trait ElementWriter {
-    /// Serializes a single [`Element`] as a top-level value.
+    /// Serializes a single [`Element`] at the current depth of the writer.
     fn write_element(&mut self, element: &Element) -> IonResult<()>;
 
-    /// Serializes a collection of [`Element`] as a series of top-level values.
+    /// Serializes a single [`Value`] at the current depth of the writer.
+    // TODO: consider extracting this to a ValueWriter trait.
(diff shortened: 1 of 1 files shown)


src/element/writer.rs
```diff
@@ -12,10 +12,17 @@ pub use TextKind::*;
...
 /// Serializes [`Element`] instances into some kind of output sink.
 pub trait ElementWriter {
-    /// Serializes a single [`Element`] as a top-level value.
+    /// Serializes a single [`Element`] at the current depth of the writer.
     fn write_element(&mut self, element: &Element) -> IonResult<()>;
 
-    /// Serializes a collection of [`Element`] as a series of top-level values.
+    /// Serializes a single [`Value`] at the current depth of the writer.
+    // TODO: consider extracting this to a ValueWriter trait.
+    fn write_value(&mut self, value: &Value) -> IonResult<()>;
```
zslayton, line 20: This makes me wonder if we need/want a `write_values(iter)` sibling method, but I'm happy to punt on that.

zslayton (approved):
Nice, thanks!

PR #580, opened 2023-06-28
Adds an `"experimental-writer"` feature

zslayton:
This PR continues the work started in #577 to address #576.

Commit 43fecfb62990e551c94e77f948a3e87cc5fca282 moves the public re-exports of:
* `IonWriter`
* `TextWriter`/`TextWriterBuilder`
* `BinaryWriter`/`BinaryWriterBuilder`

behind the feature flag `"experimental-writer"`. No behavior is changed,
and internal modules can continue to access these types as desired.

Commit 3e4a4b5919d401b8e8eab63f6d75ed60ce96b9ca downgrades the visibility of some
nested modules from `pub` to `pub(crate)` so they cannot be referenced directly by
users. Users that wish to refer to them must enable the new feature and then use the top-level
re-export.

Other cleanup changes:
* Commit 52ecbbf29d2b4897de931e952480441dc4fcd94a renames the `writer` (which contains
the `IonWriter` trait) to `ion_writer` to better align with `ion_reader`,
which contains the `IonReader` trait.
* Commit 0e687e9195bd495613c128f0c144a344f320d049 moves the `Format` and `TextKind`
enums (which are globally applicable) out of the `element` module.
*  Commit 7afe6ec0956a13cfbf0f5fe762253a0474faead2 makes the `Format` and `TextKind`
enums `#[non_exhaustive]`, so we can add other variants in the future without a breaking change.
* Commit ca9356bf5f7285c9f1a1398af7ed64e0215635c5 goes through all of the example code
in the doc tests and removes unused imports.


src/ion_writer.rs
```diff
@@ -97,29 +97,38 @@ pub trait IonWriter {
...
-    /// # roundtrip().unwrap();
-    /// ```
+    // This portion of the doc comment is always visible
+    #[doc = r##"
+Returns a reference to the writer's output.
+
+This method can be used to inspect the Ion data that the writer has produced without
+having to first drop the writer.
+    "##]
+    // This part is only visible if the "experimental-writer" feature is enabled.
```
zslayton, line 107: 🗺️ This is a pair of doc comments using [an unfamiliar syntax](https://users.rust-lang.org/t/doctests-that-require-a-non-default-feature-is-it-possible/29529/7); it allows parts of the doc comment to be hidden/shown depending on the features that are enabled.

zslayton (commented):
🗺️ PR tour

The changes in this PR are largely mechanical refactorings that are broken down by commit as detailed in the PR description, so there isn't much to walk through here.

desaikd (approved):

src/ion_writer.rs
popematt: It's cool, but why is it necessary here?

src/lib.rs
```diff
@@ -248,3 +251,21 @@ pub use result::{IonError, IonResult};
...
+}
+
+/// Supported Ion encodings.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[non_exhaustive]
+pub enum Format {
+    Text(TextKind),
+    Binary,
+    // TODO: Json(TextKind)
+}
```
popematt, line 271: Are these useful for anything if you haven't opted into the `experimental-writer` feature? If not, we should move them to a `pub(crate)` module and conditionally re-export them only if the feature is enabled.

zslayton: They're used in the `Element::write_as` API:
```rust 
let mut buffer = Vec::new();
Element::write_as(Format::Text(TextKind::Pretty), &mut buffer)?;
// ...do stuff with `buffer`
```

src/ion_writer.rs
zslayton: A confluence of a few factors:
* The `IonWriter` code gets compiled regardless of whether the `"experimental-writer"` feature is enabled. It's used to implement always-available public APIs like `Element::read_one()`.
* Doc examples can only use public APIs, so if we have a doc example that refers to an API that isn't available, the doc test will fail.
* We always build/run the doc tests on our private APIs as part of our CI. This guarantees that references to types and traits mentioned in the comments remains correct/intact.

To rephrase: if you build without `"experimental-writer"`, `IonWriter` and its doc tests still get built/run; a doc test referring to an unavailable API like `TextWriterBuilder` will cause a CI failure.

popematt (approved):

PR #584, opened 2023-06-30
Makes `IonError` variants wrap opaque types

zslayton:
This PR refactors the `IonError` enum variants to each wrap an opaque struct type. The wrapped types each offer methods and trait implementations that we can expand upon in the future without breaking changes.


examples/read_all_values.rs
```diff
@@ -10,7 +10,7 @@ fn main() {
 
 #[cfg(feature = "experimental-reader")]
 mod example {
-    use ion_rs::result::IonResult;
+    use ion_rs::IonResult;
```
zslayton, line 13: 🗺️ I have a `TODO` to limit the visibility of types to a single prescribed path within the crate. For the moment, `IonResult` can be found in both the crate root (`crate::IonResult`) and in `result` (`crate::result::IonResult`). This is true of other types as well.

src/binary/binary_writer.rs
```diff
@@ -140,7 +140,7 @@ impl<W: Write> IonWriter for BinaryWriter<W> {
                 if self.symbol_table.sid_is_valid(symbol_id) {
                     symbol_id
                 } else {
-                    return illegal_operation(format!(
+                    return IonResult::illegal_operation(format!(
```
zslayton, line 143: 🗺️ `IonFailure` (imported above on line 5) is a new crate-visible trait that provides extension methods for the `IonResult` and `IonError` types.

Previously, each `IonError` variant had one helper function in `crate::result` that would build an `IonResult`, and an accompanying function that would being an `IonError`. The names got pretty weird. For example, **`decoding_error`** would return an `IonResult` (not an error, as one could reasonably infer), and **`decoding_error_raw`** would return an `IonError`. (The "raw" indicated that it hadn't gone through the final step of wrapping the `IonError` in a `Result::Err`, which was not an obvious connection.)

The `IonFailure` trait replaces those method pairs. Now you can use `IonResult::decoding_error` to make an `IonResult` and you can use `IonError::decoding_error` to make an `IonError`. Much more straightforward.

src/binary/decimal.rs
```diff
@@ -111,12 +111,9 @@ where
...
         // the preceding type descriptor.
         let mut encoded: ArrayVec<u8, DECIMAL_BUFFER_SIZE> = ArrayVec::new();
-        encoded
-            .encode_decimal(decimal)
-            .map_err(|_e| IonError::EncodingError {
-                description: "found a decimal that was too large for the configured buffer"
-                    .to_string(),
-            })?;
+        encoded.encode_decimal(decimal).map_err(|_e| {
+            IonError::encoding_error("found a decimal that was too large for the configured buffer")
```
zslayton, line 115: 🗺️ Previously, the convenience methods for constructing errors would take either:
1. **`AsRef<str>`**, which accepts both `String` and `&str`, but required the method to make a copy of whatever was passed in so it could be stored indefinitely. If the user passed a `String`, we'd be making an unnecessary copy.
2. **`String`**, giving the user full control over how/when the `String` was constructed, and requiring at most one allocation. However, if the user didn't already have a `String`, they'd have to construct one before passing it in.

These methods now accept an `Into<String>`, allowing them to accept either a `String` or a `&str` and only making a new `String` when it's necessary.

src/element/mod.rs
```diff
@@ -663,66 +670,82 @@ impl Element {
...
         self.write_as(Format::Binary, &mut buffer)?;
         Ok(buffer)
     }
 
-    /// Serializes this [`Element`] as text Ion using the requested [`TextKind`] and returning the
-    /// output as a `String`.
-    ///
-    /// This is a convenience method; to provide your own buffer instead of allocating a `String`
-    /// on each call, see [`Element::write_as`]. To provide your own writer instead of constructing
-    /// a new `String` and writer on each call, see [`Element::write_to`].
```
zslayton, line 704: 🗺️ This doc comment referred to `write_to`, which is feature gated. If you attempted to build the docs without `"experimental-writer"` enabled, it would fail because `write_to` was not available.

This change hides that section when `write_to` is not available. The content of the comment is otherwise unchanged.

src/lazy/system_stream_item.rs
```diff
@@ -41,11 +41,11 @@ impl<'top, 'data> SystemStreamItem<'top, 'data> {
         }
     }
 
-    /// Like [`Self::version_marker`], but returns a [`crate::IonError::DecodingError`] if this item
+    /// Like [`Self::version_marker`], but returns a [`crate::IonError::Decoding`] if this item
```
zslayton, line 44: 🗺️ As you'll see later, I have removed a layer of redundancy from the naming in the `IonError` type.

Now that each error variant wraps an opaque struct named `<KIND>Error`, the variant names themselves no longer include the word "Error."

src/result.rs
zslayton, line 1: 🗺️ Because this PR introduced several new opaque types that belonged in the `result` module, I broke this file into a directory with a `mod.rs`.

src/result/decoding_error.rs
zslayton, line 1: 🗺️ We still need to add doc comments for these new types. I have opened #589 to track that.

src/result/io_error.rs
```diff
@@ -0,0 +1,27 @@
...
+    fn clone(&self) -> Self {
+        io::Error::from(self.source.kind()).into()
+    }
+}
+
+impl PartialEq for IoError {
+    fn eq(&self, other: &Self) -> bool {
+        self.source().kind() == other.source().kind()
+    }
+}
```
zslayton, line 27: 🗺️ `std::io::IoError` doesn't provide many trait implementations, so our error type has do define them manually.

src/result/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+pub mod illegal_operation;
+pub mod incomplete;
+pub mod io_error;
+pub mod position;
+
+/// A unified Result type representing the outcome of method calls that may fail.
+pub type IonResult<T> = Result<T, IonError>;
+
+/// Represents the different types of high-level failures that might occur when reading Ion data.
+#[derive(Clone, Debug, Error, PartialEq)]
```
zslayton, line 24: 🗺️ Now that we're using wrapped types that implement all the traits we need, the top-level `IonError` type can derive its implementations of `PartialEq` and `Clone`.

src/result/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+pub enum IonError {
+    /// Indicates that an IO error was encountered while reading or writing.
+    #[error("{0}")]
+    Io(#[from] IoError),
+
+    /// Indicates that the input buffer did not contain enough data to perform the requested read
+    /// operation. If the input source contains more data, the reader can append it to the buffer
+    /// and try again.
+    #[error("{0}")]
+    Incomplete(#[from] IncompleteError),
```
zslayton, line 34: 🗺️ As I mentioned earlier, I've removed the word `Error` from each of the `IonError` variants. The new nested types made it feel pretty redundant.

src/result/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+// Crate-visible convenience methods for constructing error variants and wrapping them in the
+// appropriate type: IonResult<T> or IonError. This is a trait so these methods can be added to
+// `IonResult<T>`, which is just a type alias for `Result<T, IonError>`, whose implementation
+// does not live in this crate.
+pub(crate) trait IonFailure {
+    fn incomplete(label: &'static str, position: impl Into<Position>) -> Self;
+    fn decoding_error<S: Into<String>>(description: S) -> Self;
+    fn encoding_error<S: Into<String>>(description: S) -> Self;
+    fn illegal_operation<S: Into<String>>(operation: S) -> Self;
+}
```
zslayton, line 80: 🗺️ As described earlier, this is the new crate-visible trait that makes constructing errors less weird and more readable.

It does not include an `io` method because the only way we ever construct `IonError::Io` is by converting a `std::io::IoError` with the `?` operator.

src/result/position.rs
zslayton, line 1: 🗺️ `Position` used to live in `result.rs`, which was turned into a directory. It has been broken out into its own file, but its contents are unchanged with one exception called out below.

src/result/position.rs
```diff
@@ -0,0 +1,73 @@
...
+        }
+    }
+
+    /// Returns the offset from the start of the Ion stream in bytes.
+    pub fn byte_offset(&self) -> usize {
+        self.byte_offset
+    }
+
+    /// If available returns the text position as line and column offsets.
+    pub fn line_and_column(&self) -> Option<(usize, usize)> {
```
zslayton, line 37: 🗺️ This accessor was previously called `text_position()`. I renamed it to `line_and_column()` to be consistent with the variable name and to make it more obvious what the pair of values being returned each represented.

src/text/non_blocking/raw_text_reader.rs
```diff
@@ -988,16 +987,9 @@ mod reader_tests {
...
-                        line_column: Some((line, column)),
-                        ..
-                    },
-                ..
-            }) => {
-                assert_eq!(line, 2);
-                assert_eq!(column, 0);
+            Err(IonError::Incomplete(e)) => {
+                assert_eq!(e.position().line(), Some(2));
+                assert_eq!(e.position().column(), Some(0));
```
zslayton, line 992: 🗺️ This is a good example of the updated application code for inspecting an `IonError` that has cropped up.

src/types/decimal.rs
```diff
@@ -371,16 +376,17 @@ impl TryFrom<f64> for Decimal {
 
         fn try_convert(coefficient: f64, exponent: i64) -> Result<Decimal, IonError> {
             // prefer a compact representation for the coefficient; fallback to bigint
-            let coefficient: Coefficient = if !coefficient.trunc().is_zero()
```
zslayton, line 374: 🗺️ This section is a formatting-only change.

src/types/timestamp.rs
```diff
@@ -869,29 +869,29 @@ impl TimestampBuilder {
...
-            illegal_operation_raw(format!("specified month ('{month}') is invalid"))
+            IonError::illegal_operation(format!("specified month ('{month}') is invalid"))
         })?;
         if self.precision == Precision::Month {
             return Ok(datetime);
         }
 
         // If precision >= Day, the day must be set.
         let day = self.day.expect("missing day");
-        datetime = datetime
```
zslayton, line 880: 🗺️ This section is a formatting-only change.

tests/ion_hash_tests.rs
```diff
@@ -119,7 +120,7 @@ fn ion_hash_tests() -> IonHashTestResult<()> {
 }
 
 fn test_file(file_name: &str) -> IonHashTestResult<()> {
-    let data = read(file_name).map_err(|source| ion_rs::IonError::IoError { source })?;
+    let data = read(file_name).map_err(IonError::from)?;
```
zslayton, line 123: 🗺️ Another example of trait impls making the code a bit nicer.

tests/ion_hash_tests.rs
```diff
@@ -224,7 +225,7 @@ fn expected_hash(struct_: &Struct) -> IonResult<Vec<u8>> {
     let identity = if let Some(identity) = struct_.get("identity") {
         identity.as_sequence().expect("`identity` should be a sexp")
     } else {
-        illegal_operation("only identity tests are implemented")?
+        todo!("only identity tests are implemented")
```
zslayton, line 228: 🗺️ ion-hash-tests was constructing its own `IonResult`s to describe failure modes that didn't come from Ion itself. The error constructor methods are now limited to crate-visibility. Clients like ion-hash-tests should use their own error type.

In ion-hash-tests' case, it was bubbling up the error result and panicking later on. I've changed those failures to an immediate panic. The end result is the same, and it removes the dependency on now-private functionality.

zslayton (commented):
🗺️ PR tour

src/result/decoding_error.rs
zslayton, line 1: 🗺️ We still need to add doc comments for these new types. I have opened #589 to track that.

src/element/element_stream_writer.rs
```diff
@@ -111,7 +111,9 @@ where
                 if let Some(field_name) = field_name {
                     fields.push((field_name, element));
                 } else {
-                    return illegal_operation("Values inside a struct must have a field name.");
+                    return IonResult::illegal_operation(
+                        "Values inside a struct must have a field name.",
+                    );
```
jobarr-amzn, line 116: Generally speaking an error message without context is less helpful than one that does have context. It might be helpful to provide some indication here of which value does not have a field name. I'm not suggesting that we fix it in this PR, but what sort of additional book-keeping or capabilities might we need to provide context in cases like this?

I created https://github.com/amazon-ion/ion-rust/issues/590 to track/discuss. 

Perhaps it would be good to link to #590 in places like this, so that it's easier to come back to later?

On the other hand, if we *had* a more universal context description ability then we'd likely do some general pass when the capability was added.

src/result/mod.rs
jobarr-amzn: >It does not include an `io` method because the only way we ever construct `IonError::Io` is by converting a `std::io::IoError` with the `?` operator.

Could this stand to be a comment?

jobarr-amzn (approved):

src/element/element_stream_writer.rs
zslayton: Agreed on all counts. (This particular case is a bit of an outlier--it shouldn't happen since it's serializing a fully-formed `Element`. Not setting a field name would be user error when driving an `IonWriter`.)

desaikd (approved):

PR #588, opened 2023-06-30
Clean up TimestampBuilder

popematt:
**Issue #, if available:**

None

*Description of changes:*

This changes the timestamp builder so that it uses the type-state pattern. The main benefit is that now there is only one timestamp builder in the documentation, and it's easier to see what all the functions are, etc.

I also discovered (and fixed) and integer overflow bug in the builder.

All of the meaningful changes are in `src/types/timestamp.rs`. Everything else is a mechanical refactoring change.

See 🗺️ PR tour comments for particular details.

As I was working on this, I kept seeing more things to fix, and the scope kept growing. I eventually decided that I needed to stop adding changes. So, still todo (but not in this PR):
* Organization—i.e. do we want to move the builder into `timestamp::builder`?
* Documentation of all functions
* Decide on `Timestamp` and `TimestampBuilder` integer sizes
* Clean up implementation of `build()`

For this PR, I really just want to align on the overall API. We can follow up on the other items


src/types/timestamp.rs
```diff
@@ -825,27 +769,49 @@ impl IonOrd for Timestamp {
...
-    day: Option<u8>,
-    hour: Option<u8>,
-    minute: Option<u8>,
-    second: Option<u8>,
+    year: u32,
+    month: Option<u32>,
+    day: Option<u32>,
+    hour: Option<u32>,
+    minute: Option<u32>,
+    second: Option<u32>,
```
popematt, line 787: 🗺️ As I was working on this change, I realized that there was an overflow bug. E.g.:

```rust
let ts1 = TimestampBuilder::new().with_ymd(1, 256, 256).build();
let ts2 = TimestampBuilder::new().with_ymd(1, 1, 1).build();
assert!(ts1.is_err()); // FAIL!
assert_ne!(ts1, ts2); // Also FAIL!
```

So, I changed all of the builder fields to `u32`. It made the most sense to me, especially since we were converting most of them back to `u32` when building the `Timestamp`. However, we may want to consider changing back to smaller integer types (without re-introducing the overflow bug, of course).

src/types/timestamp.rs
```diff
@@ -1007,230 +973,212 @@ impl TimestampBuilder {
         }
         Ok(timestamp)
     }
+
+    /// Like [Self::build], but the fields provided for each time unit are understood
+    /// to be in UTC rather than in the local time of the specified offset (if there is one).
+    pub(crate) fn build_utc_fields_at_offset(
```
popematt, line 979: 🗺️  This seems like a specialized function for the binary reader, so I made it `pub(crate)` so that it's not in our public API. We can re-add it to the public API if there's a large outcry for it.

src/types/timestamp.rs
```diff
@@ -1007,230 +973,212 @@ impl TimestampBuilder {
...
-/// Allows the user to set the `month` field on a builder that has already had its `year`
-/// field set. Or, if `Year` is the desired precision, they may build the [Timestamp] with an
-/// unknown offset instead.
-#[derive(Debug, Clone)]
-pub struct MonthSetter {
-    builder: TimestampBuilder,
+#[derive(Debug, Clone, Default)]
+pub struct Init;
+
+impl Default for TimestampBuilder<Init> {
```
popematt, line 992: 🗺️ Clippy was unhappy about having `new()` without implementing `Default`, so we have an implementation of `Default` _only_ for `TimestampBuilder<Init>`.

src/types/timestamp.rs
```diff
@@ -1007,230 +973,212 @@ impl TimestampBuilder {
...
 }
 
-/// Allows the user to set the `second` field on a builder that has already
-/// had its `year`, `month`, `day`, `hour`, and `minute` fields set. Or, if `HourAndMinute` is the
-/// desired precision, they may build the [Timestamp] instead, optionally specifying an offset if
-/// known.
-#[derive(Debug, Clone)]
-pub struct SecondSetter {
-    builder: TimestampBuilder,
+macro_rules! with_offset {
```
popematt, line 1089: 🗺️ We _could_ add this function to particular states by using a trait impl, but I think it makes the rust doc page more readable to just have a macro to "paste" it into the appropriate impl blocks.

src/types/timestamp.rs
zslayton: The initial motivation for using `Option<u8>` instead of `Option<u32>` was to keep the memory layout from bloating. An `Option<u32>` often takes 8 bytes to represent despite only having 5 bytes of data due to alignment.

```rust
struct Foo {
    month: Option<u8>,
    day: Option<u8>,
    hour: Option<u8>,
    minute: Option<u8>,
    second: Option<u8>,
    nanoseconds: Option<u8>,
}

struct BigFoo {
    month: Option<u32>,
    day: Option<u32>,
    hour: Option<u32>,
    minute: Option<u32>,
    second: Option<u32>,
    nanoseconds: Option<u32>,
}

pub fn main() {
    println!("sizeof Foo: {}", std::mem::size_of::<Foo>());
    println!("sizeof BigFoo: {}", std::mem::size_of::<BigFoo>());
}

// ==== OUTPUT ====

sizeof Foo: 12
sizeof Foo: 48
```
([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d64c6f640066aed3cef75ba05c71c1f1))

That said, it looks like even with the other fields added we're still looking at <100 bytes in all and I don't foresee people storing thousands of `TimestampBuilder`s, so it's probably fine. 👍

src/types/timestamp.rs
```diff
@@ -845,7 +837,52 @@ struct TimestampBuilder {
...
+state_with_markers!(HasDay, Buildable);
+state_with_markers!(HasHour);
+state_with_markers!(HasMinute, Buildable, BuildableWithOffset);
+state_with_markers!(HasSeconds, Buildable, BuildableWithOffset);
+state_with_markers!(HasFractionalSeconds, Buildable, BuildableWithOffset);
+
+impl<T> TimestampBuilder<T> {
+    fn change_state<U>(self) -> TimestampBuilder<U> {
+        // TODO: This entire function could be replaced with one line of unsafe code, but is it worthwhile?
+        // unsafe { std::mem::transmute(self) }
```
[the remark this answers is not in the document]
zslayton: > However, I don't know if it actually matters when compiling in release mode because when I tried that on godbolt, it optimized away all of my sample code. 🤷‍♂️

Relatable! 😅

I think the `--release` mode is likely to be able to see through this and turn it into something very similar to `transmute`. However, even if it's not, this is something that we should wait to see show up in a perf profile before we reach for `unsafe`.

src/types/timestamp.rs
```diff
@@ -825,27 +769,49 @@ impl IonOrd for Timestamp {
...
+    hour: Option<u32>,
+    minute: Option<u32>,
+    second: Option<u32>,
     fractional_seconds: Option<Mantissa>,
     nanoseconds: Option<u32>,
 }
 
-impl TimestampBuilder {
+// The state structs are pub in this module, but they do not appear as types
+// in the documentation, they cannot be imported, and they are not nameable from outside this crate.
```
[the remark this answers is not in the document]
zslayton: I'm not sure I understand what "state structs" means here. What exactly are we considering making `pub`? In general I'd err on the side of keeping things private until they're asked for.

src/types/timestamp.rs
zslayton: It's true that it's only used by the binary reader within our crate. I do think it's generally useful; it's comparable to [this method](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.from_utc) in the `chrono` crate, which lets you provide a set of UTC fields and an offset. That said, I'm ok with keeping it crate-visible for the time being.

src/types/timestamp.rs
```diff
@@ -1007,230 +973,212 @@ impl TimestampBuilder {
...
 }
 
-impl_time_unit_setter_for!(MonthSetter);
-impl_time_unit_setter_for!(DaySetter);
-impl_time_unit_setter_for!(HourAndMinuteSetter);
-impl_time_unit_setter_for!(SecondSetter);
-impl_time_unit_setter_for!(FractionalSecondSetter);
+#[derive(Debug, Clone)]
+pub struct HasOffset;
+// No impl for TimestampBuilder<HasOffset> because `build()` is included in TimestampBuilder<T>
```
zslayton, line 1181: What does it mean to call `build()` on a `TimestampBuilder<Init>`?

zslayton (commented):
Overall I really like this change! A few discussion points below.

src/types/timestamp.rs
popematt: You would get `0000T`. 
I could remove that if you think it's too weird.

src/types/timestamp.rs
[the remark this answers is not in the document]
popematt: The state structs are things like `HasYear`, `HasMonth`, etc. I can move/elaborate the comment to be more clear.

src/binary/non_blocking/raw_binary_reader.rs
```diff
@@ -1648,50 +1648,58 @@ mod tests {
...
         assert_eq!(
             reader.read_timestamp()?,
-            Timestamp::with_year(2022).with_month(6).build()?
+            Timestamp::builder().with_year(2022).with_month(6).build()?
         );
         expect_value(reader.next(), IonType::Timestamp);
         assert_eq!(
             reader.read_timestamp()?,
-            Timestamp::with_ymd(2022, 6, 18).build()?
+            Timestamp::builder().with_ymd(2022, 6, 18).build()?
```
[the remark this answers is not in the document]
popematt: After you finished your read through, did your opinion change? I don't mind adding `with_year()` and `with_ymd()` as entry points, but if we do, then I'd like to get rid of `builder()`.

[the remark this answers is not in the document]
zslayton: > I don't mind adding with_year() and with_ymd() as entry points, but if we do, then I'd like to get rid of builder().

Having read the whole thing, I think this is probably the way to go. Removing `builder()` eliminates the need for `Timestamp<Init>` and the oddment of creating `0000T` with `builder().build()`. We lose the ability to create a `TimestampBuilder` and start initializing it later in some other context, but that feels like a pretty niche use case. (At any rate, we can always add `builder()` later if we decide we want it.)

src/types/timestamp.rs
popematt: We could also solve the overflow issue by shrinking the argument sizes. (E.g. `with_month(month: u32)` becomes `with_month(month: u8)`.

Since timestamps always represent a specific point in time, we could forgo the `Option` and set each field to the point in time it represents if the precision is less than that unit of time. For example, if the precision is "day", then it's not a big deal if the hour, minute, and second are all set to `0` instead of `None`.

Here's another [playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3b8c83aaa84919e107d4f0a5f50805df). (I've updated `Foo` to use `u32` for `nanoseconds` because we do actually need it there.) The output is:

```
sizeof Foo: 20
sizeof BigFoo: 48
sizeof MediumFoo: 24
sizeof LittleFoo: 12
```

What do you think? Is there any reason not to shrink the function argument sizes like `with_ymd(year: u16, month: u8, day: u8)`?

zslayton: IIRC, I used `u32` everywhere because other time-centric libraries like `chrono` did that. My sense is that it's kind of rare for APIs to use the most narrow int for a type, even though that's counter-intuitive. Maybe because it forces users to do verbose casting/conversion?

zslayton (approved):
This looks great, nice work!

PR #591, opened 2023-07-03
Removes `IonDataSource` and vestigial `read()` impls

zslayton:
The original implementation of the binary reader used a trait called `IonDataSource` that provided extension methods for implementations of `BufRead`. The binary reader was later rewritten to use a type called `BinaryBuffer`, but `IonDataSource` and its methods were never removed.

This PR removes `IonDataSource` and the vestigial implementations of `read()` for various encoding primitives that relied upon `IonDataSource`.

Finally, it renames the single-method trait `ToIonDataSource` (which suggested types could be converted to an `IonDataSource`) to `IonDataSource`, reclaiming the name for a better use.

### DIFF
--- a/src/binary/non_blocking/raw_binary_reader.rs
+++ b/src/binary/non_blocking/raw_binary_reader.rs
@@ -10,7 +10,7 @@ use crate::raw_reader::{BufferedRawReader, Expandable, RawStreamItem};
 use crate::result::IonFailure;
 use crate::types::{Blob, Clob, Decimal, IntAccess, Str, SymbolId};
 use crate::{Int, IonError, IonResult, IonType, RawSymbolToken, Timestamp};
-use bytes::{BigEndian, Buf, ByteOrder};
+use bytes::{BigEndian, ByteOrder};
 use num_bigint::BigUint;
 use num_traits::Zero;
 use std::io::Read;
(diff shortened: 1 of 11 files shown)


src/binary/non_blocking/raw_binary_reader.rs
```diff
@@ -1153,11 +1153,10 @@ impl<'a> Iterator for AnnotationsIterator<'a> {
...
         }
-        // This iterator cannot be created unless the reader is currently parked on a value.
-        // If the reader is parked on a value, the complete annotations sequence is in the buffer.
-        // Therefore, reading symbol IDs from this byte slice cannot fail. This allows us to safely
-        // unwrap the result of this `read` call.
-        let var_uint = VarUInt::read(&mut self.data).unwrap();
+        let var_uint = match self.data.read_var_uint() {
+            Ok(var_uint) => var_uint,
+            Err(e) => return Some(Err(e)),
+        };
```
jobarr-amzn, line 1159: What drives the change here? Is the context about positioning or buffered bytes different now?

desaikd (approved):

jobarr-amzn (approved):

src/binary/non_blocking/raw_binary_reader.rs
zslayton: The method `VarUInt::read` was deleted. The rest of this type uses `BinaryBuffer`'s read operations; this change was changing `self.data` from being a `io::Cursor(&[u8])` that we read from using `VarUInt::read()` to being a `BinaryBuffer<&'a [u8]>` that has its own read operation.

The invariant described in the comment is still true; I think I could have kept it but didn't want to do the deep dive needed to guarantee that.

PR #593, opened 2023-07-05
Adds Decimal methods to access the coefficient and exponent

### LINKED ISSUE #405: Cannot read value of a Decimal

zslayton:
This PR:
* Makes `Coefficient` a publicly visible type.
* Adds a `Decimal::coefficient` method to access the coefficient.
* Adds a `Decimal::exponent` method to access the exponent.
* Makes the `decimal` module a directory instead of a flat file
* Moves the `Magnitude` newtype to its own module
* Conversions to/from `Int`, `UInt`, and `Decimal` now return `IonError` instead of `()` in the event of a failure.

Fixes #405.

### DIFF
--- a/src/types/decimal/coefficient.rs
+++ b/src/types/decimal/coefficient.rs
@@ -2,6 +2,7 @@ use num_bigint::{BigInt, BigUint};
 use num_traits::Zero;
 
 use crate::result::{IonError, IonFailure};
+use crate::types::decimal::magnitude::Magnitude;
 use crate::types::integer::UIntData;
 use crate::types::{Int, UInt};
 use crate::IonResult;
@@ -16,67 +17,6 @@ pub enum Sign {
     Negative,
     Positive,
 }
-
-/// A simple wrapper type around a [`UInt`]. Users are not expected to construct
-/// instances of `Magnitude` directly; it is provided for conversion ergonomics.
-///
-/// Signed integer types cannot implement `Into<UInt>`. Instead, they implement
-/// `TryInto<UInt>` and report an error if the input integer is negative.
-///
-/// Signed integers can infallibly implement `Into<Magnitude>` by using their
-/// absolute value.
-pub struct Magnitude(UInt);
-
-impl Magnitude {
-    fn new<I: Into<UInt>>(value: I) -> Self {
-        Magnitude(value.into())
-    }
-}
-
-impl From<UInt> for Magnitude {
-    fn from(value: UInt) -> Self {
-        Magnitude(value)
-    }
-}
-
-impl From<Magnitude> for UInt {
-    fn from(value: Magnitude) -> Self {
-        value.0
-    }
-}
-
-impl From<BigUint> for Magnitude {
-    fn from(value: BigUint) -> Self {
-        let uint: UInt = value.into();
-        Magnitude(uint)
-    }
-}
-
-macro_rules! impl_magnitude_from_small_unsigned_int_types {
-    ($($t:ty),*) => ($(
-        impl From<$t> for Magnitude {
-            fn from(value: $t) -> Magnitude {
-                let uint: UInt = value.into();
-                Magnitude(uint)
-            }
-        }
-    )*)
-}
-
-impl_magnitude_from_small_unsigned_int_types!(u8, u16, u32, u64, u128, usize);
-
-macro_rules! impl_magnitude_from_small_signed_int_types {
-    ($($t:ty),*) => ($(
-        impl From<$t> for Magnitude {
-            fn from(value: $t) -> Magnitude {
-                let uint: UInt = (value.unsigned_abs() as u64).into();
-                Magnitude(uint)
-            }
-        }
-    )*)
-}
-
-impl_magnitude_from_small_signed_int_types!(i8, i16, i32, i64, isize);
 
 /// A signed integer that can be used as the coefficient of a Decimal value. This type does not
 /// consider `0` and `-0` to be equal and supports magnitudes of arbitrary size.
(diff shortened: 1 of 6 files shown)


src/types/decimal/coefficient.rs
```diff
@@ -17,67 +18,6 @@ pub enum Sign {
...
 
-/// A simple wrapper type around a [`UInt`]. Users are not expected to construct
-/// instances of `Magnitude` directly; it is provided for conversion ergonomics.
-///
-/// Signed integer types cannot implement `Into<UInt>`. Instead, they implement
-/// `TryInto<UInt>` and report an error if the input integer is negative.
-///
-/// Signed integers can infallibly implement `Into<Magnitude>` by using their
-/// absolute value.
-pub struct Magnitude(UInt);
```
zslayton, line 28: 🗺️ This type was moved into its own module. You'll see its definition appear further on.

src/types/decimal/coefficient.rs
```diff
@@ -116,26 +60,26 @@ impl Coefficient {
...
+        self.is_zero_with_sign(Sign::Negative)
     }
 
     /// Returns true if the Coefficient represents positive zero.
-    pub(crate) fn is_positive_zero(&self) -> bool {
+    pub fn is_positive_zero(&self) -> bool {
+        self.is_zero_with_sign(Sign::Positive)
+    }
+
+    pub(crate) fn is_zero_with_sign(&self, test_sign: Sign) -> bool {
```
zslayton, line 73: 🗺️ This helper method is used in the implementation of both `is_negative_zero()` and `is_positive_zero()`.

Users may wish to use `is_negative_zero()` before converting the `Coefficient` to an `Int`, as `Int` cannot represent negative zero natively. `is_positive_zero()` is provided mostly for symmetry.

src/types/decimal/magnitude.rs
```diff
@@ -0,0 +1,70 @@
...
+
+/// A simple wrapper type around a [`UInt`]. Users are not expected to construct
+/// instances of `Magnitude` directly; it is provided for conversion ergonomics.
+///
+/// Signed integer types cannot implement `Into<UInt>`. Instead, they implement
+/// `TryInto<UInt>` and report an error if the input integer is negative.
+///
+/// Signed integers can infallibly implement `Into<Magnitude>` by using their
+/// absolute value.
+pub struct Magnitude(UInt);
```
zslayton, line 12: 🗺️ As mentioned above, this type was moved to its own module. It contains no new code apart from the `From<BigUint> for Magnitude` impl on line 32.

src/types/integer.rs
```diff
@@ -200,13 +200,10 @@ impl_uint_from_small_unsigned_int_types!(u8, u16, u32, u64, usize);
...
         impl TryFrom<$t> for UInt {
-            // If conversion fails, we discard the error. `TryFromIntError` provides no
-            // additional context and cannot be constructed outside of `std`. We'll return
-            // the unit type, `()`.
-            type Error = ();
+            type Error = IonError;
             fn try_from(value: $t) -> Result<Self, Self::Error> {
                 if value < 0 {
-                    return Err(());
+                    return IonResult::illegal_operation("cannot convert a negative number to a UInt");
```
zslayton, line 206: 🗺️ Previously, many conversions that would fail used `()` as their error type because the single mode of failure was obvious. (For example: trying to convert a negative `Int` to a `UInt`.) However, making these into `IonError` means they can be bubbled up with `?` alongside all other `IonResult`-producing code.

`illegal_operation` isn't ideal; I'm open to creating an `InvalidConversion` error variant, but I don't expect users to will inspect conversion errors very deeply.

zslayton (commented):
🗺️ PR tour

desaikd (approved):

src/types/decimal/coefficient.rs
```diff
@@ -285,4 +273,81 @@ mod coefficient_tests {
...
+            Int::try_from(Coefficient::new(Sign::Positive, 5)),
+            Ok(Int::from(5))
+        );
+        assert_eq!(
+            Int::try_from(Coefficient::new(Sign::Negative, 5)),
+            Ok(Int::from(-5))
+        );
+
+        // BigInt
+        let enormous_int = BigInt::from_str("1234567890123456789012345678901234567890").unwrap();
```
jobarr-amzn, line 336: Just for fun, this is a trick I sometimes use when making strings of numbers for length:
```suggestion
        let enormous_int = BigInt::from_str("1234567891023456789202345678930234567894").unwrap();
```
This allows you to more easily figure out how long the string is, find a `9x0` and the digit `x` is the 10s place not a zero. The following 1 is not valuable (inferable from `234..`), so I write `20`, `30`, what have you instead.

popematt (approved):

src/types/decimal/mod.rs
```diff
@@ -14,7 +14,31 @@ use std::convert::{TryFrom, TryInto};
 use std::fmt::{Display, Formatter};
 use std::ops::Neg;
 
+pub(crate) mod coefficient;
+mod magnitude;
+
 /// An arbitrary-precision Decimal type with a distinct representation of negative zero (`-0`).
```
jobarr-amzn, line 20: Is it worth including some comment about what negative zero is for, what it means?

jobarr-amzn (approved):

PR #594, opened 2023-07-05
Removes vestigial dependency on BigDecimal

zslayton:
Early versions of `ion-rust` used the third-party `BigDecimal` type to represent Ion `decimal` values. The library has had its own `Decimal` type for a while now that provides a correct representation of negative zero and properly implements Ion equality testing. 

This PR removes the library's remaining uses of `BigDecimal` and drops the dependency.


src/element/reader.rs
zslayton, line 1: 🗺️ Most of the changes in this file are to use implicit conversions to avoid going from Rust type -> `Value` -> `Element`. They now go directly from Rust type -> `Element`.

src/lib.rs
```diff
@@ -246,9 +246,7 @@ pub use result::{IonError, IonResult};
 /// Re-exports of third party dependencies that are part of our public API.
 ///
 /// See also: <https://github.com/amazon-ion/ion-rust/issues/302>
-pub mod external {
-    pub use bigdecimal;
-}
+pub mod external {}
```
zslayton, line 249: 🗺️ I think we still need this for dependencies behind opt-in feature gates.

zslayton (commented):
🗺️ PR tour

src/lib.rs
desaikd: (Question) Can we remove this module as it doesn't have any dependencies at the moment? Later when we need re-exports for the opt-in features, we can create the module for it.
Having this in crate docs with no dependencies in it might be confusing, so we should either add a comment or an issue link stating why we have kept it or indicating why this is empty at the moment.

desaikd (approved):

popematt (approved):
Given that crate A can't implement a trait from crate B for a type from crate C, we _might_ eventually realize that we want to provide `From<BigDecimal>` for `Decimal` or `Element` behind a feature flag.

...but I don't think that's any reason not to make the change.

src/lib.rs
zslayton: Great point, I'll take it out.

PR #595, opened 2023-07-06
Module cleanup, adds `Element::expect_*`, removes `IntAccess`

zslayton:
This PR:
* Hides several modules that were previously public.
* Makes the contents of the `types` and `element` modules visible at the root level.
* Normalizes all imports to use the root path where possible.
* Renames `Precision` to `TimestampPrecision` since it is now visible at the root level.
* Removes the `IntAccess` trait, which was introduced when we had the `borrowed` implementation of `Element`. The `as_i64()` method is now natively available on both `Element` and `Int`.
* Adds some additional `TryFrom`/`From` implementations for `Int`.


examples/read_all_values.rs
```diff
@@ -10,10 +10,10 @@ fn main() {
 
 #[cfg(feature = "experimental-reader")]
 mod example {
-    use ion_rs::IonResult;
-    use ion_rs::RawBinaryReader;
-    use ion_rs::RawStreamItem;
-    use ion_rs::{BlockingRawBinaryReader, IonReader, IonType, RawReader, StreamItem, UserReader};
+    use ion_rs::{
+        BlockingRawBinaryReader, IonReader, IonResult, IonType, RawBinaryReader, RawReader,
```
popematt, line 14: Is it intentional that we're exporting "raw" readers at the root?

src/lib.rs
```diff
@@ -165,57 +161,51 @@
...
 mod shared_symbol_table;
 mod symbol_ref;
 mod symbol_table;
 mod system_reader;
+mod text;
+
+// Publicly-visible modules with nested items which users may choose to import
+pub mod element;
+pub mod result;
+pub mod types;
```
popematt, line 186: So, just to make sure I'm understanding correctly, we're exporting everything here and at the root?

src/lib.rs
```diff
@@ -224,6 +214,8 @@ pub use {
     blocking_reader::{BlockingRawBinaryReader, BlockingRawReader, BlockingRawTextReader},
     ion_reader::IonReader,
     raw_reader::{BufferedRawReader, RawReader, RawStreamItem},
+    raw_symbol_token::RawSymbolToken,
```
popematt, line 217: I'm curious. Do you think it would be helpful, neutral, or unhelpful for everything in the `experimental-reader` feature to be exported under `ion_rs::reader` as a way of making the namespaces somewhat match the features?

src/types/timestamp.rs
```diff
@@ -15,7 +15,7 @@ use std::ops::Div;
 
 /// Indicates the most precise time unit that has been specified in the accompanying [Timestamp].
 #[derive(Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Ord, Default)]
-pub enum Precision {
+pub enum TimestampPrecision {
```
popematt, line 18: Thank you!

src/lib.rs
zslayton: > we're exporting everything here

There are some types that I don't think are commonly used enough (at least by name) to merit a spot in the crate root (`OwnedElementIterator`, `FractionalSecondSetter`, etc).

That said, there aren't _that_ many of them and they're pretty distinctively named, so I could also be convinced not to expose any modules at all.

examples/read_all_values.rs
zslayton: I haven't really thought through how I'd like to expose all of the types that are behind experimental feature flags yet. Open to suggestions.

src/lib.rs
zslayton: It depends -- if we're leaning towards a pretty flat namespace, I don't think it would help that much. If we think we'll ultimately have visible modules, it might be somewhat helpful.

popematt: I don't know if we're leaning towards a flat namespace. I do think that the most common things should probably be at the front and center.

That being said, I think figuring this out would be a blocking issue for stabilizing the reader feature, but not for the v1.0 release.

popematt (approved):

PR #598, opened 2023-07-12
Makes `element` and `types` private

zslayton:
This PR makes the `element` and `types` modules private and re-exports their commonly needed types at the root level.


popematt (approved):

jobarr-amzn (commented):
LGTM

PR #601, opened 2023-07-12
Error types now hold a Cow<str> instead of requiring a String

zslayton:
Prior to this patch, error structs like `DecodingError` each held a `String`, which meant that if a caller passed in a static string literal, it would always be unnecessarily copied to the heap.

This PR modifies each type and their corresponding constructors in the `IonFailure` trait to operate in terms of `Cow<'static, str>` and `Into<Cow<'static, str>>`, allowing static strings to avoid cloning while still supporting more complex error `String`s constructed with `format!` and the like.

### DIFF
--- a/src/catalog.rs
+++ b/src/catalog.rs
@@ -33,7 +33,6 @@ impl Catalog for MapCatalog {
         if name.is_empty() {
             return None;
         }
-      
         let versions: &BTreeMap<usize, SharedSymbolTable> = self.tables_by_name.get(name)?;
 
         let (_highest_version, table) = versions.iter().next_back()?;
--- a/src/catalog.rs
+++ b/src/catalog.rs
@@ -42,7 +42,7 @@ impl Catalog for MapCatalog {
                 ))
             })?;
 
-        let (_highest_version, table) = versions.iter().rev().next().ok_or_else(|| {
+        let (_highest_version, table) = versions.iter().next_back().ok_or_else(|| {
             IonError::illegal_operation(format!("symbol table with name: {name} does not exist"))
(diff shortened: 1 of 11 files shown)


src/result/mod.rs
```diff
@@ -80,25 +81,25 @@ pub(crate) trait IonFailure {
     // Because this trait is only crate-visible, methods can be added/changed as needed in
     // the future.
     fn incomplete(label: &'static str, position: impl Into<Position>) -> Self;
-    fn decoding_error<S: Into<String>>(description: S) -> Self;
-    fn encoding_error<S: Into<String>>(description: S) -> Self;
-    fn illegal_operation<S: Into<String>>(operation: S) -> Self;
+    fn decoding_error<S: Into<Cow<'static, str>>>(description: S) -> Self;
```
jobarr-amzn, line 84: Thinking out loud: Is it possible to typedef `Cow<'static, str>` or something? Is that even helpful?

jobarr-amzn (approved):

desaikd (approved):

src/catalog.rs
```diff
@@ -33,7 +33,7 @@ impl Catalog for MapCatalog {
         if name.is_empty() {
             return None;
         }
-
+
```
zslayton, line 36:
```suggestion
```

src/result/mod.rs
popematt: Yes! I want to see this:
```rust
type Moo = Cow<'static, str>;
```
 😉

popematt (approved):

PR #609, opened 2023-07-24
Initial raw lazy text reader (top-level nulls, bools, ints)

zslayton:
This adds an initial implementation of a `LazyRawTextReader` that can yield `LazyRawTextValue`s. At present it is limited to reading top-level `null`s, `bool`s, and `int`s of any radix. Support for other types, annotations, and containers will be added in follow-on PRs.

This implementation splits the parsing process for each value into two phases:
1. **Matching** finds a slice at the head of the input that corresponds to an element from the Ion grammar. It does not perform full semantic validation of the value.
2. **Reading** fully validates the previously matched input and turns it into a value.

This split makes it possible for the reader to avoid the cost of validation and materialization for values that the user simply skips over. To minimize rework done in the two phases, the matching phase records information it discovers along the way (subfield offsets, radix, etc) so it is available if and when the reading phase begins.


src/lazy/binary/encoding.rs
zslayton, line 1: 🗺️ This file was moved to its parent directory, not deleted. I'll call out its new location when it appears later in the diff.

src/lazy/text/as_utf8.rs
zslayton, line 1: 🗺️ This PR deals with two types that are effectively wrappers around a `&[u8]`: `TextBufferView` and `SmallVec`. Most of the time we don't need to view them as validated UTF-8 `&str`s--we can rely on the grammar to reject anything invalid--but sometimes we do. This trait defines an extension method that streamlines the "view these bytes as a `&str` or raise an `IonError`" task.

src/lazy/text/buffer.rs
```diff
@@ -0,0 +1,732 @@
...
+    pub fn len(&self) -> usize {
+        self.data.len()
+    }
+
+    /// Returns `true` if there are no bytes in the buffer. Otherwise, returns `false`.
+    pub fn is_empty(&self) -> bool {
+        self.data.is_empty()
+    }
+
+    pub fn match_whitespace(self) -> IonMatchResult<'data> {
```
zslayton, line 140: 🗺️ Note that only a handful of the `match_` parsing methods are `pub`. Most are helper-parsers.

src/lazy/text/parse_result.rs
```diff
@@ -0,0 +1,274 @@
...
+//         let mut message = String::from(value.description().unwrap_or("invalid text Ion syntax"));
+//         if let Some(label) = value.label {
+//             message.push_str(" while ");
+//             message.push_str(label.as_ref());
+//         }
+//         let position = Position::with_offset(value.input.offset()).with_length(value.input.len());
+//         let decoding_error = DecodingError::new(message).with_position(position);
+//         IonError::Decoding(decoding_error)
+//     }
+// }
```
zslayton, line 138: 🗺️ I'll remove this obsolete code after the PR is reviewed. Pushing code as I write the PR tour comments causes some number of the comments to be discarded 😐.

src/lazy/text/raw/reader.rs
```diff
@@ -0,0 +1,133 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::lazy::decoder::LazyRawValue;
+    use crate::IonType;
+
+    #[test]
+    fn test_top_level() -> IonResult<()> {
```
zslayton, line 73: 🗺️ This test demonstrates the total capability introduced in this PR. Namely, a `LazyRawTextReader` that can read top-level, unannotated nulls, bools, and ints in a text stream.

src/lazy/text/value.rs
```diff
@@ -0,0 +1,66 @@
...
+use crate::lazy::text::encoded_value::EncodedTextValue;
+use crate::lazy::text::matched::MatchedValue;
+use crate::{IonResult, IonType, RawSymbolTokenRef};
+use std::fmt;
+use std::fmt::{Debug, Formatter};
+
+/// A value that has been identified in the text input stream but whose data has not yet been read.
+///
+/// If only part of the value is in the input buffer, calls to [`LazyRawTextValue::read`] (which examines
+/// bytes beyond the value's header) may return [`IonError::Incomplete`](crate::result::IonError::Incomplete).
```
zslayton, line 15: 🗺️ I have a lingering reference to "the value's header" that was a copy/paste error from this type's binary counterpart. I'll fix that before merging.

zslayton (commented):
🗺️ PR tour

src/lazy/text/parse_result.rs
zslayton, line 1: 🗺️ Much of this was copied over from the old text reader implementation's error handling code and then updated.

src/lazy/text/parse_result.rs
```diff
@@ -0,0 +1,274 @@
...
+    }
+}
+
+pub(crate) trait AddContext<'data, T> {
+    fn with_context(
+        self,
+        label: impl Into<Cow<'static, str>>,
+        input: TextBufferView<'data>,
+    ) -> IonResult<(TextBufferView<'data>, T)>;
+}
```
zslayton, line 201: 🗺️ I plan to add a doc comment to this trait before merging. I intentionally avoided providing an `Into<IonError>` impl for `IonParseError` because some of the nom error kinds that bubble up as `IonParseError` don't retain context. This trait is an easy way for methods that could surface without context to require more information before being passed on to the user.

For an example of its usage, see `LazyRawTextReader::next()`:
image

zslayton (commented):
🗺️ couple more PR tour comments

src/lazy/text/buffer.rs
```diff
@@ -0,0 +1,732 @@
...
+        //       of further values should return an `Incomplete` or a `RawStreamItem::EndOfStream`.
+
+        // TODO: Check to see if `value` is actually an IVM.
+        //       => If it's a symbol, try the IVM parser on it and see if it succeeds.
+        //       For now, we just return the value.
+        Ok((remaining, RawStreamItem::Value(value)))
+    }
+
+    /// Matches a single scalar value or the beginning of a container.
+    pub fn match_value(self) -> IonParseResult<'data, LazyRawTextValue<'data>> {
```
desaikd, line 177: (Question) Should this be called `read_value` as it returns a parsed value? Same for `match_top_level`?

desaikd (approved):

src/result/decoding_error.rs
```diff
@@ -6,12 +7,23 @@ use thiserror::Error;
 #[error("{description}")]
 pub struct DecodingError {
     description: Cow<'static, str>,
+    position: Option<Position>,
```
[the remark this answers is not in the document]
zslayton: I'll add this in a follow-on PR to avoid needing to modify the queue.

src/lazy/text/buffer.rs
zslayton: I was reserving the `read_*` names for methods that produce one of the `RawValueRef::*` variants. The `LazyRawTextValue` that this returns has its own `read()` method that serves that purpose.

PR #612, opened 2023-07-27
Adds support for floats to the `LazyRawTextReader`

zslayton:
This PR builds on #609. Adds support for matching/reading `float` values in the `LazyRawTextReader`.


src/lazy/text/buffer.rs
```diff
@@ -192,6 +192,12 @@ impl<'data> TextBufferView<'data> {
                     EncodedTextValue::new(MatchedValue::Int(matched_int), self.offset(), length)
                 },
             ),
+            map(
+                match_and_length(Self::match_float),
+                |(matched_float, length)| {
+                    EncodedTextValue::new(MatchedValue::Float(matched_float), self.offset(), length)
+                },
```
zslayton, line 199: 🗺️ When we match a `float`, record information about the match (offset, length, what kind of float) in case we eventually parse it.

src/lazy/text/buffer.rs
```diff
@@ -372,6 +378,111 @@ impl<'data> TextBufferView<'data> {
...
         take_while1(|b: u8| b.is_ascii_hexdigit())(self)
     }
+
+    /// Matches an Ion float of any syntax
+    fn match_float(self) -> IonParseResult<'data, MatchedFloat> {
+        alt((
+            Self::match_float_special_value,
+            Self::match_float_numeric_value,
+        ))(self)
+    }
```
zslayton, line 388: 🗺️ This is the entry point for parsing`float`s, which have two kinds of syntax: keyword special values (`+inf`, `-inf`, `nan`), and numeric values. Numeric values come in a big variety of shapes.

src/lazy/text/matched.rs
```diff
@@ -107,3 +111,46 @@ impl MatchedInt {
...
+pub(crate) enum MatchedFloat {
+    /// `+inf`
+    PositiveInfinity,
+    /// `-inf`
+    NegativeInfinity,
+    /// `nan`
+    NotANumber,
+    /// Any numeric float value
+    Numeric,
+}
```
zslayton, line 126: 🗺️ PR #609 introduced an `EncodedValue` struct that stores information like offset and length, but which also has a `MatchedValue` enum field. This is one of the variants of `MatchedValue`, and remembers the kind of `float` we encountered in case we read it later. The offset and length will be available in the parent data structure.

src/lazy/text/parse_result.rs
zslayton, line 1: 🗺️ Changes in this file are reorganizing how various error types are converted into one another for easier bubbling up. No real logic changes.

zslayton (commented):
🗺️ PR tour

popematt (approved):

PR #613, opened 2023-07-27
Adds `LazyRawTextReader` support for comments

zslayton:
Builds on #609 and #612. Adds support for reading 

```rust
// rest-of-the-line
``` 
and
```rust
/* multi
    line  */
```
comments.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #617, opened 2023-08-03
Adds `LazyRawTextReader` support for reading lists

zslayton:
Builds on outstanding PRs #609, #612, #613, #614 and #616.

Adds support for reading lists to the `LazyRawTextReader`.


src/lazy/binary/raw/sequence.rs
```diff
@@ -18,11 +18,11 @@ impl<'data> LazyRawBinarySequence<'data> {
         self.value.ion_type()
     }
 
-    pub fn iter(&self) -> RawSequenceIterator<'data> {
+    pub fn iter(&self) -> RawBinarySequenceIterator<'data> {
         // Get as much of the sequence's body as is available in the input buffer.
         // Reading a child value may fail as `Incomplete`
         let buffer_slice = self.value.available_body();
-        RawSequenceIterator::new(buffer_slice)
+        RawBinarySequenceIterator::new(buffer_slice)
```
zslayton, line 25: 🗺️ Now that there are both binary and text readers, `RawSequenceIterator` has become the `RawBinarySequenceIterator`.

src/lazy/text/buffer.rs
zslayton, line 1: 🗺️ This file contains the new parsing logic needed to detect the beginning of a list, a value in the list, and the end of the list.

src/lazy/text/matched.rs
zslayton, line 1: 🗺️ All changes in this file stem from the change on line 175; see the `TODO` comment.

src/lazy/text/raw/sequence.rs
```diff
@@ -0,0 +1,126 @@
...
+    fn next(&mut self) -> Option<Self::Item> {
+        match self.input.match_list_value() {
+            Ok((remaining, Some(value))) => {
+                self.input = remaining;
+                Some(Ok(value))
+            }
+            Ok((_remaining, None)) => None,
+            Err(e) => e
+                .with_context("reading the next list value", self.input)
+                .transpose(),
```
zslayton, line 123: 🗺️ This call to `transpose()` is the `ToIteratorOutput` trait's extension method.

zslayton (commented):
🗺️ PR Tour

src/lazy/text/raw/sequence.rs
```diff
@@ -0,0 +1,126 @@
...
+impl<'a, 'data> IntoIterator for &'a LazyRawTextSequence<'data> {
+    type Item = IonResult<LazyRawTextValue<'data>>;
+    type IntoIter = RawTextSequenceIterator<'data>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.iter()
+    }
+}
+
+impl<'a> Debug for LazyRawTextSequence<'a> {
```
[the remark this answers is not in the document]
popematt: Suggestion—can we just use `self.value.input.data` (assuming it's valid utf-8) and if it fails, then we fallback to the `Debug` impl of the `TextBufferView`? This seems like an approach that could work for _all_ lazy text values. What do you think?

popematt (approved):

PR #619, opened 2023-08-10
Adds `LazyRawTextReader` support for structs

zslayton:
Builds on outstanding PRs #609, #612, #613, #614, #616, and #617.

Adds support for reading structs to the `LazyRawTextReader`.


src/lazy/encoding.rs
zslayton, line 1: 🗺️ This file housed lots of `ToDo*` types that existed to satisfy the `LazyDecoder` trait. This PR replaces them with actual implementations.

src/lazy/text/buffer.rs
```diff
@@ -331,6 +411,74 @@ impl<'data> TextBufferView<'data> {
         .parse(self)
     }
 
+    /// Matches a list.
```
zslayton, line 414: 🗺️ Introduces a matching function for the `list` type that finds its end offset.

src/lazy/text/buffer.rs
```diff
@@ -715,7 +867,13 @@ impl<'data> TextBufferView<'data> {
     /// Matches a quoted symbol (`'foo'`).
     fn match_quoted_symbol(self) -> IonParseResult<'data, MatchedSymbol> {
         delimited(char('\''), Self::match_quoted_symbol_body, char('\''))
-            .map(|(_matched, contains_escaped_chars)| MatchedSymbol::Quoted(contains_escaped_chars))
```
zslayton, line 718: 🗺️ As above, I've broken out the nested `(bool)` on the `MatchedSymbol` variant.

src/lazy/text/value.rs
```diff
@@ -46,7 +48,10 @@ impl<'data> LazyRawValue<'data, TextEncoding> for LazyRawTextValue<'data> {
     }
 
     fn read(&self) -> IonResult<RawValueRef<'data, TextEncoding>> {
-        let matched_input = self.input.slice(0, self.encoded_value.data_length());
+        let matched_input = self.input.slice(
+            self.encoded_value.data_offset() - self.input.offset(),
+            self.encoded_value.data_length(),
+        );
```
zslayton, line 54: 🗺️ Prior to this PR, the matched input slice always started with the value. Now that we have struct fields, it sometimes begins at the field name.

zslayton (commented):
🗺️ PR tour

src/lazy/text/raw/reader.rs
```diff
@@ -155,6 +157,15 @@ mod tests {
             3
         ]
         
+        {
+            // Identifier 
+            foo: 100,
+            // Quoted symbol
+            'bar': 200,
+            // Short-form string
+            "baz": 300
```
popematt, line 166: Would be good to have some comments in unexpected (but legal) locations.
```suggestion
            "baz": 300,
            /* Comments... */ qux /* comments */ : /* are */ 400 /* everywhere */,
```

src/lazy/text/buffer.rs
```diff
@@ -246,6 +248,78 @@ impl<'data> TextBufferView<'data> {
...
+                    value.input = input_including_field_name;
+                    Some(LazyRawTextField { value })
+                },
+            ),
+        ))(input_including_field_name)
+    }
+
+    /// Matches any amount of whitespace followed by a closing `}`.
+    fn match_struct_end(self) -> IonMatchResult<'data> {
+        whitespace_and_then(peek(tag("}"))).parse(self)
```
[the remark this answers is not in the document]
zslayton: Yes, it does. I wanted a more explicit name but couldn't get it short enough. `ws_and_comments_and_then`?

src/lazy/text/encoded_value.rs
```diff
@@ -60,6 +62,8 @@ pub(crate) struct EncodedTextValue {
     // value is stored. For others (e.g. a timestamp), the various components of the value are
     // recognized during matching and partial information like subfield offsets can be stored here.
     matched_value: MatchedValue,
+
+    field_name_syntax: Option<MatchedSymbol>,
```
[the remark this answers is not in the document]
zslayton: 1. Will do.
2. I'm not sure I understand the suggestion, but I can explain the approach. Both the binary version and the text version model their lazy values as a `(field_id?, annotations?, value)` tuple. That is, the field is modeled as being part of the value. (This language is loose; the data structures and approach to parsing couple the field with the value but I recognize that "the field is part of the value" doesn't feel quite right from an Ion data model perspective.) If I did introduce a `LazyRawTextSymbolToken` type, it would simply wrap a `MatchedSymbol`. Let me know if you'd like to discuss this further, I think an offline discussion would probably be helpful.
3. I think it's possible, yeah. Might even be a small performance win by shrinking the size of values that aren't in structs. I've opened #631 to track investigating this.

src/lazy/text/buffer.rs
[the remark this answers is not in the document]
popematt: I'm ambivalent about the name. I was just looking to confirm my assumption about its behavior.

popematt (approved):

src/lazy/text/encoded_value.rs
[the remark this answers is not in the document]
zslayton: Opened #632 to track investigating an optimized representation for annotations offset info.

PR #620, opened 2023-08-10
Adds `LazyRawTextReader` support for reading IVMs

zslayton:
Builds on outstanding PRs #609, #612, #613, #614, #616, #617, and #619.

Adds support for reading Ion version markers (IVMs) to the `LazyRawTextReader`.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #621, opened 2023-08-11
Initial impl of a LazyRawAnyReader

zslayton:
Builds on outstanding PRs #609, #612, #613, #614, #616, #617, #619, and #620.

Adds the `LazyRawAnyReader`, a `LazyDecoder` implementation that can read either text or binary Ion.

Because `LazyDecoder` is a family of types, using it via `Box<dyn>` would require boxing and dynamic dispatch at multiple levels. This implementation of the `LazyRawAnyReader` avoids that problem by defining several new types that hold either a text or binary reader in an `enum` and delegates method calls to the one in use.


src/lazy/any_encoding.rs
```diff
@@ -0,0 +1,535 @@
...
+            LazyRawFieldKind::Binary_1_0(f) => f.name(),
+        }
+    }
+
+    fn value(&self) -> LazyRawAnyValue<'data> {
+        match &self.encoding {
+            LazyRawFieldKind::Text_1_0(f) => f.value().into(),
+            LazyRawFieldKind::Binary_1_0(f) => f.value().into(),
+        }
+    }
```
zslayton, line 404: 🗺️ There's a subtle API challenge here.

Prior to this PR, the `value()` trait method required that implementers return a `&D::Value`; in the case of `AnyEncoding`, that would mean returning a `&LazyRawAnyValue`.

Any time a method returns a `&T` (not `&'static T`), there MUST be a `T` living inside one of the method's parameters. In this case, there would need to be a `LazyRawAnyValue` living inside `self`, which there isn't. Instead, `self` contains the raw materials to construct a `LazyRawAnyValue`: either a `LazyRawTextValue` or a `LazyRawBinaryValue`. We can't return a `&LazyRawAnyValue` because it's illegal to return a reference to a stack-allocated variable created in this method's scope.

Fortunately, both `LazyRawTextValue` and `LazyRawBinaryValue` are eligible to implement `Copy`. Neither owns any heap-allocated resources. I changed the trait to return a copy of the value instead of a reference to one, resolving this issue. There are minor ripples of this change through the rest of this PR.

zslayton (commented):
🗺️ PR tour

src/lazy/any_encoding.rs
zslayton: Now that `as_value` is returning a copy of the value and not a reference, it should be called `to_value`. I've captured that in #626 and will address it when the merge queue is empty.

src/lazy/any_encoding.rs
```diff
@@ -0,0 +1,535 @@
...
+#[derive(Debug, Clone)]
+pub struct LazyRawAnyStruct<'data> {
+    encoding: LazyRawStructKind<'data>,
+}
+
+#[derive(Debug, Clone)]
+pub enum LazyRawStructKind<'data> {
+    Text_1_0(LazyRawTextStruct<'data>),
+    Binary_1_0(LazyRawBinaryStruct<'data>),
+}
```
popematt, line 329: Are these `LazyRawAny*` structs exposed to users of the crate, or are they just `pub` in this module? If just `pub` in this module, is there any reason these can't be collapsed to remove the extra layer of indirection? (And similarly for the other struct/enum combos.)

```suggestion
#[derive(Debug, Clone)]
pub struct LazyRawAnyStruct<'data> {
    Text_1_0(LazyRawTextStruct<'data>),
    Binary_1_0(LazyRawBinaryStruct<'data>),
}
```

popematt: Offline discussion—the structs will eventually be public.

popematt (approved):

src/lazy/any_encoding.rs
zslayton: The plan is for the `struct`s to be `pub`, but the enums to be hidden. That way they don't have to be marked `non_exhaustive` and we can add other formats/versions over time.

PR #622, opened 2023-08-12
Adds lazy reader support for reading annotations

zslayton:
Builds on outstanding PRs #609, #612, #613, #614, #616, #617, #619, and #620, #621.

Adds support for reading annotations in the `LazyRawTextReader` and `LazyRawAnyReader`.


src/lazy/any_encoding.rs
```diff
@@ -37,7 +38,7 @@ impl<'data> LazyDecoder<'data> for AnyEncoding {
     type Value = LazyRawAnyValue<'data>;
     type Sequence = LazyRawAnySequence<'data>;
     type Struct = LazyRawAnyStruct<'data>;
-    type AnnotationsIterator = Box<dyn Iterator<Item = IonResult<RawSymbolTokenRef<'data>>>>;
+    type AnnotationsIterator = RawAnyAnnotationsIterator<'data>;
```
zslayton, line 41: 🗺️ The `Box<dyn _>` type was just a placeholder.

src/lazy/any_encoding.rs
```diff
@@ -181,10 +182,10 @@ impl<'data> From<RawStreamItem<'data, BinaryEncoding>> for RawStreamItem<'data,
 }
 
 impl<'data> LazyRawValuePrivate<'data> for LazyRawAnyValue<'data> {
-    fn field_name(&self) -> Option<RawSymbolTokenRef<'data>> {
+    fn field_name(&self) -> IonResult<RawSymbolTokenRef<'data>> {
```
zslayton, line 185: 🗺️ This trait method signature had to be changed. The binary raw reader only ever encounters `SymbolId` field names. The text reader, however, can encounter string and symbol field names with invalid text. (For example: illegal escape sequences or invalid unicode.) Thus, we now return an `IonResult` instead of an `Option`.

src/lazy/any_encoding.rs
```diff
@@ -204,8 +205,15 @@ impl<'data> LazyRawValue<'data, AnyEncoding> for LazyRawAnyValue<'data> {
         }
     }
 
-    fn annotations(&self) -> <AnyEncoding as LazyDecoder<'data>>::AnnotationsIterator {
-        todo!()
```
zslayton, line 208: 🗺️ As with the other `LazyRawAnyReader` methods, we branch on which encoding we're reading and delegate the method call to the appropriate reader.

src/lazy/binary/raw/value.rs
```diff
@@ -179,7 +185,7 @@ impl<'data> LazyRawBinaryValue<'data> {
 
     /// If this value is within a struct, returns its associated field name as a `Some(SymbolID)`.
     /// Otherwise, returns `None`.
-    pub(crate) fn field_name(&self) -> Option<SymbolId> {
+    pub(crate) fn field_id(&self) -> Option<SymbolId> {
```
zslayton, line 188: 🗺️ I renamed this binary reader method from `field_name` to `field_id` to avoid be confused with the raw reader trait of the same name.

The binary reader can always return a symbol ID representing the field name if there is one. However, the raw reader trait needs to accommodate failure cases in the raw text reader and returns an `IonResult<RawSymbolTokenRef>` instead of an `Option<SymbolId>`.

zslayton (commented):
🗺️ PR tour

popematt (approved):

PR #623, opened 2023-08-14
Adds lazy reader support for timestamps

zslayton:
Builds on outstanding PRs #609, #612, #613, #614, #616, #617, #619, #620, #621, and #622.

Adds support for reading text timestamps in the lazy reader.


src/lazy/any_encoding.rs
```diff
@@ -550,6 +550,10 @@ mod tests {
                 reader.next()?.expect_value()?.read()?,
                 RawValueRef::Int(5.into())
             );
+            assert_eq!(
+                reader.next()?.expect_value()?.read()?,
+                RawValueRef::Timestamp(Timestamp::with_year(2023).with_month(8).build()?)
+            );
```
zslayton, line 556: 🗺️ This simple test confirms that the `LazyRawAnyReader` can surface timestamps as expected. More rigorous correctness testing happens later in the diff.

zslayton (commented):
🗺️ PR tour

src/lazy/text/matched.rs
```diff
@@ -451,3 +455,290 @@ impl MatchedSymbol {
...
+        self
+    }
+
+    pub(crate) fn read(&self, matched_input: TextBufferView) -> IonResult<Timestamp> {
+        // The parser has already confirmed that each subfield is made of ASCII digits,
+        // so UTF-8 validation and parsing cannot fail. `unwrap()` is used in such cases
+        // throughout.
+        let year_text = matched_input.slice(0, 4).as_text().unwrap();
+        let year = u32::from_str(year_text).unwrap();
+        let timestamp = Timestamp::with_year(year);
```
popematt, line 486: Nice timestamp builder you have there. 😉

src/lazy/text/matched.rs
```diff
@@ -451,3 +455,290 @@ impl MatchedSymbol {
...
+                timestamp.with_milliseconds(milliseconds)
+            }
+            6 => {
+                let microseconds = u32::from_str(fractional_text).unwrap();
+                timestamp.with_microseconds(microseconds)
+            }
+            9 => {
+                let nanoseconds = u32::from_str(fractional_text).unwrap();
+                timestamp.with_nanoseconds(nanoseconds)
+            }
```
popematt, line 593: I think it would be nice to avoid creating a `Decimal` if we don't need to. I don't know how costly this would be, though. It's a small enough range that we could have a hard coded match for mapping `1-9` to powers of ten if need be.
```suggestion
            len if len <= 9 => {
                let fraction = u32::from_str(fractional_text).unwrap();
                let multiplier = pow(10, 9 - len);
                let nanoseconds = fraction * multiplier; 
                self.with_nanoseconds_and_precision(nanoseconds, len.to_u32().unwrap())
            }
```

src/lazy/text/matched.rs
```diff
@@ -451,3 +455,290 @@ impl MatchedSymbol {
...
+pub enum MatchedTimestampOffset {
+    Zulu,
+    HoursAndMinutes(MatchedHoursAndMinutes),
+    Unknown,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct MatchedHoursAndMinutes {
+    is_negative: bool,
+    hours_offset: usize,
```
popematt, line 620: Do you mean minutes?
```suggestion
    minutes_offset: usize,
```

zslayton: `hours` is correct; this is the text offset of the first `H` in the offset substring `HH:MM`. I'll add a comment to clarify!

src/lazy/text/buffer.rs
```diff
@@ -967,6 +978,208 @@ impl<'data> TextBufferView<'data> {
...
+
+    /// Matches the fractional seconds component of a timestamp, including a leading `.`.
+    fn match_timestamp_fractional_seconds(self) -> IonMatchResult<'data> {
+        preceded(tag("."), digit1)(self)
+    }
+
+    /// Matches a timestamp offset of any format.
+    fn match_timestamp_offset(self) -> IonParseResult<'data, MatchedTimestampOffset> {
+        alt((
+            value(MatchedTimestampOffset::Zulu, tag("Z")),
```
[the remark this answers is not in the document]
zslayton: It might be a wash; it makes reading `+00:00` cheaper at the expense of an extra branch in the general case. I've added the change for now, we can always revisit it later with benchmarks.

src/lazy/text/buffer.rs
```diff
@@ -1428,6 +1645,50 @@ mod tests {
...
+            "2023-13T",                   // Out of bounds month
+            "2023-08-41T",                // Out of bounds day
+            "2023-08+18T",                // Wrong delimiter
+            "2023-08-18T25:00Z",          // Out of bounds hour
+            "2023-08-18T14:00",           // No offset
+            "2023-08-18T14:62",           // Out of bounds minute
+            "2023-08-18T14:35:61",        // Out of bounds second
+            "2023-08-18T14:35:52.Z",      // Dot but no fractional
+            "2023-08-18T14:35:52.+24:30", // Out of bounds offset hour
+            "2023-08-18T14:35:52.+00:60", // Out of bounds offset minute
```
popematt, line 1685: These both have a dot but no fractional digits, so the test case can pass even if the out of bounds offset is not correctly rejected.

popematt (approved):
Two test cases that need a minor fix, but otherwise it looks great.

PR #627, opened 2023-08-18
Lazy reader support for s-expressions

zslayton:
Builds on outstanding PRs #612, #613, #614, #616, #617, #619, #620, #621, #622, and #623.

Adds support for reading text s-expressions in the lazy reader.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #629, opened 2023-08-20
Adds lazy reader support for blobs

zslayton:
Builds on outstanding PRs #612, #613, #614, #616, #617, #619, #620, #621, #622, #623, #627, and #628.

Adds lazy reader support for blobs.


src/lazy/bytes_ref.rs
```diff
@@ -0,0 +1,123 @@
+use crate::text::text_formatter::IonValueFormatter;
+use crate::Bytes;
+use std::borrow::Cow;
+use std::fmt::{Debug, Display, Formatter};
+use std::ops::Deref;
+
+pub struct BytesRef<'data> {
+    data: Cow<'data, [u8]>,
+}
```
zslayton, line 9: 🗺️ When there was only a binary reader, methods that read blobs could always return a `&[u8]`--a slice of the input buffer. Now that we also have a text reader, we need to accommodate base64-encoded blobs, which always require a new `Vec` to be allocated to hold the decoded data. `BytesRef` can hold either a borrowed `&[u8]` or an owned `Vec<u8>`, allowing it to be used in either situation.

This type is analogous to `StrRef` and `SymbolRef` but for blobs.

src/lazy/raw_value_ref.rs
```diff
@@ -18,7 +19,7 @@ pub enum RawValueRef<'data, D: LazyDecoder<'data>> {
     Timestamp(Timestamp),
     String(StrRef<'data>),
     Symbol(RawSymbolTokenRef<'data>),
-    Blob(&'data [u8]),
+    Blob(BytesRef<'data>),
```
zslayton, line 22: 🗺️ `RawValueRef` now returns a `BytesRef` instead of a `&[u8]` so it has the option to allocate a `Vec<u8>` when the input encoding is base64 text. The binary reader can still return a slice of the input buffer.

src/lazy/value_ref.rs
```diff
@@ -23,7 +24,7 @@ pub enum ValueRef<'top, 'data, D: LazyDecoder<'data>> {
     Timestamp(Timestamp),
     String(StrRef<'data>),
     Symbol(SymbolRef<'top>),
-    Blob(&'data [u8]),
+    Blob(BytesRef<'data>),
```
zslayton, line 27: 🗺️ Like `RawValueRef`, `ValueRef` now returns a `BytesRef` instead of a `&[u8]` so it has the option to allocate a `Vec<u8>` when the input encoding is base64 text. The binary reader can still return a slice of the input buffer.

zslayton (commented):
🗺️ PR tour

src/lazy/text/buffer.rs
```diff
@@ -2002,4 +2030,50 @@ mod tests {
...
+            "{{  aGVsbG8=}}",
+            "{{aGVsbG8=  }}",
+            "{{\taGVsbG8=\n\n}}",
+            // hello!
+            "{{aGVsbG8h}}",
+            "{{  aGVsbG8h}}",
+            "{{aGVsbG8h  }}",
+            "{{  aGVsbG8h  }}",
+            // razzle dazzle root beer
+            "{{cmF6emxlIGRhenpsZSByb290IGJlZXI=}}",
```
popematt, line 2055: Add a test case with extra white space in the middle.
```suggestion
            "{{cmF6emxlIGRhenpsZSByb290IGJlZXI=}}",
            "{{ cm F6emx lIGRhenps  ZSByb29  0IGJlZ XI= }}",
```

src/lazy/text/buffer.rs
```diff
@@ -1337,6 +1343,28 @@ impl<'data> TextBufferView<'data> {
...
         )(self)
     }
+
+    /// Matches a complete blob, including the opening `{{` and closing `}}`.
+    pub fn match_blob(self) -> IonParseResult<'data, MatchedBlob> {
+        delimited(
+            tag("{{"),
+            // Only whitespace (not comments) can appear within the blob
+            preceded(Self::match_optional_whitespace, Self::match_base64_content),
+            preceded(Self::match_optional_whitespace, tag("}}")),
```
popematt, line 1353: > Within blob values, whitespace is ignored.

That means something like `{{ cm F6emx lIGRhenps  ZSByb29  0IGJlZ XI= }}` should be matched successfully, right? 

Also, it looks like the `match_base64_content` function does not match an empty input/string, and the base64 content is not optional here, so I don't think this would match `{{}}` as it should.

src/lazy/text/buffer.rs
```diff
@@ -2002,4 +2030,50 @@ mod tests {
...
+    #[test]
+    fn test_match_blob() {
+        fn match_blob(input: &str) {
+            MatchTest::new(input).expect_match(match_length(TextBufferView::match_blob));
+        }
+        fn mismatch_blob(input: &str) {
+            MatchTest::new(input).expect_mismatch(match_length(TextBufferView::match_blob));
+        }
+        // Base64 encodings of utf-8 strings
+        let good_inputs = &[
```
popematt, line 2043: Empty blob.
```suggestion
        let good_inputs = &[
            "{{}}",
```

popematt (commented):
It looks like there are some edge cases that might need to be fixed, but I might be wrong.

popematt (approved):

Issue #618, opened 2023-08-09
`Element::write_*` APIs need a way to specify an Ion version


zslayton:
We may need a `WriteConfig` API that allows callers to specify a format and version. This overlaps with the idea of a `WriterBuilderConfig`, so we may need another factoring.

desaikd:
Completed with #685 which adds a writer configuration

PR #633, opened 2023-08-28
Adds experimental serde feature

desaikd:
*Related PR: #402*

*Description of changes:*
This PR continues on a [previous PR](#402) to implement `serde` for `ion-rs` and adds an experimental `serde` feature. 

*List of changes:*
* adds serde module which defines `ser` (serialization) and `de` (deserialization) for Ion data model.
* adds `timestamp` which is a special case for `serde` since serde data model doesn't support Ion timestamps, we use `newtype_struct` to serialize and deserialize timestamps.
* adds `timestamp` which is a special case for `serde` since serde data model doesn't support Ion timestamps, we use `newtype_struct` to serialize and deserialize timestamps.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,8 @@ experimental = [
   "experimental-lazy-reader",
   "experimental-reader",
   "experimental-writer",
-  "experimental-streaming"
+  "experimental-streaming",
+  "experimental-serde"
 ]
 
 # Experimental streaming APIs
@@ -52,6 +53,9 @@ experimental-writer = []
 # an inexpensive form of 'rewind' functionality.
 experimental-lazy-reader = []
 
+# Experimental serde API to serialize and deserialize Ion data into Rust objects using serde crate
+experimental-serde = ["dep:serde_with", "dep:serde"]
+
 [dependencies]
 base64 = "0.12"
 bytes = "0.4"
@@ -69,6 +73,8 @@ arrayvec = "0.7"
 smallvec = {version ="1.9.0", features = ["const_generics"]}
 digest = { version = "0.9", optional = true }
 sha2 = { version = "0.9", optional = true }
+serde = { version = "1.0", features = ["derive"], optional = true }
+serde_with = { verison = "2.0", optional = true }
 
 [dev-dependencies]
 rstest = "0.17.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ experimental = [
   "experimental-reader",
   "experimental-writer",
   "experimental-streaming",
-  "experimental-serde"
+  "experimental-serde",
 ]
 
 # Experimental streaming APIs
@@ -74,7 +74,7 @@ smallvec = {version ="1.9.0", features = ["const_generics"]}
 digest = { version = "0.9", optional = true }
 sha2 = { version = "0.9", optional = true }
 serde = { version = "1.0", features = ["derive"], optional = true }
-serde_with = { verison = "2.0", optional = true }
+serde_with = { version = "2.0", optional = true }
 
 [dev-dependencies]
 rstest = "0.17.0"
--- a/src/serde/de.rs
+++ b/src/serde/de.rs
@@ -1,0 +1,883 @@
+use crate::data_source::IonDataSource;
+use crate::result::IonFailure;
+use crate::serde::timestamp::ION_BINARY;
+use crate::{IonError, IonReader, IonResult, IonType, ReaderBuilder, StreamItem, Symbol};
+use chrono::{DateTime, FixedOffset};
+use serde::de;
+use serde::de::{DeserializeSeed, EnumAccess, MapAccess, SeqAccess, Visitor};
+use serde::Deserialize;
+use std::borrow::Cow;
+use std::iter::FusedIterator;
+use std::marker::PhantomData;
+
+/// Generic method that can deserialize an object from any given object
+/// that implements `ToIonDateSource`. It also enables `ION_BINARY`
+/// to ensure that `Timestamp` objects are correctly deserialized.
+pub fn from_ion<'a, T, S>(s: S) -> IonResult<T>
+where
+    T: Deserialize<'a>,
+    S: IonDataSource,
+{
+    ION_BINARY.with(move |cell| {
+        cell.set(true);
+        let mut deserializer = Deserializer {
+            reader: ReaderBuilder::new().build(s)?,
+        };
+
+        // If we are hovering over nothing call next till we have an object
+        while StreamItem::Nothing == deserializer.reader.current() {
+            deserializer.reader.next()?;
+        }
+
+        let result = T::deserialize(&mut deserializer)?;
+        cell.set(false);
+        Ok(result)
+    })
+}
+
+/// The deserializer for Ion, it doesn't care about
+/// whether the reader is reading binary or text representation.
+#[derive(Debug)]
+pub struct Deserializer<R> {
+    pub(crate) reader: R,
+}
+
+impl<'de, 'a, R> serde::de::Deserializer<'de> for &'a mut Deserializer<R>
+where
+    R: IonReader<Symbol = Symbol, Item = StreamItem>,
+{
+    type Error = IonError;
+
+    fn deserialize_any<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        if let Some(ion_type) = self.reader.ion_type() {
+            match ion_type {
+                IonType::Null => self.deserialize_unit(visitor),
+                IonType::Blob | IonType::Clob => self.deserialize_bytes(visitor),
+                IonType::String | IonType::Symbol => self.deserialize_string(visitor),
+                IonType::Float => self.deserialize_f64(visitor),
+                IonType::Int => self.deserialize_i64(visitor),
+                IonType::Decimal => self.deserialize_newtype_struct("$__ion_rs_decimal__", visitor),
+                IonType::Bool => self.deserialize_bool(visitor),
+                IonType::List => self.deserialize_seq(visitor),
+                IonType::Struct => self.deserialize_struct("", &[], visitor),
+                IonType::Timestamp => {
+                    self.deserialize_newtype_struct("$__ion_rs_timestamp__", visitor)
+                }
+                _ => IonResult::decoding_error("unexpected ion type".to_string()),
+            }
+        } else {
+            IonResult::decoding_error("unexpected end of file".to_string())
+        }
+    }
+
+    fn deserialize_bool<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        let result = visitor.visit_bool(self.reader.read_bool()?);
+        self.reader.next()?;
+        result
+    }
+
+    fn deserialize_i8<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        let result = visitor.visit_i8(self.reader.read_i64().map(|x| x as i8)?);
+        self.reader.next()?;
+        result
+    }
+
+    fn deserialize_i16<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        let result = visitor.visit_i16(self.reader.read_i64().map(|x| x as i16)?);
+        self.reader.next()?;
+        result
+    }
+
+    fn deserialize_i32<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        let result = visitor.visit_i32(self.reader.read_i64().map(|x| x as i32)?);
+        self.reader.next()?;
+        result
+    }
+
+    fn deserialize_i64<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        let result = visitor.visit_i64(self.reader.read_i64()?);
+        self.reader.next()?;
+        result
+    }
+
+    fn deserialize_u8<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        let result = visitor.visit_u8(self.reader.read_i64().map(|x| x as u8)?);
+        self.reader.next()?;
+        result
+    }
+
+    fn deserialize_u16<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        let result = visitor.visit_u16(self.reader.read_i64().map(|x| x as u16)?);
+        self.reader.next()?;
+        result
+    }
+
+    fn deserialize_u32<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        let result = visitor.visit_u32(self.reader.read_i64().map(|x| x as u32)?);
+        self.reader.next()?;
+        result
+    }
+
+    fn deserialize_u64<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        let result = visitor.visit_u64(self.reader.read_i64().map(|x| x as u64)?);
+        self.reader.next()?;
+        result
+    }
+
+    fn deserialize_f32<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        let result = visitor.visit_f32(self.reader.read_f32()?);
+        self.reader.next()?;
+        result
+    }
+
+    fn deserialize_f64<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        let result = visitor.visit_f64(self.reader.read_f64()?);
+        self.reader.next()?;
+        result
+    }
+
+    fn deserialize_char<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        let result = visitor.visit_char(
+            self.reader
+                .read_string()
+                .map(|x| x.to_string().chars().collect::<Vec<char>>()[0])?,
+        );
+        self.reader.next()?;
+        result
+    }
+
+    fn deserialize_str<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        let result = visitor.visit_string(self.reader.read_string()?.to_string());
+        self.reader.next()?;
+        result
+    }
+
+    fn deserialize_string<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        let result = visitor.visit_string(self.reader.read_str()?.to_owned());
+        self.reader.next()?;
+        result
+    }
+
+    fn deserialize_bytes<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        let result = visitor.visit_bytes(self.reader.read_blob()?.as_slice());
+        self.reader.next()?;
+        result
+    }
+
+    fn deserialize_byte_buf<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        let result = visitor.visit_byte_buf(self.reader.read_blob()?.as_slice().to_vec());
+        self.reader.next()?;
+        result
+    }
+
+    fn deserialize_option<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        if self.reader.is_null() {
+            self.reader.next()?;
+            visitor.visit_none()
+        } else {
+            visitor.visit_some(self)
+        }
+    }
+
+    fn deserialize_unit<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        if self.reader.is_null() {
+            self.reader.next()?;
+            visitor.visit_unit()
+        } else {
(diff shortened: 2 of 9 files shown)


Cargo.toml
```diff
@@ -33,7 +33,8 @@ experimental = [
   "experimental-lazy-reader",
   "experimental-reader",
   "experimental-writer",
-  "experimental-streaming"
+  "experimental-streaming",
+  "experimental-serde"
```
popematt, line 37: _extremely_ minor nitpick—please add a trailing comma so that when the next experimental feature is added, we don't don't have to add a comma to an unrelated feature.

src/serde/mod.rs
popematt, line 1: Would be good to have some module-level documentation describing the behavior of this module. E.g.:
* How do the Ion concepts map to Rust concepts?
* What limitations are there?
* What is still to-do or likely to change?
* (How) Are annotations handled?
* What is the tagging strategy for enum variants? Is it configurable (so that that data can be made compatible with e.g. `jackson-dataformat-ion`)?
* What Ion-specific knobs are available to configure the ser/de behavior?

src/serde/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+    use crate::{Decimal, Timestamp};
+    use chrono::{DateTime, FixedOffset, Utc};
+    use serde::{Deserialize, Serialize};
+    use serde_with::serde_as;
+
+    #[test]
+    fn test_struct() {
+        #[serde_as]
+        #[derive(Serialize, Deserialize)]
+        struct Test {
```
popematt, line 23: Can you add a field that is `Optional<_>`?

src/serde/timestamp.rs
```diff
@@ -0,0 +1,135 @@
...
+use serde::{self, de, ser, Deserialize, Serialize};
+use serde_with::{DeserializeAs, SerializeAs};
+use std::cell::Cell;
+use std::fmt;
+
+thread_local! {
+    /// Cell that contains a flag to determine when serialization and deserialization
+    /// is occurring with an Ion serializer. This allows us to know when to encode
+    /// Timestamps as ion timestamps
+    pub(crate) static ION_BINARY: Cell<bool> = Cell::new(false);
```
popematt, line 13: Can you provide more explanation or a link to some documentation about why this is necessary? Maybe this is a naive question, but why is this a concern in the first place? (And why does it only apply to binary?)

Also, it looks like this flag is also used in the `decimal` module, so perhaps it should belong in `mod.rs` rather than in `timestamp.rs`.

src/serde/ser.rs
```diff
@@ -0,0 +1,991 @@
...
+use serde::ser::Impossible;
+use serde::{ser, Serialize};
+
+/// Serialize an object into pretty formatted Ion text
+pub fn to_pretty<T>(value: &T) -> IonResult<String>
+where
+    T: Serialize,
+{
+    ION_BINARY.with(move |cell| {
+        cell.set(true);
```
popematt, line 21: Is `ION_BINARY` named correctly? Here you are setting it to `true` for Ion text.

src/serde/ser.rs
```diff
@@ -0,0 +1,991 @@
...
+
+    fn serialize_newtype_struct<T: ?Sized>(
+        self,
+        name: &'static str,
+        value: &T,
+    ) -> Result<Self::Ok, Self::Error>
+    where
+        T: Serialize,
+    {
+        if name == "$__ion_rs_timestamp__" {
```
popematt, line 209: Can you add a `const` for `"$__ion_rs_timestamp__"` and `"$__ion_rs_decimal__"`? (See [this SO answer](https://softwareengineering.stackexchange.com/a/365344) if you want more information about why magic strings are generally not a good idea.)

src/serde/ser.rs
desaikd: Renamed as this is used by both binary and text as a flag to indicate serde to treat the serialization or deserialization in terms of Ion format since timestamp and decimal representation in Ion differs from serde.

src/serde/mod.rs
desaikd: Yes, you can optional field and that will be considered as `null` in terms of Ion representation

src/serde/mod.rs
```diff
@@ -0,0 +1,185 @@
+//! # Serialization and deserialization of Ion data
+//!
+//! This module offers APIs for serialization of Rust data structures into Ion data and deserialization of Ion data into Rust data structures.
+//! The APIs uses serde framework for serialization and deserialization. See the Serde website <https://serde.rs/> for additional documentation and usage examples.
+//! This feature doesn't support [Ion annotations] for serialization and deserialization.
```
popematt, line 5: How are enum variants handled—or is that not supported yet? E.g.:
```rust
#[derive(Serialize, Deserialize)]
enum IpAddr {
    V4(String),
    V6(String),
}
```

What about tuple structs? Does this work? Is the serialized representation a s-expression?
```rust
#[derive(Serialize, Deserialize)]
struct Point(i32, i32);
```

src/serde/decimal.rs
```diff
@@ -0,0 +1,79 @@
...
+                    type Value = Decimal;
+
+                    fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+                        formatter.write_str("an Ion Decimal")
+                    }
+
+                    fn visit_str<E>(self, s: &str) -> Result<Self::Value, E>
+                    where
+                        E: Error,
+                    {
```
zslayton, line 53: I expected `DecimalVisitor`'s implementation of `de::Visitor` to override the definition of [`visit_newtype_struct`](https://docs.rs/serde/latest/serde/de/trait.Visitor.html#method.visit_newtype_struct) instead of `visit_str`. 

See [this snippet from a `serde` GH issue](https://github.com/serde-rs/serde/issues/2106#issue-1031792899).

zslayton (commented):
This is a partial review. There are a couple of questions I'd like to talk through before I review the rest.

popematt:
One more thing I want to mention—I've asked a lot of questions about what is and is not supported. If S-Expressions and annotations are not supported, that's fine because this is still an experimental feature. I just want the documentation to mention what is and is not supported. Please don't take my questions to mean that these things need to be implemented in the current PR.

Same thing with catching early returns and unwind panics to reset the thread-local variable–if you want to create an issue for it and punt that to a later PR, that's fine with me.

src/serde/de.rs
```diff
@@ -0,0 +1,883 @@
...
+                IonType::Blob | IonType::Clob => self.deserialize_bytes(visitor),
+                IonType::String | IonType::Symbol => self.deserialize_string(visitor),
+                IonType::Float => self.deserialize_f64(visitor),
+                IonType::Int => self.deserialize_i64(visitor),
+                IonType::Decimal => self.deserialize_newtype_struct(ION_DECIMAL, visitor),
+                IonType::Bool => self.deserialize_bool(visitor),
+                IonType::List => self.deserialize_seq(visitor),
+                IonType::Struct => self.deserialize_struct("", &[], visitor),
+                IonType::Timestamp => self.deserialize_newtype_struct(ION_TIMESTAMP, visitor),
+                _ => IonResult::decoding_error("unexpected ion type".to_string()),
```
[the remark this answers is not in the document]
desaikd: This only lacks SExpressions at this point. So unexpected type is just SExp (I will modify the error message). Also as per the offline discussion, Serialization for sexpression requires to handle serialization of an `Element` as those are the values we get when iterating over an sexpression or we might need a trick similar to `newtype_struct` for this. For now, SExpressions are not supported but we will work on adding it in future after investigating on a better approach for it.

src/serde/de.rs
```diff
@@ -0,0 +1,883 @@
...
+        self.reader.next()?;
+        result
+    }
+
+    fn deserialize_option<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        if self.reader.is_null() {
+            self.reader.next()?;
```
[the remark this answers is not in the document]
desaikd: Yes, for `None` we have a null value so we juts do next here but for `Some` value we will go to the particualr type's visitor where it will do `read_next`

src/serde/de.rs
```diff
@@ -0,0 +1,883 @@
...
+
+    fn deserialize_newtype_struct<V>(
+        self,
+        name: &'static str,
+        visitor: V,
+    ) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        if name == ION_TIMESTAMP {
```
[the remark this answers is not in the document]
desaikd: Removed with new changes.

src/serde/decimal.rs
desaikd: Removed `visit_str` from both decimal and timestamp deserializers as now we are using [transmute_copy](https://doc.rust-lang.org/std/mem/fn.transmute_copy.html) to treat a timestamp/decimal as `V::Value` which is visitor's value data type as defined in deserialization methods.

desaikd:
> One more thing I want to mention—I've asked a lot of questions about what is and is not supported. If S-Expressions and annotations are not supported, that's fine because this is still an experimental feature. I just want the documentation to mention what is and is not supported. Please don't take my questions to mean that these things need to be implemented in the current PR.
> 
> Same thing with catching early returns and unwind panics to reset the thread-local variable–if you want to create an issue for it and punt that to a later PR, that's fine with me.

I have added some documentation comments for what is supported and what is not supported here. `SExpressions` and `annotations` are not supported currently.

src/serde/timestamp.rs
```diff
@@ -0,0 +1,130 @@
...
+                deserializer.deserialize_newtype_struct("$__ion_rs_timestamp__", TimestampVisitor)
+            } else {
+                let string_rep = String::deserialize(deserializer)?;
+                let fixed_date: DateTime<FixedOffset> =
+                    match DateTime::parse_from_rfc3339(string_rep.as_str()) {
+                        Ok(data) => Ok(data),
+                        Err(e) => Err(de::Error::custom(e.to_string())),
+                    }?;
+                Ok(Timestamp::from(fixed_date))
+            }
```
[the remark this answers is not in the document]
desaikd: This has been removed with the new change with removal of `SERDE_AS_ION` flag.

src/serde/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+//!
+//! For deserialization `from_ion` API is provided through this module.
+//!
+//! ## Mapping of Ion data types to Rust types
+//!
+//!| Ion data type | Rust data structure |
+//!|---------------|---------------------|
+//!| int           | u64, i64, u32, i32, u16, i16, u8, i8  |
+//!| string        | String               |
+//!| struct        | struct              |
```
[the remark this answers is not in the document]
desaikd: added all types in this table for documentation

src/serde/de.rs
```diff
@@ -0,0 +1,931 @@
...
+
+    fn deserialize_i8<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        let result = visitor.visit_i8(
+            self.reader
+                .read_i64()
+                .map(i8::try_from)?
+                .map_err(|e| IonError::decoding_error(e.to_string()))?,
```
zslayton, line 100: For each of these, the `e.to_string()` makes something not super informative for the user: `Err(TryFromIntError(()))`. ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c040d5c87ffc358953831e6830261eca))

I suggest adding your own text:
```
.map_err(|_| IonError::decoding_error(found an integer was out of bounds for an `i8`"))
```

src/serde/de.rs
```diff
@@ -0,0 +1,931 @@
...
+    {
+        if name == TUNNELED_TIMESTAMP_TYPE_NAME {
+            let timestamp = self.reader.read_timestamp()?;
+            assert_eq!(
+                std::mem::size_of::<V::Value>(),
+                std::mem::size_of::<Timestamp>()
+            );
+            // compiler doesn't understand that the generic Timestamp here is actually V::Value here
+            let visitor_value =
+                unsafe { std::mem::transmute_copy::<Timestamp, V::Value>(&timestamp) };
```
zslayton, line 315: [By convention](https://std-dev-guide.rust-lang.org/policy/safety-comments.html), any use of `unsafe` is accompanied by a `// SAFETY:` comment that explains why you're using `unsafe` (which you've got), why what you're doing is actually safe even though the compiler disagrees, and what other invariants/assumptions others need to ensure for it to keep working as intended.

src/serde/mod.rs
```diff
@@ -0,0 +1,271 @@
...
+//!| byte-array | blob |
+//!| option | None - null, Some - based on other mappings |
+//!| unit, unit_struct | null |
+//!| seq, tuple, tuple_struct, tuple_variant | list |
+//!| newtype_struct ,newtype_variant, map, struct, struct_variant | struct |
+//!
+//! _Note: Since the serde framework doesn't support [Ion decimal] and [Ion timestamp] types, distinct serialization and deserialization of these types are defined in this module.
+//! It uses `newtype_struct` with `$__ion_rs_decimal__` and `$__ion_rs_timestamp__` as struct names from [serde data model],
+//! to indicate serde framework to use Ion's implementation of decimal and timestamp serialization and deserialization.
+//! If one wants to use [chrono::DateTime], it needs to be tagged with `#[serde_as(as = crate::Timestamp)]`._
```
zslayton, line 47: Do we still support `DateTime`? Since it adds a public-facing dependency, I think it should be an opt-in feature flag when we add it.

src/serde/mod.rs
```diff
@@ -0,0 +1,271 @@
...
+//!
+//! fn main() -> IonResult<()> {
+//!     // data structure for representing address
+//!     let address = Address {
+//!         street: "10 Downing Street".to_owned(),
+//!         city: "London".to_owned(),
+//!     };
+//!
+//!     // serialize it to an Ion data
+//!     let ion = to_string(&address)?;
```
zslayton, line 69: If it's meant to be a free-standing function, how about a name like `to_ion_text()`? Is there a way to pass in a writer of your own so you can configure it for `pretty`/`lines`/whatever?

src/serde/mod.rs
```diff
@@ -0,0 +1,271 @@
...
+//!
+//!| Serde data type | Ion data type |
+//!|---------------|---------------------|
+//!| u64, i64, u32, i32, u16, i16, u8, i8 | int |
+//!| char, string, unit_variant | string |
+//!| byte-array | blob |
+//!| option | None - null, Some - based on other mappings |
+//!| unit, unit_struct | null |
+//!| seq, tuple, tuple_struct, tuple_variant | list |
+//!| newtype_struct ,newtype_variant, map, struct, struct_variant | struct |
```
zslayton, line 42:
```suggestion
//!| newtype_struct, newtype_variant, map, struct, struct_variant | struct |
```

src/serde/mod.rs
```diff
@@ -0,0 +1,271 @@
...
+//! ## Example of serialization and deserialization for Decimal
+//!```
+//! use serde::{Deserialize, Serialize};
+//! use ion_rs::Decimal;
+//! use ion_rs::IonResult;
+//! use ion_rs::serde::from_ion;
+//!
+//! #[derive(Serialize, Deserialize)]
+//! struct Product {
+//!     name: String,
```
zslayton, line 159: It'd be good to have an example that shows deserializing strings and byte arrays to a `&'a str` and `&'a [u8]` respectively.

src/serde/ser.rs
```diff
@@ -0,0 +1,621 @@
...
+use crate::types::Int;
+use crate::{
+    BinaryWriterBuilder, Decimal, IonError, IonResult, IonType, TextKind, TextWriterBuilder,
+    Timestamp,
+};
+use serde::ser::Impossible;
+use serde::{ser, Serialize};
+
+/// Serialize an object into pretty formatted Ion text
+pub fn to_pretty<T>(value: &T) -> IonResult<String>
```
zslayton, line 16: We have a `Format` enum you could take:
```rust
to_ion(Format::Text(TextKind::Pretty))?;
to_ion(Format::Text(TextKind::Lines))?;
to_ion(Format::Text(TextKind::Compact))?;
to_ion(Format::Binary)?;
```
Of course, that doesn't capture the upcoming availability of Ion 1.1...

I think we need a write config builder API that we can use for things like these. See #618.

src/serde/ser.rs
```diff
@@ -0,0 +1,621 @@
...
+    type SerializeMap = Impossible<String, IonError>;
+    type SerializeStruct = Impossible<String, IonError>;
+    type SerializeStructVariant = Impossible<String, IonError>;
+
+    fn serialize_bool(self, _v: bool) -> Result<Self::Ok, Self::Error> {
+        Err(key_must_be_a_string())
+    }
+
+    fn serialize_i8(self, v: i8) -> Result<Self::Ok, Self::Error> {
+        Ok(v.to_string())
```
zslayton, line 495: I wonder if there's a way to avoid allocating a string for every key?

src/serde/ser.rs
```diff
@@ -0,0 +1,621 @@
...
+    }
+
+    /// Serialize all integer types using the `Integer` intermediary type.
+    fn serialize_u32(self, v: u32) -> Result<Self::Ok, Self::Error> {
+        self.writer.write_int(&Int::from(v))
+    }
+
+    /// Serialize all integer types using the `Integer` intermediary type.
+    fn serialize_i64(self, v: i64) -> Result<Self::Ok, Self::Error> {
+        self.writer.write_int(&Int::from(v))
```
zslayton, line 133: For all of these that fit in an `i64` (which is everything but `u64`) you can use `self.writer.write_i64()` to avoid making an intermediate `Int`.

src/serde/ser.rs
```diff
@@ -0,0 +1,621 @@
...
+    serializer.writer.flush()?;
+    drop(serializer);
+
+    Ok(cursor.get_ref().clone())
+}
+
+/// Implements a standard serializer for Ion
+pub struct Serializer<E> {
+    pub(crate) writer: E,
+}
```
zslayton, line 79: Having a way to pass in a writer of my own would be helpful.

src/serde/de.rs
```diff
@@ -0,0 +1,931 @@
...
+                IonType::Decimal => {
+                    self.deserialize_newtype_struct(TUNNELED_DECIMAL_TYPE_NAME, visitor)
+                }
+                IonType::Bool => self.deserialize_bool(visitor),
+                IonType::List => self.deserialize_seq(visitor),
+                IonType::Struct => self.deserialize_struct("", &[], visitor),
+                IonType::Timestamp => {
+                    self.deserialize_newtype_struct(TUNNELED_TIMESTAMP_TYPE_NAME, visitor)
+                }
+                _ => IonResult::decoding_error("unexpected ion type"),
```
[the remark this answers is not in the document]
desaikd: I can add deserialize_seq for SExp but we don't support serialization for SExp. This might be confusing for users which is why I wanted to keep SExp as not supported for both serialization and deserialization. Thoughts?

src/serde/mod.rs
desaikd: currently we have different methods for to_string, to_pretty, ot_binary. but yes, I can modify it to be a single method with configuration to the writer.

src/serde/de.rs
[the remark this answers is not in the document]
popematt: I am not sure that we should deserialize lists and sexps as if they are interchangeable. I am sure that we should eventually have support for deserialization _and_ serialization of s-expressions. (Possibly configurable using `#[ion(sexp)]` or something like that on a struct or field.) I think that the configuration for serializing as a sexp might affect our decisions about how to deserialize a sexp. Therefore, I am in favor of punting both serialization and deserialization of sexps. Let's create an issue for it, and just make sure it gets resolved before we remove the "experimental" status of this feature.

src/serde/ser.rs
desaikd: https://github.com/amazon-ion/ion-rust/issues/646

src/serde/ser.rs
desaikd: https://github.com/amazon-ion/ion-rust/issues/646

src/serde/mod.rs
desaikd: https://github.com/amazon-ion/ion-rust/issues/646

src/serde/de.rs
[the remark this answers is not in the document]
desaikd: https://github.com/amazon-ion/ion-rust/issues/646

src/serde/de.rs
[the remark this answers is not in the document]
zslayton: `SExp` is a materialized s-expression, but I in this context we're detecting an s-expression in the input bytes. If the user's data type is
```rust
struct MyData {
   numbers: Vec<i64>,
}
```
and the deserializer sees:
```
{
  numbers: (1 2 3 4 5)
}
```
I think it would be reasonable for it to just treat it the same way it would treat:
```
{
  numbers: [1, 2, 3, 4, 5]
}
```

src/serde/de.rs
```diff
@@ -321,7 +308,9 @@ where
                 std::mem::size_of::<V::Value>(),
                 std::mem::size_of::<Decimal>()
             );
+            // # Safety
             // compiler doesn't understand that the generic Decimal here is actually V::Value here
+            // the assert statement above that compares sizes of both the Decimal and V::Value ensures the conversion performed below is safe
```
zslayton, line 313: I'd avoid saying that the check ensures the conversion is safe; we don't have a way to know for a fact that `V::Value` is actually a `Decimal`.

```suggestion
            // The assert statement above that compares the sizes of the Decimal and V::Value types
            // as a safeguard. If a project actually had a newtype struct with the unlikely name of 
            // `$__ion_rs_decimal__`, it would be possible for `V::Value` to be that type instead.
            // We cannot confirm the actual type, so we confirm the expected size as a proxy.
```

zslayton (approved):
Take a look at my comment about s-expressions. It may be trivial enough to squeeze in, but I'm fine with adding it in a follow-on PR too if that's easier.

popematt (approved):
🚢

desaikd:
> Take a look at my comment about s-expressions. It may be trivial enough to squeeze in, but I'm fine with adding it in a follow-on PR too if that's easier.

I will follow up on that with separate PR, here's the issue for it: #646

Issue #635, opened 2023-08-29
Incorporate feedback from `LazyRawTextReader` PRs


zslayton:
The `LazyRawTextReader` implementation was broken into over a dozen PRs. To keep the merge process simple, outstanding minor requests from reviewers were deferred until after the queue was merged. Once #634 is fixed, I'll open a new PR that addresses those requests.

PR #638, opened 2023-09-01
Adds lazy reader support for reading clobs

### LINKED ISSUE #634: Implement lazy reader support for clobs
The text implementation does not yet have support for clobs.

zslayton:
Adds `LazyRawTextReader` support for matching and reading clobs.

Fixes #634. 


src/lazy/text/buffer.rs
```diff
@@ -1169,6 +1175,13 @@ impl<'data> TextBufferView<'data> {
                 contains_escaped_chars = true;
                 continue;
             }
+            if *byte == b'\r' {
+                // If the text contains an unescaped carriage return, we may need to normalize it.
+                // In some narrow cases, setting this flag to true may result in a sanitization buffer
+                // being allocated when it isn't strictly necessary.
+                contains_escaped_chars = true;
+                continue;
+            }
```
zslayton, line 1184: 🗺️ In long-form clobs and long-form strings, we need to normalize unescaped `\r` and `\r\n` to `\n`. This throws the naming off a bit; `contains_escapes` should really be something like `requires_substitutions`. However, I think `escapes` is a more obvious/suggestive name. Open to input here; I left it as-is because a consistent rename across usages/modules would touch a lot of lines and I'd rather do it in another PR.

src/lazy/text/matched.rs
```diff
@@ -338,7 +338,14 @@ impl MatchedString {
...
         // There are no escaped characters, so we can just validate the string in-place.
         let mut sanitized = Vec::with_capacity(matched_input.len());
-        escape_text(body, &mut sanitized)?;
+        decode_text_containing_escapes(
+            body,
+            &mut sanitized,
+            // Normalize newlines
+            true,
+            // Support unicode escapes
+            true,
```
zslayton, line 347: 🗺️ I considered enums for these two `bool`s to make them self-documenting, but as they're not part of the public API I decided to just comment the handful of places where this method is called.

zslayton (commented):
🗺️ PR tour

popematt (approved):
Sorry, I meant to approve this because none of my latest comments are things that would block the PR.

src/lazy/text/matched.rs
```diff
@@ -856,8 +965,92 @@ impl MatchedBlob {
     }
 }
 
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum MatchedClob {
+    Short,
+    Long,
```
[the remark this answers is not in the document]
zslayton: There's a narrow case that benefits: single-segment clobs that only contain ASCII. Every other case requires a sanitization/decoding buffer anyway. I concluded that I'd wait to see if anyone actually uses clobs outside of `ion-tests` before worrying about optimizing it further.

zslayton:
> Sorry, I meant to approve this because none of my latest comments are things that would block the PR.

Thanks, I've got another PR out that depends on this one (#639), so I'll go ahead and merge this and address the comments as part of #635.

Issue #636, opened 2023-08-29
Add `ion-tests` integration for the LazyReader


zslayton:
Add the `LazyReader` to the `ion-tests` integration to demonstrate spec conformance.

PR #639, opened 2023-09-06
Adds `ion-tests` integration for the lazy reader

zslayton:
Adds `ion-tests` integration for the lazy reader. Also addresses a number bugs surfaced by the conformance tests.

Fixes issue #636. 


src/lazy/text/buffer.rs
```diff
@@ -1009,7 +1060,7 @@ impl<'data> TextBufferView<'data> {
     /// Returns a matched buffer and a boolean indicating whether any escaped characters were
     /// found in the short string.
     pub(crate) fn match_short_string_body(self) -> IonParseResult<'data, (Self, bool)> {
-        Self::match_text_until_unescaped(self, b'\"')
+        Self::match_text_until_unescaped(self, b'\"', false)
```
zslayton, line 1063: 🗺️ Callers to this function must now specify whether unescaped newlines are permitted (long-form string literals) or not (short-form string literals).

src/lazy/text/buffer.rs
```diff
@@ -1156,41 +1218,76 @@ impl<'data> TextBufferView<'data> {
...
+                // Other escape sequences that are followed by more than one byte (e.g. \u and \U)
+                // are always followed by ASCII letters, which aren't used as delimiters.
                 contains_escaped_chars = true;
+                // Peek at the next two bytes to see if this is a \r\n
+                let next_two_bytes = self.bytes().get(index + 1..index + 3);
+                let bytes_to_skip = if next_two_bytes == Some(&[b'\r', b'\n']) {
+                    2
+                } else {
+                    1
+                };
```
zslayton, line 1247: 🗺️ Took a while to track this one down.

src/lazy/text/buffer.rs
```diff
@@ -1156,41 +1218,76 @@ impl<'data> TextBufferView<'data> {
...
+                } else {
+                    self.validate_string_control_character(byte, index, allow_unescaped_newlines)?;
+                }
+            }
         }
         Err(nom::Err::Incomplete(Needed::Unknown))
     }
 
+    #[cold]
+    fn validate_string_control_character(
```
zslayton, line 1272: 🗺️ This is broken into its own `#[cold]` method because unescaped control characters in a string aren't that common, and the code to build up the errors adds up.

src/lazy/text/buffer.rs
```diff
@@ -1759,7 +1859,7 @@ mod tests {
         /// contents of the input are considered a complete token.
         fn new(input: &str) -> Self {
             MatchTest {
-                input: format!("{input}\n0"), // add whitespace and a trailing value
```
zslayton, line 1762: 🗺️ Now that we're using `complete_` parsers where necessary, we no longer have to append extra values to our test data.

src/lazy/text/matched.rs
```diff
@@ -128,7 +150,11 @@ impl MatchedInt {
             }
         };
 
-        Ok(int)
+        if self.is_negative {
+            Ok(-int)
+        } else {
+            Ok(int)
+        }
```
zslayton, line 157: 🗺️ I had tests to match negative ints, but didn't have tests to read negative ints.

src/lazy/text/matched.rs
```diff
@@ -178,7 +204,7 @@ pub(crate) struct MatchedDecimal {
     is_negative: bool,
     digits_offset: u16,
     digits_length: u16,
-    trailing_digits_length: u16,
+    num_trailing_digits: u16,
```
zslayton, line 207: 🗺️ Now that underscores have been filtered out of the trailing digits, this has been renamed to reflect that this is a digit count, not a length count.

src/lazy/text/matched.rs
```diff
@@ -412,7 +438,8 @@ impl MatchedString {
             // Support Unicode escapes
             true,
         )?;
-        let text = String::from_utf8(sanitized).unwrap();
+        let text = String::from_utf8(sanitized);
+        let text = text.unwrap();
```
zslayton, line 442: 🗺️ This was a debugging leftover; I'll fix it before merging.

src/lazy/text/matched.rs
```diff
@@ -588,38 +636,28 @@ fn decode_hex_digits_escape<'data>(
...
+    // We just confirmed all of the digits are ASCII hex digits, so this step cannot fail.
     let hex_digits = std::str::from_utf8(hex_digit_bytes).unwrap();
-    // If this was a '\x' escape, we cannot interpret the hex digits as a Unicode scalar. We treat
-    // it as a byte literal instead.
-    if num_digits == 2 {
-        let byte = u8::from_str_radix(hex_digits, 16).unwrap();
-        sanitized.push(byte);
+
+    if !support_unicode_escapes {
+        // Inside a clob, \x is a byte literal, not a Unicode code point.
```
zslayton, line 643: 🗺️ This was surprising/confusing to me, but here we are.

src/lazy/text/matched.rs
```diff
@@ -634,6 +672,57 @@ fn decode_hex_digits_escape<'data>(
     Ok(remaining_input)
 }
 
+/// Reads another escaped code point from the buffer, treating it as the low surrogate to be paired
+/// with the specified high surrogate. Appends the UTF-8 encoding of the resulting Unicode scalar
+/// to `sanitized` and returns the remaining text in the buffer.
+fn complete_surrogate_pair<'data>(
```
zslayton, line 678: 🗺️ I added support for these to pass the `ion-tests` around strings/symbols. See amazon-ion/ion-tests#637

src/lazy/text/matched.rs
```diff
@@ -1050,9 +1139,48 @@ impl MatchedClob {
...
+            let actual = matched.read(buffer).unwrap();
+            assert_eq!(
+                actual, expected,
+                "Actual didn't match expected for input '{}'.\n{:?}\n!=\n{:?}",
+                data, actual, expected
+            );
+        }
+
+        let tests = [
+            ("-5", Int::from(-5)),
```
zslayton, line 1165: 🗺️ Added int reading tests that could have caught the bug that discarded the sign information.

src/lazy/text/matched.rs
```diff
@@ -1259,14 +1396,11 @@ mod tests {
             (r"'''he''' '''llo'''", "hello"),
             (r#""😎🙂🙃""#, "😎🙂🙃"),
             (r"'''😎🙂''' '''🙃'''", "😎🙂🙃"),
-            // The below bytes are the UTF-8 encoding of Unicode code points: U+2764 U+FE0F
-            (r#""\xe2\x9d\xa4\xef\xb8\x8f""#, "❤️"),
-            (r"'''\xe2\x9d\xa4\xef\xb8\x8f'''", "❤️"),
```
zslayton, line 1264: 🗺️ In `clob`s, these escapes produce the correct UTF-8 data. In `string`s, they do not. Because the code points needed to represent this heart emoji are two bytes each, they cannot be represented via `\x` escapes, each of which can only represent 1-byte unicode scalar values.

src/lazy/text/raw/reader.rs
```diff
@@ -55,6 +55,14 @@ impl<'data> LazyRawTextReader<'data> {
...
             .match_top_level_item()
             .with_context("reading a top-level value", buffer_after_whitespace)?;
+
+        if let RawStreamItem::VersionMarker(major, minor) = matched {
+            if (major, minor) != (1, 0) {
+                return IonResult::decoding_error(format!(
+                    "Ion version {major}.{minor} is not supported"
+                ));
+            }
+        }
```
zslayton, line 65: 🗺️ There are `ion-tests` which expect the reader to reject future Ion versions.

src/lazy/text/raw/reader.rs
```diff
@@ -191,8 +199,8 @@ mod tests {
             // Second item
             2 /*comment before comma*/,
             // Third item
-            3
-        ]
+            3 ,]
```
zslayton, line 202: 🗺️ I added a trailing comma while debugging.

src/lazy/text/raw/sequence.rs
```diff
@@ -114,9 +115,17 @@ impl<'data> RawTextListIterator<'data> {
...
+            input_after_last
+                .match_optional_comments_and_whitespace()
+                .with_context("seeking the end of a list", input_after_last)?;
+        // Skip an optional comma and more whitespace
+        if input_after_ws.bytes().first() == Some(&b',') {
+            (input_after_ws, _) = input_after_ws
+                .slice_to_end(1)
+                .match_optional_comments_and_whitespace()
+                .with_context("skipping a list's trailing comma", input_after_ws)?;
+        }
```
zslayton, line 128: 🗺️ If a list had a trailing comma, this would break.

src/lazy/text/raw/struct.rs
```diff
@@ -39,9 +39,17 @@ impl<'data> RawTextStructIterator<'data> {
...
+            input_after_last
+                .match_optional_comments_and_whitespace()
+                .with_context("seeking the end of a struct", input_after_last)?;
+        // Skip an optional comma and more whitespace
+        if input_after_ws.bytes().first() == Some(&b',') {
+            (input_after_ws, _) = input_after_ws
+                .slice_to_end(1)
+                .match_optional_comments_and_whitespace()
+                .with_context("skipping a list's trailing comma", input_after_ws)?;
+        }
```
zslayton, line 52: 🗺️ If the struct had a trailing comma, this would break.

src/lazy/value_ref.rs
```diff
@@ -97,6 +97,15 @@ impl<'top, 'data, D: LazyDecoder<'data>> TryFrom<ValueRef<'top, 'data, D>> for V
...
 }
 
+impl<'top, 'data, D: LazyDecoder<'data>> TryFrom<ValueRef<'top, 'data, D>> for Element {
+    type Error = IonError;
+
+    fn try_from(value_ref: ValueRef<'top, 'data, D>) -> Result<Self, Self::Error> {
+        let value: Value = value_ref.try_into()?;
+        Ok(value.into())
+    }
+}
```
zslayton, line 107: 🗺️ This enables: `let element: Element = reader.next()?.read()?.try_into()?;`

zslayton (commented):
🗺️ PR tour

src/lazy/binary/immutable_buffer.rs
zslayton, line 1: 🗺️ Previously, the binary reader had a monolithic `peek_value` method that contained logic for reading fields, annotations, NOPs, and the values themselves. I refactored this so that each has its own code path with clearer invariants. This made it easier to track down a NOP-related buffer indexing problem.

src/lazy/reader.rs
zslayton, line 1: 🗺️ I tried to normalize the reader level names a bit more. There are three levels: `Raw`, `System`, and `Application`. Each level has three supported decoders: `Text`, `Binary`, or `Any`, where `Any` uses enum dispatch to abstract over `Text` or `Binary`.
* `LazyRawBinaryReader`, `LazyRawTextReader`, `LazyRawAnyReader`
* `LazySystemBinaryReader`, `LazySystemTextReader`, `LazySystemAnyReader`
* `LazyApplicationBinaryReader`, `LazyApplicationSystemTextReader`, `LazyApplicationSystemAnyReader`

Because users will almost always interact with the `Application` level, that is the default and type aliases are provided to allow the level to be omitted.

* `LazyBinaryReader`, `LazyTextReader`

Similarly, most users won't want to have to specify the format, so they can just use the `LazyReader` to get a `LazyApplicationAnyReader`.

zslayton (commented):
🗺️ Restoring PR tour comments that GitHub decided to remove/hide as "outdated."

src/lazy/reader.rs
```diff
@@ -55,11 +56,11 @@ use crate::{IonError, IonResult};
 ///# Ok(())
 ///# }
 /// ```
-pub struct LazyReader<'data, D: LazyDecoder<'data>> {
+pub struct LazyApplicationReader<'data, D: LazyDecoder<'data>> {
```
popematt, line 59: Doc comments need to be updated. It still refers to this as a "binary" reader.

src/lazy/text/buffer.rs
```diff
@@ -357,7 +366,7 @@ impl<'data> TextBufferView<'data> {
     /// input bytes where the field name is found, and the value.
     pub fn match_struct_field_name_and_value(
         self,
-    ) -> IonParseResult<'data, ((MatchedSymbol, Range<usize>), LazyRawTextValue<'data>)> {
+    ) -> IonParseResult<'data, ((MatchedFieldName, Range<usize>), LazyRawTextValue<'data>)> {
```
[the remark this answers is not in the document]
popematt:
```
{'''f''' '''o''' '''o''': '''b''' '''a''' '''r'''}
```
🤯

src/lazy/text/buffer.rs
```diff
@@ -701,7 +722,7 @@ impl<'data> TextBufferView<'data> {
 
     /// Matches the three parts of an int--its base, its sign, and its digits--without actually
     /// constructing an Int from them.
-    fn match_int(self) -> IonParseResult<'data, MatchedInt> {
+    pub fn match_int(self) -> IonParseResult<'data, MatchedInt> {
```
popematt, line 725: Just curious—why does this need to be `pub` now?

popematt (approved):
Overall it looks good. I do have one question though.

You've made a "Binary", "Text", and "Any" implementation for the "Raw", "System", and "Application" readers. Is there any reason that the text and binary distinction has to be pushed up that far? Would it be possible to create a `RawAnyReader` that is an enum of binary and text raw readers so that higher levels of readers can have just one implementation?

src/lazy/reader.rs
popematt: Oh, I missed this comment when I was reviewing the code. However, I think my question is still valid. Why do we need the encoding distinction at every level of reader?

zslayton:
> Overall it looks good. I do have one question though.
> 
> You've made a "Binary", "Text", and "Any" implementation for the "Raw", "System", and "Application" readers. Is there any reason that the text and binary distinction has to be pushed up that far? Would it be possible to create a `RawAnyReader` that is an enum of binary and text raw readers so that higher levels of readers can have just one implementation?

The old reader uses a `Box<dyn IonReader>` to abstract over the raw reader and format, and unfortunately it adds a really substantial amount of overhead. A quick test I threw together showed it was about 25% slower than using static dispatch with a single format. The `Any` readers use enum dispatch, and my assumption is that it will be far faster than `dyn` (branch predictors are great!), but I haven't quantified the performance impact yet. If it's a slowdown, users focused on performance will probably want the option to work with a format-specific reader.

Note that `LazyApplicationReader<D>` is `pub` no matter what, so we're really just talking about what type aliases to provide.

src/lazy/text/buffer.rs
zslayton: I wanted each of the methods that matched a full Ion type to be reusable from outside the buffer (but not outside the crate without more consideration).

PR #643, opened 2023-09-09
Stub out types for RawTextReader_1_1

zslayton:
This PR makes superficial changes that affect a large number of lines of code:
* It stubs out the types necessary to introduce a raw text Ion 1.1 reader, providing those stub types with minimal implementations that forward calls to the existing 1.0 types/implementations. These types will be replaced by actual 1.1 implementations in subsequent PRs.
* It splits the `TextEncoding` type into `TextEncoding_1_0` and `TextEncoding_1_1`
* It renames `BinaryEncoding` to `BinaryEncoding_1_0` to make way for an eventual `BinaryEncoding_1_1`.

This PR contains no logic changes.

### DIFF
--- a/src/lazy/encoding.rs
+++ b/src/lazy/encoding.rs
@@ -1,39 +1,53 @@
+#![allow(non_camel_case_types)]
 use crate::lazy::binary::raw::annotations_iterator::RawBinaryAnnotationsIterator;
 use crate::lazy::binary::raw::r#struct::LazyRawBinaryStruct;
 use crate::lazy::binary::raw::reader::LazyRawBinaryReader;
 use crate::lazy::binary::raw::sequence::{LazyRawBinaryList, LazyRawBinarySExp};
 use crate::lazy::binary::raw::value::LazyRawBinaryValue;
-use crate::lazy::decoder::LazyDecoder;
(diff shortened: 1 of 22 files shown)


src/lazy/encoding.rs
```diff
@@ -1,39 +1,53 @@
...
-impl<'data> LazyDecoder<'data> for BinaryEncoding {
+impl<'data> LazyDecoder<'data> for BinaryEncoding_1_0 {
     type Reader = LazyRawBinaryReader<'data>;
     type Value = LazyRawBinaryValue<'data>;
     type SExp = LazyRawBinarySExp<'data>;
     type List = LazyRawBinaryList<'data>;
     type Struct = LazyRawBinaryStruct<'data>;
     type AnnotationsIterator = RawBinaryAnnotationsIterator<'data>;
+    // Macros are not supported in Ion 1.0
+    type MacroInvocation = ();
```
popematt, line 36: question—instead of `()`, could you use `!` here?

popematt (approved):

src/lazy/encoding.rs
zslayton: I've done a version of that in the follow on PR in working on 👍

Issue #653, opened 2023-10-05
`TextViewBuffer` does not correctly parse macros in structs


popematt:
Here are the inputs that are currently failing to parse. I did some investigation, and it seems like it is parsing the content just fine, but it is failing somewhere in the raw struct reader/iterator.
```rust
    use rstest::rstest;
    #[rstest]
    #[case::e_exp_in_struct("{(:foo)}")]
    #[case::e_exp_in_struct_with_space_before("{ (:foo)}")]                       // FAILS
    #[case::e_exp_in_struct_with_space_after("{(:foo) }")]
    #[case::e_exp_in_struct_field("{a:(:foo)}")]                                  // FAILS
    #[case::e_exp_in_struct_field_with_comma("{a:(:foo),}")]                      // FAILS
    #[case::e_exp_in_struct_field_with_comma_and_second_field("{a:(:foo), b:2}")]
    #[case::e_exp_in_struct_field_with_space_before("{ a:(:foo)}")]               // FAILS
    #[case::e_exp_in_struct_field_with_space_after("{a:(:foo) }")]                // FAILS
    fn test_match_macro_invocation_in_context(#[case] input: &str) {
        MatchTest::new(input).expect_match(match_length(TextBufferView::match_top_level_item_1_1));
    }
```

zslayton:
This is working now.

PR #654, opened 2023-10-05
Fixes parsing of macro invocations in Lists and SExps and some cases …

popematt:
…in Structs

**Issue #, if available:**

None

**Description of changes:**

As I was starting to write test cases in `ion-tests` for macros, I discovered that macros were working when they appeared at the top level of an Ion stream, or when they were nested no more than one layer deep. I've fixed that, though some issues still remain with macros in structs (see #653).

See specific notes in 🗺️ comments.

### DIFF
--- a/src/lazy/text/buffer.rs
+++ b/src/lazy/text/buffer.rs
@@ -32,7 +32,7 @@ use crate::lazy::text::matched::{
 };
 use crate::lazy::text::parse_result::{InvalidInputError, IonParseError};
 use crate::lazy::text::parse_result::{IonMatchResult, IonParseResult};
-use crate::lazy::text::raw::r#struct::LazyRawTextField_1_0;
+use crate::lazy::text::raw::r#struct::{LazyRawTextField_1_0, RawTextStructIterator_1_0};
 use crate::lazy::text::raw::sequence::{RawTextListIterator_1_0, RawTextSExpIterator_1_0};
 use crate::lazy::text::raw::v1_1::reader::{
     EncodedTextMacroInvocation, MacroIdRef, RawTextListIterator_1_1, RawTextMacroInvocation,
@@ -117,6 +117,7 @@ pub(crate) struct TextBufferView<'a> {
     //                          offset: 6
     data: &'a [u8],
     offset: usize,
+    // encoding: AnyTextEncoding,
 }
 
 impl<'data> TextBufferView<'data> {
@@ -334,15 +335,31 @@ impl<'data> TextBufferView<'data> {
     pub fn match_sexp_value_1_1(
         self,
     ) -> IonParseResult<'data, Option<LazyRawValueExpr<'data, TextEncoding_1_1>>> {
-        alt((
-            whitespace_and_then(
-                Self::match_e_expression
-                    .map(|matched| Some(RawValueExpr::MacroInvocation(matched))),
-            ),
-            Self::match_sexp_value.map(|maybe_matched| {
+        whitespace_and_then(alt((
+            Self::match_e_expression.map(|matched| Some(RawValueExpr::MacroInvocation(matched))),
+            value(None, tag(")")),
+            pair(
+                opt(Self::match_annotations),
+                // We need the s-expression parser to recognize the input `--3` as the operator `--` and the
+                // int `3` while recognizing the input `-3` as the int `-3`. If `match_operator` runs before
+                // `match_value`, it will consume the sign (`-`) of negative number values, treating
+                // `-3` as an operator (`-`) and an int (`3`). Thus, we run `match_value` first.
+                whitespace_and_then(alt((Self::match_value_1_1, Self::match_operator))),
+            )
+            .map(|(maybe_annotations, mut value)| {
+                if let Some(annotations) = maybe_annotations {
+                    value.encoded_value = value
+                        .encoded_value
+                        .with_annotations_sequence(annotations.offset(), annotations.len());
+                    // Rewind the value's input to include the annotations sequence.
+                    value.input = self.slice_to_end(annotations.offset() - self.offset());
+                }
+                Some(value)
+            })
+            .map(|maybe_matched| {
                 maybe_matched.map(|matched| RawValueExpr::ValueLiteral(matched.into()))
             }),
-        ))
+        )))
         .parse(self)
     }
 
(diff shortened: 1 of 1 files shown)


src/lazy/text/buffer.rs
```diff
@@ -334,15 +335,31 @@ impl<'data> TextBufferView<'data> {
...
-                Self::match_e_expression
-                    .map(|matched| Some(RawValueExpr::MacroInvocation(matched))),
-            ),
-            Self::match_sexp_value.map(|maybe_matched| {
+        whitespace_and_then(alt((
+            Self::match_e_expression.map(|matched| Some(RawValueExpr::MacroInvocation(matched))),
+            value(None, tag(")")),
+            pair(
+                opt(Self::match_annotations),
+                // We need the s-expression parser to recognize the input `--3` as the operator `--` and the
```
popematt, line 343: 🗺️ Because this delegated to `match_sexp_value` (implied `_1_0`), it would use the Ion 1.0 matchers for matching anything that is not an E-Expression inside a S-Expression, which caused this to happen: 
```
outer::(     // Handled by match_sexp_1_1
  abc        // Handled by match_value_1_0
  (:foo)     // Handled by match_e_expression
  inner::(   // Handled by match_sexp_1_0
    def      // Handled by match_value_1_0
    (:bar)   // Invalid (for match_sexp_1_0)
  )
)
```
The solution was to copy/paste the content from `match_sexp_value`, changing it to use `match_value_1_1` instead of `match_value_1_0`.

src/lazy/text/buffer.rs
```diff
@@ -117,6 +117,7 @@ pub(crate) struct TextBufferView<'a> {
     //                          offset: 6
     data: &'a [u8],
     offset: usize,
+    // encoding: AnyTextEncoding,
```
popematt, line 120: Oops.
```suggestion
```

src/lazy/text/buffer.rs
```diff
@@ -790,7 +807,7 @@ impl<'data> TextBufferView<'data> {
                 },
             ),
             map(
-                match_and_length(Self::match_struct),
+                match_and_length(Self::match_struct_1_1),
```
popematt, line 810: 🗺️ I added `match_struct_1_1`, so it got updated here, inside `match_value_1_1`.

src/lazy/text/buffer.rs
```diff
@@ -931,18 +948,22 @@ impl<'data> TextBufferView<'data> {
...
-            Self::match_list_value.map(|maybe_matched| {
+        whitespace_and_then(alt((
+            terminated(
+                Self::match_e_expression,
+                Self::match_delimiter_after_list_value,
+            )
+            .map(|matched| Some(RawValueExpr::MacroInvocation(matched))),
+            value(None, tag("]")),
+            terminated(
+                Self::match_annotated_value_1_1.map(Some),
```
popematt, line 959: 🗺️ Same problem here as in `match_sexp_value_1_1`. Fixed by copy/pasting the content of `match_list_value`, and altering this line to point to `match_annotated_value_1_1`.

src/lazy/text/buffer.rs
```diff
@@ -998,7 +1019,7 @@ impl<'data> TextBufferView<'data> {
         }
         // Scan ahead to find the end of this struct.
         let struct_body = self.slice_to_end(1);
-        let struct_iter = RawTextStructIterator_1_1::new(struct_body);
+        let struct_iter = RawTextStructIterator_1_0::new(struct_body);
```
popematt, line 1022: 🗺️ Because `match_ion_struct` was using `RawTextStructIterator_1_1`, it was theoretically possible that Ion 1.0 would accept E-Expressions as long as they were found inside a struct.

src/lazy/text/buffer.rs
```diff
@@ -1021,6 +1042,37 @@ impl<'data> TextBufferView<'data> {
         Ok((remaining, matched))
     }
 
+    pub fn match_struct_1_1(self) -> IonMatchResult<'data> {
```
popematt, line 1045: 🗺️ This is basically a copy/paste of `match_struct`, but it uses `RawTextStructIterator_1_1`.

src/lazy/text/buffer.rs
```diff
@@ -2276,8 +2328,9 @@ mod tests {
             P: Parser<TextBufferView<'data>, O, IonParseError<'data>>,
         {
             let result = self.try_match(parser);
-            let (_remaining, match_length) = result
-                .unwrap_or_else(|_| panic!("Unexpected parse fail for input '{}'", self.input));
+            let (_remaining, match_length) = result.unwrap_or_else(|e| {
```
popematt, line 2331: 🗺️ This is a developer experience improvement. It causes a test failure to print the whole error, not just "Unexpected parse fail for input".

src/lazy/text/buffer.rs
```diff
@@ -2709,6 +2762,56 @@ mod tests {
             mismatch_sexp(input);
         }
     }
+    #[test]
+    fn test_match_sexp_1_1() {
```
popematt, line 2766: 🗺️ I added some tests for Ion 1.1 SExps and Lists, as well as some test cases for E-Expressions that are nested inside other things. This is far from comprehensive, but it's at least an improvement.

popematt (commented):
I think the long term solution needs to be some way of extracting the matcher logic from the `TextBufferView` and then having the `TextBufferView` keep track of what version it is currently reading. We could attach the matcher logic to `TextEncoding_1_0` and `TextEncoding_1_1` by way of a trait impl (with most of the logic being in the default implementation, and only places where it diverges having specific implementations), and then store the version in `TextBufferView` as an enum or a type-state. A `change_version` function on `TextBufferView` would allow a Reader to switch between Ion 1.0 and Ion 1.1 as needed.

zslayton:
> I think the long term solution needs to be some way of extracting the matcher logic from the TextBufferView and then having the TextBufferView keep track of what version it is currently reading. We could attach the matcher logic to TextEncoding_1_0 and TextEncoding_1_1 by way of a trait impl (with most of the logic being in the default implementation, and only places where it diverges having specific implementations), and then store the version in TextBufferView as an enum or a type-state. A change_version function on TextBufferView would allow a Reader to switch between Ion 1.0 and Ion 1.1 as needed.

I think that's a great idea.

PR #673, opened 2023-10-17
Adds integration testing for Ion 1.1

popematt:
**Issue #, if available:**

None

**Description of changes:**

* There are a few clippy changes because a new version of rust was released.
* A large portion of the diff is `iontestdata` -> `iontestdata_1_0`.
* Adds an `ion_1_1` module to `element_test_vectors.rs` integration testing for the Ion 1.1 reader.
* Adds support for reading equivs and non-equivs test cases that are defined as a struct, and for reading embedded documents that are a blob.

### DIFF
--- a/tests/element_test_vectors.rs
+++ b/tests/element_test_vectors.rs
@@ -4,7 +4,7 @@
 
 use ion_rs::{
     BinaryWriterBuilder, Element, ElementReader, ElementWriter, Format, IonData, IonError,
-    IonResult, IonWriter, Sequence, TextKind, TextWriterBuilder,
+    IonResult, IonWriter, Sequence, TextKind, TextWriterBuilder, Value,
 };
 
 use std::fs::read;
@@ -147,6 +147,7 @@ trait ElementApi {
     /// embedded_documents::(
     ///   "a b c"
     ///   "'a' 'b' 'c'"
+    ///   {{ 4AEA6u6OgYPbhnEDh7aBYYFigWNxCnELcQw= }}
     /// )
     /// ```
     ///
@@ -160,9 +161,13 @@ trait ElementApi {
         let group_res: IonResult<Vec<_>> = raw_group
             .elements()
             .map(|elem| {
-                Self::make_reader(elem.as_text().unwrap().as_bytes())?
-                    .elements()
-                    .collect()
+                Self::make_reader(match elem.value() {
+                    Value::String(text) => text.text().as_bytes(),
+                    Value::Blob(bytes) => bytes.as_ref(),
+                    _ => panic!("Expected embedded document to be an Ion String or Ion Blob"),
+                })?
+                .elements()
+                .collect()
             })
             .collect();
         let group = group_res?;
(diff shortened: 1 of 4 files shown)


tests/element_test_vectors.rs
```diff
@@ -196,26 +201,54 @@ trait ElementApi {
...
     where
         // group index, value 1 index, value 1, value 2 index, value 2
-        F1: Fn(usize, usize, &Element, usize, &Element),
+        F1: Fn(usize, String, &Element, String, &Element),
         F2: Fn(&Vec<Element>, &Vec<Element>),
     {
         let group_lists = Self::read_file(file_name)?;
-        for (group_index, group_list) in group_lists.iter().enumerate() {
+        for (group_index, group_container) in group_lists.iter().enumerate() {
             // every grouping set is a list/sexp
```
zslayton, line 209: I think this comment needs an update to include structs?

tests/element_test_vectors.rs
```diff
@@ -196,26 +201,54 @@ trait ElementApi {
...
+                (Value::List(group), true) | (Value::SExp(group), true) => {
                     Self::read_group_embedded(group, &group_assert)?;
                 }
-                (Some(group), false) => {
+                (Value::Struct(group), true) => {
+                    let sequence: Sequence = group
+                        .to_owned()
+                        .fields()
+                        .map(|(_, v)| v.to_owned())
+                        .collect();
```
zslayton, line 221: I think you can skip the `to_owned()` call on `group`.

Checking my understanding: this is discarding the field names (which are names for the different tests) because we don't really have a use for them at the moment, yes?

zslayton (approved):
👍 

There are a few unrelated clippy warnings on the diff.

tests/element_test_vectors.rs
```diff
@@ -150,38 +149,19 @@ trait ElementApi {
...
-        let group = group_res?;
-        for this in group.iter() {
-            for that in group.iter() {
-                group_assert(this, that);
-            }
+    fn read_embedded_doc_as_sequence(elem: &Element) -> IonResult<Sequence> {
+        match elem.value() {
+            Value::String(text) => Element::read_all(text.text().as_bytes()),
+            Value::Blob(bytes) => Element::read_all(bytes.as_ref()),
+            _ => panic!("Expected embedded document to be an Ion String or Ion Blob"),
```
zslayton, line 156: Nice!

tests/element_test_vectors.rs
```diff
@@ -198,30 +178,29 @@ trait ElementApi {
...
-                (Value::Struct(group), true) => {
-                    let sequence: Sequence = group
-                        .to_owned()
-                        .fields()
-                        .map(|(_, v)| v.to_owned())
-                        .collect();
-                    Self::read_group_embedded(&sequence, &group_assert)?;
-                }
-                (Value::List(group), false) | (Value::SExp(group), false) => {
+            match group_container.value().to_owned() {
```
zslayton, line 191: Is the `to_owned()` needed? It looks like we're just iterating over whatever we get.

zslayton (approved):

tests/element_test_vectors.rs
popematt: A `to_owned()` is necessary somewhere because I can't have the `if` expression returning `Sequence` from one branch and `&Sequence` from another, but I was able to move the `to_owned()` call so that it is used as late as possible.

PR #674, opened 2023-10-18
Evaluation of template macros

### LINKED ISSUE #658: Evaluating user-defined template macros
Ion 1.1 allows users to define _templates_--reusable data shapes that can be invoked as many times as needed--that expand to zero or more Ion values. 

This issue builds on #656 and #657, and tracks building the machinery needed to evaluate invocations of a user-defined template macro. Such invocations can appear in two places:
1. In the data stream, encoded using an e-expression.
1. In the body of another template, encoded as an s-expression.

The latter requires implementing the ability to pass named parameters (i.e. variables) from the host template's environment to the macro being invoked.

For example:

```ion_1_1
// Template definition
(macro
    greet              // Template name
    (part_of_day name) // Parameter names
    (make_string       // Template body
        "Good "
        part_of_day
        ", "
        name
        "."))

(:greet "morning" "Mr. Cartwright")

// Notionally expands to the intermediate equivalent:
(:make_string "Good " "morning" ", " "Mr. Cartwright" ".")

// Which evaluates to:
"Good morning, Mr. Cartwright." 
```

In this case, the named parameters `part_of_day` and `name` are passed from the environment of the `greet` invocation to the environment of the `make_string` invocation.

zslayton:
Adds:
* a `TemplateExpansion` macro kind that can evaluate invocations of a user-specified macro definition.
* a `TemplateCompiler` that reads a macro definition expression and emits a `TemplateMacro`. 

Fixes #658.


src/lazy/expanded/macro_evaluator.rs
```diff
@@ -10,208 +10,345 @@
...
-pub struct MacroEvaluator<
-    'data,
-    // The Ion format of the data stream
-    D: LazyDecoder<'data>,
-    // The syntactic element representing the e-expression or template macro invocation
-    M: MacroInvocation<'data, D>,
-    // The storage being used to store the macro expansion stack. (Either a long-lived `Vec` or
-    // a bumpalo [`Bump`](BumpVec) whose contents only live as long as the reader is parked on
-    // the same top-level value).
-    S: Stack<MacroExpansion<'data, D, M>>,
```
zslayton, line 142: 🗺️ Previously, `MacroEvaluator` was generic over its `Stack` implementation, allowing it to hold either a static `Vec` or a `BumpVec`. This was done so that top level macros could produce multiple values and the `Vec` lived long enough to persist their expansion state between calls to `next()`.

Thanks to a small amount of `unsafe` code, it is now possible for the root `MacroEvaluator` that used to live in the `LazyExpandingReader` to instead live in the bump allocator. The `LazyExpandingReader` holds a raw pointer to it.

This enables the evaluator (and its macro stack) to hold references to other items in the bump allocator, symbols in the symbol table, macros in the macro table, and even other evaluators. They all have the same lifetime, `'top`, which refers to the duration for which the reader is parked on the current top level expression.

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -273,22 +408,34 @@ impl<
...
+                    let completed_kind = &self.macro_stack.last().unwrap().kind;
+                    if matches!(completed_kind, MacroExpansionKind::Template(_)) {
+                        // NB: Here and below, we use `truncate()` instead of `pop()` so the value can
+                        // be dropped in place without incurring a move. That move runs afoul of the
+                        // aliasing requirements that `miri` looks for, though I'm unsure why.
+                        // Once Polonius lands and we are able to remove the `unsafe` usages in
+                        // the LazyExpandingReader, this will be unnecessary.
+                        self.env_stack.truncate(self.env_stack.len() - 1);
+                    }
+                    self.macro_stack.truncate(self.macro_stack.len() - 1);
```
zslayton, line 435: 🗺️ There is some `unsafe` code in the `LazyExpandingReader` that exists purely as a workaround to a limitation in `rustc` that will be fixed some time in the coming months. `miri` reported that `pop()` was problematic, but I'm skeptical that it mattered. Running `miri` again with its more sophisticated [tree-borrow model](https://perso.crans.org/vanille/treebor/) enabled allowed it to pass without complaint.

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -303,113 +450,46 @@ impl<
...
+    where
+        'data: 'top,
+        Self: Sized,
+    {
+        self.push(context, invocation.into())?;
         Ok(EvaluatingIterator::new(self, context))
     }
 }
 
-// ===== Type aliases for commonly used flavors of `MacroEvaluator` =====
```
zslayton, line 316: 🗺️ `MacroEvaluator` is now a `struct` instead of a `trait`, so we no longer need these implementations.

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -432,47 +512,33 @@ impl<
 ///   (:values 1)                 => 1
 ///   (:values 1 2 3)             => 1 2 3
 ///   (:values 1 2 (:values 3 4)) => 1 2 3 4
-pub struct ValuesExpansion<'data, D: LazyDecoder<'data>, M: MacroInvocation<'data, D>> {
+pub struct ValuesExpansion<'top, 'data, D: LazyDecoder<'data>> {
     // Which argument the macro is in the process of expanding
-    arguments: M::ArgumentsIterator,
+    arguments: MacroExprArgsIterator<'top, 'data, D>,
```
zslayton, line 517: 🗺️ Rather than holding an iterator over the syntactic elements from the invocation, the newly added `'top` lifetime allows us to hold an iterator over the _resolved_ arguments of the invocation. This handles variable resolution for us, which simplifies the implementation of each macro kind a good bit.

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -432,47 +512,33 @@ impl<
...
-        match arg_expr {
-            // If the argument is a value, return it.
-            ArgumentKind::ValueLiteral(value) => Ok(MacroExpansionStep::ExpandedValue(value)),
-            ArgumentKind::Variable(_variable) => todo!("variable expansion"),
-            // If the argument is a macro invocation, yield it that so the evaluator can push it onto the stack.
-            ArgumentKind::MacroInvocation(invocation) => {
-                Ok(MacroExpansionStep::AnotherMacroToEvaluate(invocation))
-            }
-        }
+        self.arguments.next().transpose()
```
zslayton, line 541: 🗺️ The implementation of `(:values)`  is now especially straightforward.

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -606,17 +663,68 @@ impl<'data, D: LazyDecoder<'data>, M: MacroInvocation<'data, D>> MakeStringExpan
...
+            Some(expr) => expr,
+        };
+        self.step_index += 1;
+
+        let step = match value_expr {
+            TemplateBodyValueExpr::Element(e) => {
+                match e.value() {
+                    TemplateValue::List(range)
+                    | TemplateValue::SExp(range)
+                    | TemplateValue::Struct(range) => self.step_index += range.len(),
```
zslayton, line 701: 🗺️ If the value in the template body is a container, it will be able to iterate over its child elements from the template body. This iterator can skip over them.

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -606,17 +663,68 @@ impl<'data, D: LazyDecoder<'data>, M: MacroInvocation<'data, D>> MakeStringExpan
...
+                    environment,
+                    TemplateElement::new(self.template, e),
+                ))
+            }
+            TemplateBodyValueExpr::Variable(variable) => {
+                *environment.get_expected(variable.signature_index())?
+            }
+            TemplateBodyValueExpr::MacroInvocation(raw_invocation) => {
+                let invocation = raw_invocation.resolve(self.template, context)?;
+                self.step_index += invocation.arg_expressions().len();
```
zslayton, line 715: 🗺️ Similarly, if this is a macro invocation it will handle its own arguments. When this template expansion is back at the top of the stack (after this new macro is evaluated), it needs to skip the arguments to resume processing.

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -642,31 +750,226 @@ mod tests {
...
-        let invocation = Element::read_one(invocation)?;
-        let actuals = evaluator.evaluate(context, invocation.expect_sexp()?)?;
+    fn eval_template_invocation(
+        template_definition: &str,
+        invocation: &str,
+        expected: &str,
+    ) -> IonResult<()> {
+        let mut reader = LazyTextReader_1_1::new(invocation.as_bytes())?;
+        let _macro_address = reader.register_template(template_definition)?;
+        let actuals = reader.read_all_elements()?;
```
zslayton, line 760: 🗺️ We no longer have to manually construct an `EncodingContext` and build an evaluator. We can just use a `LazyTextReader_1_1`. For now, though, we have to manually hand it our template definition text.

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -642,31 +750,226 @@ mod tests {
...
+                (:lst ["foo", "bar", "baz"]) $10 $11 $12
+            "#,
+            r#"
+                foo bar baz
+            "#,
+        )
+    }
+
+    #[test]
+    fn context_changes_happen_between_top_level_expressions() -> IonResult<()> {
```
zslayton, line 806: 🗺️ The spec requires that a single expression use the same encoding for its entirety. If an expression produces multiple values, any system values will not be applied until the expression has been evaluated in full.

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -757,8 +1069,7 @@ mod tests {
     }
 
     // TODO: macros_inside_a_tdl_sexp()
-    // This requires an implementation of TDL's `(make_sexp)` or `(quote)`. Without these,
-    // a sexp is always considered a TDL macro invocation.
+    // This requires an implementation of TDL's `(make_sexp)`.
```
zslayton, line 1072: 🗺️ You actually can't use `quote` to make an s-expression if you want it to have nested macro invocations. `quote` will cause the nested invocations to be considered s-expressions too.

src/lazy/expanded/mod.rs
```diff
@@ -131,84 +134,334 @@ impl<'top, 'data, D: LazyDecoder<'data>> ExpandedStreamItem<'top, 'data, D> {
...
 /// The source of data backing a [`LazyExpandedValue`].
-#[derive(Debug, Clone)]
+#[derive(Copy, Clone)]
 pub enum ExpandedValueSource<'top, 'data, D: LazyDecoder<'data>> {
     /// This value was a literal in the input stream.
     ValueLiteral(D::Value),
     /// This value was part of a template definition.
-    Template(&'top Element),
-    /// This value was the computed result of a macro invocation like `(:make_string ...)`.
+    Template(Environment<'top, 'data, D>, TemplateElement<'top>),
```
zslayton, line 441: 🗺️ We no longer use an `Element` (which allocates on the heap and uses atomic operations) as our template value. Instead, we use `TemplateElement`, which is either a plain scalar value or a container type holding a number of child elements.

src/lazy/expanded/mod.rs
```diff
@@ -131,84 +134,334 @@ impl<'top, 'data, D: LazyDecoder<'data>> ExpandedStreamItem<'top, 'data, D> {
...
         //       it to `Never` and the compiler can eliminate this code path where applicable.
-        (
-            // A collection of bump-allocated annotation strings
-            BumpVec<'top, &'top str>,
-            ExpandedValueRef<'top, 'data, D>,
-        ),
+        // Constructed data stored in the bump allocator. Holding references instead of the data
+        // itself allows this type (and those that contain it) to impl `Copy`.
+        &'top [&'top str],
+        &'top ExpandedValueRef<'top, 'data, D>,
```
zslayton, line 449: 🗺️ These values all live in the bump allocator. Switching them to references allowed `LazyExpandedValue` to implement `Copy`.

src/lazy/expanded/sequence.rs
```diff
@@ -1,16 +1,70 @@
...
 use crate::lazy::expanded::{
     EncodingContext, ExpandedAnnotationsIterator, ExpandedAnnotationsSource, ExpandedValueSource,
     LazyExpandedValue,
 };
-use crate::{Annotations, IonResult, IonType, Sequence};
+use crate::result::IonFailure;
+use crate::{IonError, IonResult, IonType};
+
+#[derive(Copy, Clone, Debug)]
+pub struct Environment<'top, 'data, D: LazyDecoder<'data>> {
```
zslayton, line 17: 🗺️ The series of (not yet evaluated) expressions passed to a macro as arguments.

`TODO` for me: doc comment

src/lazy/expanded/stack.rs
zslayton, line 1: 🗺️ Now that the `MacroEvaluator` is not generic over `Vec`/`BumpVec`, this isn't necessary.

src/lazy/expanded/tdl_macro.rs
zslayton, line 1: 🗺️ Everything living in this file became obsolete as a result of the refactor.

src/lazy/system_reader.rs
```diff
@@ -77,15 +70,7 @@ const SYMBOLS: RawSymbolTokenRef = RawSymbolTokenRef::SymbolId(7);
...
-    // TODO: Remove this RefCell when the Polonius borrow checker is available.
-    //       See: https://github.com/rust-lang/rust/issues/70255
-    expanding_reader: RefCell<LazyExpandingReader<'data, D>>,
-    // TODO: Make the symbol and macro tables traits on `D` such that they can be configured
-    //       statically. Then 1.0 types can use `Never` for the macro table.
-    symbol_table: SymbolTable,
-    macro_table: MacroTable,
-    allocator: BumpAllocator,
-    pending_lst: PendingLst,
+    pub(crate) expanding_reader: LazyExpandingReader<'data, D>,
```
zslayton, line 73: 🗺️ The `LazyExpandingReader` can only apply changes to the `EncodingContext` between top-level expressions, which meant it was much easier for it to own all of these resources. Having moved them all over to the `LazyExpandingReader`, it became clear that there wasn't anything left for the `LazySystemReader` to do. I'll merge them in another PR later. Tracking this in #675.

zslayton (commented):
🗺️ PR tour

src/lazy/expanded/mod.rs
```diff
@@ -131,84 +134,334 @@ impl<'top, 'data, D: LazyDecoder<'data>> ExpandedStreamItem<'top, 'data, D> {
...
+    ) -> IonResult<Option<SystemStreamItem<'top, 'data, D>>> {
+        let evaluator_ptr = match self.evaluator_ptr.get() {
+            // There's not currently an evaluator.
+            None => return Ok(None),
+            // There's an evaluator in the process of expanding a macro.
+            Some(ptr) => ptr,
+        };
+        let evaluator = Self::ptr_to_evaluator(evaluator_ptr);
+
+        match evaluator.next(self.context(), 0) {
```
popematt, line 420: What does the `0` represent here? (Maybe add a comment so that it doesn't look like a magic number.)

src/lazy/expanded/mod.rs
```diff
@@ -131,84 +134,334 @@ impl<'top, 'data, D: LazyDecoder<'data>> ExpandedStreamItem<'top, 'data, D> {
...
         // TODO: Make this an associated type on the LazyDecoder trait so 1.0 types can set
         //       it to `Never` and the compiler can eliminate this code path where applicable.
-        (
-            // A collection of bump-allocated annotation strings
-            BumpVec<'top, &'top str>,
-            ExpandedValueRef<'top, 'data, D>,
-        ),
+        // Constructed data stored in the bump allocator. Holding references instead of the data
+        // itself allows this type (and those that contain it) to impl `Copy`.
+        &'top [&'top str],
```
popematt, line 448: Is this the annotations for the constructed value?

benches/read_many_structs.rs
```diff
@@ -0,0 +1,120 @@
...
+use ion_rs::lazy::r#struct::LazyStruct;
+use ion_rs::lazy::reader::{LazyApplicationReader, LazyTextReader_1_0, LazyTextReader_1_1};
+use ion_rs::lazy::value::LazyValue;
+use ion_rs::lazy::value_ref::ValueRef;
+use ion_rs::IonResult;
+use nom::AsBytes;
+
+pub fn criterion_benchmark(c: &mut Criterion) {
+    const NUM_VALUES: usize = 10_000;
+    let data_1_0 = concat!("{",
```
zslayton, line 13: 🗺️ Slightly awkward syntax here to produce a compact struct instead of a pretty-printed one.

benches/read_many_structs.rs
```diff
@@ -0,0 +1,120 @@
...
+        for field in lazy_struct {
+            count += count_value_and_children(&field?.value())?;
+        }
+        Ok(count)
+    }
+
+    c.bench_function("text 1.0: scan all", |b| {
+        b.iter(|| {
+            let mut reader =
+                LazyApplicationReader::<'_, TextEncoding_1_1>::new(data_1_0.as_bytes()).unwrap();
```
zslayton, line 82: 🗺️ The 1.0 benchmark tests are using the 1.1 reader because child expression caching is hugely impactful and has not yet been added to the 1.0 decoders. Because Ion 1.1 text is a superset of 1.0 text, the 1.1 reader can process it fine. The measurement _should_ be basically the same as using the 1.0 reader since there are just a few more branches that never get taken, but we'll get this changed over to a proper 1.0 reader once that optimization is backported.

I'll add a comment for this.

src/lazy/text/buffer.rs
```diff
@@ -107,37 +109,52 @@ fn dbg_parse<I: Debug, O: Debug, E: Debug, P: Parser<I, O, E>>(
...
     // `offset` is the absolute position in the overall input stream where that slice begins.
     //
     // input: 00 01 02 03 04 05 06 07 08 09
     //                          └────┬────┘
     //                          data: &[u8]
     //                          offset: 6
-    data: &'a [u8],
+    data: &'top [u8],
     offset: usize,
+    allocator: &'top BumpAllocator,
```
zslayton, line 123: 🗺️ `TextBufferView` now holds a reference to the bump allocator, allowing it to cheaply store structural information about containers and e-expressions. Data living in the bump allocator has the `'top` lifetime.

Passing an allocator into each parsing method as an argument would cause those methods to violate the `Parser` contract, which says they must only accept a single parameter of some type `I: Input`. Adding that field allows us to use it everywhere that a `TBV` is already accepted.

src/lazy/text/buffer.rs
```diff
@@ -107,37 +109,52 @@ fn dbg_parse<I: Debug, O: Debug, E: Debug, P: Parser<I, O, E>>(
...
 }
 
-impl<'data> TextBufferView<'data> {
+impl<'a> PartialEq for TextBufferView<'a> {
+    fn eq(&self, other: &Self) -> bool {
+        self.offset == other.offset && self.data == other.data
+    }
+}
```
zslayton, line 130: 🗺️ Adding a `BumpAllocator` to `TextBufferView` meant that we could no longer derive an implementation of `PartialEq`.

src/lazy/text/buffer.rs
```diff
@@ -383,38 +404,42 @@ impl<'data> TextBufferView<'data> {
             // If the next thing in the input is a `}`, return `None`.
             value(None, Self::match_struct_end),
             // Otherwise, match a name/value pair and turn it into a `LazyRawTextField`.
-            Self::match_struct_field_name_and_value.map(
```
zslayton, line 378: 🗺️ As you'll see below, `clippy` (rightfully) complained about the complexity of the type returned by `match_struct_field_name_and_value` among others. I've introduced a `MatchedFieldName` type that simplifies the signatures quite a bit.

src/lazy/text/buffer.rs
```diff
@@ -441,55 +466,53 @@ impl<'data> TextBufferView<'data> {
                 whitespace_and_then(alt((tag(","), peek(tag("}"))))),
             )
             .map(|invocation| Ok(Some(RawFieldExpr::MacroInvocation(invocation)))),
-            Self::match_struct_field_name_and_e_expression_1_1.map(
-                |((matched_name, name_span), invocation)| {
```
zslayton, line 437: 🗺️ These closure parameters were consolidated, making this line shorter and causing `rustfmt` to reflow the lines that followed. 😞 This is not a logic change.

src/lazy/text/buffer.rs
```diff
@@ -499,16 +522,10 @@ impl<'data> TextBufferView<'data> {
...
     pub fn match_struct_field_name_and_e_expression_1_1(
         self,
-    ) -> IonParseResult<
-        'data,
-        (
-            (MatchedFieldName, Range<usize>),
-            RawTextEExpression_1_1<'data>,
-        ),
-    > {
+    ) -> IonParseResult<'top, (MatchedFieldName, RawTextEExpression_1_1<'top>)> {
```
zslayton, line 525: 🗺️ This is the return type simplification. `MatchedFieldName` now contains the equivalent of a `Range<usize>`, allowing this to be reduced.

src/lazy/text/buffer.rs
```diff
@@ -533,7 +556,9 @@ impl<'data> TextBufferView<'data> {
     }
 
     /// Matches an optional annotation sequence and a trailing value.
-    pub fn match_annotated_value(self) -> IonParseResult<'data, MatchedRawTextValue<'data>> {
+    pub fn match_annotated_value(
+        self,
+    ) -> IonParseResult<'top, MatchedRawTextValue<'top, TextEncoding_1_0>> {
```
zslayton, line 561: 🗺️ `MatchedRawTextValue` is now generic over an `E: TextEncoding<'top>`, which allows it to contain a cache of raw value expressions of that encoding.

src/lazy/text/buffer.rs
```diff
@@ -575,20 +602,19 @@ impl<'data> TextBufferView<'data> {
     /// * An identifier
     /// * A symbol ID
     /// * A short-form string
-    pub fn match_struct_field_name(self) -> IonParseResult<'data, MatchedFieldName> {
-        alt((
-            Self::match_string.map(MatchedFieldName::String),
-            Self::match_symbol.map(MatchedFieldName::Symbol),
-        ))(self)
+    pub fn match_struct_field_name(self) -> IonParseResult<'top, MatchedFieldName> {
+        match_and_span(alt((
```
zslayton, line 606: 🗺️ Previously, every call to this method would wrap it in a `match_and_span(` call. Now the method performs that step itself, simplifying a few call sites.

src/lazy/text/buffer.rs
```diff
@@ -705,19 +731,40 @@ impl<'data> TextBufferView<'data> {
...
                 |(matched_list, length)| {
-                    EncodedTextValue::new(MatchedValue::List, matched_list.offset(), length)
+                    // TODO: Cache child expressions found in 1.0 list
+                    let not_yet_used_in_1_0 =
+                        bumpalo::collections::Vec::new_in(self.allocator).into_bump_slice();
+                    EncodedTextValue::new(
+                        MatchedValue::List(not_yet_used_in_1_0),
+                        matched_list.offset(),
+                        length,
+                    )
```
zslayton, line 741: 🗺️ The 1.0 container types have caching stubbed out for trait conformance, but it needs to be implemented. I'll open an issue for this.

src/lazy/text/buffer.rs
```diff
@@ -869,7 +930,15 @@ impl<'data> TextBufferView<'data> {
...
-    pub fn match_list_1_1(self) -> IonMatchResult<'data> {
+    pub fn match_list_1_1(
+        self,
+    ) -> IonParseResult<
+        'top,
+        (
+            TextBufferView<'top>,
+            &'top [LazyRawValueExpr<'top, TextEncoding_1_1>],
+        ),
+    > {
```
zslayton, line 941: 🗺️ Matching a list now returns both the matched text buffer and a bump-allocated collection of child value expressions.

src/lazy/text/buffer.rs
```diff
@@ -878,64 +947,76 @@ impl<'data> TextBufferView<'data> {
...
-            // If invalid syntax was encountered, return a failure to prevent nom from trying
-            // other parser kinds.
-            Err(e) => {
-                return {
-                    let error = InvalidInputError::new(self)
-                        .with_label("matching a v1.1 list")
-                        .with_description(format!("{}", e));
-                    Err(nom::Err::Failure(IonParseError::Invalid(error)))
+        let (span, child_exprs) =
+            match TextListSpanFinder_1_1::new(self.allocator, sequence_iter).find_span() {
```
zslayton, line 951: 🗺️ The `_______SpanFinder` types wrap a parsing iterator and return the matched span along with any nested expressions.

src/lazy/text/matched.rs
```diff
@@ -58,31 +60,84 @@ pub(crate) enum MatchedValue {
...
     Clob(MatchedClob),
-    List,
-    SExp,
-    Struct,
+    List(&'top [LazyRawValueExpr<'top, D>]),
+    SExp(&'top [LazyRawValueExpr<'top, D>]),
+    Struct(&'top [LazyRawFieldExpr<'top, D>]),
+}
+
+impl<'top, D: LazyDecoder> PartialEq for MatchedValue<'top, D> {
```
zslayton, line 68: 🗺️ Storing slices of `LazyRawFieldExpr` meant that we could no longer derive an implementation of `PartialEq`.

src/lazy/text/raw/reader.rs
```diff
@@ -97,6 +98,34 @@ mod tests {
 
     use super::*;
 
+    struct TestReader<'data> {
+        allocator: BumpAllocator,
+        reader: LazyRawTextReader_1_0<'data>,
+    }
```
zslayton, line 104: 🗺️ A consequence of expression caching is that using the raw reader requires the application to supply a `BumpAllocator` to each call to `next()`. Since this is not an API many people will interact with--most won't need to and it will be behind a feature flag--I'm ok with that tradeoff.

src/lazy/text/raw/v1_1/reader.rs
```diff
@@ -543,6 +627,80 @@ impl<'top> RawTextStructIterator_1_1<'top> {
...
+    }
+
+    /// Scans ahead to find the end of this s-expression and reports the input span that it occupies.
+    /// As it scans, it records lazy references to the S-expression's child expressions.
+    ///
+    /// The `initial_bytes_skipped` parameter indicates how many bytes of input that represented the
+    /// beginning of the expression are not in the buffer. For plain s-expressions, this will always
+    /// be `1` as they begin with a single open parenthesis `(`. For e-expressions (which are used
+    /// to invoke macros from the data stream), it will always be a minimum of `3`: two bytes for
+    /// the opening `(:` and at least one for the macro identifier. (For example: `(:foo`.)
```
zslayton, line 650: `TODO` for me: fix this comment that refers to s-expressions

zslayton (commented):
🗺️ PR tour for most recent commits

src/lazy/text/raw/reader.rs
popematt: This is fine, but is there any particular reason it can't be stored in the reader?

src/lazy/expanded/template.rs
```diff
@@ -35,67 +154,717 @@ impl<'top, 'data, D: LazyDecoder<'data>> Iterator for TemplateSequenceIterator<'
...
+                            .body
+                            .expressions()
+                            .get(body_invocation.arg_expr_range().ops_range())
+                            .unwrap(),
+                    );
+                    self.index += invocation.arg_expressions.len();
+                    match self.evaluator.push(self.context, invocation) {
+                        Ok(_) => continue,
+                        Err(e) => return Some(Err(e)),
+                    };
```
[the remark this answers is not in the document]
zslayton: This method is where an e-expression's arguments are evaluated for the first time. If an argument is missing, malformed, or invokes a non-existent macro, those errors will surface here.

I'm open to renaming this if you can think of a name that preserves what `push()` communicates (i.e. we're adding to the stack of evaluations in progress) while also indicating that some validation is happening.

src/lazy/expanded/template.rs
```diff
@@ -35,67 +154,717 @@ impl<'top, 'data, D: LazyDecoder<'data>> Iterator for TemplateSequenceIterator<'
...
+                        ValueExpr::ValueLiteral(value) => {
+                            return Some(Ok(*value));
+                        }
+                        ValueExpr::MacroInvocation(invocation) => {
+                            match self.evaluator.push(self.context, *invocation) {
+                                Ok(_) => continue,
+                                Err(e) => return Some(Err(e)),
+                            }
+                        }
+                    }
```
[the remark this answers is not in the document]
zslayton: `rustfmt` preserved the match-arm-braces in the latter case, but accepted the other changes.

src/lazy/expanded/mod.rs
```diff
@@ -131,84 +134,334 @@ impl<'top, 'data, D: LazyDecoder<'data>> ExpandedStreamItem<'top, 'data, D> {
...
+    fn apply_pending_lst(pending_lst: &mut PendingLst, symbol_table: &mut SymbolTable) {
+        // If the symbol table's `imports` field had a value of `$ion_symbol_table`, then we're
+        // appending the symbols it defined to the end of our existing local symbol table.
+        // Otherwise, we need to clear the existing table before appending the new symbols.
+        if !pending_lst.is_lst_append {
+            // We're setting the symbols list, not appending to it.
+            symbol_table.reset();
+        }
+        // `drain()` empties the pending symbols list
+        for symbol in pending_lst.symbols.drain(..) {
```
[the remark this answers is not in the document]
zslayton: Yep, the variations on the range literal each construct one of [these types](https://doc.rust-lang.org/std/ops/index.html#structs).

src/lazy/text/raw/v1_1/reader.rs
```diff
@@ -168,13 +168,29 @@ impl<'top> RawTextListIterator_1_1<'top> {
             has_returned_error: false,
         }
     }
+}
 
-    pub(crate) fn find_span(&self) -> IonResult<Range<usize>> {
+pub(crate) struct TextListSpanFinder_1_1<'top> {
```
popematt, line 173: Doc comments, please.

src/lazy/text/raw/v1_1/reader.rs
```diff
@@ -422,34 +504,36 @@ impl<'top> LazyRawSequence<'top, TextEncoding_1_1> for LazyRawTextSExp_1_1<'top>
...
+        RawTextSequenceCacheIterator_1_1::new(child_exprs)
     }
 
     fn as_value(&self) -> LazyRawTextValue_1_1<'top> {
         self.value.matched.into()
     }
 }
 
-impl<'top> Iterator for RawTextSExpIterator_1_1<'top> {
+impl<'top> Iterator for RawTextListIterator_1_1<'top> {
```
popematt, line 518: Did `impl<'top> Iterator for RawTextListIterator_1_1` and `impl<'top> Iterator for RawSExpListIterator_1_1` get swapped for any particular reason?

src/lazy/text/raw/v1_1/reader.rs
```diff
@@ -168,13 +168,29 @@ impl<'top> RawTextListIterator_1_1<'top> {
...
+}
 
-    pub(crate) fn find_span(&self) -> IonResult<Range<usize>> {
+pub(crate) struct TextListSpanFinder_1_1<'top> {
+    pub(crate) allocator: &'top bumpalo::Bump,
+    pub(crate) iterator: RawTextListIterator_1_1<'top>,
+}
+
+impl<'top> TextListSpanFinder_1_1<'top> {
+    pub(crate) fn find_span(
```
popematt, line 179: As far as I can tell, this new function isn't called anywhere, so I'm having a hard time understanding how the caching actually helps anything.

src/lazy/expanded/template.rs
```diff
@@ -35,67 +154,717 @@ impl<'top, 'data, D: LazyDecoder<'data>> Iterator for TemplateSequenceIterator<'
...
+                    expression_address,
+                    invoked_macro_address,
+                    expr_range,
+                ),
+            ))
+    }
+}
+
+#[derive(Copy, Clone, Debug)]
+pub struct TemplateValueExpr<'top> {
```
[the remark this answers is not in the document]
zslayton: This type was a holdover from an earlier factoring; I've deleted it.

src/lazy/expanded/template.rs
```diff
@@ -35,67 +154,717 @@ impl<'top, 'data, D: LazyDecoder<'data>> Iterator for TemplateSequenceIterator<'
...
+    start: u32,
+    end: u32,
+}
+
+impl SmallRange {
+    pub fn empty() -> Self {
+        Self { start: 0, end: 0 }
+    }
+
+    pub fn new(range: Range<usize>) -> Self {
```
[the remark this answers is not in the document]
zslayton: We could do either of those. In practice (at least for the moment), every use of the `new()` method looks something like this:
image

`new(start..end)` ends up being a pretty trivial difference from `new(start, end)` or `(start..end).into()`.

src/lazy/expanded/sequence.rs
```diff
@@ -1,16 +1,70 @@
...
+    expressions: &'top [ValueExpr<'top, 'data, D>],
+}
+
+pub const fn empty_args<'top, 'data, D: LazyDecoder<'data>>() -> &'top [ValueExpr<'top, 'data, D>] {
+    &[]
+}
```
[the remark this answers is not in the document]
zslayton: For anyone curious: `const` and `static` variables cannot use generics from their containing function signature.

Since this was the only usage of `empty_args()`, I made it a nested function for easier reading / improved locality.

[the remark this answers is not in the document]
zslayton: Turns out I was able to simplify this to just
```rust
    pub fn empty() -> Environment<'top, D> {
        Environment { expressions: &[] }
    }
```
I was sure I needed a `static` or `const` to prevent the `[]` from living on the stack, but I guess not!

src/lazy/expanded/template.rs
```diff
@@ -35,67 +154,717 @@ impl<'top, 'data, D: LazyDecoder<'data>> Iterator for TemplateSequenceIterator<'
...
+        }
+    }
+}
+
+/// A macro invocation in a template body.
+#[derive(Copy, Clone)]
+pub struct TemplateMacroInvocation<'top> {
+    context: EncodingContext<'top>,
+    // The definition of the template in which this macro invocation appears
+    host_template: TemplateMacroRef<'top>,
```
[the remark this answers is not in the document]
zslayton: > Is this just for debugging?
Yes. Added a comment.

src/lazy/expanded/mod.rs
```diff
@@ -131,84 +134,334 @@ impl<'top, 'data, D: LazyDecoder<'data>> ExpandedStreamItem<'top, 'data, D> {
...
+    ///
+    /// This method will consume and process as many system-level values as possible until it
+    /// encounters an application-level value or the end of the stream.
+    pub fn next_value<'top>(&'top mut self) -> IonResult<Option<LazyValue<'top, 'data, D>>> {
+        loop {
+            match self.next_item()? {
+                SystemStreamItem::VersionMarker(_, _) => {
+                    // TODO: Handle version changes 1.0 <-> 1.1
+                }
+                SystemStreamItem::SymbolTable(_) => {}
```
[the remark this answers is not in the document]
zslayton: No, any necessary symbol table processing happened in the call to `next_item` before it returned the symbol table. I've added a comment stating that explicitly.

src/lazy/expanded/mod.rs
zslayton: Yes, added a comment.

src/lazy/text/raw/reader.rs
zslayton: `LazyRawReader::next()` mutably borrows the raw reader for `'top`, and the expanding reader needs access to the allocator after calling `LazyRawReader::next()` which would require another borrow.

An alternative to passing the allocator into `next()` is having `next()` return a reference to the allocator with each value. That's possible (and maybe preferable?) but I'd like to defer implementing that until after this PR.

src/lazy/text/raw/v1_1/reader.rs
zslayton: I think this is a misleading diff that came from cutting the `RawTextSExpIterator` and pasting it closer to the definition of `struct LazyRawTextSExp_1_1`.

src/lazy/expanded/mod.rs
```diff
@@ -131,84 +134,334 @@ impl<'top, 'data, D: LazyDecoder<'data>> ExpandedStreamItem<'top, 'data, D> {
...
+
+        let macro_table = self.macro_table.get_mut();
+        macro_table.add_macro(template_macro)
+    }
+
+    fn context(&self) -> EncodingContext<'_> {
+        // SAFETY: The only time that the macro table, symbol table, and allocator can be modified
+        // is in the body of the method `between_top_level_expressions`. As long as nothing holds
+        // a reference to the `EncodingContext` we create here when that method is running,
+        // this is safe.
```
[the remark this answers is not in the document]
popematt: I guess it depends on how "correct" we want things to be. Is it going to be easier to understand and maintain this later if we mark them as unsafe? I will defer to your judgment on the matter.

benches/read_many_structs.rs
popematt: If you feel like it, you could probably use something like
```rust
r"
{
  foo:1,
  bar:2,
}
".split("\n".into()).map(|it| it.trim()).collect::<Vec<_>>().join("");
```

It's still kind of awkward, but it moves the awkwardness somewhere else. I don't have a preference.

zslayton: I'm adding another commit that writes out the struct in full (as you've done) and then just roundtrips it to a `String` of compact text.

The same commit also makes the `IonEq` trait externally visible so the benchmark can confirm that the 1.0 and 1.1 test data are equivalent before measurement begins.

src/lazy/expanded/mod.rs
zslayton: Addressed in [this commit](https://github.com/amazon-ion/ion-rust/pull/674/commits/1db5b308edd4ecdb83f53667f0e4e6fe6dd9ef37).

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -642,31 +750,226 @@ mod tests {
...
+                foo bar baz
+            "#,
+        )
+    }
+
+    #[test]
+    fn context_changes_happen_between_top_level_expressions() -> IonResult<()> {
+        eval_template_invocation(
+            r#"
+                (macro lst (symbols)
```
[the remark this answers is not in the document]
zslayton: Yes, definitely. We haven't settled on a final set of system-provided macros, but something like `lst` or `macros` is a shoo-in

src/ion_data/mod.rs
```diff
@@ -1,4 +1,4 @@
-mod ion_eq;
+pub(crate) mod ion_eq;
```
popematt, line 1: If you wanted, we could leave this mod the way it is, and you can use `IonData::eq(a, b)` to compare to values.

src/lib.rs
```diff
@@ -204,6 +204,7 @@ pub use element::{
     reader::ElementReader, writer::ElementWriter, Annotations, Element, IntoAnnotatedElement,
     IntoAnnotations, Sequence, Value,
 };
+pub use ion_data::ion_eq::IonEq;
```
popematt, line 207: Let's not expose this publicly (at least not yet). There was a lot of back and forth about this when it was implemented, and I can't remember all of the details right now. I don't think we should export this without revisiting the prior discussion.

src/ion_data/mod.rs
zslayton: Good call, I've done this in [158348e](https://github.com/amazon-ion/ion-rust/pull/674/commits/158348ea62837ceb94983db0863eca2de6105c25).

popematt (approved):

PR #680, opened 2023-11-14
Introduces a new writer API

zslayton:
Introduces a friendlier writer API that will supplant the existing `IonWriter`. For now, it's called the `LazyWriter` for symmetry with the `LazyReader`. However, I plan to revisit that naming scheme once these implementations are ready to become the primary (read: only) implementations in the library.

The new API uses traits instead of methods to handle writing different data types, which results less noisy code. For example, this code:
```rust
let mut writer = LazyRawTextWriter_1_0::new(&mut buffer);
writer
    .write(1)?
    .write(false)?
    .write(3f32)?
    .write("foo")?
    .write("bar".as_symbol_ref())?
    .write(Timestamp::with_ymd(2023, 11, 9).build()?)?
    .write([0xE0u8, 0x01, 0x00, 0xEA])?
    .write([1, 2, 3])?
    .flush()?;
```
produces this output:
```
1
false
3e0
"foo"
bar
2023-11-09T
{{4AEA6g==}}
[1, 2, 3]
```


src/lazy/encoder/annotate.rs
```diff
@@ -0,0 +1,73 @@
...
+    /// let actual = Element::read_one(&buffer)?;
+    ///
+    /// assert!(IonData::eq(&expected, &actual));
+    ///# Ok(())
+    ///# }
+    /// ```
+    fn annotate<'a, A: AsRawSymbolTokenRef>(
+        &'a self,
+        annotations: &'a [A],
+    ) -> Annotated<'a, Self, A>;
```
zslayton, line 40: 🗺️ This extension method allows values being passed into `writer.write(...)` to add annotations to themselves:
```rust
writer.write(30_000.annotate("lbs_bananas"))?;
```

src/lazy/encoder/annotate.rs
```diff
@@ -0,0 +1,73 @@
...
+    T: WriteAsIonValue,
+    A: AsRawSymbolTokenRef,
+{
+    fn write_as_ion<'a, W: Write + 'a, E: LazyEncoder<W>, V: AnnotatedValueWriter<'a, W, E>>(
+        &self,
+        annotations_writer: V,
+    ) -> IonResult<()> {
+        let value_writer = annotations_writer.write_annotations(self.annotations.iter())?;
+        self.value.write_as_ion_value(value_writer)
+    }
```
zslayton, line 72: 🗺️ `AnnotatedValueWriter` is a staged writer. You call a method to either write annotations or `no_annotations()` and upon success it returns a `ValueWriter` that can be used to write the value itself.

src/lazy/encoder/mod.rs
```diff
@@ -0,0 +1,565 @@
...
+pub trait AnnotatedValueWriter<'a, W: Write, E: LazyEncoder<W>> {
+    /// Writes the provided annotations to the output stream and returns a [`ValueWriter`] that can
+    /// be used to serialize the value itself.
+    ///
+    /// If there are no annotations, use [`Self::no_annotations`] instead. This method will loop
+    /// over the empty sequence, incurring minor performance overhead while `no_annotations`
+    /// is a true no-op.
+    fn write_annotations<
+        SymbolType: AsRawSymbolTokenRef,
+        IterType: Iterator<Item = SymbolType> + Clone,
```
zslayton, line 75: 🗺️ An earlier version of this used much more generics; I started spelling out what each generic was (e.g. `SymbolType` instead of `A`). I think I kind of like it?

src/lazy/encoder/mod.rs
```diff
@@ -0,0 +1,565 @@
...
+        }
+        Ok(self.value_writer)
+    }
+
+    fn no_annotations(self) -> TextValueWriter_1_0<'a, W> {
+        self.value_writer
+    }
+}
+
+impl<'a, W: Write> ValueWriter<'a, W, TextEncoding_1_0> for TextValueWriter_1_0<'a, W> {
```
zslayton, line 250: 🗺️ The bodies of these methods are either copy/pasted from the existing raw text writer or invoke one of its methods.

src/lazy/encoder/mod.rs
```diff
@@ -0,0 +1,565 @@
...
+
+impl<'a, W: Write> Drop for TextListWriter_1_0<'a, W> {
+    fn drop(&mut self) {
+        // If the user didn't call `end`, the closing delimiter was not written to output.
+        // It's too late to call it here because we can't return a Result.
+        if !self.has_been_closed {
+            panic!("List writer was dropped without calling `end()`.");
+        }
+    }
+}
```
zslayton, line 450: 🗺️ This is my compromise in the absence of linear types. We can't `close()` in `Drop` because there's no way for the application to handle a failure there. However, that also means we cannot force the writer to close automatically. This at least makes contract violations noisy -- it's not possible to forget to to close the container and keep going.

src/lazy/encoder/mod.rs
```diff
@@ -0,0 +1,565 @@
...
+    fn write_f32(self, value: f32) -> IonResult<()>;
+    fn write_f64(self, value: f64) -> IonResult<()>;
+    fn write_decimal(self, value: &Decimal) -> IonResult<()>;
+    fn write_timestamp(self, value: &Timestamp) -> IonResult<()>;
+    fn write_string<A: AsRef<str>>(self, value: A) -> IonResult<()>;
+    fn write_symbol<A: AsRawSymbolTokenRef>(self, value: A) -> IonResult<()>;
+    fn write_clob<A: AsRef<[u8]>>(self, value: A) -> IonResult<()>;
+    fn write_blob<A: AsRef<[u8]>>(self, value: A) -> IonResult<()>;
+    fn list_writer(self) -> IonResult<E::ListWriter<'a>>;
+    fn sexp_writer(self) -> IonResult<E::StructWriter<'a>>;
```
zslayton, line 99: `TODO` for me: change the return type to `SExpWriter`.

src/lazy/encoder/mod.rs
```diff
@@ -0,0 +1,565 @@
...
+}
+
+impl<W: Write> LazyEncoder<W> for TextEncoding_1_0 {
+    type Writer = LazyRawTextWriter_1_0<W>;
+    type ValueWriter<'a> = TextValueWriter_1_0<'a, W> where W: 'a;
+    type AnnotatedValueWriter<'a> = TextAnnotatedValueWriter_1_0<'a, W> where W: 'a;
+    type ListWriter<'a> = TextListWriter_1_0<'a, W> where W: 'a;
+    type SExpWriter<'a> = ();
+    type StructWriter<'a> = ();
+    type EExpressionWriter<'a> = ();
```
zslayton, line 62: `TODO` for me: mark these as `TODO`.

src/lazy/encoder/mod.rs
```diff
@@ -0,0 +1,565 @@
...
+        let test = |writer: &mut LazyRawTextWriter_1_0<&mut Vec<u8>>| {
+            writer
+                .write(1)?
+                .write(false)?
+                .write(3f32)?
+                .write("foo")?
+                .write("bar".as_symbol_ref())?
+                .write(Timestamp::with_ymd(2023, 11, 9).build()?)?
+                .write([0xE0u8, 0x01, 0x00, 0xEA])?
+                .write([1, 2, 3])?;
```
zslayton, line 503: 🗺️ There are two ways to write a list: if you have a type that represents the list and your type implements `WriteAsIon[Value]`, you can pass it to `.write(...)` as is done here. For a streaming list writer that can emit heterogeneous values, you can use `ListWriter`.

If `WriteAsIon[Value]` is implemented, it would be implemented using `ListWriter`.

src/lazy/encoder/write_as_ion.rs
```diff
@@ -0,0 +1,273 @@
...
+{
+    fn write_as_ion<'a, W: Write + 'a, E: LazyEncoder<W>, V: AnnotatedValueWriter<'a, W, E>>(
+        &self,
+        writer: V,
+    ) -> IonResult<()> {
+        self.write_as_ion_value(writer.no_annotations())
+    }
+}
+
+// ===== WriteAsIonValue implementations for common types =====
```
zslayton, line 56: 🗺️ There are common types not covered here, like `Vec<T: WriteAsIon>` and several more integer types. I'd like to punt on those.

src/text/raw_text_writer.rs
zslayton, line 1: 🗺️ Changes in this file make its inner workings crate-visible so the `LazyRawTextWriter_1_0` can re-use them. Down the road, this type will be removed and I'll consolidate the logic.

src/types/null.rs
```diff
@@ -0,0 +1,7 @@
+use crate::IonType;
+
+/// An in-memory representation of a typed `null`.
+// This type allows us to define custom behavior for `null` via trait implementations.
+// For an example, see the `WriteAsIonValue` trait.
+#[derive(Debug, PartialEq, Copy, Clone)]
+pub struct Null(pub IonType);
```
zslayton, line 7: 🗺️ We have similar wrapper types in the `Element` API for `Blob(Bytes)`, `Clob(Bytes)`, `List(Sequence)` and `SExp(Sequence)`.

zslayton (commented):
I suggest reading the files in this order:
1. `write_as_ion.rs`
1. `annotate.rs`
1. `encoder/mod.rs`
1. `text/raw_text_writer.rs`

and the rest in any order.

🗺️ PR tour

src/text/raw_text_writer.rs
zslayton, line 1: 🗺️ Changes in this file make its inner workings crate-visible so the LazyRawTextWriter_1_0 can re-use them. Down the road, this type will be removed and I'll consolidate the logic.

src/lazy/encoder/mod.rs
```diff
@@ -0,0 +1,567 @@
...
+        todo!()
+    }
+
+    fn write_blob<A: AsRef<[u8]>>(mut self, value: A) -> IonResult<()> {
+        // Rust format strings escape curly braces by doubling them. The following string is:
+        // * The opening {{ from a text Ion blob, with each brace doubled to escape it.
+        // * A {} pair used by the format string to indicate where the base64-encoded bytes
+        //   should be inserted.
+        // * The closing }} from a text Ion blob, with each brace doubled to escape it.
+        write!(self.output(), "{{{{{}}}}}", base64::encode(value))?;
```
popematt, line 356: `"{{{{{}}}}}"` 😆 

Thanks for the explanatory comment.

src/lazy/encoder/mod.rs
```diff
@@ -0,0 +1,567 @@
...
+        let test = |writer: &mut LazyRawTextWriter_1_0<&mut Vec<u8>>| {
+            let mut list = writer.list_writer()?;
+            list.write(1)?
+                .write(false)?
+                .write(3f32)?
+                .write("foo")?
+                .write("bar".as_symbol_ref())?
+                .write(Timestamp::with_ymd(2023, 11, 9).build()?)?
+                .write(&[0xE0u8, 0x01, 0x00, 0xEA][..])?;
+            list.end()?;
```
popematt, line 562: Just want to confirm—we can chain `end()` after `write()`, like this, right?
```rust
list.write(1)?
     // ...
    .write(&[0xE0u8, 0x01, 0x00, 0xEA][..])?
    .end()?;
```

src/lazy/encoder/write_as_ion.rs
```diff
@@ -0,0 +1,273 @@
...
+
+impl WriteAsIonValue for Null {
+    fn write_as_ion_value<'a, W: Write + 'a, E: LazyEncoder<W>, V: ValueWriter<'a, W, E>>(
+        &self,
+        writer: V,
+    ) -> IonResult<()> {
+        writer.write_null(self.0)
+    }
+}
```
popematt, line 65: You could DRY up a bit with a macro...
```rust
macro_rules! impl_write_as_ion_value {
    ($t:ty, $e:expr) => {
        impl WriteAsIonValue for $t {
            fn write_as_ion_value<'a, W: Write + 'a, E: LazyEncoder<W>, V: ValueWriter<'a, W, E>>(
                &self,
                writer: V,
            ) -> IonResult<()> {
                $e
            }
        }
    };
}

impl_write_as_ion_value!(bool, writer.write_bool(*self));
impl_write_as_ion_value!(i32, writer.write_i64(*self as i64));
impl_write_as_ion_value!(i64, writer.write_i64(*self));
impl_write_as_ion_value!(Decimal, writer.write_decimal(self));
impl_write_as_ion_value!(Timestamp, writer.write_timestamp(self));
// etc.
```
Thoughts?

popematt (commented):
Looks pretty good. I do wonder whether these traits will work for the binary writer, but that doesn't necessarily need to be solved right now.

src/lazy/encoder/write_as_ion.rs
zslayton: I'm for using a macro, but as written this won't work due to hygiene rules. When the compiler sees `writer.write_bool(...)`, it looks for a `writer` variable in the current scope.

src/lazy/encoder/mod.rs
zslayton: Sadly, no. `write` returns a `&mut Self` and `end` takes (moves) `self`. The alternative is for `write` to _also_ take `self` and return `self`, so we have to choose between one of these:

**Option 1**
```rust
fn write(&mut self) -> IonResult<&mut Self> {...}

// Writing part of a list in a single statement looks nice...
let mut list = writer.list_writer()?;
list.write(1)?
     .write(2)?
     .write(3)?;

//...but we can't write the whole list in a single statement. `end` is on its own line.
let mut list = writer.list_writer()?;
list.write(1)?
    .write(2)?
    .write(3)?
list.end()?;
```

**Option 2**
```rust
fn write(self) -> IonResult<Self> {...}

// Writing the entire list in a single statement looks nice...
writer.list_writer()?
    .write(1)?
     .write(2)?
     .write(3)?
     .end()?;

//...but writing part of a list in a statement requires an extra `list =` prefix
let mut list = writer.list_writer()?;
list = list.write(1)?
    .write(2)?
    .write(3)?
list.end()?;
```

I went with option 1 partly because option 2 makes writing list values in a `for` loop (presumably common) require the extra `list = `.

Open to other ideas! To see another case in which a crate struggled with this tradeoff, see [the `README` for `derive_builder`](https://docs.rs/derive_builder/latest/derive_builder/#owned-aka-consuming).

src/lazy/encoder/annotate.rs
```diff
@@ -0,0 +1,73 @@
...
+    /// writer.write(42_usize.annotate(&["foo", "bar", "baz"]))?.flush()?;
+    ///
+    /// let expected = Element::read_one("foo::bar::baz::42")?;
+    /// let actual = Element::read_one(&buffer)?;
+    ///
+    /// assert!(IonData::eq(&expected, &actual));
+    ///# Ok(())
+    ///# }
+    /// ```
+    fn annotate<'a, A: AsRawSymbolTokenRef>(
```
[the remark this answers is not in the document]
zslayton: Good point! Unfortunately, `with_annotations` is already being used by the `Element` API and constructs a materialized `Element` tree with the given annotations set. (That is: `5.with_annotations(...)` already has a meaning.)

I can do `5.annotate_with(&["foo", "bar"])`, but the distinction will end up being pretty subtle. Thoughts?

src/lazy/encoder/mod.rs
```diff
@@ -0,0 +1,565 @@
...
+pub trait LazyEncoder<W: Write>: 'static + Sized + Debug + Clone + Copy {
+    // XXX: ^-- This is named 'Lazy' for symmetry with the `LazyDecoder`. In reality, there's nothing
+    //      lazy about it. We should revisit the Lazy* naming scheme, as eventually it will be the
+    //      only implementation of a reader and won't need the word 'Lazy' to distinguish itself.
+
+    /// A writer that serializes Rust values as Ion, emitting the resulting data to an implementation
+    /// of [`Write`].
+    type Writer: LazyRawWriter<W, Self>;
+
+    /// A single-use type that can emit an Ion value.
```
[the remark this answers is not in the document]
zslayton: `ValueWriter` consumes `self`, so whatever method you call on it will prevent it from being used a second time.

src/lazy/encoder/mod.rs
```diff
@@ -0,0 +1,565 @@
...
+    fn write_f64(self, value: f64) -> IonResult<()>;
+    fn write_decimal(self, value: &Decimal) -> IonResult<()>;
+    fn write_timestamp(self, value: &Timestamp) -> IonResult<()>;
+    fn write_string<A: AsRef<str>>(self, value: A) -> IonResult<()>;
+    fn write_symbol<A: AsRawSymbolTokenRef>(self, value: A) -> IonResult<()>;
+    fn write_clob<A: AsRef<[u8]>>(self, value: A) -> IonResult<()>;
+    fn write_blob<A: AsRef<[u8]>>(self, value: A) -> IonResult<()>;
+    fn list_writer(self) -> IonResult<E::ListWriter<'a>>;
+    fn sexp_writer(self) -> IonResult<E::StructWriter<'a>>;
+    fn struct_writer(self) -> IonResult<E::StructWriter<'a>>;
```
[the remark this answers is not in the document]
zslayton: Punting for the moment.

src/lazy/encoder/mod.rs
popematt: Thanks for the explanation. I agree that option 1 is better.

src/lazy/encoder/annotate.rs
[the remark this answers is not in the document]
zslayton: Opened #681 to track this.

popematt (approved):

PR #684, opened 2023-11-19
Adds sexp and struct writers to the `LazyRawTextWriter_1_0`

zslayton:
* Introduces a `TextContainerWriter_1_0` that houses information and behavior common to the container writer.
* Adds an s-expression writer, `TextSExpWriter_1_0`.
* Adds a struct writer, `TextStructWriter_1_0`.

### DIFF
--- a/src/lazy/encoder/mod.rs
+++ b/src/lazy/encoder/mod.rs
@@ -559,7 +559,6 @@ impl<'a, W: Write, E: LazyEncoder<W>> SequenceWriter<'a, W, E> for TextSExpWrite
     }
 }
 
-
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


src/lazy/encoder/mod.rs
```diff
@@ -564,4 +708,64 @@ mod tests {
         };
         writer_test(expected, test)
     }
+
+    #[test]
+    fn write_sexp() -> IonResult<()> {
```
popematt, line 713: Not urgent, but it would be good to have a test case demonstrating that multiple levels of nested containers will work properly.

popematt (approved):

Issue #353, opened 2022-02-02
Fix the CI pipeline for windows-latest


zslayton:
Last month [GitHub introduced a new virtual environment](https://github.blog/changelog/2022-01-11-github-actions-jobs-running-on-windows-latest-are-now-running-on-windows-server-2022/), `windows-2022`, and updated the `windows-latest` alias to point to it. Our pipeline worked for `windows-2019` but [fails now that the alias has been redirected](https://github.com/amzn/ion-rust/runs/5043208962?check_suite_focus=true):

![image](https://user-images.githubusercontent.com/611616/152241774-132257c7-ae3c-4a05-bf7b-0f8210fbef69.png)

I've modified the GitHub Action configuration to point to `windows-2019` for now, but we should make it work for both platforms.

zslayton:
Fixed in #688.

PR #690, opened 2023-12-19
Implements reading/writing `FlexInt`, `FlexUInt`

zslayton:
Implements reading and writing of Ion 1.1 encoding primitives (`FlexInt`, `FlexUInt`).

Also adds benchmarks comparing the time needed to read/write 10k integers with that of the corresponding Ion 1.0 encoding primitives (`VarInt`, `VarUInt`). Benchmark results on my 2019 16" Intel MBP:
image


.github/workflows/miri.yml
```diff
@@ -11,7 +11,6 @@ jobs:
     if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != 'amazon-ion/ion-rust'
     strategy:
       matrix:
-        # See https://github.com/amazon-ion/ion-rust/issues/353
```
zslayton, line 14: 🗺️ Issue #353 was fixed in #688, which changed the CI task back to `windows-latest`.

.github/workflows/rust.yml
```diff
@@ -18,10 +18,6 @@ jobs:
       checks: write
 
     steps:
-      - name: Remove MSys64 MingW64 Binaries
-        if: runner.os == 'Windows'
-        # remove this because there is a bad libclang.dll that confuses bindgen
-        run: Remove-Item -LiteralPath "C:\msys64\mingw64\bin" -Force -Recurse
```
zslayton, line 24: 🗺️ This was necessary for `ion-c-sys`, which was removed.

benches/encoding_primitives.rs
```diff
@@ -0,0 +1,213 @@
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use rand::prelude::StdRng;
+use rand::{distributions::Uniform, Rng, SeedableRng};
+use std::io;
```
zslayton, line 4: 🗺️ This file is a suite of benchmarks comparing the time needed to read or write 10k random integers using `VarUInt`, `VarInt`, `FlexUInt`, or `FlexInt` encoding.

src/binary/non_blocking/type_descriptor.rs
```diff
@@ -9,7 +9,7 @@ use crate::IonType;
 /// [Typed Value Formats](https://amazon-ion.github.io/ion-docs/docs/binary.html#typed-value-formats)
 /// section of the binary Ion spec.
 #[derive(Copy, Clone, Debug, PartialEq)]
-pub(crate) struct TypeDescriptor {
+pub struct TypeDescriptor {
```
zslayton, line 12: 🗺️ This PR makes a couple of types `pub` while leaving their module `pub(crate)` or even private. A later change in this diff re-exports these `pub` types if the `experimental-lazy-reader` feature is enabled, allowing the benchmarks to use some of these implementation details.

src/lib.rs
```diff
@@ -242,6 +242,16 @@ pub use {
...
+// Exposed to allow benchmark comparisons between the 1.0 primitives and 1.1 primitives
+#[cfg(feature = "experimental-lazy-reader")]
+pub use {
+    binary::int::DecodedInt, binary::non_blocking::type_descriptor::Header,
+    binary::uint::DecodedUInt, binary::var_int::VarInt, binary::var_uint::VarUInt,
+    lazy::binary::immutable_buffer::ImmutableBuffer,
+    lazy::encoder::binary::v1_1::flex_int::FlexInt,
+    lazy::encoder::binary::v1_1::flex_uint::FlexUInt,
+};
+
```
zslayton, line 254: 🗺️ Feature-gated re-exports of encoding-primitive-related types.

src/lazy/binary/immutable_buffer.rs
```diff
@@ -157,6 +159,257 @@ impl<'a> ImmutableBuffer<'a> {
...
+    fn read_small_flex_uint(bytes: &[u8]) -> FlexUInt {
+        debug_assert!(bytes.len() >= 8);
+        let num_encoded_bytes = bytes[0].trailing_zeros() as usize + 1;
+        let num_encoded_bits = 8 * num_encoded_bytes;
+        // Get a mask with the low 'n' bits set
+        // TODO: Should this be a const cache of num_encoded_bits -> mask?
+        let mask = 1u64
+            .checked_shl(num_encoded_bits as u32)
+            .map(|v| v - 1)
+            .unwrap_or(u64::MAX);
```
zslayton, line 224: 🗺️ I think making a `const` cache for this would get us a modest speedup, but the performance is already much better than `VarUInt`.

zslayton (commented):
🗺️ PR tour

src/lazy/binary/immutable_buffer.rs
```diff
@@ -157,6 +159,257 @@ impl<'a> ImmutableBuffer<'a> {
...
+        self,
+        support_sign_extension: bool,
+        bytes: &'a [u8],
+        second_byte: u8,
+    ) -> IonResult<FlexUInt> {
+        // There are 10 prefix (continuation) bits, 64 bits of magnitude, and 6 bits of sign
+        // extension (if enabled). We cannot store the highest 6 bits, so this method just checks
+        // to make sure that they do not modify the meaning of the value in the lower 64 bits.
+        // For signed values, this means the 6 extra bits must all be the same as the 64th bit.
+        // For unsigned values, this means that the 6 extra bits must all be `1`.
```
popematt, line 357: Hold on... why are the extra bits always `1` for unsigned values? Why not `0`?

src/lazy/encoder/binary/v1_1/flex_uint.rs
```diff
@@ -0,0 +1,229 @@
...
+                let mut buffer: [u8; 9] = [0; 9];
+
+                // The first byte will always be 0x00, indicating that 8 more bytes follow.
+                //
+                // We need to leave a `1` in the low bit of the second byte to be the End flag. Because
+                // we need fewer than 64 bits for magnitude, we can encode the remainder of the data
+                // in a u64.
+                let encoded_value = (value << 1) + 1; // Leave a trailing `1` in the lowest bit
+                buffer[1..].copy_from_slice(&encoded_value.to_le_bytes()[..]);
+                output.write_all(buffer.as_slice())?;
```
popematt, line 77: Does it really make a big difference being able to write all 9 bytes at one time? It seems to me that it would be simpler to implement like this:

```rust
const zero: [u8; 1] = [0];
output.write_all(&zero)?;
let encoded_value = (value << 1) + 1;
output.write_all(&encoded_value.to_le_bytes()[..])?;
```

src/lazy/encoder/binary/v1_1/flex_uint.rs
```diff
@@ -0,0 +1,229 @@
...
+                // the whole value is 10 bytes long. We can fit 6 bits of magnitude in this second byte.
+                let second_byte = ((value & 0b111111) << 2) as u8 | 0b10u8;
+                buffer[1] = second_byte;
+
+                // The remaining 58 bits of magnitude can be encoded in a u64.
+                let remaining_magnitude: u64 = value >> 6;
+                buffer[2..].copy_from_slice(&remaining_magnitude.to_le_bytes()[..]);
+
+                // Call `write_all()` once with our complete encoding.
+                output.write_all(buffer.as_slice()).unwrap();
```
popematt, line 95:
```suggestion
                output.write_all(buffer.as_slice())?;
```

src/lazy/encoder/binary/v1_1/flex_uint.rs
```diff
@@ -0,0 +1,229 @@
...
+            _ => unreachable!(
+                "write_large_u64() is only called for values whose encoded size is 9 or 10 bytes"
+            ),
+        }
+    }
+
+    pub fn value(&self) -> u64 {
+        self.value
+    }
+    pub fn size_in_bytes(&self) -> usize {
```
popematt, line 107: minor nitpick
```suggestion
    }

    pub fn size_in_bytes(&self) -> usize {
```

src/lazy/encoder/binary/v1_1/flex_uint.rs
```diff
@@ -0,0 +1,229 @@
...
+    size_in_bytes: usize,
+}
+
+impl FlexUInt {
+    pub fn new(size_in_bytes: usize, value: u64) -> Self {
+        Self {
+            value,
+            size_in_bytes,
+        }
+    }
```
popematt, line 37: Considering that the `write_u64` function does not have `&self` as one of it's parameters, is there really a need to have instances of `FlexUInt`?

src/lazy/encoder/binary/v1_1/flex_uint.rs
```diff
@@ -0,0 +1,229 @@
...
+
+    #[inline]
+    pub fn write_u64<W: Write>(output: &mut W, value: u64) -> IonResult<usize> {
```
popematt, line 40: Is there any case where we would write a `FlexUInt` from something smaller than a `u64`? If so, it may be worth adding a second function for `u32` since anything `u32` or smaller can be written without branching.

src/lazy/binary/immutable_buffer.rs
```diff
@@ -157,6 +159,257 @@ impl<'a> ImmutableBuffer<'a> {
...
+
+        Ok((
+            FlexInt::new(flex_uint.size_in_bytes(), signed_value),
+            remaining,
+        ))
+    }
+
+    /// Reads a [`FlexUInt`] from the buffer.
+    #[inline]
+    pub fn read_flex_uint(self) -> ParseResult<'a, FlexUInt> {
```
popematt, line 192: minor organizational suggestion—it took me a little bit to figure out why `FlexUInt` needed data or a constructor, since the `write_u64` function doesn't use instances of `FlexUInt` at all. Consider extracting the decoding logic into `FlexUInt` so that we can keep related things close together.

E.g.: add `read(&[u8]) -> SomeResult<FlexUInt>` to `FlexUInt`, and then this function can be reduced to:
```rust
pub fn read_flex_uint(self) -> ParseResult<'a, FlexUInt> {
    let flex_uint = FlexUInt::read(self.bytes())?;
    let remaining = self.consume(flex_uint.size_in_bytes());
    Ok((flex_uint, remaining))
}
```

src/lazy/binary/immutable_buffer.rs
```diff
@@ -157,6 +159,257 @@ impl<'a> ImmutableBuffer<'a> {
         }
     }
 
+    /// Reads a [`FlexInt`] from the buffer.
+    pub fn read_flex_int(self) -> ParseResult<'a, FlexInt> {
```
popematt, line 163: See comment on `read_flex_uint()`. I think we could apply the same thing here.

src/lazy/binary/immutable_buffer.rs
```diff
@@ -157,6 +159,257 @@ impl<'a> ImmutableBuffer<'a> {
...
+        // We need to inspect the second byte to see how many more prefix bits there are.
+        if bytes_available < 2 {
+            return incomplete();
+        }
+        let second_byte = bytes[1];
+
+        if second_byte & 0b11 == 0b00 {
+            // The flag bits in the second byte indicate at least two more bytes, meaning the total
+            // length is more than 10 bytes. We're not equipped to handle this.
+            return IonResult::decoding_error("found a >10 byte VarUInt too large to fit in a u64");
```
popematt, line 310:
```suggestion
            return IonResult::decoding_error("found a >10 byte FlexUInt too large to fit in a u64");
```
...although this will still be an inaccurate message if you try to read a `FlexInt` (which delegates to this implementation) of more than 10 bytes.

src/lazy/encoder/binary/v1_1/flex_uint.rs
```diff
@@ -0,0 +1,229 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::lazy::binary::immutable_buffer::ImmutableBuffer;
+    use crate::lazy::encoder::binary::v1_1::flex_uint::FlexUInt;
+    use crate::{IonError, IonResult};
+
+    const FLEX_UINT_TEST_CASES: &[(u64, &[u8])] = &[
+        (0u64, &[0b00000001]),
```
popematt, line 119: The `u64` suffix is unnecessary because it is specified in `&[(u64, &[u8])]`.

src/lazy/encoder/binary/v1_1/flex_uint.rs
```diff
@@ -0,0 +1,229 @@
...
+            9223372036854775807u64,
+            &[
+                0b00000000, 0b11111111, 0b11111111, 0b11111111, 0b11111111, 0b11111111, 0b11111111,
+                0b11111111, 0b11111111,
+            ],
+        ),
+    ];
+
+    #[test]
+    fn decode_flex_uint() -> IonResult<()> {
```
popematt, line 195: Can you add test cases for overpadded FlexUInts (such as `&[0b00000110, 0b00000000]`)—or are we just going to say that FlexUInts must be encoded with the smallest possible number of bytes?

src/lazy/encoder/binary/v1_1/flex_uint.rs
zslayton: I adapted this impl from a prototype implementation I had lying around. In benchmarking the prototype, I found that avoiding two calls to `write_all` was a minor net benefit because it avoided the additional error handling branches involved. I agree that the version you wrote is simpler, and this code path may not benchmark the same way as my prototype.

I'll try the simpler impl and see if it moves the needle either way in the benchmarks.

popematt: Actually, as I've looked at it more, I've realized that the cost of copying 8 bytes is probably less than the cost of a function call, and there's not much difference in terms of readability. Don't spend more than a few minutes trying it out if you try it at all.

src/lazy/binary/immutable_buffer.rs
zslayton: Oops! The code does the right thing, but this comment got it backwards. Will fix, thanks.

src/lazy/encoder/binary/v1_1/flex_uint.rs
zslayton: > or are we just going to say that FlexUInts must be encoded with the smallest possible number of bytes?

This is as of yet an unresolved question, though I think there are advocates for allowing padding. I'll add some tests for it.

src/lazy/encoder/binary/v1_1/flex_uint.rs
zslayton: Yes, though I'd like to punt on this for the sake of this PR. Tracking it in #693.

src/lazy/binary/immutable_buffer.rs
zslayton: Done in commit c572b6d7c81a2ccdfd06f5d025aa2effa120d8f1.

popematt (approved):

PR #697, opened 2024-01-16
Implements writing v1.1 decimals

zslayton:

### DIFF
--- a/src/lazy/encoder/binary/v1_1/value_writer.rs
+++ b/src/lazy/encoder/binary/v1_1/value_writer.rs
@@ -172,20 +172,23 @@ impl<'value, 'top> BinaryValueWriter_1_1<'value, 'top> {
         let opcode_index = self.encoding_buffer.len();
         self.push_byte(0x60);
 
-        // A positive zero coefficient, unspecified exponent
+        // Whether the decimal has a positive zero coefficient (of any exponent). This value is needed in two places
+        // and is non-trivial, so we compute it up front and store the result.
         let is_positive_zero = value.coefficient().is_positive_zero();
 
+        // If the value is 0.0, then the encoding has no body. The 0x60 opcode is the complete encoding.
         if value.exponent() == 0 && is_positive_zero {
-            // The 0x60 opcode is the complete encoding.
             return Ok(());
         }
 
+        // For any value that is not 0.0, the encoding begins with a FlexInt representing the exponent.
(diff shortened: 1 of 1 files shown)


src/lazy/encoder/binary/v1_1/value_writer.rs
```diff
@@ -161,8 +166,69 @@ impl<'value, 'top> BinaryValueWriter_1_1<'value, 'top> {
         Ok(())
     }
 
-    pub fn write_decimal(self, _value: &Decimal) -> IonResult<()> {
-        todo!()
+    pub fn write_decimal(mut self, value: &Decimal) -> IonResult<()> {
+        // Insert a placeholder opcode; we'll overwrite the length nibble with the appropriate value when the encoding
+        // is complete.
+        let opcode_index = self.encoding_buffer.len();
+        self.push_byte(0x60);
```
popematt, line 173: How expensive is it to splice into the array? (Is it common enough for us to even care?) Is there any particular reason we can't calculate the size using `FlexInt::encoded_size_i64` for the exponent and something like `FixedInt::encoded_size_i64` for the coefficient so that we can write the correct length to begin with?

Maybe answering these questions is something that we punt for later if we don't think it's going to be very common.

popematt (approved):

src/lazy/encoder/binary/v1_1/value_writer.rs
zslayton: > How expensive is it to splice into the array?

It can (potentially) involve an allocation.

> (Is it common enough for us to even care?)

I don't think I've ever seen a 16+ byte decimal in the wild, only in tests.

>  Is there any particular reason we can't calculate the size using FlexInt::encoded_size_i64 for the exponent and something like FixedInt::encoded_size_i64 for the coefficient so that we can write the correct length to begin with?

We could do this, though we'd need to support writing `FixedInt`s that were `BigInt` under the hood (not just `i64`). Estimating a `BigInt`'s size is weirdly tricky -- going through [the docs](https://docs.rs/num-bigint/latest/num_bigint/struct.BigInt.html), the only way I could see to do it was to call `to_bytes_le()` (which allocates a new `Vec`) and check its length. This felt similar in cost to just splicing, so I didn't look into it much further.

Since I think this doesn't come up much, I'd like to punt on optimizing it until I find out why I'm wrong. :P

PR #699, opened 2024-01-19
Implements writing v1.1 timestamps

zslayton:
While this diff looks big, some ~900 of the ~1,200 lines are just generously spaced/commented unit tests.

### DIFF
--- a/src/lazy/encoder/binary/v1_1/value_writer.rs
+++ b/src/lazy/encoder/binary/v1_1/value_writer.rs
@@ -172,20 +172,23 @@ impl<'value, 'top> BinaryValueWriter_1_1<'value, 'top> {
         let opcode_index = self.encoding_buffer.len();
         self.push_byte(0x60);
 
-        // A positive zero coefficient, unspecified exponent
+        // Whether the decimal has a positive zero coefficient (of any exponent). This value is needed in two places
+        // and is non-trivial, so we compute it up front and store the result.
         let is_positive_zero = value.coefficient().is_positive_zero();
 
+        // If the value is 0.0, then the encoding has no body. The 0x60 opcode is the complete encoding.
         if value.exponent() == 0 && is_positive_zero {
-            // The 0x60 opcode is the complete encoding.
             return Ok(());
         }
 
+        // For any value that is not 0.0, the encoding begins with a FlexInt representing the exponent.
         let encoded_exponent_size = FlexInt::write_i64(self.encoding_buffer, value.exponent())?;
+
         let encoded_coefficient_size = if is_positive_zero {
-            // Write nothing; an implicit zero coefficient is positive.
+            // If the coefficient is zero but the exponent was non-zero, write nothing; an implicit zero is positive.
             0
         } else if value.coefficient.is_negative_zero() {
-            // Write a zero byte; an explicit zero is negative.
+            // If the coefficient is negative zero (of any exponent), write a zero byte; an explicit zero is negative.
             self.push_byte(0x00);
             1
         } else {
@@ -852,7 +855,7 @@ mod tests {
             ),
             (Decimal::new(7, 4), &[0x62, 0x09, 0x07]),
             (
-                // ~PI
+                // ~Pi
                 Decimal::new(3_1415926535i64, -10),
                 &[0x66, 0xED, 0x07, 0xFF, 0x88, 0x50, 0x07],
             ),
--- a/src/lazy/encoder/binary/v1_1/value_writer.rs
+++ b/src/lazy/encoder/binary/v1_1/value_writer.rs
@@ -231,8 +231,156 @@ impl<'value, 'top> BinaryValueWriter_1_1<'value, 'top> {
         Ok(())
     }
 
-    pub fn write_timestamp(self, _value: &Timestamp) -> IonResult<()> {
-        todo!()
+    pub fn write_timestamp(mut self, value: &Timestamp) -> IonResult<()> {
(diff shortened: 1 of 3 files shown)


src/types/timestamp.rs
```diff
@@ -604,19 +604,30 @@ impl Timestamp {
...
-    /// (e.g. a timestamp with fractional seconds of `0.000000000999` would be returned as `0`)
+    /// (e.g. a timestamp with fractional seconds of `0.000000000999` would return `0`)
     pub fn nanoseconds(&self) -> u32 {
         self.fractional_seconds_as_nanoseconds().unwrap_or_default()
     }
 
+    /// Returns this Timestamp's fractional seconds in microseconds
+    ///
+    /// NOTE: This is a potentially lossy operation. A Timestamp with picoseconds would return a
+    /// number of microseconds, losing precision. If it loses precision then truncation is preformed.
```
popematt, line 615:
```suggestion
    /// number of microseconds, losing precision. If it loses precision then truncation is performed.
```
This spelling mistake is present in the `nanoseconds` and `milliseconds` doc comments as well.

src/lazy/encoder/binary/v1_1/value_writer.rs
```diff
@@ -231,8 +232,253 @@ impl<'value, 'top> BinaryValueWriter_1_1<'value, 'top> {
...
+        // reader), we calculate a mask of which bits are relevant to the current opcode and set any bits not in use
+        // to `0`.
+        // TODO: Remove this logic pending resolution of https://github.com/amazon-ion/ion-docs/issues/295, which
+        //       suggests requiring readers to ignore bits not used by the specified opcode
+        let mask = 1u64
+            .checked_shl(bits_populated)
+            .unwrap_or(0)
+            .checked_sub(1)
+            .unwrap_or(u64::MAX);
+        encoding &= mask;
```
popematt, line 361: If we had a If we had a precomputed `const SHORT_TS_BIT_MASKS: [u64;65] = ...`, then this could be simplified to...
```rust
encoding &= SHORT_TS_BIT_MASKS[bits_populated];
```
...which might be cheap enough that we don't need to care about the performance aspect of ion-docs/295.

src/lazy/encoder/binary/v1_1/value_writer.rs
```diff
@@ -231,8 +232,253 @@ impl<'value, 'top> BinaryValueWriter_1_1<'value, 'top> {
...
+        let precision = value.precision();
+        let scale = value.fractional_seconds_scale().unwrap_or(0);
+
+        // The encoded bytelength of all components *except* subseconds.
+        let (mut encoded_length, num_bits_in_use) = match precision {
+            Year => (2, 14),
+            Month => (3, MONTH_BIT_OFFSET + NUM_MONTH_BITS),
+            Day => {
+                let bits_in_use = DAY_BIT_OFFSET + NUM_DAY_BITS;
+                // Set the "day-not-month" bit, just beyond the day bits.
```
popematt, line 428: I'm not sure I understand why this is necessary. If there's "day" precision, then the day will already be non-zero. If anything, shouldn't we be zero-ing the day if it's month precision?

src/lazy/encoder/binary/v1_1/value_writer.rs
```diff
@@ -231,8 +232,253 @@ impl<'value, 'top> BinaryValueWriter_1_1<'value, 'top> {
...
+        // irrelevant to the final encoding. To simplify unit testing (and in the current absence of a binary 1.1
+        // reader), we calculate a mask of which bits are relevant to the current opcode and set any bits not in use
+        // to `0`.
+        // TODO: Remove this logic pending resolution of https://github.com/amazon-ion/ion-docs/issues/295, which
+        //       suggests requiring readers to ignore bits not used by the specified opcode
+        let mask = 1u64
+            .checked_shl(num_bits_in_use)
+            .unwrap_or(0)
+            .checked_sub(1)
+            .unwrap_or(u64::MAX);
```
popematt, line 447: FYI, this could be replaced with [`wrapping_sub`](https://doc.rust-lang.org/stable/std/primitive.u64.html#method.wrapping_sub).

zslayton: Oh yeah, good call 👌

src/lazy/encoder/binary/v1_1/value_writer.rs
zslayton: I agree, but I'd like to leave this for a future PR.

src/lazy/encoder/binary/v1_1/value_writer.rs
zslayton: In the same spirit, I tried to replace `checked_shl`/`unwrap_or` with `wrapping_shl` and learned that trying to left shift by more bits than the type uses is undefined behavior (and not just `0`). Wild.

src/lazy/encoder/binary/v1_1/value_writer.rs
zslayton: > If anything, shouldn't we be zero-ing the day if it's month precision?

This is related to the question raised by https://github.com/amazon-ion/ion-docs/issues/295; if the writer MUST zero out unused bits, then the reader could use a `day` of `0` as an indication. However, if the reader MUST ignore unused bits, then we need a dedicated bit.

I've added your question to that issue and made the issue more general.

tgregg (approved):

PR #700, opened 2024-01-22
Adds implementation for reader builder with catalog

desaikd:
*Description of changes:*
This PR adds implementation of reader builder which allows adding catalog information to the reader. 

*List of changes:*
- Changes `ReaderBuilder` to store catalog information
- adds `with_catalog()` to add catalog information to builder

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -34,9 +34,9 @@ impl ReaderBuilder {
     }
 
     /// Constructs a [ReaderBuilder] with catalog information.
-    pub fn with_catalog(self, catalog: Box<dyn Catalog>) -> ReaderBuilder {
+    pub fn with_catalog(self, catalog: impl Catalog + 'static) -> ReaderBuilder {
         ReaderBuilder {
-            catalog: Some(catalog),
+            catalog: Some(Box::new(catalog)),
         }
     }
 
@@ -316,7 +316,9 @@ mod tests {
     use crate::binary::constants::v1_0::IVM;
     use crate::reader::{BlockingRawBinaryReader, StreamItem::Value};
 
+    use crate::catalog::MapCatalog;
     use crate::result::IonResult;
+    use crate::shared_symbol_table::SharedSymbolTable;
     use crate::types::IonType;
 
     type TestDataSource = io::Cursor<Vec<u8>>;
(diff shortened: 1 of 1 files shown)


zslayton (commented):
This looks good! We'll need unit tests showing that the reader successfully uses the catalog when it encounters a shared symbol table. 

As a `TODO`, we also need to add API support for this to the lazy reader. After this PR, `Element` serialization will be done using the lazy writer API while `Element` deserialization will be done using the non-lazy reader. Eventually we'll need everything on the same API so we can remove the old implementations.

src/reader.rs
```diff
@@ -20,13 +21,22 @@ use std::fmt::{Display, Formatter};
 
 use crate::types::Str;
 /// Configures and constructs new instances of [Reader].
-pub struct ReaderBuilder {}
+pub struct ReaderBuilder {
+    // This will be set to to `Some` catalog when the reader builder is created with catalog information.
+    // Default value for this field will be set to `None`.
+    catalog: Option<Box<dyn Catalog>>,
```
zslayton, line 27: I think we can use [the `EmptyCatalog` type](https://github.com/amazon-ion/ion-rust/blob/ef641f48cdbc4da2ba947ae02edd540bdd12ef98/src/catalog.rs#L71) you wrote as the default instead of `None`. Because `EmptyCatalog` is a zero-sized type, `Box::new(EmptyCatalog)` won't actually allocate.

zslayton (approved):

src/reader.rs
desaikd: I tried using that but the issue with using `EmptyCatalog` was we can't distinguish between a reader with an actually empty catalog vs a reader without a catalog. Which is why I chose the `Option` here. This distinction will be useful for creating the reader from builder in `make_*_reader()`.

zslayton: Can't you replace:
```rust
    fn make_binary_reader<'a, I: 'a + IonDataSource>(self, data: I) -> IonResult<Reader<'a>> {
        let raw_reader = Box::new(BlockingRawBinaryReader::new(data)?);
        if let Some(catalog) = self.catalog {
            return Ok(Reader::with_catalog(raw_reader, catalog));
        }
        Ok(Reader::new(raw_reader))
    }
```
with:
```rust
    fn make_binary_reader<'a, I: 'a + IonDataSource>(self, data: I) -> IonResult<Reader<'a>> {
        let raw_reader = Box::new(BlockingRawBinaryReader::new(data)?);
        Ok(Reader::with_catalog(raw_reader, self.catalog))
    }
```
or am I missing something?

desaikd: Thats right. My bad. I thought this distinction was required at system reader level but even that uses empty catalog as default.

PR #706, opened 2024-02-02
Fix to support finding an IVM after a NOP in binary v1.0

zslayton:
Currently, the reader will fail if an IVM is encountered following a NOP pad at the top level. This patch fixes that.

### DIFF
--- a/src/binary/non_blocking/raw_binary_reader.rs
+++ b/src/binary/non_blocking/raw_binary_reader.rs
@@ -1218,7 +1218,8 @@ impl<'a, A: AsRef<[u8]>> TxReader<'a, A> {
         if type_descriptor.is_nop() {
             if let Some(item) = self.consume_nop_padding(&mut type_descriptor)? {
                 // We may encounter the end of the file or container while reading NOP padding,
-                // in which case `item` will be RawStreamItem::Nothing.
+                // in which case `item` will be RawStreamItem::Nothing. At the top level, it is
+                // possible to encounter an IVM at the end of the NOP pad.
                 return Ok(item);
             }
             // Note that if `consume_nop_padding` reads NOP bytes but doesn't hit EOF, it will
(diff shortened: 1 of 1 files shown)


src/binary/non_blocking/raw_binary_reader.rs
```diff
@@ -1218,7 +1218,8 @@ impl<'a, A: AsRef<[u8]>> TxReader<'a, A> {
         if type_descriptor.is_nop() {
             if let Some(item) = self.consume_nop_padding(&mut type_descriptor)? {
                 // We may encounter the end of the file or container while reading NOP padding,
-                // in which case `item` will be RawStreamItem::Nothing.
+                // in which case `item` will be RawStreamItem::Nothing. At the top level, it is
```
rmarrowstone, line 1221: Can't suggest change outside of diff, but looks like the fn comment needs updating.

src/binary/non_blocking/raw_binary_reader.rs
```diff
@@ -1361,6 +1362,9 @@ impl<'a, A: AsRef<[u8]>> TxReader<'a, A> {
                 return Ok(Some(RawStreamItem::Nothing));
             }
             *type_descriptor = self.tx_buffer.peek_type_descriptor()?;
+            if type_descriptor.is_ivm_start() {
+                return self.read_ivm().map(Some);
```
rmarrowstone, line 1366: I assume IVM at non-top-level is caught elsewhere?

rmarrowstone (approved):

JustinDaleGray (approved):

src/binary/non_blocking/raw_binary_reader.rs
zslayton: Yeah, `read_ivm()` returns an error if the reader isn't at the top level.

PR #708, opened 2024-02-09
Fixes ivm-after-nop in the lazy reader

zslayton:
#706 fixed the raw binary reader (which is used in the `ion-cli` and `Element` API) to allow it to recognize an IVM following a NOP pad. This PR makes the same change but for the lazy binary reader that will eventually replace the existing reader.

### DIFF
--- a/src/lazy/binary/raw/reader.rs
+++ b/src/lazy/binary/raw/reader.rs
@@ -70,14 +70,25 @@ impl<'data> LazyRawBinaryReader<'data> {
     where
         'data: 'top,
     {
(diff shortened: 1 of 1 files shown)


desaikd (approved):
Looks good! 
Minor note: Clippy suggests some warning for IVM constant, although it doesn't seem to be related to this change. We might need to suppress the warning by putting under a feature probably.

zslayton:
> Looks good! Minor note: Clippy suggests some warning for IVM constant, although it doesn't seem to be related to this change. We might need to suppress the warning by putting under a feature probably.

Here's a small follow-on PR to address that: #710

PR #709, opened 2024-02-12
Implements writing length-prefixed and delimited structs

zslayton:
This PR diverges from the specification-as-written to align with ion-java's implementation.

Rather than having two separate opcodes for structs with and without inline field names, we only use `0xC_` and start each struct using `FlexUInt` symbol ID field names. If/when a field name needs to be written with inline text, the write emits a `0x00` byte to communicate to the reader that we are switching from `FlexUInt` field names to `FlexSym` field names for all field names that follow.

This allows Ion 1.1 structs that do not leverage inline field names to be as compact as Ion 1.0 structs while preserving more opcode space for other uses. It also obviates the need for writers to guess which struct encoding they'll need at the outset of the struct since they'll be able to "upgrade" the struct's encoding on demand.

### DIFF
--- a/src/lazy/encoder/binary/v1_1/container_writers.rs
+++ b/src/lazy/encoder/binary/v1_1/container_writers.rs
@@ -1,5 +1,3 @@
-use std::mem;
-
 use bumpalo::collections::Vec as BumpVec;
 use bumpalo::Bump as BumpAllocator;
 use delegate::delegate;
@@ -9,7 +7,7 @@ use crate::lazy::encoder::value_writer::internal::MakeValueWriter;
 use crate::lazy::encoder::value_writer::{SequenceWriter, StructWriter};
 use crate::lazy::encoder::write_as_ion::WriteAsIon;
 use crate::raw_symbol_token_ref::AsRawSymbolTokenRef;
-use crate::{FlexUInt, IonResult, RawSymbolTokenRef};
+use crate::IonResult;
 
 /// A helper type that holds fields and logic that is common to [`BinaryListWriter_1_1`],
 /// [`BinarySExpWriter_1_1`], and [`BinaryStructWriter_1_1`].
--- a/src/lazy/encoder/binary/v1_1/container_writers.rs
+++ b/src/lazy/encoder/binary/v1_1/container_writers.rs
@@ -1,3 +1,5 @@
+use std::mem;
+
 use bumpalo::collections::Vec as BumpVec;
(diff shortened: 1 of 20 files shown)


nirosys (approved):

src/lazy/encoder/binary/v1_1/container_writers.rs
```diff
@@ -110,20 +110,76 @@ impl<'value, 'top> SequenceWriter for BinarySExpWriter_1_1<'value, 'top> {
...
+            flex_uint_encoding: true,
+            container_writer,
+        }
+    }
+
+    fn enable_flex_sym_encoding(&mut self) {
+        if self.flex_uint_encoding {
+            // Write a zero byte out to signal to future readers that we are switching from
+            // FlexUInt to FlexSym.
+            self.container_writer.buffer().push(0x00);
```
tgregg, line 134: In Java we ended up using FlexInt zero (0x01) instead of 0x00. See https://github.com/amazon-ion/ion-java/pull/673/files

tgregg: In Java, if nothing had been written yet we swapped out the opcode (to 0xFD: variable length, FlexSym fields) instead of writing the switch byte. I think the idea is that if you're using FlexSyms you're more likely to exceed the lower nibble anyway, so you can use the byte for the length instead and potentially save a byte. We can discuss whether this is worthwhile though. If it's not, then maybe 0xFD can be reclaimed. See https://github.com/amazon-ion/ion-java/pull/665/files#diff-1903a8a68f333503cde63de27c671d623ff68da4accf35aa1db835d2c3e6f783R339

src/lazy/encoder/binary/v1_1/value_writer.rs
```diff
@@ -2454,4 +2521,282 @@ mod tests {
...
+            // Empty struct
+            (&[], &[0xF3, 0xF0]),
+            // Struct with a single FlexUInt field name
+            (
+                &[field(4, "foo")],
+                &[
+                    // Delimited struct
+                    0xF3,
+                    // FlexUInt symbol ID 4
+                    0x09,
```
tgregg, line 2683: I think Matt and I were still operating under the assumption that delimited structs always had FlexSym field names: https://github.com/amazon-ion/ion-java/pull/665/files#diff-c491f83b80fd84cfed9e62a411f61789c10d9731d4afbb7fefe538738dd3cd27R407

PR #714, opened 2024-02-21
Fixes bug in reading shared symbol table import

desaikd:
*Issue:*
The issue is in importing shared symbol import. SST import used `RawSymbolToken#text()` method to get a field name of SST import which resulted in not loading the symbol table import properly. `RawSymbolToken#text()` returns `None` if the token is an SID. Instead it should look up in the symbol table to get that SID's text. 
Another issue was with importing a system symbol with a shared symbol table import. While reading Ion data it should add this system table again to current symbol table and but while writing it should return the smallest SID(the system SID). 

*Description of changes:*
This PR works on fixing a bug in shared symbol table import.

*List of changes:*
* Modified how SST import read symbol text so that in case of SID it
gets corresponding text first
* Modified adding symbols to current symbol table using
`add_symbol_or_placeholder` 

*Test:*
Added a test that verifies SST import in binary Ion data. It also adds a system symbol to make sure the reading works as expected.

### DIFF
--- a/src/symbol_table.rs
+++ b/src/symbol_table.rs
@@ -53,18 +53,12 @@ impl SymbolTable {
         }
 
         // Otherwise, intern it and return the new ID.
-        let id = self.symbols_by_id.len();
-        let arc: Arc<str> = Arc::from(text);
-        let symbol = Symbol::shared(arc);
-        self.symbols_by_id.push(symbol.clone());
-        self.ids_by_text.insert(symbol, id);
-        id
+        self.add_symbol(text)
     }
 
     /// adds `text` to the symbol table and returns the newly assigned [SymbolId].
     pub(crate) fn add_symbol<A: AsRef<str>>(&mut self, text: A) -> SymbolId {
         let text = text.as_ref();
-        // Otherwise, intern it and return the new ID.
         let id = self.symbols_by_id.len();
         let arc: Arc<str> = Arc::from(text);
         let symbol = Symbol::shared(arc);
--- a/src/symbol_table.rs
+++ b/src/symbol_table.rs
@@ -61,12 +61,36 @@ impl SymbolTable {
         id
     }
 
+    /// adds `text` to the symbol table and returns the newly assigned [SymbolId].
+    pub(crate) fn add_symbol<A: AsRef<str>>(&mut self, text: A) -> SymbolId {
+        let text = text.as_ref();
+        // Otherwise, intern it and return the new ID.
+        let id = self.symbols_by_id.len();
+        let arc: Arc<str> = Arc::from(text);
+        let symbol = Symbol::shared(arc);
+        self.symbols_by_id.push(symbol.clone());
+        self.ids_by_text.insert(symbol, id);
+        id
+    }
+
     /// Assigns unknown text to the next available symbol ID. This is used when an Ion reader
     /// encounters null or non-string values in a stream's symbol table.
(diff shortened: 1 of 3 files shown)


src/system_reader.rs
```diff
@@ -343,13 +343,16 @@ impl<R: RawReader> SystemReader<R> {
...
-                        if let Some(field_name) = field_name_token.text() {
-                            if field_name == "name" {
-                                self.lst.state = AtImportName;
-                            } else if field_name == "version" {
-                                self.lst.state = AtImportVersion;
-                            } else if field_name == "max_id" {
-                                self.lst.state = AtImportMaxId;
+                        let field_name = match field_name_token {
+                            RawSymbolToken::SymbolId(sid) => self.symbol_table.text_for(sid),
+                            RawSymbolToken::Text(_) => field_name_token.text(),
```
zslayton, line 348: Could/should this `match` arm be
```rust
RawSymbolToken::Text(text_token) => Some(text_token.text())
```
?

src/symbol_table.rs
```diff
@@ -61,6 +61,18 @@ impl SymbolTable {
         id
     }
 
+    /// adds `text` to the symbol table and returns the newly assigned [SymbolId].
+    pub(crate) fn add_symbol<A: AsRef<str>>(&mut self, text: A) -> SymbolId {
+        let text = text.as_ref();
+        // Otherwise, intern it and return the new ID.
```
zslayton, line 67: I think this comment was copy/pasted from `fn intern` above? Consider having `fn intern` call `add_symbol` so we cut down on the duplicated code.

src/symbol_table.rs
```diff
@@ -69,6 +81,18 @@ impl SymbolTable {
...
+    ) -> SymbolId {
+        match maybe_text {
+            Some(text) => self.add_symbol(text),
+            None => self.add_placeholder(),
+        }
+    }
+
     /// If `maybe_text` is `Some(text)`, this method is equivalent to `intern(text)`.
     /// If `maybe_text` is `None`, this method is equivalent to `add_placeholder()`.
     pub(crate) fn intern_or_add_placeholder<A: AsRef<str>>(
```
zslayton, line 98: Is this method still being used anywhere?

src/system_reader.rs
```diff
@@ -1100,6 +1113,43 @@ mod tests {
...
+                0xde, 0x94, // 21 bytes struct
+                0x84, // Field $4 `name`
+                0x8c, 0x73, 0x68, 0x61, 0x72, 0x65, 0x64, 0x5f, 0x74, 0x61, 0x62, 0x6c,
+                0x65, // "shared_table"
+                0x85, // Field $5 `version`
+                0x21, 0x01, // INT 1
+                0x88, // $8 `max_id`
+                0x21, 0x02, // INT 2
+                0x71, 0x04, // $4 `name`
+                0x71, 0x0b, // $11 `foo`
```
zslayton, line 1140: Please also have this test that `$10` turns into `name`. The comment mentions that it should be added, but I don't think it's being tested here.

src/symbol_table.rs
desaikd: Yes its being used in `SymbolTable#initalize`.

zslayton (approved):
Make sure that the catalog and shared symbol table modules themselves are not public and that we're only re exporting the necessary types at the top level.

PR #725, opened 2024-03-06
Adds Amazon Linux (arm64 and x86_64) to the build matrix

### LINKED ISSUE #104: AArch64 Continuous Integration Pipeline
#102 illustrates that we should have some kind of CI for AArch64 if we want to ensure that things build/run on this platform.  A quick survey on Github Actions doesn't make it clear that it is an option here, but regardless, we should do something for ARM.

popematt:
**Issue #, if available:**

Fixes #104

**Description of changes:**

Adds Amazon Linux on Aarch64 and x86_64 for workflows that run in the `amazon-ion/ion-rust` (but not for forks).


zslayton (approved):
🚀

PR #727, opened 2024-03-14
Adds a `StreamingRawReader`, switches `Element` to the `LazyReader`

zslayton:
This PR:
* Introduces an `IonDataSource` trait with two implementations:
   * `IonSlice<impl AsRef<[u8]>`, which reads from byte-array backed fixed streams with no overhead.
   * `IonStream<impl io::Read>`, which reads incrementally from a streaming input source with minimal overhead.
* Introduces an `IonInput` trait for types that can be converted to an `IonDataSource`, including `&[u8]`, `Vec<u8>`, `&str`, `File`, etc.
* Introduces a `StreamingRawReader` that wraps an `IonDataSource` and exposes one top-level value at a time as a `LazyValue`.
* Modifies the `Element::read*` and `Element::iter` APIs to use the LazyReader.
* Modifies the `ion-tests` to exercise the `LazyReader`.
* Adds `range()` and `span()` methods for viewing the input bytes that correspond to a given value. There's a lot more to do for this API, but I stopped because this PR was getting too big. I'll continue it in a separate PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ edition = "2021"
 rust-version = "1.65"
 
 [features]
-default = []
+default = ["experimental"]
 experimental-ion-hash = ["digest"]
 
 # Feature for indicating particularly bleeding edge APIs or functionality in the library.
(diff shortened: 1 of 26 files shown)


zslayton (commented):
🗺️ PR tour

nirosys (approved):

PR #730, opened 2024-03-28
Adds trait implementations

zslayton:
Prior to this PR, the `AnnotatableValueWriter` trait provided many of the same methods as `ValueWriter` for ergonomics. Calls to these methods were delegated to the type returned from `without_annotations()`.

This patch reifies the relationship between `AnnotatableValueWriter` and `ValueWriter` by moving the delegated methods into a blanket implementation of `ValueWriter` for all `AnnotatableValueWriter` implementations. Existing uses of these methods continue to work as expected, but instances of an `AnnotatableValueWriter` implementation can now also be passed into methods that expect an `impl ValueWriter`.

The PR also adds `WriteAsIon` implementations for a variety of Rust int types.

### DIFF
--- a/src/element/mod.rs
+++ b/src/element/mod.rs
@@ -1494,7 +1494,7 @@ mod value_tests {
         let list: Element = ion_list![1, 2, 3].into();
         thread::scope(|_| {
             // Move `list` into this scope, demonstrating `Send`
-            let elements = vec![list];
+            let elements = [list];
             // Trivial assertion to use `elements`
             assert_eq!(elements.len(), 1);
         });
--- a/src/lazy/encoder/value_writer.rs
+++ b/src/lazy/encoder/value_writer.rs
@@ -38,46 +38,6 @@ pub trait AnnotatableValueWriter: Sized {
 
     /// Performs no operations and returns a [`ValueWriter`].
     fn without_annotations(self) -> Self::ValueWriter;
-
-    // Users can call `ValueWriter` methods on the `AnnotatedValueWriter` directly. Doing so
-    // will implicitly call `without_annotations`.
-
-    delegate! {
-        to self.without_annotations() {
-            fn write_null(self, ion_type: IonType) -> IonResult<()>;
-            fn write_bool(self, value: bool) -> IonResult<()>;
-            fn write_i64(self, value: i64) -> IonResult<()>;
-            fn write_int(self, value: &Int) -> IonResult<()>;
-            fn write_f32(self, value: f32) -> IonResult<()>;
-            fn write_f64(self, value: f64) -> IonResult<()>;
-            fn write_decimal(self, value: &Decimal) -> IonResult<()>;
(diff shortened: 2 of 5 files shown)


src/element/mod.rs
```diff
@@ -1494,7 +1494,7 @@ mod value_tests {
         let list: Element = ion_list![1, 2, 3].into();
         thread::scope(|_| {
             // Move `list` into this scope, demonstrating `Send`
-            let elements = vec![list];
+            let elements = [list];
```
zslayton, line 1497: 🗺️ Clippy noticed that this `vec!` (and the associated heap allocation) is not necessary.

src/lazy/encoder/value_writer.rs
```diff
@@ -38,46 +38,6 @@ pub trait AnnotatableValueWriter: Sized {
 
     /// Performs no operations and returns a [`ValueWriter`].
     fn without_annotations(self) -> Self::ValueWriter;
-
-    // Users can call `ValueWriter` methods on the `AnnotatedValueWriter` directly. Doing so
-    // will implicitly call `without_annotations`.
```
zslayton, line 43: 🗺️ These default delegated methods mimicked those on `ValueWriter`, allowing users to call them on an `AnnotatableValueWriter` without first calling `with_annotations` or `without_annotations`. However, an `AnnotatableValueWriter` could not be passed to methods that required an implementation of `ValueWriter`.

src/lazy/encoder/value_writer.rs
```diff
@@ -239,6 +199,20 @@ macro_rules! delegate_value_writer_to_self {
...
+{
+    type ListWriter<'a> = <V::ValueWriter as ValueWriter>::ListWriter<'a>;
+    type SExpWriter<'a> = <V::ValueWriter as ValueWriter>::SExpWriter<'a>;
+    type StructWriter<'a> = <V::ValueWriter as ValueWriter>::StructWriter<'a>;
+    type MacroArgsWriter<'a> = <V::ValueWriter as ValueWriter>::MacroArgsWriter<'a>;
+
+    delegate_value_writer_to!(closure |self_: Self| {
+       self_.without_annotations()
+    });
+}
```
zslayton, line 214: 🗺️ The new blanket implementation of `ValueWriter` for all `AnnotatableValueWriter`s.

src/lazy/never.rs
```diff
@@ -78,87 +78,3 @@ impl MakeValueWriter for Never {
 }
 
 impl MacroArgsWriter for Never {}
-
-impl ValueWriter for Never {
```
zslayton, line 82: 🗺️ `Never` implements many traits so it can be used as a marker to statically indicate unreachable/unused code paths. Previously, `Never` implemented both `ValueWriter` and `AnnotatableValueWriter`; the new blanket impl allows us to eliminate the explicit implementation of `ValueWriter`.

zslayton (commented):
🗺️ PR tour

popematt (approved):

Issue #444, opened 2022-10-13
Using `Symbol` as a `HashMap` key 


zslayton:
`Symbol` is an enum that represents either `Rc<str>`, `String`, or `UnknownText`. You can think of it as an `Option<String>` with an extra branch.

A `HashMap` with a key type `K` can only be queried by types `T` for which `K` implements `Borrow<T>`. This allows you to query for a `String` using a `&str` (instead of having to allocate a new `String` just for the query). However, it doesn't work for many other types because of lifetime restrictions. (See [this URLO question](https://users.rust-lang.org/t/querying-hashmap-when-borrow-t-cant-be-satisfied/82682/3) for some discussion.)

Currently, `ion-rust` works around this by simply having `Symbol` implement `Borrow<str>` (like `String` does) and panicking if you ever try to borrow text from an `UnknownText` variant. This works; however, I believe that if some text happened to produce the same hash code as unknown text, `HashMap::get` might try to borrow from the unknown text.

We need to prevent this improbable-but-possible panic from occurring by:
1. Defining `Symbol`'s hash function to prevent this from ever occurring (and living with the wart of `Borrow<str>` panicking if it ever happens in some other context).
2. Write a wrapper type for `Symbol` that can be used more safely as a key in a `HashMap`. (The options for this are clunky -- see the URLO thread linked above.) We could also write a `SymbolMap<T>` that wraps a `HashMap<SymbolWrapper, T>` to hide a lot of the grossness.
3. Use an [`indexmap`](https://docs.rs/indexmap/latest/indexmap/index.html) instead of a `HashMap` and implement its custom [`Equivalent` trait](https://docs.rs/indexmap/latest/indexmap/trait.Equivalent.html) for `Symbol`. This is appealing for `Struct` because the implementation of `indexmap` mirrors our own `Fields` implementation. Research is needed to see if it could also feasibly be used for `SymbolTable`, which also uses a `HashMap<Symbol, _>`.

PR #732, opened 2024-04-01
Simplifies the `ValueWriter` trait family

zslayton:
The `ValueWriter` trait's container-writing methods (`write_list`, `write_sexp`, etc) take a closure using [HRTB](https://doc.rust-lang.org/nomicon/hrtb.html), allowing the user to call whatever methods they want on the provided container writer.

Unfortunately, due to a limitation in the compiler's implementation of HRTB evaluation, HRTB lifetimes are over-conservatively [assumed to be `'static`](https://blog.rust-lang.org/2022/10/28/gats-stabilization.html#implied-static-requirement-from-higher-ranked-trait-bounds). This constraint led to a number of workaround warts in the `ValueWriter` trait family.

This PR replaces those closures with a collection of new traits.

*Old signatures*
```rust
fn write_list<F: for<'a> FnOnce(&mut Self::ListWriter<'a>) -> IonResult<()>>(
    self,
    list_fn: F,
) -> IonResult<()>;

fn write_sexp<F: for<'a> FnOnce(&mut Self::SExpWriter<'a>) -> IonResult<()>>(
    self,
    sexp_fn: F,
) -> IonResult<()>;

fn write_struct<F: for<'a> FnOnce(&mut Self::StructWriter<'a>) -> IonResult<()>>(
    self,
    struct_fn: F,
) -> IonResult<()>;

fn write_eexp<'macro_id, F: for<'a> FnOnce(&mut Self::MacroArgsWriter<'a>) -> IonResult<()>>(
    self,
    _macro_id: impl Into<MacroIdRef<'macro_id>>,
    _macro_fn: F,
) -> IonResult<()>;
```

*New signatures*
```rust
fn write_list(self, list_fn: impl ListFn<Self>) -> IonResult<()>;

fn write_sexp(self, sexp_fn: impl SExpFn<Self>) -> IonResult<()>;

fn write_struct(self, struct_fn: impl StructFn<Self>) -> IonResult<()>;

fn write_eexp<'macro_id>(
    self,
    _macro_id: impl Into<MacroIdRef<'macro_id>>,
    _macro_fn: impl MacroArgsFn<Self>,
) -> IonResult<()>;
```

These traits are intentionally very close in meaning to the closures they replace. However, introducing them simplifies things quite a bit; in particular:
* We are able to remove a lifetime from each of `ValueWriter`'s associated types. For example: `ValueWriter::ListWriter<'a>` is now `ValueWriter::ListWriter`.
* Because the traits' generics do not involve HRTB (`for <'a> ...`), the compiler is able to accept the traits anywhere we were already accepting closures.
* We can automatically implement the new traits for the old closures, so there's no change to user code needed.
* Passing around values that implement these types is much less likely to lead to lifetime errors.
* The method signatures are less likely to cause your eyes to bleed.

This patch also modifies the behavior of `Symbol`'s implementation of `Borrow`, which allows `Symbol`s to be used as keys in a `HashMap`. Previously, the `Symbol` would panic if it had no text and the user called `borrow()` to get a `&str`. (See #444 for a more detailed write-up.) This PR encountered a spot in our implementation of `Struct`/`Fields` that needs to look up fields which have unknown text, causing the tests to trigger the panic. For now, I've modified the `Borrow` impl to return the empty string if no text is available. This allows symbols with unknown text to be used as HashMap keys, but it could be surprising behavior on the off chance that someone finds another use case for `Symbol`'s `Borrow` implementation. #444 discusses some long-term options that are more robust.

### DIFF
--- a/src/lazy/encoder/binary/mod.rs
+++ b/src/lazy/encoder/binary/mod.rs
@@ -1,3 +1,2 @@
 pub mod v1_0;
 pub mod v1_1;
-
--- a/src/lazy/encoder/binary/mod.rs
+++ b/src/lazy/encoder/binary/mod.rs
@@ -1,21 +1,3 @@
 pub mod v1_0;
 pub mod v1_1;
 
-/// Takes a series of `TYPE => METHOD` pairs, generating a function for each that calls the host
-/// type's `encode_annotated` method to encode an annotations sequence and then delegates encoding
-/// the value to the corresponding value writer method.
-// This macro is used in the v1_0 and v1_1 binary writer implementations, which both define an
-// `encode_annotated` method. That method is not codified (for example: in a trait); this relies
-// solely on convention between the two.
-macro_rules! annotate_and_delegate {
-    // End of iteration
-    () => {};
-    // Recurses one argument pair at a time
-    ($value_type:ty => $method:ident, $($rest:tt)*) => {
-        fn $method(self, value: $value_type) -> IonResult<()> {
-            self.encode_annotated(|value_writer| value_writer.$method(value))
-        }
-        annotate_and_delegate!($($rest)*);
-    };
-}
-pub(crate) use annotate_and_delegate;
--- a/src/lazy/encoder/binary/v1_0/container_writers.rs
+++ b/src/lazy/encoder/binary/v1_0/container_writers.rs
@@ -1,16 +1,14 @@
+use bumpalo::Bump as BumpAllocator;
+use bumpalo::collections::Vec as BumpVec;
+
+use crate::{IonError, IonResult, RawSymbolTokenRef};
 use crate::binary::var_uint::VarUInt;
-use crate::lazy::encoder::binary::v1_0::value_writer::{
-    BinaryAnnotatableValueWriter_1_0, MAX_INLINE_LENGTH,
-};
-use crate::lazy::encoder::value_writer::internal::MakeValueWriter;
+use crate::lazy::encoder::binary::v1_0::value_writer::BinaryAnnotatableValueWriter_1_0;
 use crate::lazy::encoder::value_writer::{SequenceWriter, StructWriter};
+use crate::lazy::encoder::value_writer::internal::MakeValueWriter;
 use crate::lazy::encoder::write_as_ion::WriteAsIon;
 use crate::raw_symbol_token_ref::AsRawSymbolTokenRef;
 use crate::result::EncodingError;
-use crate::{IonError, IonResult, RawSymbolTokenRef};
-use bumpalo::collections::Vec as BumpVec;
-use bumpalo::Bump as BumpAllocator;
-use delegate::delegate;
 
 /// A helper type that holds fields and logic that is common to [`BinaryListWriter_1_0`],
 /// [`BinarySExpWriter_1_0`], and [`BinaryStructWriter_1_0`].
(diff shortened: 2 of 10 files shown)


src/lazy/encoder/binary/mod.rs
```diff
@@ -1,21 +1,2 @@
...
-    () => {};
-    // Recurses one argument pair at a time
-    ($value_type:ty => $method:ident, $($rest:tt)*) => {
-        fn $method(self, value: $value_type) -> IonResult<()> {
-            self.encode_annotated(|value_writer| value_writer.$method(value))
-        }
-        annotate_and_delegate!($($rest)*);
-    };
-}
-pub(crate) use annotate_and_delegate;
```
zslayton, line 21: 🗺️ This macro was split into `1_0` and `1_1` versions; the new macros live closer to their usage sites.

src/lazy/encoder/binary/v1_0/container_writers.rs
```diff
@@ -22,43 +20,31 @@ pub struct BinaryContainerWriter_1_0<'value, 'top> {
     // The buffer containing the parent's encoded body. When this list writer is finished encoding
     // its own data, a header will be written to the parent and then the list body will be copied
     // over.
-    parent_buffer: &'value mut BumpVec<'top, u8>,
+    buffer: &'value mut BumpVec<'top, u8>,
```
zslayton, line 23: 🗺️ Removal of the `[List/SExp]ValuesWriter*` and `StructFieldsWriter` family of types meant that this variable name was no longer ambiguous.

src/lazy/encoder/binary/v1_0/container_writers.rs
```diff
@@ -72,108 +58,48 @@ pub struct BinaryContainerValuesWriter_1_0<'value> {
 // reallocations/copies.
 const DEFAULT_CONTAINER_BUFFER_SIZE: usize = 512;
 
-impl<'value> BinaryContainerValuesWriter_1_0<'value> {
```
zslayton, line 75: 🗺️ Notice that these types had a single lifetime (`'value`); this was done to work around the compiler's HRTB limitation. Now that we're using the new traits, the writers can have two lifetimes (`'value, 'top`), meaning we don't need these anymore.

src/lazy/encoder/container_fn.rs
```diff
@@ -0,0 +1,56 @@
...
+///```text
+///     fn populate(self, writer: &mut V::ListWriter) -> IonResult<()>;
+///```
+///
+/// Types that implement this trait can be invoked as though they were closures.
+/// Closures with the signature `(&mut V::ListWriter) -> IonResult<()>` automatically implement
+/// the trait.
+///
+/// This circular arrangement allows the compiler to accept closures that would otherwise
+/// require explicit Higher-Rank Trait Bounds, a feature which currently has limitations.
```
zslayton, line 28: 🗺️ This felt absurd to try, but hey! it worked! 🙃😵‍💫

src/types/symbol.rs
```diff
@@ -34,7 +34,7 @@ impl Hash for SymbolText {
         match self {
             SymbolText::Shared(text) => text.hash(state),
             SymbolText::Owned(text) => text.hash(state),
-            SymbolText::Unknown => 0.hash(state),
+            SymbolText::Unknown => "".hash(state),
```
zslayton, line 37: 🗺️ The hashing behavior of `Symbol` now also uses the empty string if there's no text for agreement with `Borrow`.

zslayton (commented):
🗺️ PR tour

src/lazy/encoder/write_as_ion.rs
```diff
@@ -151,7 +151,7 @@ macro_rules! impl_write_as_ion_value_for_iterable {
             for<'a> &'a $item: WriteAsIon,
         {
             fn write_as_ion_value<V: ValueWriter>(&self, writer: V) -> IonResult<()> {
-                writer.write_list(|list| {
+                writer.write_list(|list: &mut V::ListWriter| {
```
desaikd, line 154: Is this declaration of type required? I didn't see it in other places e.g. `write_sexp`.

src/lazy/encoder/container_fn.rs
```diff
@@ -0,0 +1,56 @@
+//! Container population traits that allow closures to be used in places where the borrow checker
+//! would normally balk due to point-in-time limitations.
+//! See: <https://blog.rust-lang.org/2022/10/28/gats-stabilization.html#implied-static-requirement-from-higher-ranked-trait-bounds>
+
+use crate::lazy::encoder::value_writer::ValueWriter;
+use crate::IonResult;
+
+/// Generates a trait with a single method that takes one of [`ValueWriter`]'s associated types as
+/// an argument. The trait extends `FnOnce(TheAssociatedType) -> IonResult<()>`, allowing it to be
```
desaikd, line 9: This is just a suggestion, feel free to take it or leave it! I think it might be worth adding whats the functionality of `populate` for this workaround. 

```suggestion
/// Generates a trait with a single method that takes one of [`ValueWriter`]'s associated types as
/// an argument. This `populate` method has the same functionality as the closure and it helps populate the container by adding new fields/values. The trait extends `FnOnce(TheAssociatedType) -> IonResult<()>`, allowing it to be
```

desaikd (approved):
LGTM! Thanks for adding detailed comments on the workaround macro `container_fn_trait!`.

src/lazy/encoder/write_as_ion.rs
zslayton: Good eye--there was a brief time during development when it was needed but it's no longer necessary. I'll remove it.

zslayton:
Per our offline conversation, I'll be incorporating these comments into my next PR.

popematt:
> For now, I've modified the Borrow impl to return the empty string if no text is available.

Could the function return `Option<String>` instead, and then return `None` if no text is available?

PR #733, opened 2024-04-01
Adds support for parameterized Rust enums in serde

desaikd:
*Description of changes:*
This PR works on adding support for parameterized Rust enums with `experimental-serde` feature.

*Example:*
Below is an example parameterized enum value in Rust:
```rust
#[serde_as]
#[derive(Serialize, Deserialize)]
enum ParameterizedEnum {
    Unit,
    Newtype(u32),
    Tuple(u32, u32),
    Struct { a: u32 },
}
```
Ion representation:
```ion
// For each of the enum variant following are corresponding Ion representation
Unit // simply a symbol value 
NewType::1
Tuple::[1, 2]
Struct::{a: 1}
```

*List of changes:*
- Modified mapping table for serde data types to Ion types.
- Modified serialization to remove adding a new struct for `newtype_variant`, `struct_variant` and `tuple_variant`. Instead add annotation with variant name.
- Modified deserialization to check for annotation with `newtype_variant`, `struct_variant` and `tuple_variant`.

*Test:*
Added tests for enum variants and struct types.

### DIFF
--- a/src/serde/de.rs
+++ b/src/serde/de.rs
@@ -401,8 +401,8 @@ impl<'a, 'de> de::Deserializer<'de> for ValueDeserializer<'a, 'de> {
         use IonType::*;
         match self.value.ion_type() {
             String => visitor.visit_enum(UnitVariantAccess::new(self)),
-            Struct => visitor.visit_enum(VariantAccess::new(self)),
-            _ => IonResult::decoding_error("expected a string or struct enum representation"),
+            // All the parameterized Rust enums uses annotations for representing enum variant
+            _ => visitor.visit_enum(VariantAccess::new(self)),
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/serde/de.rs
```diff
@@ -401,16 +401,21 @@ impl<'a, 'de> de::Deserializer<'de> for ValueDeserializer<'a, 'de> {
...
+            _ => visitor.visit_enum(VariantAccess::new(self)),
         }
     }
 
     fn deserialize_identifier<V>(self, visitor: V) -> Result<V::Value, Self::Error>
     where
         V: Visitor<'de>,
     {
-        visitor.visit_str(self.value.read()?.expect_text()?)
+        // annotations are currently only supported with parameterized Rust enums
```
popematt, line 413: What exactly does this mean? Does the rustdoc for this feature need to be updated as well?

desaikd: This means that a serialized Ion data will only contain annotation in the case of parameterized Rust enums. And yes, I did add more comments in the rustdocs for it updating the table of serde data type to Ion type mappings: https://github.com/amazon-ion/ion-rust/pull/733/files#diff-de73824c75ee6203daf57ccbc5aee2ab3e2b46fbd9219d494c190298a7a8e745R43

popematt: Oh, I saw that but didn't make the connection. It might be worth renaming the table to say "Ion representation" instead of "Ion data type" now that we're including more information than just the data type.

popematt (approved):
LGTM, but I'll defer to @zslayton in case he has anything else to add.

src/serde/de.rs
desaikd: Sure. I will change that.

zslayton (approved):

PR #737, opened 2024-04-09
Add support for binary 1.1 reader

nirosys:
*Issue #, if available:*

*Description of changes:*
This PR adds the initial wiring-up of a 1.1 binary reader. The bulk of the additions are in `lazy/binary/raw/v1_1`, with some type renaming and implementations to satisfy the encoding struct for binary 1.1.

In addition to the base reader, base support for null.null and IVMs are also added, with basic unit tests.

More comments posted in the 🗺️ PR Tour.

---


src/lazy/binary/raw/v1_1/type_code.rs
nirosys, line 1: (🗺️  PR Tour) This gets fleshed out a little better over the next few PRs. I've mostly just tried to lay out the IonTypecode values here. The From implementations still need to be implemented.

src/lazy/binary/raw/v1_1/type_descriptor.rs
nirosys, line 1: (🗺️  PR Tour) This is all based on the 1.0 `TypeDescriptor`, with 1.1 type codes. Noticing a few comments I missed as well. I'll followup to remove them.

src/lazy/binary/raw/v1_1/value.rs
nirosys, line 1: (🗺️  PR Tour) Mostly stubbed out 1.1 value. `null.null`, IVM, and NOPs are parsing. I defaulted to nested imports, before realizing that's not what the rest of the codebase really does. I can follow up to change that.

src/lazy/binary/raw/value.rs
nirosys, line 1: (🗺️  PR Tour) More type name changes.

src/lazy/encoder/binary/v1_1/flex_uint.rs
```diff
@@ -36,7 +36,7 @@ pub struct FlexUInt {
 }
 
 impl FlexUInt {
-    fn new(size_in_bytes: usize, value: u64) -> Self {
+    pub(crate) fn new(size_in_bytes: usize, value: u64) -> Self {
```
nirosys, line 39: Thoughts on this? In a future PR I needed to do similar to FixedInt, and opted to just implement `From<i64>`. In retrospect implementing `From<u64>` here would probably be better, imo.

src/lazy/encoding.rs
nirosys, line 1: (🗺️  PR Tour) Added new encoding struct for binary 1.1.

src/lazy/binary/raw/v1_1/reader.rs
nirosys, line 1: (🗺️ PR Tour) This is mostly a copy from the 1.0 implementation, with some 1.1 specific changes and some 1.0 specifics removed.

Basic unit tests are also added here to verify the implementation.

nirosys (commented):
(🗺️ PR Tour)

src/lazy/binary/raw/v1_1/type_code.rs
```diff
@@ -0,0 +1,147 @@
...
+/// Represents the type information found in the header byte of each binary Ion value.
+/// While this value can be readily mapped to a user-level [`IonType`], it is a distinct concept.
+/// The IonTypeCode enum captures system-level information that is not exposed to end users of the
+/// library, including:
+/// * Whether the cursor is positioned over whitespace that needs to be skipped.
+/// * Whether the integer value being read is positive or negative.
+/// * Whether the next type code is reserved.
+///
+/// See the
+/// [Typed Value Formats](https://amazon-ion.github.io/ion-docs/docs/binary.html#typed-value-formats)
```
popematt, line 16: FYI—this link refers to the 1.0 type codes.

src/lazy/binary/raw/v1_1/type_code.rs
```diff
@@ -0,0 +1,147 @@
...
+                return IonResult::decoding_error(format!(
+                    "Attempted to make an IonType from an invalid type code: {ion_type_code:?}"
+                ));
+            }
+        };
+        Ok(ion_type)
+    }
+}
+
+impl TryFrom<u8> for IonTypeCode {
```
popematt, line 82: Is this relevant for Ion 1.1?

src/lazy/binary/raw/v1_1/type_code.rs
```diff
@@ -0,0 +1,147 @@
...
+/// library, including:
+/// * Whether the cursor is positioned over whitespace that needs to be skipped.
+/// * Whether the integer value being read is positive or negative.
+/// * Whether the next type code is reserved.
+///
+/// See the
+/// [Typed Value Formats](https://amazon-ion.github.io/ion-docs/docs/binary.html#typed-value-formats)
+/// section of the spec for more information.
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
+pub enum IonTypeCode {
```
popematt, line 19: nit—Ion 1.1 refers to these as `OpCode`s.

src/lazy/encoder/binary/v1_1/flex_uint.rs
popematt: Do you need to create an instance of `FlexUInt` from the reader other than what you can create using `FlexUInt::read()`?

src/lazy/binary/raw/v1_1/reader.rs
```diff
@@ -0,0 +1,141 @@
...
+        &mut self,
+        buffer: ImmutableBuffer<'data>,
+    ) -> IonResult<LazyRawStreamItem<'top, BinaryEncoding_1_1>>
+    where
+        'data: 'top,
+    {
+        let ((major, minor), _buffer_after_ivm) = buffer.read_ivm()?;
+        if (major, minor) != (1, 1) {
+            return IonResult::decoding_error(format!(
+                "unsupported version of Ion: v{}.{}; only 1.1 is supported by this reader",
```
popematt, line 37: Doesn't need to be fixed now, but you should probably consider whether this design will make it possible to read any IVM. There needs to be some way to (mid stream) read either IVM and then switch to the correct reader.

popematt (commented):
The module structure is making this more difficult to review than it needs to be. It's hard to know what code is version specific and what code is version agnostic. It would be _nice_ if you could create a new PR that reorganizes the code in preparation for having two Ion versions, and then rebase this on top of that PR. However, I know that's probably a lot of tedious work, so a follow up PR would also be fine.

Did you consider making some of these types generic w.r.t. Ion versions? (I.e. `<const MinorVersion: u8>` or use a `PhantomData<Ion_1_1>` somewhere?) Then we might be able to avoid duplicating as much stuff, and _maybe_ when we encounter a version

nirosys:
> The module structure is making this more difficult to review than it needs to be. It's hard to know what code is version specific and what code is version agnostic. It would be _nice_ if you could create a new PR that reorganizes the code in preparation for having two Ion versions, and then rebase this on top of that PR. However, I know that's probably a lot of tedious work, so a follow up PR would also be fine.

I'm not sure I follow. The only changes in this PR that _aren't_ 1.1 specific implementations, or part of 1.1 encoding wire-up, are the 1.0 type renames. The module layout should mirror the raw text reader, with some added types in the `v1_1` module, to satisfy the encoding type requirements.

> Did you consider making some of these types generic w.r.t. Ion versions? (I.e. `<const MinorVersion: u8>` or use a `PhantomData<Ion_1_1>` somewhere?) Then we might be able to avoid duplicating as much stuff, and _maybe_ when we encounter a version

I had put some thought into parameterizing but wasn't sure exactly where the boundaries would be, yet. Most of the duplicated types (like the ones in `lazy::binary:raw::v1_1::sequence` and `struct`), are mainly to satisfy the type requirements, until I can revisit parameterizing on either the encoding, or a version number.

src/lazy/encoder/binary/v1_1/flex_uint.rs
nirosys: Yes. For simplicity, I handle the lengths of values via FlexUInt, so I need to create instances for when the header provides enough information for the length.

I probably should have provided more context. The question is more to gain consensus around whether `new` is sufficiently user-friendly for the crate. The `size_in_bytes` argument makes it seem a little more 'internal', which would make me think a `From<u64>` or similar added method would be a better alternative.

src/lazy/binary/raw/v1_1/struct.rs
```diff
@@ -0,0 +1,152 @@
...
+}
+
+impl<'top> LazyRawBinaryStruct_1_1<'top> {
+    fn annotations(&self) -> RawBinaryAnnotationsIterator<'top> {
+        self.value.annotations()
+    }
+
+    pub fn iter(&self) -> RawBinaryStructIterator_1_1<'top> {
+        // Get as much of the struct's body as is available in the input buffer.
+        // Reading a child value may fail as `Incomplete`
```
zslayton, line 53: This comment refers to a no-longer-true invariant.

At one time, the binary container types used to store however much of their body had been available in the input buffer at read time. Because 1.0 containers are always length-prefixed, only the header's encoding needed to be available in the input for the creation of the lazy container value (hereon: LCV) to succeed.

Conceptually, this allowed LCVs to avoid needing to buffer entire top-level values. The LCV could successfully visit and read however many of its child values' encodings were fully available and eventually fail at an incomplete value. In practice, however, this offers little benefit. It's much easier for applications (including the streaming reader wrapper) to handle early-bound incompleteness errors; writing data processing logic that can transactionally roll back and try again when more data is available is not fun.

Additionally, the introduction of delimited binary containers meant that skipping to the next value required that the end of the container be found via scanning. This could either happen when the container was first encountered (guaranteeing that the entire container was available in the process) or on demand when the container's next sibling was requested from the parent. Finding the end of the container at the outset means that:
1. We can guarantee the container is fully available
2. We can cache the lazy child values that we encountered
3. Future iterators over the container's contents can iterate over the cache instead of re-reading the data each time

Collectively, this model does mean that:
1. Truncated Ion data (i.e. partial data that will never be complete) cannot be read with this API. (We could extend the API to support this in the future with "advanced" methods.)
2.  Top level values must fit in the buffer. This was once a problem in `ion-java` because Java uses signed 32-bit integers as its array indices, limiting arrays (and thus buffers) to a size of ~2GB. Rust does not have that limitation; applications wishing to structure their data this way are free to do so at the expense of RAM.

We need to add unit tests for the 1.0 and 1.1 binary readers that demonstrate early-bound availability errors for delimited and length-prefixed container types .

src/lazy/binary/raw/v1_1/type_code.rs
```diff
@@ -0,0 +1,147 @@
...
+    EExpression(bool), // 0x00-0x4F - With, or without, address in the opcode.
```
zslayton, line 20: I think this would be more self-documenting as two variants; something like:
```suggestion
    EExpressionWithAddress, // 0x00-0x3F
    EExpressionAddressFollows // 0x40-0x4F
```

Of course, we're in the processing of changing the opcodes and the meaning of "address follows", so this can be a `TODO` for later.

src/lazy/binary/raw/v1_1/type_code.rs
```diff
@@ -0,0 +1,147 @@
...
+            // 15 => Reserved,
+            _ => {
+                return IonResult::decoding_error(format!(
+                    "{type_code:?} is not implemented, or an invalid type code."
+                ));
+            } // _ => {
+              //     return IonResult::decoding_error(format!(
+              //         "{type_code:?} is not a valid header type code."
+              //     ));
+              // }
```
zslayton, line 116: Old comment?

src/lazy/binary/raw/v1_1/type_descriptor.rs
```diff
@@ -0,0 +1,141 @@
...
+            ion_type,
+            ion_type_code: self.ion_type_code,
+            length_code: self.length_code,
+        };
+        Some(header)
+    }
+}
+
+pub enum LengthType {
+    InHeader(usize),
```
zslayton, line 94: Often the phrase "header" includes the trailing length when present. Maybe:
```suggestion
    InOpcode(usize),
```
?

src/lazy/binary/raw/v1_1/value.rs
```diff
@@ -0,0 +1,251 @@
...
+    /// fully buffered before reading begins.
+    pub(crate) fn available_body(&self) -> ImmutableBuffer<'top> {
+        let value_total_length = self.encoded_value.total_length();
+        let value_body_length = self.encoded_value.value_length();
+        let value_offset = value_total_length - value_body_length;
+
+        let bytes_needed = std::cmp::min(self.input.len() - value_offset, value_body_length);
+        let buffer_slice = self.input.slice(value_offset, bytes_needed);
+        buffer_slice
+    }
```
zslayton, line 179: This is also part of that no-longer-true invariant. We should (as part of that `TODO`) find places where this is used and use `fn value_body` above instead.

src/lazy/encoder/binary/v1_1/flex_uint.rs
zslayton: I think it's fine as long as it's `pub(crate)`. There's not a reason for (most) users to manually create a type representing an encoding artifact. If we want to eventually make `Flex[U]Int` available to the public (say, for use in the `ion-cli`), we would add an experimental feature gate for it anyway.

Having `From<i64/u64>` is great for simplicity as long as the type stays out of the public API.

src/lazy/encoding.rs
```diff
@@ -134,6 +141,18 @@ impl LazyDecoder for TextEncoding_1_1 {
...
 
+impl LazyDecoder for BinaryEncoding_1_1 {
+    type Reader<'data> = LazyRawBinaryReader_1_1<'data>;
+    type ReaderSavedState = ();
+    type Value<'top> = LazyRawBinaryValue_1_1<'top>;
+    type SExp<'top> = LazyRawBinarySExp_1_1<'top>;
+    type List<'top> = LazyRawBinaryList_1_1<'top>;
+    type Struct<'top> = LazyRawBinaryStruct_1_1<'top>;
+    type AnnotationsIterator<'top> = RawBinaryAnnotationsIterator_1_1<'top>;
+    // Macros are not supported in Ion 1.0
```
zslayton, line 152:
```suggestion
    // TODO: implement macros in 1.1
```

zslayton (commented):
Nice! Some comments below, mostly nits and a couple of `TODO`s for later.

src/lazy/binary/raw/v1_1/struct.rs
zslayton: Opened #739 to track this.

popematt:
> I'm not sure I follow. The only changes in this PR that aren't 1.1 specific implementations, or part of 1.1 encoding wire-up, are the 1.0 type renames. The module layout should mirror the raw text reader, with some added types in the v1_1 module, to satisfy the encoding type requirements.

What I mean is that the existing module structure (for reading binary) assumes 1.0 everywhere. The new modules are all contained in a flat `v1_1` module. We know, just by looking at the file path, which files are 1.1 specific. We cannot say the same about 1.0 or version agnostic modules. It's not something for which I would block a PR, but it is something that needs to be cleaned up at some point.

src/lazy/binary/raw/v1_1/type_descriptor.rs
popematt: The size in the opcode is never more than 15, right? If so, we can use `u8`.

popematt (approved):
LGTM. However, I would like to defer to @zslayton for final approval.

src/lazy/binary/raw/v1_1/immutable_buffer.rs
```diff
@@ -375,7 +344,8 @@ impl<'a> ImmutableBuffer<'a> {
         &mut self,
         parser: F,
     ) -> IonResult<Option<LazyRawBinaryValue_1_1<'a>>> {
-        let buffer = self.advance_to_next_item()?;
+        // let buffer = self.advance_to_next_item()?;
```
zslayton, line 347:
```suggestion
```

src/lazy/binary/raw/v1_1/immutable_buffer.rs
```diff
@@ -386,7 +356,7 @@ impl<'a> ImmutableBuffer<'a> {
         // If the value we read doesn't start where we began reading, there was a NOP.
         let num_nop_bytes = lazy_value.input.offset() - buffer.offset();
         self.consume(num_nop_bytes);
-        self.bytes_to_skip = lazy_value.encoded_value.total_length();
+        // self.bytes_to_skip = lazy_value.encoded_value.total_length();
```
zslayton, line 359:
```suggestion
```

zslayton (approved):
🚀

Issue #740, opened 2024-04-10
Container writers should panic if dropped without calling `end()`


zslayton:
Dropping a container writer without calling `end()` can result in lost or corrupted Ion data. We should `panic!` to noisily prevent that. For example:
https://github.com/amazon-ion/ion-rust/blob/e3f0eaff67e736f1d57c622e056f38d49638487b/src/lazy/encoder/text/value_writer.rs#L110-L121

PR #741, opened 2024-04-10
Drops closure-based container writers

zslayton:
Recent work on #732 surfaced lifetime issues in the closure-centric writer APIs for containers and e-expressions. In particular, the workarounds for [the HRTBs-are-'static GATS impl limitation](https://blog.rust-lang.org/2022/10/28/gats-stabilization.html#implied-static-requirement-from-higher-ranked-trait-bounds) ended up making it impossible to write nested containers because the inner closure's lifetime was unsatisfiable.

This patch makes several simplifications to the writer API.

-----

### Writing containers and e-expressions no longer uses closures

Methods have been added to the `ValueWriter` trait that provide writers for the corresponding container/expression.
```rust
trait ValueWriter: ... {
  // ...

  fn list_writer(&mut self) -> IonResult<Self::ListWriter>;
  fn sexp_writer(&mut self) -> IonResult<Self::SExpWriter>;
  fn struct_writer(&mut self) -> IonResult<Self::StructWriter>;
  fn eexp_writer(&mut self) -> IonResult<Self::EExpWriter>;
}
```

For example, what was previously achieved via `write_list`:
```rust
writer.write_list(|list| {
  list
    .write(1)?
    .write(2)?
    .write(3)?;
  Ok(())
})
```
would now be done via `list_writer`:
```rust
let mut list = writer.list_writer()?;
list
  .write(1)?
  .write(2)?
  .write(3)?;
list.end()
```
or more succinctly:
```rust
let mut list = writer.list_writer()?;
list.write_all(&[1, 2, 3])?;
list.end()
```
or even more succinctly, when all values are the same type:
```rust
writer.write_list([1, 2, 3])?;
```

Dropping a container without calling `end()` is illegal as it can corrupt the stream, but it is not currently enforced via `panic!`. I've opened #740 to track that.

-----

### There is no longer an `AnnotatableValueWriter` trait

The `AnnotatableValueWriter` trait provided a `with_annotations()` method that allowed the user to get a separate `ValueWriter` implementation that could handle annotations and from which it was statically impossible to try and set annotations again.

`ValueWriter` implementations now have a `with_annotations(...)` method that returns another `ValueWriter` prepared to encode the provided symbols. Calling `with_annotations()` on a `ValueWriter` that already has annotations returns a new `ValueWriter` with the annotations replaced (it does not e.g. concatenate the annotation sequences).

Accepting this behavioral change allowed the indirection and code introduced by `AnnotatableValueWriter` to be removed. The resulting API is the same for the annotated case:
```rust
writer
  .value_writer()
  .with_annotations(&["foo", "bar", "baz"])
  .write_bool(true)?;
```

but the traits/types involved are simpler; `AnnotatableValueWriter`'s functionality has been folded into `ValueWriter`.

-----

### The `WriteAsIon` and `WriteAsIonValue` traits have been consolidated into `WriteAsIon`

The elimination of the `AnnotatableValueWriter` trait made it possible to merge the `WriteAsIonValue` trait (designed to serialize types that don't use annotations) with the `WriteAsIon` trait (designed to serialize types that might use annotations).

-----

### Delimited container settings are propagated to all child containers

In the Ion 1.1 binary writer API, calling `with_delimited_containers()` on a `ValueWriter` causes all nested containers to also use delimited containers by default. For example:

```rust
// Create a delimited StructWriter
let mut struct_ = writer
  .value_writer()
  .with_delimited_containers()
  .struct_writer()?;

// Write a nested delimited list
struct_
  .write("foo", &["bar", "baz"])?; // This list will be delimited

struct_ .end()
```

This allows users to set the desired encoding mode once at the top with the expectation that that will often be what they would like to use throughout that value. This is easily overridden:

```rust
// Create a delimited StructWriter
let mut struct_ = writer
  .value_writer()
  .with_delimited_containers()
  .struct_writer()?;

// Write a nested length-prefixed list
struct_
  field_writer()
    .with_length_prefixed_containers()
    .write("foo", &["bar", "baz"])? // This list will be length-prefixed
  .end()
```


src/lazy/encoder/binary/v1_0/value_writer.rs
```diff
@@ -621,13 +460,12 @@ mod tests {
...
-        mut test: impl FnMut(&mut LazyRawBinaryWriter_1_0<&mut Vec<u8>>) -> IonResult<()>,
+        test: impl FnOnce(&mut LazyRawBinaryWriter_1_0<Vec<u8>>) -> IonResult<()>,
     ) -> IonResult<()> {
         let expected = Element::read_all(expected)?;
-        let mut buffer = Vec::new();
-        let mut writer = LazyRawBinaryWriter_1_0::new(&mut buffer)?;
+        let mut writer = LazyRawBinaryWriter_1_0::new(Vec::new())?;
         test(&mut writer)?;
-        writer.flush()?;
+        let buffer = writer.finish()?;
```
zslayton, line 468: 🗺️ Top-level writers implement the `SequenceWriter` trait, which has an `end()` method.

Curious to hear others' takes on this: we had a `finish()` method already, but now it also has an `end()`. Can I get rid of `finish()`?

Alternatively: should this be `close()` everywhere instead of `end()` or `finish()`?

src/lazy/encoder/binary/v1_1/container_writers.rs
```diff
@@ -22,23 +21,90 @@ use crate::{FlexUInt, IonResult, RawSymbolTokenRef, SymbolId};
...
-    //      container.
-    //   2. a delimited container, this will be the parent's own encoding buffer, to which the delimited
-    //      container start opcode has already been written.
+    encoder: ContainerEncodingKind<'value, 'top>,
+}
+
+enum ContainerEncodingKind<'value, 'top> {
+    Delimited(DelimitedEncoder<'value, 'top>),
+    LengthPrefixed(LengthPrefixedEncoder<'value, 'top>),
+}
```
zslayton, line 30: 🗺️ To enable containers to be either delimited or length prefixed, `BinaryContainerWriter_1_1` now holds an enum of the required encoding resources instead of a pointer to a buffer.

src/lazy/encoder/binary/v1_1/container_writers.rs
```diff
@@ -123,110 +275,121 @@ pub struct BinaryStructWriter_1_1<'value, 'top> {
...
-    #[inline]
-    pub fn write<A: AsRawSymbolTokenRef, V: WriteAsIon>(
-        &mut self,
-        name: A,
-        value: V,
-    ) -> IonResult<&mut Self> {
-        use RawSymbolTokenRef::*;
+impl<'value, 'top> FieldEncoder for BinaryStructWriter_1_1<'value, 'top> {
+    fn encode_field_name(&mut self, name: impl AsRawSymbolTokenRef) -> IonResult<()> {
+        use crate::RawSymbolTokenRef::*;
```
zslayton, line 316: 🗺️ `StructWriter` impls now must also define the (private) method `encode_field_name`; they can then defer value serialization to `make_value_writer()`.

src/lazy/encoder/text/value_writer.rs
```diff
@@ -114,6 +104,7 @@ struct TextContainerWriter_1_0<'a, W: Write> {
     // The Ion type of the container using this TextContainerWriter_1_0. This value is only
     // used for more informative error messages.
     ion_type: IonType,
+    value_delimiter: &'static str,
```
zslayton, line 107: 🗺️ When `StuctWriter`s go to serialize a field, they use `FieldEncoder::encode_field_name(...)` and a `ValueWriter` to encode it. In text, this means that the `ValueWriter` is responsible for writing its trailing delimiter where applicable.

I considered adding a `write_delimiter` method to `StructWriter`/`FieldEncoder`, but didn't like that it wouldn't be applicable to the binary side. Whether this is better though is subjective. 🤷

zslayton (commented):
🗺️ PR tour

src/lazy/encoder/binary/v1_0/value_writer.rs
popematt: I think `close()` is a good choice that works well for both.

zslayton: Done! See commit https://github.com/amazon-ion/ion-rust/pull/741/commits/020e078febe7c4384087362f84a1e32b122485f7.

nirosys (approved):

src/lazy/encoder/value_writer.rs
```diff
@@ -215,33 +292,50 @@ macro_rules! delegate_and_return_self {
...
+        for value in values {
+            self.write(value)?;
+        }
+        Ok(self)
+    }
+
+    /// Closes out the sequence being written. Delimited writers can use this opportunity to emit
+    /// a sentinel value, and length-prefixed writers can flush any buffered data to the output
+    /// buffer.
+    fn close(self) -> IonResult<Self::Resources>;
```
popematt, line 337: Nice!

src/lazy/encoder/binary/v1_0/container_writers.rs
```diff
@@ -32,25 +36,123 @@ impl<'value, 'top> BinaryContainerWriter_1_0<'value, 'top> {
...
-        self.buffer.as_slice()
+    pub fn child_values_buffer(&self) -> &[u8] {
+        self.child_values_buffer.as_slice()
     }
-}
 
-pub struct BinaryContainerValuesWriter_1_0<'value> {
-    allocator: &'value BumpAllocator,
-    buffer: BumpVec<'value, u8>,
+    pub fn end(mut self) -> IonResult<()> {
```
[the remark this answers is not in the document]
zslayton: I've fixed this (and `StructWriter::end`) in a follow-on PR I hope to have out next week.

src/lazy/encoder/binary/v1_0/container_writers.rs
```diff
@@ -149,10 +316,26 @@ impl<'value, 'top> StructWriter for BinaryStructWriter_1_0<'value, 'top> {
...
+    fn make_value_writer(&mut self) -> Self::ValueWriter<'_> {
+        BinaryValueWriter_1_0::new(
+            self.container_writer.allocator,
+            &mut self.container_writer.child_values_buffer,
+        )
+    }
+}
+
+impl<'value, 'top> StructWriter for BinaryStructWriter_1_0<'value, 'top> {
+    fn end(self) -> IonResult<()> {
```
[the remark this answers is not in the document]
zslayton: This will be fixed in my next PR.

src/lazy/encoder/binary/v1_0/value_writer.rs
```diff
@@ -67,7 +64,7 @@ impl<'value, 'top> BinaryValueWriter_1_0<'value, 'top> {
         self.encoding_buffer.as_slice()
     }
 
-    pub fn write_symbol_id(&mut self, symbol_id: SymbolId) -> IonResult<()> {
+    pub fn write_symbol_id(mut self, symbol_id: SymbolId) -> IonResult<()> {
```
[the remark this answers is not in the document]
zslayton: All of the methods in the `ValueWriter` trait consume `self` to prevent users from misusing it when (e.g.) they go to implement `WriteAsIon`:

```rust
impl WriteAsIon for MyType {
    fn write_as_ion<V: ValueWriter>(&self, writer: V) -> IonResult<()> {
        writer.write_string("foo")?
          .write_i64(1)?
          .write_bool(true)
    }
}
```

This is important for (e.g.) the APIs for writing annotated values and struct fields where there's metadata attached to the single value being produced. (Note that `SequenceWriter` and `StructWriter` implementations take `&mut self` to allow writing many nested values.)

src/lazy/encoder/binary/v1_1/container_writers.rs
```diff
@@ -22,23 +21,90 @@ use crate::{FlexUInt, IonResult, RawSymbolTokenRef, SymbolId};
...
+    pub fn allocator(&self) -> &'top BumpAllocator {
+        self.allocator
+    }
+
+    /// The buffer to which this ContainerWriter encodes child values.
+    pub fn child_values_buffer(&mut self) -> &'_ mut BumpVec<'top, u8> {
+        self.encoder.target_buffer()
+    }
+
+    pub fn has_delimited_containers(&self) -> bool {
```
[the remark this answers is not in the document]
zslayton: Both. The current container is using a delimited encoding, but that also means that all nested containers will use a delimited encoding by default. This allows the user to opt into delimited containers once at the top but still have the ability to use length prefixing where they'd care to.

popematt (approved):

PR #742, opened 2024-04-10
Update CodeBuild Runner integration to include the run id and attempt

popematt:
**Issue #, if available:**

None

**Description of changes:**

According to https://docs.aws.amazon.com/codebuild/latest/userguide/action-runner.html, we need to start including the run id and run attempt in the "runs on" configuration for CodeBuild-hosted runners.

I've also created a job that functions as a "join" operation for all of the forks from the build matrix. We can use the build confirmation job as the required job for the branch protection rule since it has a stable name.

The codecov workflow is failing for this PR. It looks like it's related to [this issue](https://community.codecov.com/t/upload-issues-unable-to-locate-build-via-github-actions-api/3954), so in theory, it's to be expected for any fork->source PRs (since the workflows for the PR can't access GitHub secrets).

If this PR is approved, I will also update the branch protection rule accordingly.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -3,25 +3,59 @@ name: CI Build
 on: [push, pull_request]
 
 jobs:
+  setup:
+    # This job sets up the runners to be used in the matrix for the build workflow.
+    # It provides a list of available runners with stable, human-friendly names and a mapping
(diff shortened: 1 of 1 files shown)


desaikd (approved):
LGTM!

PR #743, opened 2024-04-16
Getting Ion-Rust to Build and Run Examples Locally

JustinDaleGray:
*Issue #, if available:*

 N/A

*Description of changes:*
- Added a clean-rebuild.ps1 for Windows. 
- Updated timestamp to use updated Chrono APIs. 
- Updated BigUint usage in flex_uint.rs. 
- Got write_log_events example to run, but is not right.

### DIFF
--- a/clean-rebuild.ps1
+++ b/clean-rebuild.ps1
@@ -1,0 +1,22 @@
+# Fail on error
+$ErrorActionPreference = "Stop"
+
+# Update the index
+cargo update
+
+# Default build
+cargo clean
+cargo test --workspace
+
+# All features
+cargo clean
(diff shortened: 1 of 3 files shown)


nirosys:
What do you mean when you say the write_log_events isn't right?

JustinDaleGray:
> What do you mean when you say the write_log_events isn't right?

I don't think that it is using the actual offset that is being set and using 0.

zslayton:
@JustinDaleGray -- the crate has evolved quite a bit in the last couple of weeks, so I'm going to close this out for the moment. Let's schedule a time to meet up and I can walk you through the new APIs.

PR #745, opened 2024-04-18
Introduces an application-level lazy writer

zslayton:
Adds an `ApplicationWriter` that wraps the various lazy raw writer types and creates new symbol table entries as needed. The new writer is now being used in our `ion-tests` integration.

I intend to rename the `ApplicationWriter` once the legacy writer is removed.

### DIFF
--- a/src/lazy/encoder/annotate.rs
+++ b/src/lazy/encoder/annotate.rs
@@ -1,17 +1,17 @@
-use crate::lazy::encoder::value_writer::ValueWriter;
+use crate::lazy::encoder::annotation_seq::AnnotationSeq;
+use crate::lazy::encoder::value_writer::{AnnotatableWriter, ValueWriter};
 use crate::lazy::encoder::write_as_ion::WriteAsIon;
-use crate::raw_symbol_token_ref::AsRawSymbolTokenRef;
 use crate::IonResult;
 
 /// Associates a value to serialize with a sequence of annotations.
-pub struct Annotated<'a, T: ?Sized, A> {
+pub struct Annotated<'a, T: ?Sized, A: 'a> {
     value: &'a T,
-    annotations: &'a [A],
+    annotations: A,
 }
 
-/// Provides implementors with an extension method ([`annotate`](Annotate::annotated_with)) that allows
+/// Provides implementors with an extension method ([`annotate`](Annotatable::annotated_with)) that allows
 /// them to be serialized with an associated sequence of annotations.    
-pub trait Annotate {
+pub trait Annotatable {
     /// Pairs a reference to the provided value with a slice containing annotations.
     ///
     /// ```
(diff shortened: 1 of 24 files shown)


tests/ion_tests/mod.rs
```diff
@@ -47,19 +50,19 @@ pub fn serialize(format: Format, elements: &Sequence) -> IonResult<Vec<u8>> {
...
+                std::str::from_utf8(buffer.as_slice()).unwrap()
+            );
         }
         Format::Binary => {
-            let mut binary_writer = BinaryWriterBuilder::new().build(&mut buffer)?;
+            let mut binary_writer = ApplicationWriter::<BinaryEncoding_1_0, _>::new(buffer)?;
             binary_writer.write_elements(elements)?;
-            binary_writer.flush()?;
+            buffer = binary_writer.close()?;
         }
```
zslayton, line 66: 🗺️ The application writer is now used as the serializer in the Ion tests, ensuring its compliance for all (1.0) formats.

src/lazy/encoder/annotate.rs
```diff
@@ -1,30 +1,30 @@
...
     /// use ion_rs::{Element, IonData};
     /// use ion_rs::lazy::encoder::text::LazyRawTextWriter_1_0;
-    /// use ion_rs::lazy::encoder::annotate::Annotate;
+    /// use ion_rs::lazy::encoder::annotate::Annotatable;
     ///
     /// let mut buffer = vec![];
     /// let mut writer = LazyRawTextWriter_1_0::new(&mut buffer);
     ///
-    /// writer.write(42_usize.annotated_with(&["foo", "bar", "baz"]))?.flush()?;
+    /// writer.write(42_usize.annotated_with(["foo", "bar", "baz"]))?.flush()?;
```
popematt, line 27: Does it have to take ownership of the annotations now? Based on the trait impls for `AnnotationSequence`, it looks like it should still work to borrow the slice here.

src/lazy/encoder/binary/v1_0/writer.rs
```diff
@@ -126,6 +128,14 @@ impl<W: Write> LazyRawWriter<W> for LazyRawBinaryWriter_1_0<W> {
             fn flush(&mut self) -> IonResult<()>;
         }
     }
+
+    fn output(&self) -> &W {
+        &self.output
+    }
+
+    fn output_mut(&mut self) -> &mut W {
```
popematt, line 136: Any tradeoffs considered between `output_mut(&mut self)` vs `to_output(self)`? (Does `to_output` have any use case?)

src/lazy/encoder/writer.rs
```diff
@@ -0,0 +1,501 @@
...
+    pub fn new(output: Output) -> IonResult<Self> {
+        Self::with_config(E::default_write_config(), output)
+    }
+
+    /// Constructs a writer for the requested encoding using the provided configuration.
+    pub fn with_config(config: WriteConfig<E>, output: Output) -> IonResult<Self> {
+        let directive_writer = E::Writer::build(config.clone(), vec![])?;
+        let mut data_writer = E::Writer::build(config, vec![])?;
+        // Erase the IVM that's created by default
+        data_writer.output_mut().clear();
```
popematt, line 62: Could you put the initial version into the config, so that the writer always starts with the correct IVM instead of having to clear a default IVM?

zslayton: Now that there's an application-level writer that can automate calling a `write_ivm()` method, I think we could remove the automatic functionality from the raw-level writers.

There's a minor design question around what to do for the 1.0 text writer. When the stream starts, it could leave off the `$ion_1_0`. However, if the user or application writer is calling `write_ivm()` to reset the encoding context, it could not leave it out. We could either have it _always_ include a `$ion_1_0` (unnecessary but pretty harmless) or we could pass an `implicit_ivm_permitted: bool` that other writers could just ignore.

zslayton: I opened #746 to track this.

popematt (approved):

PR #747, opened 2024-04-22
Modifies the serde serializer to use the new writer

zslayton:
This PR replaces usages of the old `UserWriter` with the new `ApplicationWriter` introduced in #745 (currently under review).

It also modifies the encoding of various serde data types to include type annotations:
<table >
  <tr>
    <td>Serde type</td>
    <td>Example Rust</td>
    <td>Old serialization</td>
    <td>New serialization</td>
  </tr>
  <tr>
    <td>Unit struct</td>
    <td>

```rust
struct Foo;
```

</td>
<td>

<pre>
"Foo"
</pre>

</td>

<td>

<pre>
Foo
</pre>

</td>
</tr>

<tr>
    <td>Newtype</td>
    <td>

```rust

struct Foo(i32);
```

</td>
<td>

<pre>
1
</pre>

</td>

<td>

<pre>
Foo::1
</pre>

</td>
</tr>

<tr>
    <td>Tuple struct</td>
    <td>

```rust
struct Foo(i32, i32);
```

</td>
<td>

<pre>
[1, 2]
</pre>

</td>

<td>

<pre>
Foo::[1, 2]
</pre>

</td>
</tr>

<tr>
    <td>Struct</td>
    <td>

```rust

struct Foo {
  x: i32,
  y: i32                    
}
```

</td>
<td>

<pre>
{
  x: 1,
  y: 2,
}
</pre>

</td>

<td>

<pre>
Foo::{
  x: 1,
  y: 2,
}
</pre>

</td>
</tr>

</table>

Similarly, enum variant kinds now also include the name of the enum type:

<table >
  <tr>
    <td>Serde type</td>
    <td>Example Rust</td>
    <td>Old serialization</td>
    <td>New serialization</td>
  </tr>
  <tr>
    <td>Unit struct variant</td>
    <td>

```rust
enum Foo {
  Bar // No associated data        
}
```

</td>
<td>

<pre>
"Bar"
</pre>

</td>

<td>

<pre>
Foo::Bar
</pre>

</td>
</tr>

<tr>
    <td>Newtype variant</td>
    <td>

```rust

enum Foo {
  Bar(i32)                          
}
```

</td>
<td>

<pre>
Bar::1
</pre>

</td>

<td>

<pre>
Foo::Bar::1
</pre>

</td>
</tr>

<tr>
    <td>Tuple variant</td>
    <td>

```rust

enum Foo {
  Bar(i32, i32)                          
}
```

</td>
<td>

<pre>
Bar::[1, 2]
</pre>

</td>

<td>

<pre>
Foo::Bar::[1, 2]
</pre>

</td>
</tr>

<tr>
    <td>Struct variant</td>
    <td>

```rust

enum Foo {
  Bar {
    x: i32,
    y: i32       
  }                   
}
```

</td>
<td>

<pre>
Bar::{
  x: 1,
  y: 2,
}
</pre>

</td>

<td>

<pre>
Foo::Bar::{
  x: 1,
  y: 2,
}
</pre>

</td>
</tr>

</table>

This clarity comes at the expense of a bit of data size, but macros make it possible to reduce the size again if you're so motivated. The type names are not strictly needed as `serde` can determine the expected type from context. However, it may be helpful for humans.

Feedback on this tradeoff is welcome!

### DIFF
--- a/src/binary/raw_binary_writer.rs
+++ b/src/binary/raw_binary_writer.rs
@@ -13,7 +13,7 @@ use crate::ion_writer::IonWriter;
 use crate::raw_symbol_token_ref::{AsRawSymbolTokenRef, RawSymbolTokenRef};
 use crate::result::{IonFailure, IonResult};
 use crate::types::integer::IntData;
-use crate::types::ContainerType;
+use crate::types::ParentType;
 use crate::{Decimal, Int, IonType, SymbolId, Timestamp};
 
 use super::decimal::DecimalBinaryEncoder;
@@ -34,7 +34,7 @@ impl RawBinaryWriterBuilder {
     pub fn build<W: Write>(self, out: W) -> IonResult<RawBinaryWriter<W>> {
         let mut levels = Vec::with_capacity(INITIAL_ENCODING_LEVELS_CAPACITY);
         // Create an EncodingLevel to represent the top level. It has no annotations.
-        levels.push(EncodingLevel::new(ContainerType::TopLevel, None, 0, 0));
+        levels.push(EncodingLevel::new(ParentType::TopLevel, None, 0, 0));
         // Create an empty IoRange for top-level leading scalar values.
         let mut io_ranges = Vec::with_capacity(INITIAL_IO_RANGE_CAPACITY);
         io_ranges.push(0usize..0);
@@ -105,7 +105,7 @@ type IoRange = Range<usize>;
 // annotations, field_id, and container type.
 #[derive(Debug)]
 struct EncodingLevel {
-    container_type: ContainerType,
+    container_type: ParentType,
     field_id: Option<SymbolId>,
     // Annotations are stored in a common Vec on the BinarySystemWriter. Each EncodingLevel tracks
     // how many annotations it had, allowing that Vec to be treated as a stack. Stepping into
@@ -119,7 +119,7 @@ struct EncodingLevel {
 
 impl EncodingLevel {
     fn new(
-        container_type: ContainerType,
+        container_type: ParentType,
         field_id: Option<SymbolId>,
         num_annotations: u8,
         td_io_range_index: usize,
@@ -205,7 +205,7 @@ impl<W: Write> RawBinaryWriter<W> {
     fn is_in_struct(&self) -> bool {
         self.levels
             .last()
-            .map(|level| level.container_type == ContainerType::Struct)
+            .map(|level| level.container_type == ParentType::Struct)
             .unwrap_or(false)
     }
 
@@ -729,9 +729,9 @@ impl<W: Write> IonWriter for RawBinaryWriter<W> {
     fn step_in(&mut self, ion_type: IonType) -> IonResult<()> {
         use IonType::*;
         let container_type = match ion_type {
-            List => ContainerType::List,
-            SExp => ContainerType::SExpression,
-            Struct => ContainerType::Struct,
+            List => ParentType::List,
+            SExp => ParentType::SExp,
+            Struct => ParentType::Struct,
             _ => return IonResult::illegal_operation("Cannot step into a scalar Ion type."),
         };
 
@@ -785,10 +785,10 @@ impl<W: Write> IonWriter for RawBinaryWriter<W> {
         // `self.levels` always has at least one value: the top level.
         // This means we can `unwrap()` the last value safely.
         match self.levels.last().unwrap().container_type {
-            ContainerType::TopLevel => None,
-            ContainerType::Struct => Some(IonType::Struct),
-            ContainerType::List => Some(IonType::List),
-            ContainerType::SExpression => Some(IonType::SExp),
+            ParentType::TopLevel => None,
+            ParentType::Struct => Some(IonType::Struct),
+            ParentType::List => Some(IonType::List),
+            ParentType::SExp => Some(IonType::SExp),
         }
     }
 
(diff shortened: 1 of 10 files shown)


src/serde/ser.rs
```diff
@@ -204,172 +210,206 @@ where
...
     }
 }
 
-impl<'a, E> ser::SerializeSeq for &'a mut Serializer<E>
-where
-    E: IonWriter,
-{
+pub struct SeqWriter<V: ValueWriter> {
+    seq_writer: V::ListWriter,
+}
```
zslayton, line 314: 🗺️ This type and `MapWriter` below exist because it's not possible to implement a trait (in this case: `ser::SerializeSeq` and `ser::SerializeMap`) for a generic like `V::ListWriter` because it violates [the orphan rules](https://doc.rust-lang.org/reference/items/implementations.html#orphan-rules).

These shim types are locally defined, which means that we're not also implementing the trait for other `V: ValueWriter` types we don't own. (Though `ValueWriter` is effectively sealed, sealed traits aren't an official feature and so can't be considered when resolving the orphan rules.)

src/text/raw_text_writer.rs
```diff
@@ -32,7 +32,7 @@ impl RawTextWriterBuilder {
     /// ```
     pub fn compact() -> RawTextWriterBuilder {
         RawTextWriterBuilder {
-            whitespace_config: COMPACT_WHITESPACE_CONFIG.clone(),
+            whitespace_config: COMPACT_WHITESPACE_CONFIG,
```
zslayton, line 35: 🗺️ These types don't own any heap data, so they're now `Copy`.

src/text/raw_text_writer.rs
```diff
@@ -227,11 +227,11 @@ impl<W: Write> RawTextWriter<W> {
 
     /// Called after each value is written to emit an appropriate delimiter before the next value.
     fn write_value_delimiter(&mut self) -> IonResult<()> {
-        use ContainerType::*;
+        use ParentType::*;
         let delimiter = match self.parent_level().container_type {
             TopLevel => "",
             Struct | List => ",",
-            SExpression => "",
+            SExp => "",
```
zslayton, line 234: 🗺️ This variant was renamed for consistency with `IonType::SExp`.

src/text/raw_text_writer.rs
```diff
@@ -679,10 +679,10 @@ impl<W: Write> IonWriter for RawTextWriter<W> {
 
     fn parent_type(&self) -> Option<IonType> {
```
zslayton, line 680: 🗺️ The `RawTextWriter` will be removed in a near-future PR, so I'm not worried about changing this `Option<IonType>` to a `ParentType`.

zslayton (commented):
🗺️ PR tour

src/binary/raw_binary_writer.rs
```diff
@@ -13,7 +13,7 @@ use crate::ion_writer::IonWriter;
 use crate::raw_symbol_token_ref::{AsRawSymbolTokenRef, RawSymbolTokenRef};
 use crate::result::{IonFailure, IonResult};
 use crate::types::integer::IntData;
-use crate::types::ContainerType;
+use crate::types::ParentType;
```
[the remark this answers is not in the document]
popematt: It's `IonDatagram` all over again! 😉

src/lazy/encoder/text/mod.rs
```diff
@@ -50,7 +48,8 @@ impl<W: Write> LazyRawTextWriter_1_0<W> {
         TextValueWriter_1_0::new(
             self,
             0,
-            self.whitespace_config.space_between_top_level_values,
+            "", // No delimiter between values at the top level
```
popematt, line 51: I'm missing something here. I don't see how the other changes allow us to have no delimiter between top level values here. I also don't know why the parent type here is `TopLevel`. I thought this _is_ the top level writer?

src/serde/de.rs
```diff
@@ -410,11 +417,26 @@ impl<'a, 'de> de::Deserializer<'de> for ValueDeserializer<'a, 'de> {
...
     {
-        // annotations are currently only supported with parameterized Rust enums
-        if let Some(annotation) = self.value.annotations().next() {
-            visitor.visit_str(annotation?.text().unwrap())
-        } else {
-            visitor.visit_str(self.value.read()?.expect_text()?)
+        let mut annotations = self.value.annotations();
+        let first_annotation = annotations.next().transpose()?;
+        let second_annotation = annotations.next().transpose()?;
+        match (first_annotation, second_annotation) {
```
popematt, line 423: If I understand this correctly, it will fail if you include _only_ the enum variant annotation. While I could be convinced that it's okay to always write the type annotation, I don't think we should require a type annotation on read because it makes this harder to use it with object mappers in other language ecosystems.

src/text/raw_text_writer.rs
popematt: The various `*_CONFIG` are constants that could just be borrowed, right?

src/lazy/encoder/text/mod.rs
zslayton: > I'm missing something here. I don't see how the other changes allow us to have no delimiter between top level values here.

This code was adding a single space between top level values, but the parameter was actually expecting the caller to provide a _delimiter_ between values -- `,` for list/struct, or the empty string for sexp/top-level. Doing both resulted in too much whitespace in the output; not invalid, but not we wanted.

> I also don't know why the parent type here is TopLevel.

Each parent type (top level, list, sexp, struct) can construct a `*ValueWriter` that will encode the provided value into that context. Because this value writer is being created by the `LazyRawTextWriter_1_0` (and not by a container writer), it is going to write the value it encodes into the top level. Thus, the value's parent is the top level.

src/serde/de.rs
zslayton: I think I'll take your original suggestion and remove the type annotations that I added in this PR. I'm going to do that after my next PR lands though.

src/text/raw_text_writer.rs
zslayton: It was actually already `Box`ed in the `RawTextWriter`, it was only inline in the builder so the fields could be modified. 😅 

Either way, this builder and its writer are being deleted in the next PR. 😛

zslayton:
Merging this with the intent to roll back the type annotations after #748 lands.

PR #748, opened 2024-04-22
Removes the `IonWriter` trait and its implementations

zslayton:
This PR builds on #747.

Removes:
* `IonWriter`
* `RawBinaryWriter`
* `BinaryWriter`
* `RawTextWriter`
* `TextWriter`
* `ElementStreamWriter`

Uses of these types have all been updated to use the new writer API in previous PRs.

Any code additions in this PR are moves, not new code.


src/text/whitespace_config.rs
```diff
@@ -0,0 +1,46 @@
...
+    // Top-level values are independent of other values in the stream, we may separate differently
+    pub(crate) space_between_top_level_values: &'static str,
+    // Non-top-level values are within a container
+    pub(crate) space_between_nested_values: &'static str,
+    // Indentation is repeated before nested values, corresponding to the level of nesting
+    pub(crate) indentation: &'static str,
+    // e.g. after 'foo:' in "{foo: bar}"
+    pub(crate) space_after_field_name: &'static str,
+    // Between the container open and any value in it
+    pub(crate) space_after_container_start: &'static str,
```
popematt, line 12: Is there any way to configure the space between the last value and the container end?

popematt (approved):

src/text/whitespace_config.rs
zslayton: There's not currently. The new writer has a minor formatting "regression" in that it always writes a `", "` after values in a list and fields in a struct. Not incorrect, but the old writer would omit the trailing comma from the last value. I'm open to adding that back in the future.

PR #751, opened 2024-04-24
Values are no longer serialized with a type identifier annotation.

zslayton:
Addresses [this comment](https://github.com/amazon-ion/ion-rust/pull/747#issuecomment-2069824799) from PR #747.

Enum variants and struct types are no longer annotated with their Rust type name. Enums _are_ still annotated with their variant name, which can be modified using standard `serde` config attributes.

### DIFF
--- a/src/serde/de.rs
+++ b/src/serde/de.rs
@@ -10,7 +10,6 @@ use crate::lazy::value_ref::ValueRef;
 use crate::result::IonFailure;
 use crate::serde::decimal::TUNNELED_DECIMAL_TYPE_NAME;
 use crate::serde::timestamp::TUNNELED_TIMESTAMP_TYPE_NAME;
-use crate::symbol_ref::AsSymbolRef;
 use crate::{Decimal, IonError, IonResult, IonType, Timestamp};
 
 /// Generic method that can deserialize an object from any given type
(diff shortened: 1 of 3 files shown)


src/serde/de.rs
```diff
@@ -395,22 +387,14 @@ impl<'a, 'de> de::Deserializer<'de> for ValueDeserializer<'a, 'de> {
...
     ) -> Result<V::Value, Self::Error>
     where
         V: Visitor<'de>,
     {
-        use IonType::*;
-        if self.value.annotations().next() != Some(Ok(name.as_symbol_ref())) {
-            return IonResult::decoding_error("expected an instance of enum {name}");
-        }
-        match self.value.ion_type() {
-            Symbol => visitor.visit_enum(UnitVariantAccess::new(self)),
```
desaikd, line 410: (Question) Shouldn't we still be using this `Symbol` match for `UnitVariantAccess`? I see that the tests are passing, so just want to understand how this change works.

zslayton: This check now happens in `deserialize_identifier`, which helps the enum visitor determine which variant it's working with:
https://github.com/amazon-ion/ion-rust/blob/2f97e99914f4a05100008e6d2945b1456239c5c4/src/serde/de.rs#L408

It now only requires it to be a symbol when it is unannotated, as it's possible for an enum to be defined like this:
```rust
enum Foo {
    Unit,                    // Serializes as an unannotated symbol
    NewType(ion_rs::Symbol), // Serializes as an annotated symbol
}
```
and we don't want the dispatch to see an `IonType::Symbol` enum variant and assume it's a unit type.

desaikd (approved):

src/serde/de.rs
desaikd: Nice. Thanks for explaining! Looks good to me.

PR #752, opened 2024-04-24
Removes the `IonReader` trait and its implementations.

zslayton:
The `LazyReader` is complete/spec-compliant, so we can now remove all of the legacy reader implementation machinery.

With the exception of a couple of unit test updates, there is no new code in this diff, only removals.


desaikd (approved):
LGTM 🚀

Issue #662, opened 2023-10-16
Binary support for Ion 1.1 scalars


zslayton:
Support parsing the new encodings for:
* [ ] Nulls
* [ ] Bools
* [ ] Ints
* [ ] Floats
* [ ] Decimals
* [ ] Timestamps
* [ ] Strings
* [ ] Symbols
* [ ] Blobs
* [ ] Clobs

nirosys:
Base reader PR: #737

PR #753, opened 2024-04-24
Add 1.1 binary reader support for bools

nirosys:
*Issue #, if available:* #662

*Description of changes:*
This PR is the first in the chain of scalar value implementations, implementing the parsing of boolean values.

I've tried to keep each value type separated into its own PR, in order to keep the changes trivial and quick to review.. this is proving to be irritating, so I might merge a few.

---

### DIFF
--- a/src/lazy/binary/raw/v1_1/value.rs
+++ b/src/lazy/binary/raw/v1_1/value.rs
@@ -9,7 +9,7 @@ use crate::{
             raw::{
                 v1_1::{
                     annotations_iterator::RawBinaryAnnotationsIterator_1_1,
-                    immutable_buffer::ImmutableBuffer, Header,
+                    immutable_buffer::ImmutableBuffer, Header, OpcodeType,
                 },
                 value::ValueParseResult,
             },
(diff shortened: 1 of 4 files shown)


src/lazy/binary/raw/v1_1/value.rs
```diff
@@ -186,7 +186,20 @@ impl<'top> LazyRawBinaryValue_1_1<'top> {
...
+        let header = &self.encoded_value.header();
+        let representation = header.type_code();
+        let value = match (representation, header.length_code) {
+            (OpcodeType::Boolean, 0xE) => true,
+            (OpcodeType::Boolean, 0xF) => false,
+            invalid => {
+                return IonResult::decoding_error(format!(
+                    "found a boolean value with an illegal type code representation: {:?}",
+                    invalid
+                ))
```
zslayton, line 199: I assume that this branch can't be reached unless there's a bug, in which case I think it's ok for this to be an `unreachable!("...")` panic.

zslayton (approved):

PR #756, opened 2024-05-01
Add 1.1 binary reader support for floats

nirosys:
*Issue #, if available:* #662

*Description of changes:*
Add support for floats to the 1.1 binary raw reader.

---

### DIFF
--- a/src/lazy/binary/raw/v1_1/reader.rs
+++ b/src/lazy/binary/raw/v1_1/reader.rs
@@ -377,18 +377,6 @@ mod tests {
         Ok(())
     }
 
-    macro_rules! assert_closely_eq {
-        ($x:expr, $y:expr, $d:expr) => {
-            let (a, b) = ($x, $y);
-            if ((a - b).abs() >= $d) {
-                panic!(
-                    "float not close enough to expected value: left: {}, right: {}",
-                    a, b
-                );
-            }
-        };
-    }
-
     #[test]
     #[allow(clippy::approx_constant)]
     fn floats() -> IonResult<()> {
@@ -417,16 +405,14 @@ mod tests {
         // TODO: Implement Half-precision.
         // assert_eq!(reader.next()?.expect_value()?.read()?.expect_float()?, 3.14);
 
-        assert_closely_eq!(
+        assert_eq!(
             reader.next()?.expect_value()?.read()?.expect_float()? as f32,
             3.1415927f32,
-            f32::EPSILON
         );
 
-        assert_closely_eq!(
+        assert_eq!(
             reader.next()?.expect_value()?.read()?.expect_float()?,
             std::f64::consts::PI,
-            f64::EPSILON
         );
 
         Ok(())
--- a/src/lazy/binary/raw/v1_1/reader.rs
+++ b/src/lazy/binary/raw/v1_1/reader.rs
@@ -377,6 +377,61 @@ mod tests {
         Ok(())
     }
 
+    macro_rules! assert_closely_eq {
+        ($x:expr, $y:expr, $d:expr) => {
+            let (a, b) = ($x, $y);
+            if ((a - b).abs() >= $d) {
+                panic!(
+                    "float not close enough to expected value: left: {}, right: {}",
+                    a, b
+                );
+            }
+        };
+    }
+
+    #[test]
+    #[allow(clippy::approx_constant)]
+    fn floats() -> IonResult<()> {
+        #[rustfmt::skip]
+        let data: Vec<u8> = vec![
+            // IVM
+            0xe0, 0x01, 0x01, 0xea,
+            // 0e0
+            0x5A,
+
+            // 3.14 (half-precision)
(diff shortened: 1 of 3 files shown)


src/lazy/binary/raw/v1_1/value.rs
```diff
@@ -222,7 +222,33 @@ impl<'top> LazyRawBinaryValue_1_1<'top> {
...
+                let mut buffer = [0; 4];
+                let val_bytes = self.input.slice(1, 4);
+                buffer[..4].copy_from_slice(val_bytes.bytes());
+
+                f32::from_le_bytes(buffer).into()
+            }
+            2 => todo!("implement half-precision floats"),
+            0 => 0.0f64,
+            n => {
+                return IonResult::decoding_error(format!(
```
nirosys, line 245: Just realized I left this in here, could probably turn this into an `unreachable!`.

src/lazy/binary/raw/v1_1/value.rs
```diff
@@ -222,7 +222,33 @@ impl<'top> LazyRawBinaryValue_1_1<'top> {
...
+                f64::from_le_bytes(buffer)
+            }
+            4 => {
+                let mut buffer = [0; 4];
+                let val_bytes = self.input.slice(1, 4);
+                buffer[..4].copy_from_slice(val_bytes.bytes());
+
+                f32::from_le_bytes(buffer).into()
+            }
+            2 => todo!("implement half-precision floats"),
```
popematt, line 242: Can we create an issue to track this? It would be great for users (and our own benchmarks) if we could use 2-byte floats for "common" numbers like `1.0`.

src/lazy/binary/raw/v1_1/reader.rs
```diff
@@ -300,4 +300,59 @@ mod tests {
...
+        ($x:expr, $y:expr, $d:expr) => {
+            let (a, b) = ($x, $y);
+            if ((a - b).abs() >= $d) {
+                panic!(
+                    "float not close enough to expected value: left: {}, right: {}",
+                    a, b
+                );
+            }
+        };
+    }
```
popematt, line 314: For serialization/deserialization, I don't think we should be considering any epsilon value. A round trip of a float value should result in a value that is equivalent to the original value, and (for non-NaN values) I that means we should expect actual equality of the actual and expected values.

zslayton: +1. If roundtripping from `f64 -> f32 -> f64` results in a different value, then `f32` isn't a valid encoding of that value. You can use `IonEq::ion_eq` to compare floats with the expected `NaN` handling.

nirosys: Ah yea, thank you for calling this out. I had originally intended to round-trip "friendly" float literals and use this so I wasn't using the true float value for comparison. I didn't go that route, and forgot to remove this.

src/lazy/binary/raw/v1_1/value.rs
nirosys: Created: #764

src/lazy/binary/raw/v1_1/type_descriptor.rs
```diff
@@ -109,6 +110,8 @@ impl Header {
         use LengthType::*;
         match (self.ion_type_code, self.length_code) {
             (OpcodeType::Boolean, 0xE..=0xF) => InOpcode(0),
+            (OpcodeType::Float, 0xA) => InOpcode(0),
+            (OpcodeType::Float, 0xB..=0xD) => InOpcode(1 << (self.length_code - 0xA)),
```
zslayton, line 114: > 1 << (self.length_code - 0xA)

Nice!

src/lazy/binary/raw/v1_1/value.rs
```diff
@@ -237,7 +237,28 @@ impl<'top> LazyRawBinaryValue_1_1<'top> {
...
     /// Helper method called by [`Self::read`]. Reads the current value as a float.
     fn read_float(&self) -> ValueParseResult<'top, BinaryEncoding_1_1> {
-        todo!();
+        debug_assert!(self.encoded_value.ion_type() == IonType::Float);
+
+        let value = match self.encoded_value.value_body_length {
+            8 => {
+                let mut buffer = [0; 8];
+                let val_bytes = self.input.slice(1, 8);
+                buffer[..8].copy_from_slice(val_bytes.bytes());
```
zslayton, line 246: Here and below, `ImmutableBuffer` has a method called `bytes_range` (that I really should have named `slice_bytes`) that skips making an intermediate `ImmutableBuffer`.
```suggestion
                let val_bytes = self.input.bytes_range(1, 8);
                buffer[..8].copy_from_slice(val_bytes);
```

zslayton (approved):

src/lazy/binary/raw/v1_1/value.rs
nirosys: Oh awesome, I will add this change in, ty!

PR #757, opened 2024-05-01
Add 1.1 binary reader support for decimals

nirosys:
*Issue #, if available:* #662

*Description of changes:*
Add support for decimals to the 1.1 binary raw reader.

---


src/lazy/binary/raw/v1_1/value.rs
```diff
@@ -227,7 +227,31 @@ impl<'top> LazyRawBinaryValue_1_1<'top> {
...
+            let coefficient_size = self.encoded_value.value_length() - exponent.size_in_bytes();
+            let coefficient = if coefficient_size > 0 {
+                FixedInt::read(
+                    &value_bytes[exponent.size_in_bytes()..],
+                    coefficient_size,
+                    0,
+                )?
+            } else {
+                0i64.into()
+            };
```
popematt, line 250: This makes me think that `FixedInt::read` should return zero if the size is zero. This seems like something that could come up in other places, and rather than handle this condition everywhere, we should probably just move it to `FixedInt::read`.

Issue #631, opened 2023-08-22
Explore moving struct field info out of `Lazy*Value` and into `Lazy*Field`


zslayton:
Not only would this be a more sensible factoring, but `EncodedTextValue`s without a field would occupy less space in memory.

PR #760, opened 2024-05-04
Adds APIs for accessing encoding of raw stream items

zslayton:
In aggregate, the changes in this PR make it straightforward to access a "lower level" view of a given item in an Ion stream, determining whether it was an input literal, part of a template, or a constructed value. This enables the development of a variety of tools.

Changes:

* Uses the bump allocator to handle text escape processing, allowing `RawSymbolTokenRef` to hold a reference to a `&'bump str` instead of potentially owning a `String`. This change allows the `RawSymbolTokenRef` type to implement `Copy`, which in turn allows all of the `LazyExpandedValue`- and `LazyValue`-related types to also implement `Copy`.
* Removes the `RawSymbolToken` type, which is now redundant to the `RawSymbolTokenRef` type.
* Adds a `Span` type that provides access to the input bytes that comprised various raw stream items.
* Adds a `LazyRawVersionMarker` trait and per-encoding impls that can provide a `Span` upon request.
* Moves the field encoding information from the `LazyDecoder::Value` types into implementations of a new `LazyRawFieldName` trait. This fixes issue #631.
* Adds a `LazyRawField` trait and per-encoding impls that can provide a `Span` upon request.
* Adds an `UnexpandedField` type that can represent both raw struct fields and struct fields from a template body. This simplified the code for expanding structs.
* Adds methods to convert container types back to the general value type.
* Adds `EncodedBinaryValueData_1_0` and `EncodedBinaryAnnotations_1_0` types that can be used to access spans and ranges for the various components of a binary 1.0 value.

This patch exposes many functions and types which we likely wish to feature gate, but that change is being left for a future PR.


examples/write_log_events.rs
```diff
@@ -194,9 +194,9 @@ mod example {
                 .write(11, event.thread_id)?
                 .write(12, &event.thread_name)?
                 //                 v--- The fixed strings from the log statement are also SIDs
-                .write(13, RawSymbolToken::SymbolId(17))? // logger name
-                .write(14, RawSymbolToken::SymbolId(18))? // log level
-                .write(15, RawSymbolToken::SymbolId(19))? // format
+                .write(13, RawSymbolTokenRef::SymbolId(17))? // logger name
+                .write(14, RawSymbolTokenRef::SymbolId(18))? // log level
+                .write(15, RawSymbolTokenRef::SymbolId(19))? // format
```
zslayton, line 199: 🗺️ Now that `RawSymbolTokenRef` can hold a reference to a decoded string in the bump allocator, there's no longer a need for a separate `RawSymbolToken` type that can own a string.

Down the line, I'd like to rename `RawSymbolTokenRef` to simply `RawSymbol`.

src/lazy/any_encoding.rs
```diff
@@ -62,23 +72,99 @@ impl LazyDecoder for AnyEncoding {
...
+    encoding: LazyRawAnyVersionMarkerKind<'top>,
 }
+
+#[derive(Debug, Copy, Clone)]
+pub enum LazyRawAnyVersionMarkerKind<'top> {
+    Text_1_0(LazyRawTextVersionMarker_1_0<'top>),
+    Binary_1_0(LazyRawBinaryVersionMarker_1_0<'top>),
+    Text_1_1(LazyRawTextVersionMarker_1_1<'top>),
+    Binary_1_1(LazyRawBinaryVersionMarker_1_1<'top>),
+}
```
zslayton, line 92: 🗺️ There's now a per-encoding version marker type that can provide a `Span` of its input bytes. Previously we only surfaced the `(major: u8, minor: u8)` pair, but didn't have information about its encoding (particularly its location).

src/lazy/any_encoding.rs
```diff
@@ -62,23 +72,99 @@ impl LazyDecoder for AnyEncoding {
...
+
+impl<'top> HasSpan<'top> for LazyRawAnyVersionMarker<'top> {
+    fn span(&self) -> Span<'top> {
+        use LazyRawAnyVersionMarkerKind::*;
+        match self.encoding {
+            Text_1_0(marker) => marker.span(),
+            Binary_1_0(marker) => marker.span(),
+            Text_1_1(marker) => marker.span(),
+            Binary_1_1(marker) => marker.span(),
+        }
```
zslayton, line 102: 🗺️ This file (`LazyRawAny_____` types) is entirely enum dispatch. A while ago I looked to see if I could use the `enum_dispatch` crate to eliminate all the repetition, but the use of lifetimes was an obstacle. I have not checked recently to see if that limitation has been overcome.

src/lazy/any_encoding.rs
```diff
@@ -280,6 +384,13 @@ pub struct LazyRawAnyValue<'top> {
     encoding: LazyRawValueKind<'top>,
 }
 
+impl<'top> LazyRawAnyValue<'top> {
+    // TODO: feature gate
+    pub fn kind(&self) -> LazyRawValueKind<'top> {
+        self.encoding
+    }
+}
```
zslayton, line 392: 🗺️ Allows an application to access the encoding-specific data type that underlies this value.

src/lazy/any_encoding.rs
```diff
@@ -1211,7 +1412,7 @@ mod tests {
 
             assert!(matches!(
                 reader.next(&allocator)?,
-                LazyRawStreamItem::<AnyEncoding>::EndOfStream
+                LazyRawStreamItem::<AnyEncoding>::EndOfStream(_)
```
zslayton, line 1415: 🗺️ The `EndOfStream` item now returns an `EndPosition` that can provide its range and span. This is not especially useful in and of itself, but it allows the `RawStreamItem` type that contains it to also provide a range and span for every item.

src/lazy/binary/encoded_value.rs
```diff
@@ -53,33 +52,29 @@ pub(crate) struct EncodedValue<HeaderType: EncodedHeader> {
     // and IonType.
     pub(crate) header: HeaderType,
 
-    // Each encoded value has up to five components, appearing in the following order:
+    // Each encoded value has up to four components, appearing in the following order:
     //
-    // [ field_id? | annotations? | header (type descriptor) | header_length? | value ]
+    // [ annotations? | header (type descriptor) | header_length? | value_body ]
```
zslayton, line 57: 🗺️ The `field_id` offset information (and related methods) have been removed from the `EncodedValue` type and now lives in each encoding's `LazyRawFieldName` implementation. This means that we only pay the data size and offset processing cost of field IDs when we're actually processing a struct field.

src/lazy/binary/encoded_value.rs
```diff
@@ -89,9 +84,9 @@ pub(crate) struct EncodedValue<HeaderType: EncodedHeader> {
     pub length_length: u8,
     // The number of bytes used to encode the value itself, not including the header byte
     // or length fields.
-    pub value_length: usize,
+    pub value_body_length: usize,
```
zslayton, line 87: 🗺️ Several fields/methods with `value` in the name do not include information about the value's opcode or its trailing length. I've renamed them to make this more explicit.

src/lazy/binary/immutable_buffer.rs
```diff
@@ -143,15 +153,16 @@ impl<'a> ImmutableBuffer<'a> {
...
             .peek_n_bytes(IVM.len())
             .ok_or_else(|| IonError::incomplete("an IVM", self.offset()))?;
 
         match bytes {
             [0xE0, major, minor, 0xEA] => {
-                let version = (*major, *minor);
-                Ok((version, self.consume(IVM.len())))
+                let matched = ImmutableBuffer::new_with_offset(bytes, self.offset);
+                let marker = LazyRawBinaryVersionMarker_1_0::new(matched, *major, *minor);
+                Ok((marker, self.consume(IVM.len())))
```
zslayton, line 165: 🗺️ We now surface a proper version marker type that holds the buffer slice that was originally matched.

src/lazy/binary/raw/reader.rs
```diff
@@ -188,12 +200,41 @@ impl<'data> DataSource<'data> {
         self.bytes_to_skip = lazy_value.encoded_value.total_length();
         Ok(Some(lazy_value))
     }
+
+    /// Runs the provided parsing function on this DataSource's buffer.
+    /// If it succeeds, marks the `DataSource` as ready to advance by the 'n' bytes
+    /// that were consumed.
+    /// If it does not succeed, the `DataSource` remains unchanged.
+    pub(crate) fn try_parse_next_field<
```
zslayton, line 208: 🗺️ Now that struct fields are their own type (and not just a different "view" of a value), they need their own parsing method on the reader.

src/lazy/binary/raw/v1_1/sequence.rs
```diff
@@ -27,6 +29,12 @@ impl<'top> LazyContainerPrivate<'top, BinaryEncoding_1_1> for LazyRawBinaryList_
     }
 }
 
+impl<'top> LazyRawContainer<'top, BinaryEncoding_1_1> for LazyRawBinaryList_1_1<'top> {
+    fn as_value(&self) -> <BinaryEncoding_1_1 as LazyDecoder>::Value<'top> {
+        self.sequence.value
+    }
+}
+
```
zslayton, line 37: 🗺️ All lazy container types can now be converted back to the original (less specific) 'value' representation.

src/lazy/binary/raw/value.rs
```diff
@@ -78,20 +116,160 @@ impl<'top> LazyRawValue<'top, BinaryEncoding_1_0> for LazyRawBinaryValue_1_0<'to
...
         self.read()
     }
+}
 
-    fn range(&self) -> Range<usize> {
-        self.encoded_value.annotated_value_range()
+#[derive(Copy, Clone)]
+pub struct EncodedBinaryAnnotations_1_0<'a, 'top> {
+    value: &'a LazyRawBinaryValue_1_0<'top>,
+}
```
zslayton, line 124: 🗺️ This type offers methods for accessing the various encoding components of a binary 1.0 annotations sequence.

src/lazy/binary/raw/value.rs
```diff
@@ -78,20 +116,160 @@ impl<'top> LazyRawValue<'top, BinaryEncoding_1_0> for LazyRawBinaryValue_1_0<'to
...
+        let bytes = &self.span().bytes()[local_start..];
+        let stream_start = range.start + local_start;
+        Span::with_range(stream_start..stream_start + sequence_length, bytes)
+    }
+}
+
+#[derive(Copy, Clone)]
+pub struct EncodedBinaryValueData_1_0<'a, 'top> {
+    value: &'a LazyRawBinaryValue_1_0<'top>,
+}
```
zslayton, line 185: 🗺️ This type offers methods for accessing the various encoding components of a binary 1.0 value.

src/lazy/binary/raw/value.rs
```diff
@@ -173,39 +343,27 @@ impl<'top> LazyRawBinaryValue_1_0<'top> {
...
-                "only part of the requested value is available in the buffer",
-                self.input.offset(),
-            );
-        }
-        let value_body_length = self.encoded_value.value_length();
+        debug_assert!(self.input.len() >= value_total_length);
+        let value_body_length = self.encoded_value.value_body_length();
         let value_offset = value_total_length - value_body_length;
-        Ok(self.input.bytes_range(value_offset, value_body_length))
+        self.input.bytes_range(value_offset, value_body_length)
```
zslayton, line 351: 🗺️ An earlier version of the lazy reader API did not have the invariant that the top level value was fully in the buffer, so the `value_body` method would confirm it on each call. Now it's always true -- it's checked when the lazy value is constructed.

src/lazy/bytes_ref.rs
```diff
@@ -1,25 +1,24 @@
...
 use crate::Bytes;
-use std::borrow::Cow;
 use std::fmt::{Debug, Display, Formatter};
 use std::ops::Deref;
 
-#[derive(Clone)]
+#[derive(Copy, Clone)]
 pub struct BytesRef<'data> {
-    data: Cow<'data, [u8]>,
+    data: &'data [u8],
```
zslayton, line 8: 🗺️ Having types like this implement `Copy` (because decoding can use the bump allocator instead of the heap) should offer a nice speed boost. It also makes a lot of the APIs easier to work with -- every 'lazy' item type is now also `Copy`.

src/lazy/encoder/text/v1_0/value_writer.rs
```diff
@@ -28,67 +28,13 @@ pub struct TextValueWriter_1_0<'value, W: Write + 'value> {
...
-/// Unlike other symbols, identifiers don't have to be wrapped in quotes.
-fn token_is_identifier(token: &str) -> bool {
-    if token.is_empty() {
-        return false;
-    }
-    let mut chars = token.chars();
-    let first = chars.next().unwrap();
-    (first == '$' || first == '_' || first.is_ascii_alphabetic())
-        && chars.all(|c| c == '$' || c == '_' || c.is_ascii_alphanumeric())
-}
```
zslayton, line 47: 🗺️ The methods in this section were originally copied from the `IonValueFormatter`. This PR introduces the `IoFmtShim` type that allows them to be re-used instead of copied.

src/lazy/expanded/struct.rs
```diff
@@ -320,25 +393,7 @@ impl<'top, D: LazyDecoder> Iterator for ExpandedStructIterator<'top, D> {
...
-                            RawFieldExpr::NameValuePair(name, RawValueExpr::ValueLiteral(value)),
-                        ),
-                        RawFieldExpr::MacroInvocation(invocation) => {
-                            let resolved_invocation = invocation.resolve(context)?;
-                            Ok(RawFieldExpr::MacroInvocation(resolved_invocation.into()))
-                        }
-                    },
-                );
-                Self::next_field_from(context, state, e_exp_evaluator, &mut iter_adapter)
+                Self::next_field_from(context, state, e_exp_evaluator, raw_struct_iter)
```
zslayton, line 396: 🗺️ The introduction of the `UnexpandedField` enum made it possible to eliminate a lot of the complexity and generics involved in struct expansion.

src/lazy/text/buffer.rs
```diff
@@ -86,22 +86,6 @@ const WHITESPACE_CHARACTERS: &[char] = &[
...
-    mut parser: P,
-) -> impl Parser<I, O, E> {
-    move |input: I| {
-        let result = parser.parse(input);
-        #[cfg(debug_assertions)]
-        println!("{}: {:?}", label, result);
-        result
-    }
-}
-
```
zslayton, line 104: 🗺️ This method would cause a clippy warning in release builds because `label` wouldn't be used. The text reader is spec compliant, so I feel we don't really need this debug helper anymore.

src/lazy/text/buffer.rs
```diff
@@ -462,56 +451,20 @@ impl<'top> TextBufferView<'top> {
...
-                Self::match_e_expression,
+                Self::match_e_expression.map(|eexp| Ok(Some(LazyRawFieldExpr::EExp(eexp)))),
                 whitespace_and_then(alt((tag(","), peek(tag("}"))))),
-            )
-            .map(|invocation| Ok(Some(RawFieldExpr::MacroInvocation(invocation)))),
+            ),
             Self::match_struct_field_name_and_e_expression_1_1.map(|(field_name, invocation)| {
-                // TODO: We're discarding the name encoding information here. When we revise our field name
-                //       storage strategy[1], we should make sure to capture this for tooling's sake.
-                //       [1]: https://github.com/amazon-ion/ion-rust/issues/631
```
zslayton, line 472: 🗺️ 🎉

src/lazy/text/matched.rs
```diff
@@ -383,19 +389,25 @@ impl MatchedString {
     // Strings longer than 64 bytes will allocate a larger space on the heap.
     const STACK_ALLOC_BUFFER_CAPACITY: usize = 64;
 
-    pub fn read<'data>(&self, matched_input: TextBufferView<'data>) -> IonResult<StrRef<'data>> {
+    pub fn read<'data>(
+        &self,
+        allocator: &'data BumpAllocator,
```
zslayton, line 394: 🗺️ Text escapes and base64 decoding now happen in bump-allocated scratch space instead of heap-allocated `String`s and `Vec`s.

zslayton (commented):
🗺️ PR tour

src/lazy/binary/immutable_buffer.rs
```diff
@@ -745,15 +753,12 @@ impl<'a> ImmutableBuffer<'a> {
...
-            // If applicable, these are populated by the caller: `peek_field()`
-            field_id_length: 0,
-            field_id: None,
             // If applicable, these are populated by the caller: `read_annotated_value()`
             annotations_header_length: 0,
             annotations_sequence_length: 0,
             header_offset,
             length_length,
-            value_length,
+            value_body_length: value_length,
```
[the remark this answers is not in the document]
zslayton: I changed the field name on the type and my IDE very conservatively kept the constructor parameter name the same during the refactor.

src/lazy/expanded/mod.rs
```diff
@@ -477,11 +480,31 @@ impl<'top, V: RawValueLiteral, Encoding: LazyDecoder<Value<'top> = V>> From<V>
...
+}
+
 /// A value produced by expanding the 'raw' view of the input data.
 #[derive(Copy, Clone)]
 pub struct LazyExpandedValue<'top, Encoding: LazyDecoder> {
     pub(crate) context: EncodingContext<'top>,
     pub(crate) source: ExpandedValueSource<'top, Encoding>,
+    // If this value came from a variable reference in a template macro expansion, the
+    // template and the name of the variable can be found here.
+    pub(crate) variable: Option<TemplateVariableReference<'top>>,
```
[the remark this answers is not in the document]
zslayton: Yep. It turns out it's quite large, though--it takes up ~half of the layout of `LazyExpandedValue`. In a future PR I'll be looking to pare it back.

src/lazy/expanded/template.rs
```diff
@@ -335,25 +325,19 @@ impl<'top, D: LazyDecoder> Iterator for TemplateStructRawFieldsIterator<'top, D>
...
-                    }
-                    ValueExpr::MacroInvocation(invocation) => {
-                        RawValueExpr::MacroInvocation(*invocation)
-                    }
-                }
+            TemplateBodyValueExpr::Variable(variable) => {
+                let arg_expr = self
+                    .environment
+                    .get_expected(variable.signature_index())
+                    .expect("reference to non-existent parameter");
```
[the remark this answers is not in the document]
zslayton: If it happens it's a bug, not an input error. It means that the template compiler stored a reference to a parameter that doesn't exist.

src/lazy/span.rs
```diff
@@ -0,0 +1,57 @@
...
+
+/// Represents a slice of input data.
+///
+/// Various items in the Reader API implement the [`HasSpan`](crate::lazy::decoder::HasSpan) trait,
+/// allowing the byte slice that represented that item in the input by calling `span()`.
+#[derive(Debug, Copy, Clone)]
+pub struct Span<'a> {
+    bytes: &'a [u8],
+    start: usize,
+    end: usize,
```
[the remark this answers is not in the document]
zslayton: The `bytes` field contains only those input bytes which represent the item implementing `HasSpan`.

`Span` began its life as a `(Range<usize>, &[u8])`. However, `Range<usize>` doesn't implement `Copy`, (which is hopefully [changing in edition 2024](https://github.com/rust-lang/rfcs/pull/3550)! 🤞), so I split it into distinct `start` and `end` fields. However, looking back at it now I see that having both `start` and `end` is redundant when we're also holding a slice. I'll reduce them to a single absolute `offset` indicating where the slice begins in the overall stream. The range can be derived from the slice length.

src/lazy/struct.rs
```diff
@@ -78,6 +77,28 @@ impl<'top, D: LazyDecoder> LazyStruct<'top, D> {
         }
     }
 
+    //TODO: Feature gate
+    pub fn lower(&self) -> LazyExpandedStruct<'top, D> {
+        self.expanded_struct
+    }
+    pub fn as_value(&self) -> LazyValue<'top, D> {
```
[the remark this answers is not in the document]
zslayton: The whole reader API will be behind a feature gate for the near future. However, I think `as_value` can be exposed when the reader API is because `LazyValue` and the `LazyList`/`LazyStruct/`LazySExp` types are all part of the reader API.

src/lazy/span.rs
```diff
@@ -0,0 +1,57 @@
...
+        self.expect_text().ok()
+    }
+
+    pub fn expect_text(&self) -> IonResult<&'a str> {
+        std::str::from_utf8(self.bytes)
+            .map_err(|_| IonError::decoding_error("span text was not valid UTF-8"))
+    }
+}
+
+pub trait ToRelativeRange {
```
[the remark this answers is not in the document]
zslayton: Good catch, it's a holdover. Removed.

src/lazy/text/buffer.rs
```diff
@@ -201,6 +185,11 @@ impl<'top> TextBufferView<'top> {
         self.data.len()
     }
 
+    /// Returns the stream byte offset range that this buffer contains.
+    pub fn range(&self) -> Range<usize> {
+        self.offset..self.offset + self.len()
+    }
```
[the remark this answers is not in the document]
zslayton: It could, sure. This method predates those traits. 

`TextBufferView` is not a public-facing type though, so there's not really a need for it IMO.

src/lazy/text/encoded_value.rs
```diff
@@ -42,17 +43,14 @@ pub(crate) struct EncodedTextValue<'top, E: TextEncoding<'top>> {
     data_offset: usize,
     // The number of bytes _before_ `data_offset` at which the field name begins. If this value
     // does not have a field name, this value will be zero.
-    field_name_offset: u32,
+    // field_name_offset: u32,
```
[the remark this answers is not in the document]
zslayton: I've removed this in a follow-on PR that cleans up a lot of the text reader data structures

src/lazy/raw_stream_item.rs
```diff
@@ -109,28 +109,35 @@ impl<M: Copy + Debug, V: Copy + Debug, E: Copy + Debug> RawStreamItem<M, V, E> {
     }
 }
 
+/// Represents the end of a [`RawInputStream`].
```
popematt, line 112: FYI—there's a GHA annotation on this line of this comment that says that `RawInputStream` is an unresolved link.

popematt (approved):

PR #762, opened 2024-05-07
Ensure that FixedInt and FixedUInt handle zero-length integers

popematt:
**Issue #, if available:**

None

**Description of changes:**

As I was looking at #757, I realized that there are cases where we may have a zero-width FixedInt (and maybe FixedUInt), so I'm updating their implementations to be able to read that correctly. While I was at it, I managed to get rid of the branching for positive and negative numbers in `FlexInt::read`. No changes were needed in `FlexUInt::read`, but I added a test case for it anyway.


zslayton (approved):
Nice.

PR #763, opened 2024-05-07
Text reader implementation cleanup

zslayton:
Many of the types that comprise the text reader API are quite large. This PR makes a variety of small changes to shrink their layout, hopefully helping to minimize the need for `memcpy`s and reducing bump allocated memory usage. Benchmarks showed modest improvements in the 1-3% range. I suggest reviewing the commits individually.

Following PR #760, field name information is no longer stored in the `LazyRawTextValue`; it has its own type that is only constructed in the context of a struct. Commit 3f68fab3b45a3be577bbc7a577da85cc97cb07b0 modifies text values so that their `TextBufferView` layout starts with the annotations sequence (if any) and ends with the final byte of the value. Prior to this change, many values held a slice that contained the rest of the available input. This change makes it possible to use the `TextBufferView`'s offset and length as a proxy for the value's own. It also eliminates most of the metadata stored in the `LazyRawTextValue`, leaving only a `data_offset: u16` that indicates where the first byte of the value is.

The same commit also modified the `MacroEvaluator` to avoid storing an extra copy of the allocator when one could be fetched from a nested `BumpVec`. Similarly, the `MacroEvaluator` can now reference the `EncodingContext` that lives on each `MacroExpansion`.

Commit 054df229fab481c6f95c1f166c60fcd798e41088 causes the `MacroEvaluator` to use a capacity hint when bump-allocating space to store macro argument expressions.

Commit a8514b600a86f8353738f24fbdfc846364543e59 moves the `EncodingContext` that was previously copied into  instances of most types into the bump allocator and passes around a reference to it instead. This shrank the size of many types by 16 bytes.

Commit c0256c078ead34dd3166529da44011f6946ace1d removes the superfluous `MatchedRawTextValue` type, which was intended to be a shim allowing the 1.0 and 1.1 text value types to use the same structure. I was able to remove the need for it via generics.

Commit 375915d43d9ccd492e42ae4f1a045ead6d4144e0 shrinks the size of the macro parameter index stored in each `TemplateVariableReference` from `usize` to `u16`. This made space for the `TemplateVariableReference`'s enum discriminant, saving 8 bytes.

-------

Here are some example data type size reductions for `AnyEncoding`:

| Data type | Size Before | Size After |
| ---------- |:-----------:|:---------:|
| `LazyValue` | 128 | 96 |
| `LazyField` | 208 | 160 |
| `StructIterator` | 320 | 224|

While some types were specifically target for improvement, many types benefitted from the `EncodingContext` change.

### DIFF
--- a/src/lazy/binary/raw/v1_1/value.rs
+++ b/src/lazy/binary/raw/v1_1/value.rs
@@ -175,7 +175,6 @@ impl<'top> LazyRawBinaryValue_1_1<'top> {
     fn value_body(&self) -> IonResult<&'top [u8]> {
         let value_total_length = self.encoded_value.total_length();
         if self.input.len() < value_total_length {
-            eprintln!("[value_body] Incomplete {:?}", self);
             return IonResult::incomplete(
                 "only part of the requested value is available in the buffer",
                 self.input.offset(),
--- a/src/lazy/binary/raw/value.rs
+++ b/src/lazy/binary/raw/value.rs
@@ -231,7 +231,7 @@ impl<'a, 'top> EncodedBinaryValueData_1_0<'a, 'top> {
     pub fn trailing_length_span(&self) -> Span<'top> {
         let stream_range = self.trailing_length_range();
         let offset = self.value.input.offset();
-        let local_range = stream_range.start - offset .. stream_range.end - offset;
+        let local_range = stream_range.start - offset..stream_range.end - offset;
         let bytes = &self.value.input.bytes()[local_range];
         Span::with_offset(stream_range.start, bytes)
     }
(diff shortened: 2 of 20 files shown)


popematt (approved):

popematt:
Given that [`mem::size_of`](https://doc.rust-lang.org/std/mem/fn.size_of.html#) returns the size in bytes, I assume that the sizes given in your table are also in bytes. If that's the case, I'm a little bit surprised at how large some of those types are, but my point of comparison is the `IonValue` and `IonElement` types, which are for JVM... so it's not really a fair comparison. 

For my own interest, do you know if there are any tools that can be used that show how the bytes are used for the various parts of a type?

PR #765, opened 2024-05-13
Add 1.1 binary reader support for timestamps.

nirosys:
*Issue #, if available:* #662

*Description of changes:*
This PR adds support for timestamps to the binary ion 1.1 reader.

---


src/lazy/binary/raw/v1_1/value.rs
```diff
@@ -265,9 +268,298 @@ impl<'top> LazyRawBinaryValue_1_1<'top> {
...
+            offset as i32
+        };
+        let ts_builder = ts_builder.with_hour_and_minute(hour, minute);
+
+        if value_length == 6 {
+            if offset != 0x0FFF {
+                let ts_builder = ts_builder.with_offset(offset);
+                return Ok(RawValueRef::Timestamp(ts_builder.build()?));
+            }
+            return Ok(RawValueRef::Timestamp(ts_builder.build()?));
```
popematt, line 513: You can get rid of the double `return`s like this:
```suggestion
            let ts_builder = if offset == 0x0FFF { ts_builder } else {
                ts_builder.with_offset(offset);
            }
            return Ok(RawValueRef::Timestamp(ts_builder.build()?));
```

If you take my previous suggestion to use `Option<i32>` for the offset, then it can be further simplified to
```rust
let ts_builder = offset.map_or(ts_builder, |o| ts_builder.with_offset(o));
return Ok(RawValueRef::Timestamp(ts_builder.build()?));
```

...which, I suppose, is a signal that the `TimestampBuilder.with_offset` should probably be changed to accept `Option<i32>`.

src/lazy/binary/raw/v1_1/value.rs
```diff
@@ -265,9 +268,298 @@ impl<'top> LazyRawBinaryValue_1_1<'top> {
...
+                "invalid scale for timestamp fractional second",
+            ));
+        }
+
+        let frac_sec = Decimal::new(coefficient, -(scale.value() as i64));
+        if frac_sec.is_greater_than_or_equal_to_one() {
+            return Err(IonError::decoding_error(
+                "invalid fractional second for timestamp",
+            ));
+        }
```
popematt, line 543: `Timestamp` ensures its own invariants for fractional seconds, so these checks are not needed.
```suggestion
        let frac_sec = Decimal::new(coefficient, -(scale.value() as i64));
```

src/lazy/binary/raw/v1_1/value.rs
```diff
@@ -265,9 +268,298 @@ impl<'top> LazyRawBinaryValue_1_1<'top> {
...
     /// Helper method called by [`Self::read`]. Reads the current value as a timestamp.
     fn read_timestamp(&self) -> ValueParseResult<'top, BinaryEncoding_1_1> {
-        todo!();
+        debug_assert!(self.encoded_value.ion_type() == IonType::Timestamp);
+
+        match self.encoded_value.header.type_code() {
+            OpcodeType::TimestampShort => self.read_timestamp_short(),
+            OpcodeType::TimestampLong => self.read_timestamp_long(),
+            _ => unreachable!("invalid timestamp type_code"),
+        }
```
popematt, line 562: Rather than matching on `IonType` in the `read` function, and then matching on `OpcodeType` here, maybe we should consider matching on `OpcodeType` directly in `read`. What do you think?

If you agree, please create an issue for it. I don't want to blow up the scope of this PR.

src/types/mod.rs
```diff
@@ -14,7 +14,7 @@ mod sexp;
 mod string;
 mod r#struct;
 pub(crate) mod symbol;
-mod timestamp;
+pub(crate) mod timestamp;
```
popematt, line 17: Is this needed so that you can use `TimestampBuilder` in a function signature? If so, I wonder if we should actually change L29 to include `TimestampBuilder`.

src/lazy/binary/raw/v1_1/value.rs
```diff
@@ -265,9 +268,298 @@ impl<'top> LazyRawBinaryValue_1_1<'top> {
...
+        let year = (value_bytes[0] as u32) | ((value_bytes[1] & 0x3F) as u32) << 8;
+        let ts_builder = Timestamp::with_year(year);
+        if value_length == 2 {
+            return Ok(RawValueRef::Timestamp(ts_builder.build()?));
+        }
+
+        let month = ((value_bytes[1] & 0xC0) as u32) >> 6 | ((value_bytes[2] & 0x03) as u32) << 2;
+        let day = ((value_bytes[2] & 0x7C) as u32) >> 2;
+        let ts_builder = ts_builder.with_month(month);
+        if value_length == 3 && day == 0 {
```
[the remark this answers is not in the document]
nirosys: This is the month precision case, when the day component is 0. If it's not 0, then it gets added to the timestamp on 493.

src/lazy/binary/raw/v1_1/value.rs
nirosys: Unfortunately, TimestampBuilder's precision is managed through a generic parameter, so each precision change also changes the type. So the if-else fails to compile because its branches produce 2 different types. That's the bulk of the reason for carrying the offset and setting it only when the timestamp is built.

nirosys:
Moving this over to 'ready for review', going to add the comments and tidying I had planned for while in draft to the updates addressing feedback.

src/lazy/binary/raw/v1_1/value.rs
popematt: Oh, right. I forgot about that.

src/lazy/binary/raw/v1_1/value.rs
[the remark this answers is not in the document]
popematt: What I'm trying to say is that I don't think we need to check `value_length` here. We already know that it's not years precision, and year and month precisions are the only precision that could have `day == 0`. All other precisions must have a non-zero day.

If `day == 0` then `value_length` must be `3` for any valid Ion... but now that I write it out, I realize that keeping the check would actually protect against invalid data, so we should probably keep it.

popematt (approved):

nirosys:
Rebased to get rid of conflicts. Looks like some previous changes "broke" the `From<FixedUInt>` for Coefficient that I had, so I re-worked it to use `TryFrom`.

PR #766, opened 2024-05-13
Add 1.1 binary reader support for typed nulls

nirosys:
*Issue #, if available:* #662

*Description of changes:*
This PR adds support for typed nulls to the binary ion 1.1 reader.

---

### DIFF
--- a/src/lazy/binary/raw/v1_1/type_descriptor.rs
+++ b/src/lazy/binary/raw/v1_1/type_descriptor.rs
@@ -14,6 +14,20 @@ pub struct Opcode {
 /// A statically defined array of TypeDescriptor that allows a binary reader to map a given
 /// byte (`u8`) to a `TypeDescriptor` without having to perform any masking or bitshift operations.
 pub(crate) static ION_1_1_OPCODES: &[Opcode; 256] = &init_opcode_cache();
+pub(crate) static ION_1_1_TYPED_NULL_TYPES: &[IonType; 12] = &[
+    IonType::Bool,
+    IonType::Int,
+    IonType::Float,
+    IonType::Decimal,
+    IonType::Timestamp,
+    IonType::String,
+    IonType::Symbol,
+    IonType::Blob,
+    IonType::Clob,
+    IonType::List,
+    IonType::SExp,
+    IonType::Struct,
+];
 
 static ION_1_1_TIMESTAMP_SHORT_SIZE: [u8; 13] = [1, 2, 2, 4, 5, 6, 7, 8, 5, 5, 7, 8, 9];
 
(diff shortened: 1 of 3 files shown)


src/lazy/binary/raw/v1_1/type_descriptor.rs
```diff
@@ -14,6 +14,20 @@ pub struct Opcode {
...
+    IonType::Decimal,
+    IonType::Timestamp,
+    IonType::String,
+    IonType::Symbol,
+    IonType::Blob,
+    IonType::Clob,
+    IonType::List,
+    IonType::SExp,
+    IonType::Struct,
+];
```
popematt, line 30: Do you think there may be any benefit to using a `match` instead of a lookup table? E.g.:
```rust
#[inline]
fn get_type_of_null(null_body: u8) -> IonType {
    match null_body {
        0x0 => IonType::Bool,
        0x1 => IonType::Int,
        // ...
        0xB => IonType::Struct,
        _ => unreachable!(),
    }
}
```
If it isn't obvious which one is better and/or there is no consensus in the Rust community, then we should just create an issue to revisit `match` vs lookup tables at some later time. No need to go deep into a rabbit hole on this now.

src/lazy/binary/raw/v1_1/value.rs
```diff
@@ -150,8 +151,13 @@ impl<'top> LazyRawBinaryValue_1_1<'top> {
     /// or [`LazyStruct`](crate::lazy::struct::LazyStruct) that can be traversed to access the container's contents.
     pub fn read(&self) -> ValueParseResult<'top, BinaryEncoding_1_1> {
         if self.is_null() {
-            let raw_value_ref = RawValueRef::Null(self.ion_type());
-            return Ok(raw_value_ref);
+            let ion_type = if self.encoded_value.header.ion_type_code == OpcodeType::TypedNull {
+                let body = self.value_body()?;
+                ION_1_1_TYPED_NULL_TYPES[body[0] as usize]
+            } else {
+                self.ion_type()
```
popematt, line 158: Since we know that the raw value is null but it's not using the `TypedNull` op code in this branch, then I think we could just return `IonType::Null` here. Maybe the compiler/optimizer is smart enough to figure that out, but I think that explicitly returning `IonType::Null` makes the meaning of the code a little more clear to the reader.

src/lazy/binary/raw/v1_1/type_descriptor.rs
popematt: Wow, that was a very thorough answer for my idle curiosity.

> If the values for null_body were the same as the values used in IonType, then rust is smart enough to just return null_body and not lookup, or cmp anything. This would be awesome to make possible, but the binary encoded values would need to match the definition order of IonType (or we'd have to set the values).

Nice!

However, we already expect typed nulls to be a relatively cold path, so I guess it's probably not worth putting a lot of effort into optimizing it.

zslayton (approved):
LGTM pending Matt's comments

src/lazy/binary/raw/v1_1/type_descriptor.rs
```diff
@@ -48,6 +62,7 @@ impl Opcode {
             (0xE, 0x0) => (IonVersionMarker, low_nibble, None),
             (0xE, 0x1..=0x3) => (SymbolAddress, low_nibble, Some(IonType::Symbol)),
             (0xE, 0xA) => (NullNull, low_nibble, Some(IonType::Null)),
+            (0xE, 0xB) => (TypedNull, low_nibble, Some(IonType::Null)),
```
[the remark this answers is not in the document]
nirosys: This is a little more involved. Right now all of the data needed for the `Header` is expected to be available from the first byte. So, any headers created by opcodes without an IonType would cause a panic.

This does make sense, it would just probably be better as a follow-up to allow types to be deferred. I'll create an issue for it.

popematt (approved):

Issue #759, opened 2024-05-03
Add text reader support for iterating through incomplete/invalid containers


sajidanw:
### Description

Add support in the Ion text lazy readers to be able to step into and iterate through incomplete/invalid containers (i.e., the closing token is missing or there's a syntax error within).

Some Ion readers in other languages, as well as the `ion-rust` reader that's present in the pre-1.0.0 releases, can support this through the `next`/`step_in`/`step_out` methods, but they don't report spans/ranges of the source text (requirement for me) and this doesn't map onto the APIs of the lazy readers that are in the current codebase.

### Background

I've been experimenting with `ion-rust` for use within a language server for Ion text. As a language server, the parser I'm looking to develop on top of `ion-rust` needs to be tolerant to parsing errors, since a user may be in the middle of writing code but still expects _some_ amount of language server functionality (completions, find references, etc.) to be available.

For my purposes, that generally means that I need to be able to parse "incomplete" containers — as in, lists, sexps, and structs whose closing tokens aren't present in the document; or which have an Ion value in the container that is unparseable (due to a syntax error or its own incompleteness). This is because I want to be able to semantically analyze all of the parseable data before the error occurred.

> For example, imagine the user is developing the Ion Schema below:
> 
> ```
> type::{
>   name: Person,
>   type: struct,
>   fields: {
>     age: {
>        type: int,
>        valid_values: range::[1, // <-- The user's cursor is here!
>     }
>   }
> }
> ```
>
> At this cursor location, the user requests completions to know which values are available for them to specify (such as `max`). To an Ion parser that only supports complete values, this whole document would fail to parse entirely, even though there's enough information prior to the parsing error to build semantic knowledge and know what values can be suggested.

The lazy reader that's currently in the codebase doesn't support incomplete containers, since it's designed to match the container's full span from the opening `[`/`(`/`{` to the closing `]`/`)`/`}`. This design allows a client of the library to know that a matched value is fully complete and syntactically valid, but prevents a tolerant language server use-case in its current state.

### Suggested Proposals

I think we _don't_ want to change the default behavior of the list/sexp/struct parsers to return the container without identifying it as being incomplete or invalid (such as by returning an `Ok` match). Most clients are likely just in the business of parsing valid Ion, or at most surfacing which position in the input stream caused the violation, so changing this default behavior would be disruptive to the majority of users.

A couple of ideas I've thought about so far:

- **Allow the lazy reader to be configured to match on incomplete/invalid containers.**
[...]

PR #767, opened 2024-05-14
Removes the crate's dependency on `BigInt`, `BigUint`

zslayton:
This PR limits the range of integers that can be read and written to those that can be represented by [the `i128` type](https://doc.rust-lang.org/std/primitive.i128.html#).

This reduces the size of `Int`, `UInt`, `Decimal`, `Timestamp`, `RawValueRef`, and `ValueRef`. It also allows those types to implement `Copy`.

It also allows us to remove several custom methods to compare across mixed `i64`/`BigInt` and `u64`/`BigUint` representations for equality and ordering.

Here's the benchmark output following this change:

image


Cargo.toml
```diff
@@ -18,9 +18,10 @@ exclude = [
 ]
 version = "1.0.0-rc.3"
 edition = "2021"
-# We need at least 1.65 for GATs
-# https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html
-rust-version = "1.65"
+# We need at least 1.65 for GATs[1] and 1.67 for `ilog`[2]
+# [1] https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html
+# [2] https://blog.rust-lang.org/2023/01/26/Rust-1.67.0.html#stabilized-apis
+rust-version = "1.67"
```
zslayton, line 24: 🗺️ Small bump to the minimum Rust version to enable us to use the stabilized `ilog10` method for counting decimal digits in integers.

Cargo.toml
```diff
@@ -62,7 +63,6 @@ chrono = { version = "0.4", default-features = false, features = ["clock", "std"
 delegate = "0.12.0"
 thiserror = "1.0"
 nom = "7.1.1"
-num-bigint = "0.4.3"
```
zslayton, line 65: 🗺️ 👋

benches/encoding_primitives.rs
```diff
@@ -99,7 +99,7 @@ pub fn criterion_benchmark(c: &mut Criterion) {
         b.iter(|| {
             let mut encoded_length: usize = 0;
             for value in &unsigned_values {
-                encoded_length += black_box(FlexUInt::write_u64(&mut output, *value).unwrap());
+                encoded_length += black_box(FlexUInt::write(&mut output, *value).unwrap());
```
zslayton, line 102: 🗺️ The `FlexUInt::write` method can now accept any int size. I ran the `encoding_primitives` benchmark and confirmed that this is the same or faster than the `u64`-specific method in all cases.

src/binary/decimal.rs
```diff
@@ -20,9 +18,7 @@ const DECIMAL_BUFFER_SIZE: usize = 32;
 const DECIMAL_POSITIVE_ZERO: Decimal = Decimal {
     coefficient: Coefficient {
         sign: Sign::Positive,
-        magnitude: UInt {
-            data: UIntData::U64(0),
-        },
+        magnitude: UInt::ZERO,
```
zslayton, line 21: 🗺️ I added `ZERO` and `NEGATIVE_ZERO` `const` values to the `Decimal` and `Coefficient` types for easier reference.

src/binary/uint.rs
```diff
@@ -69,22 +75,28 @@ impl DecodedUInt {
...
 
-/// A buffer for storing a UInt's Big Endian bytes. UInts that can fit in a `u64` will use the
-/// `Stack` storage variant, meaning that no heap allocations are required in the common case.
+/// A buffer for storing a UInt's Big Endian bytes.
 #[derive(Clone, Debug, PartialEq, Eq)]
-pub enum UIntBeBytes {
-    Stack([u8; mem::size_of::<u64>()]),
-    Heap(Vec<u8>),
+pub struct UIntBeBytes {
+    bytes: [u8; mem::size_of::<u128>()],
```
zslayton, line 93: 🗺️ This type now always uses stack storage.

src/types/decimal/mod.rs
```diff
@@ -202,83 +198,9 @@ impl Decimal {
...
-            (magnitude, (BigInt::zero(), 0))
-        } else if self.exponent.is_negative() {
-            let divisor = BigInt::from(10u64).pow((-self.exponent) as u32);
-            let (i, f) = magnitude.div_rem(&divisor);
-            (i, (f, self.exponent))
-        } else {
-            let multiplicand = BigInt::from(10u64).pow(self.exponent as u32);
-            (magnitude * multiplicand, (BigInt::zero(), 0))
-        }
-    }
```
zslayton, line 233: 🗺️ This and related methods existed solely to enable conversion of decimal components to/from `BigInt`.

src/types/decimal/mod.rs
```diff
@@ -615,132 +570,47 @@ mod decimal_tests {
...
+    // Values with very small magnitudes
+    #[case(0.000_000_000_000_001, Decimal::new(1, -15))]
+    #[case(-0.000_000_000_000_001, Decimal::new(-1, -15))]
     fn test_decimal_try_from_f64_ok(#[case] value: f64, #[case] expected: Decimal) {
         let actual: Decimal = value.try_into().unwrap();
-        assert_eq!(actual, expected);
-    }
-
-    #[test]
-    fn test_difference_by_parts_lossy() {
```
zslayton, line 634: 🗺️ Removed tests for the also-removed `difference_by_parts_lossy` & co.

zslayton (commented):
🗺️ PR Tour

src/types/timestamp.rs
```diff
@@ -48,7 +46,7 @@ pub enum Mantissa {
     Digits(u32),
     /// Specifies the fractional seconds precisely as a `Decimal` in the range `>= 0` and `< 1`.
     /// The Decimal will have the correct precision; the complete value can and should be used.
-    /// This representation should only be used for precisions greater than nanoseconds as can
+    /// This representation should only be used for precisions greater than nanoseconds as it can
```
popematt, line 49: > it can require allocations

As is "the implementation may be changed without warning such that it allocates"?

src/types/decimal/mod.rs
```diff
@@ -408,54 +332,86 @@ impl TryFrom<f64> for Decimal {
...
+        // For example, this 40 digit number:
+        //
+        //    1234567890_1234567890_1234567890_1234567890
+        //
+        // would be turned into a decimal whose coefficient was:
+        //
+        //    1234567890_1234567890_1234567890_12345678XX
+        //                               discarded ----^^
+        //
+        // and its exponent would be increased by 2 to retain the scale of discarded digits.
```
jobarr-amzn, line 384: Really appreciating the added commentary :)

src/lazy/value_ref.rs
```diff
@@ -15,6 +15,7 @@ use std::fmt::{Debug, Formatter};
 /// Unlike a [Value], a `ValueRef` avoids heap allocation whenever possible, choosing to point instead
```
popematt, line 15:
```suggestion
/// Unlike a [Value], a `ValueRef` avoids heap allocation, choosing to point instead
```
Is this now true?

src/lazy/encoder/binary/v1_1/value_writer.rs
```diff
@@ -537,10 +532,10 @@ impl<'value, 'top> BinaryValueWriter_1_1<'value, 'top> {
             // 65,792 and higher
             _ => {
                 cold_path! {
-                    xl_symbol_id_value => {
+                    xl_symbol_id_value => {{
```
popematt, line 535: Is the double `{` intentional?

jobarr-amzn (approved):

src/lazy/encoder/binary/v1_1/flex_uint.rs
```diff
@@ -337,74 +271,25 @@ impl FlexUInt {
         // back to the general-purpose.
         cold_path! { encode_opcode_and_length_general_case => {
             output.push(opcode);
-            FlexUInt::encode_u64_general_case(output, length)
+            let _ = FlexUInt::write(output, length).unwrap();
         }}
     }
 
-    pub fn write_u64<W: Write>(output: &mut W, value: u64) -> IonResult<usize> {
+    const MAX_FLEX_UINT_SIZE_IN_BYTES: usize = 16;
```
popematt, line 278: Encoded or decoded size?

src/binary/uint.rs
```diff
@@ -42,13 +41,20 @@ impl DecodedUInt {
     }
 
     /// Interprets all of the bytes in the provided slice as big-endian unsigned integer bytes.
-    pub(crate) fn big_uint_from_slice(uint_bytes: &[u8]) -> BigUint {
-        BigUint::from_bytes_be(uint_bytes)
+    pub(crate) fn big_uint_from_slice(uint_bytes: &[u8]) -> u128 {
```
popematt, line 44: This name might be a little misleading since it no longer returns a `BigUint`.

rmarrowstone (approved):
This is a great simplification. Particularly for the Decimal and Timestamp implementations. Love it. 

I'm not concerned for the potential limiting of decimals and timestamps, but for Ion Integers I don't have a clear picture of what a users options are (could be none) when they have a value stream with an "overlarge" integer. I'm good with an answer like "they can access a vec of the two's complement bytes from this API", and am likely fine with they can't. But I'm not familiar enough with ion-rust to see it from this PR. Thanks.

src/types/integer.rs
```diff
@@ -1,25 +1,27 @@
...
+#[derive(Debug, Copy, Clone)]
 pub struct UInt {
-    pub(crate) data: UIntData,
-}
-
-#[derive(Debug, Clone)]
-pub(crate) enum UIntData {
-    U64(u64),
-    BigUInt(BigUint),
+    pub(crate) data: u128,
```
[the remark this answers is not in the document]
zslayton: > `U512([u8;64])`

I like this idea! Good thinking.

> Keeping the enum ensures that all of our code continues to support the possibility of there being enum variants.

I believe this is already true as the `UInt` struct's `data` field is not public. If/when we need `U512` we can swap out the inner data and retain the same interface.

src/lazy/encoder/binary/v1_1/value_writer.rs
zslayton: It is--the `cold_path!` macro takes a single expression. In order to mark multiple statements as the cold path, you need to pass in a block. The outer `{}` is part of the macro's syntax, the inner `{}` is grouping the statements that follow into a block.

src/types/timestamp.rs
zslayton: RE: "without warning", this is not a public-facing type. (Its host module is private.)

src/lazy/encoder/binary/v1_1/flex_uint.rs
zslayton: Encoded.

src/binary/uint.rs
zslayton: Yeah, fair. It's paired with another method above (`small_uint_from_slice`) that returns a `u64`. They should be merged anyway.

src/binary/int.rs
```diff
@@ -61,8 +61,8 @@ impl DecodedInt {
     /// Encodes a negative zero as an `Int` and writes it to the provided `sink`.
     /// Returns the number of bytes written.
     ///
-    /// This method is similar to [Self::write_i64]. However, because an i64 cannot represent a negative
-    /// zero, a separate method is required.
+    /// This method is similar to [Self::write]. However, because Rust's native integer types cannot
+    /// represent a negative zero, a separate method is required.
     pub fn write_negative_zero<W: Write>(sink: &mut W) -> IonResult<usize> {
```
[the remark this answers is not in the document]
zslayton: This (confusingly named) type is the [binary 1.0 int encoding primitive](https://amazon-ion.github.io/ion-docs/docs/binary.html#uint-and-int-fields) rather than [the user-level `int` data type](https://amazon-ion.github.io/ion-docs/docs/binary.html#0x2-and-0x3-int).

The encoding primitive is used to [encode a decimal's coefficient](https://amazon-ion.github.io/ion-docs/docs/binary.html#0x5-decimal):

image

and the coefficient can be `-0`.

src/lazy/encoder/binary/v1_1/flex_uint.rs
popematt: So, the max supported encoded size is actually 18 bytes, right? 2 bytes worth of continuation bits, and then 16 bytes of data? Or are we limiting it to 16 bytes in order to simplify the implementation?

src/types/integer.rs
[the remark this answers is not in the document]
popematt: Another thought... if we (someday) build the machinery to use XL integers, we could also add in another crate feature to use smaller integers (e.g. i32) for when you know that your application doesn't need any extremely large numbers and you want to have a smaller memory footprint.

src/binary/decimal.rs
```diff
@@ -177,19 +140,4 @@ mod binary_decimal_tests {
...
-        // fixed-size, stack-allocated buffer to encode the body of the decimal.
-        // If it's asked to encode a decimal that won't fit (>32 bytes), it should
-        // return an error.
-        let precise_pi = Decimal::new(
-            BigUint::from_str("31415926535897932384626433832795028841971693993751058209749445923078164062862089986280348253421170679").unwrap(),
-            -99
-        );
-        let result = buf.encode_decimal_value(&precise_pi);
-        assert!(result.is_err());
-    }
```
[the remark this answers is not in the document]
zslayton: This sent me down a useful rabbit hole.

Because I had removed the constructor that accepted a `BigUint` coefficient, it became statically impossible to trigger this case. However, because the `Coefficient` type was a `(UInt, Sign)` tuple, it was possible for it to store magnitudes outside the `i128` range but inside the `u128` range.

In my latest commits I've modified `Coefficient` to an `(Int, Sign)` pair where the `Int`'s sign information always agrees with `Sign` _except_ when the coefficient is a negative zero. This is essentially the same logic as before but with a constructor that's now capped at `Int::MAX`. This had the nice side effect of also eliminating the need for constructors that take an explicit `Sign` parameter because that would only ever be necessary to construct a `-0`. There are now `Coefficient::new(some_int_value)` and `Coefficient::negative_zero()` constructors.

All of that too say, it's now statically impossible to programmatically construct a `Decimal` that's too big.

zslayton:
> This is a great simplification. Particularly for the Decimal and Timestamp implementations. Love it.
> 
> I'm not concerned for the potential limiting of decimals and timestamps, but for Ion Integers I don't have a clear picture of what a users options are (could be none) when they have a value stream with an "overlarge" integer. I'm good with an answer like "they can access a vec of the two's complement bytes from this API", and am likely fine with they can't. But I'm not familiar enough with ion-rust to see it from this PR. Thanks.

For the moment, there's not a (stable) option. An upcoming release will expose the `span` APIs, allowing users to access a value's underlying data.

Longer-term, I've been thinking about augmenting the `IonError` type with variants to expose values that were outside the happy path but potentially usable. For example, returning an `IonError::OversizedInt(oversized_int)` variant where `oversized_int` had methods to access its bytes and determine its byte order. (This is related to #759, where users may wish to access top-level containers that are found to be incomplete or contain invalid child data--CC @sajidanw.) I haven't fully baked this idea though; it seems like it would require adding a lifetime to `IonError`, which could be a big change. Another alternative would be adding a `tolerant_next()` method (name tbd) alongside `next()` which returned a different enum of its own, allowing users to opt into handling such cases.

jobarr-amzn (approved):

Issue #664, opened 2023-10-16
Binary read support for length-prefixed containers


zslayton:
* [x] Lists - #769 
* [x] S-expressions - #769 
* [x] Structs
   * [x] `Symbol` field encoding - #768 
   * [x] `FlexSym` field encoding - #768
* [ ] Follow-up to merge Symbol & FlexSym kinds. - #768

nirosys:
Need to follow-up with a PR to merge the separate struct kinds, and handle mid-struct encoding changes.

PR #768, opened 2024-05-15
Add 1.1 binary reader support for length-prefixed structs

nirosys:
*Issue #, if available:* #664

*Description of changes:*
This PR adds support to the 1.1 raw binary reader for length-prefixed structs.

---

### DIFF
--- a/src/lazy/binary/raw/v1_1/reader.rs
+++ b/src/lazy/binary/raw/v1_1/reader.rs
@@ -169,6 +169,7 @@ impl<'data> LazyRawReader<'data, BinaryEncoding_1_1> for LazyRawBinaryReader_1_1
 #[cfg(test)]
 mod tests {
     use crate::lazy::binary::raw::v1_1::reader::LazyRawBinaryReader_1_1;
+    use crate::raw_symbol_ref::RawSymbolRef;
     use crate::{IonResult, IonType};
     use rstest::*;
 
@@ -310,8 +311,6 @@ mod tests {
 
     #[test]
     fn symbols() -> IonResult<()> {
-        use crate::RawSymbolRef;
-
         #[rustfmt::skip]
         let data: Vec<u8> = vec![
             // IVM
@@ -756,7 +755,6 @@ mod tests {
         Ok(())
     }
 
-
     #[test]
     fn nulls() -> IonResult<()> {
         #[rustfmt::skip]
@@ -785,12 +783,16 @@ mod tests {
 
     fn structs() -> IonResult<()> {
         use crate::lazy::decoder::{LazyRawFieldExpr, LazyRawFieldName};
-        use crate::raw_symbol_ref::RawSymbolRef;
 
         #[rustfmt::skip]
         #[allow(clippy::type_complexity)]
         let tests: &[(&[u8], &[(RawSymbolRef, IonType)])] = &[
             // Symbol Address
+            (
+                // {}
+                &[0xC0],
+                &[],
+            ),
             (
                 // { $10: 1, $11: 2 }
                 &[0xC6, 0x15, 0x51, 0x01, 0x17, 0x51, 0x02],
@@ -813,6 +815,26 @@ mod tests {
                 &[
                     (11usize.into(), IonType::Float),
                 ],
+            ),
+            (
+                // { $10: 1, $11: <NOP> } - with nops at end of struct.
+                &[ 0xC5, 0x15, 0x51, 0x01, 0x17, 0xEC ],
+                &[
+                    (10usize.into(), IonType::Int),
+                ],
+            ),
+            (
+                // { $10: { $11: "foo" }, $11: 2 }
+                &[ 0xC6, 0x15, 0xC4, 0x83, 0x66, 0x6F, 0x6F, 0x17, 0x51, 0x02 ],
+                &[
+                    (10usize.into(), IonType::Struct),
+                    (11usize.into(), IonType::Int),
+                ],
+            ),
+            (
+                // {}
+                &[ 0xFC, 0x01 ],
+                &[],
             ),
             (
                 // { $10: "variable length struct" }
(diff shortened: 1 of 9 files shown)


src/lazy/binary/raw/v1_1/struct.rs
```diff
@@ -117,20 +120,167 @@ impl<'top> LazyRawStruct<'top, BinaryEncoding_1_1> for LazyRawBinaryStruct_1_1<'
     }
 }
 
+enum StructType {
+    Invalid,
+    FlexSym,
+    SymbolAddress,
+}
```
zslayton, line 127: Why not limit it to `FlexSym`/`SymbolAddress` and communicate `Invalid` via a `Result<StructType>`?

src/lazy/binary/raw/v1_1/struct.rs
```diff
@@ -117,20 +120,167 @@ impl<'top> LazyRawStruct<'top, BinaryEncoding_1_1> for LazyRawBinaryStruct_1_1<'
...
+        use crate::IonError;
+        use std::cmp::Ordering;
+
+        if buffer.is_empty() {
+            return Ok(None);
+        }
+
+        let (flex_sym, after) = buffer.read_flex_int()?;
+        let sym_value = flex_sym.value();
+        let (sym, after) = match sym_value.cmp(&0) {
```
zslayton, line 168: Could you move this bit of logic to [the `FlexSym` type](https://github.com/amazon-ion/ion-rust/blob/9fd213c591c8cf95a5cf50b6f0f2f9fc5908a0b4/src/lazy/encoder/binary/v1_1/flex_sym.rs#L10)?

src/lazy/binary/raw/v1_1/struct.rs
```diff
@@ -117,20 +120,167 @@ impl<'top> LazyRawStruct<'top, BinaryEncoding_1_1> for LazyRawBinaryStruct_1_1<'
...
+            Ok((None, after_nops))
+        } else {
+            buffer
+                .read_value(opcode)
+                .map(|v| (Some(v), v.input.consume(v.encoded_value.total_length)))
+        }
+    }
+
+    /// Helper function called from [`Self::next`] to parse the current field and value from the
+    /// struct. On success, returns an both the field pair via [`LazyRawFieldExpr`] as well as the
```
zslayton, line 230:
```suggestion
    /// struct. On success, returns both the field pair via [`LazyRawFieldExpr`] as well as the
```

src/lazy/binary/raw/v1_1/reader.rs
```diff
@@ -457,4 +457,81 @@ mod tests {
...
+                    (11usize.into(), IonType::Float),
+                ],
+            ),
+            (
+                // { $10: <NOP>, $11: 0e0 } - with nops, skip the NOP'd fields.
+                &[ 0xC4, 0x15, 0xEC, 0x17, 0x5A ],
+                &[
+                    (11usize.into(), IonType::Float),
+                ],
+            ),
```
zslayton, line 492: Could you add a test case where the NOP(s) are at the end of the container?

src/lazy/binary/raw/v1_1/reader.rs
```diff
@@ -457,4 +457,81 @@ mod tests {
...
+                    0x62, 0x6C, 0x65, 0x20, 0x6c, 0x65, 0x6E, 0x67, 0x74, 0x68,
+                    0x20, 0x73, 0x74, 0x72, 0x75, 0x63, 0x74
+                ],
+                &[ (10usize.into(), IonType::String) ],
+            ),
+            // FlexSym
+            (
+                // { "foo": 1, $11: 2 }
+                &[ 0xD9, 0xFB, 0x66, 0x6F, 0x6F, 0x51, 0x01, 0x17, 0x91, 0x02],
+                &[ ("foo".into(), IonType::Int), (11usize.into(), IonType::Symbol)],
```
zslayton, line 506: I realize you're implementing an older copy of the spec, but the most recent version merged the distinct struct "kinds" into a single kind that can switch from symbol addresses to flexsym partway through a struct by emitting a `FlexUInt 0` in field name position. This can wait for a follow-on PR since we have lots of updates to make around opcodes.

src/lazy/binary/raw/v1_1/reader.rs
```diff
@@ -457,4 +457,81 @@ mod tests {
...
+        let tests: &[(&[u8], &[(RawSymbolTokenRef, IonType)])] = &[
+            // Symbol Address
+            (
+                // { $10: 1, $11: 2 }
+                &[0xC6, 0x15, 0x51, 0x01, 0x17, 0x51, 0x02],
+                &[
+                    (10usize.into(), IonType::Int),
+                    (11usize.into(), IonType::Int),
+                ]
+            ),
```
zslayton, line 477: Could you add test cases for the empty struct case and a nested struct?

src/lazy/binary/raw/v1_1/struct.rs
nirosys: I didn't use a result, because I'd have to somehow work a Result into the different `iter` functions, and `into_iter` which felt like a much bigger ripple than I wanted to happen. I've reworked that now, so that it only holds `FlexSym`, and `SymbolAddress`. If the constructor has an unknown opcode it will panic, since the struct iterator shouldn't be created without a valid struct.

zslayton (approved):

src/lazy/encoder/binary/v1_1/flex_sym.rs
```diff
@@ -43,4 +53,48 @@ impl FlexSym {
...
         output.extend_from_slice_copy(encoding);
     }
+
+    /// Reads a [`FlexSym`] from the beginning of `input`.
+    ///
+    /// `input` is the byte slice from which to read a [`FlexSym`].
+    /// `offset` is the position of the slice in some larger input stream. It is only used to
+    ///          populate an appropriate error message if reading fails.
+    pub fn read(input: &'top [u8], offset: usize) -> IonResult<FlexSym<'top>> {
+        use crate::{result::IonFailure, IonError};
```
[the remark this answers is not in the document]
nirosys: I do it manually. I try to keep it local to where it's used if it's not something that's going to be used elsewhere.

I would love it if my editor could promote the imports to higher scope if there was an existing local import though, or even if clippy could just point it out.

PR #771, opened 2024-05-22
Fix serde string deserialization for Ion symbols

sajidanw:
I noticed an issue with the serde module for Ion symbol to Rust string deserialization, which the module documentation says is supported. Because the current deserializer uses `expect_string()`, it panics on a symbol. I've updated this to use the existing `expect_text()` method to support both strings and symbols.

A couple other notes about this proposed change:

- I've deleted the comment about the `StrRef` possibly having an owned string. It looks like that was written before a recent change that replaced the `StrRef`'s `Cow` with `&str` (https://github.com/amazon-ion/ion-rust/commit/892ef2d084ac359f01cd10bc2710dee2a26725d6). I think this means that a new owned string will always have to be created.
- I updated the module documentation to reflect that unit structs are serialized as symbols (which currently is the only way to serialize a symbol, I believe?).

### DIFF
--- a/src/serde/mod.rs
+++ b/src/serde/mod.rs
@@ -39,7 +39,8 @@
 //!| char, string, unit_variant                                   | string                                      |
 //!| byte-array                                                   | blob                                        |
 //!| option                                                       | None - null, Some - based on other mappings |
-//!| unit, unit_struct                                            | null                                        |
+//!| unit                                                         | null                                        |
+//!| unit_struct                                                  | symbol                                      |
 //!| seq, tuple, tuple_struct                                     | list                                        |
 //!| newtype_struct, map, struct                                  | struct                                      |
 //!| newtype_variant                                              | variant value with annotation               |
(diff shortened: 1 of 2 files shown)


src/serde/mod.rs
```diff
@@ -340,4 +341,19 @@ mod tests {
             Element::read_first(to_string(&expected).unwrap())
         );
     }
+
+    #[test]
+    fn test_symbol() {
+        let i = r#"inches"#;
+        let expected = String::from("inches");
+        assert_eq!(expected, from_ion::<String, _>(i).unwrap());
```
sajidanw, line 349: Forgot to add, not sure why but with `assert_eq!`, the compiler can't seem to infer `String` as the type: "cannot infer type of the type parameter `T` declared on the function `from_ion`".

zslayton (approved):
Nice! Thanks for the PR!

PR #772, opened 2024-05-22
Feature gating and renames

zslayton:
Nearly all of this PR is renames and feature gating. More substantive
changes are called out in the PR tour.

The `Element` methods `to_binary` and `to_string` (which only
supported serializing to Ion 1.0) have been replaced with an
`encode_as` method that allows you to provide a `WriteConfig`.

The `Element` method `write_to` has been replaced by `encode_to`,
which also takes a `WriteConfig`.

API items that belong to a Cargo feature are now gated appropriately.
Doc tests that involve such items have also been modified to have
their content skipped when the feature is disabled, allowing the
tests to pass.

Several types have been renamed now that the old reader and writer
have been removed, including:
* `ApplicationReader` -> `IonReader`
* `LazyReader` -> `Reader`
* `LazySystemReader` -> `SystemReader`
* `LazyTextReader` -> `TextReader`
* `LazyBinaryReader` -> `BinaryReader`
* `ApplicationWriter` -> `IonWriter`
* `RawSymbolTokenRef` -> `RawSymbolRef`

I have also set up `v1_0` and `v1_1` modules at the top level where
version specific entities are re-exported with friendlier names.

* `BinaryEncoding_1_0` -> `v1_0::Binary`
* `BinaryEncoding_1_1` -> `v1_1::Binary`
* `TextEncoding_1_0` -> `v1_0::Text`
* `TextEncoding_1_1` -> `v1_1::Text`

This means that encoding an `Element` with the stabilizing API would look like:

```rust
use ion_rs::{Element, v1_0::Binary};

let element = Element::string("hello");
let ion_bytes = element.encode_as(Binary).unwrap();
```

I have also updated the benchmarks, examples, and unit tests to use the new types.
On my machine, I have rewritten `ion beta inspect` to use these APIs, so I know they're
sufficient at least as a starting point.


zslayton (commented):
🗺️ PR Tour

jobarr-amzn (approved):

Issue #776, opened 2024-05-28
Wire up `Catalog` in `SystemReader`


zslayton:
The lazy `SystemReader` (now the only implementation in the crate) does not have shared symbol tables implemented. The [old implementation](https://github.com/amazon-ion/ion-rust/blob/v1.0.0-rc.3/src/system_reader.rs) supported them. This must be addressed before we can release 1.0.

zslayton:
Fixed in #777.

PR #777, opened 2024-05-29
Restores shared symbol table import functionality

zslayton:
Issue: #776.

> [!TIP]
> A large chunk of this PR's diff comes from mechanical renaming/refactoring that occurred in cf6aacc422970758960eb4b7699d012d6c26ee53. I suggest skimming that one and then reviewing the changes in efd8e22504a401756224f8174c7bd746384e248a more closely.

The recently removed `IonReader` API's `SystemReader` had the ability to import shared symbol tables from a given `Catalog` implementation. This PR adds the same functionality to the lazy reader API's `SystemReader`.

It also normalizes the constructors used for `Reader` and `Writer`. There are no longer exported type aliases for `BinaryReader`, `TextReader`, etc. They are all `Reader<D: Decoder, I: IonInput>`. As part of this change, `Reader::new` now accepts an `Into<ReadConfig<D>>` that allows the user to specify the decoder, `Catalog`, and other settings to use.

For example:
```rust
let mut reader = Reader::new(AnyEncoding, ion_data)?;
```
or
```rust
let mut reader = Reader::new(v1_1::Text.with_catalog(my_catalog), ion_data)?;
```

This arrangement mirrors `Writer`'s constructors:
```rust
let mut writer = Writer::new(v1_1::Text.with_format(TextFormat::Pretty), output)?;
```

The patch reworks the `PendingLst` type to hold `Symbol` instances instead of `Option<String>`. Storing `Option<String>` meant that we would heap-allocate a `String` for every symbol, and then later heap-allocate an `Arc<str>` when we added it to the symbol table. In the new arrangement, the `PendingLst` creates `Symbol` instances backed by `Arc<str>`s from the start, halving the allocations required.

Finally, this PR adds methods to the `SystemReader` that will allow tooling to "see" the symbols that are going to be added to the symbol table while the reader is parked on the LST. Without this, the reader would need to advance to the next top level value in the stream to see the updated symbol table (or manually interpret the `$ion_symbol_table::{...}` to learn what would be added). This is helpful for programs like `ion inspect`.

### DIFF
--- a/src/lazy/system_reader.rs
+++ b/src/lazy/system_reader.rs
@@ -1,15 +1,18 @@
 #![allow(non_camel_case_types)]
 
-use crate::lazy::any_encoding::AnyEncoding;
 use crate::lazy::decoder::Decoder;
-use crate::lazy::encoding::{BinaryEncoding_1_0, TextEncoding_1_0, TextEncoding_1_1};
 use crate::lazy::expanded::{ExpandedValueRef, ExpandingReader, LazyExpandedValue};
 use crate::lazy::streaming_raw_reader::{IonInput, StreamingRawReader};
 use crate::lazy::system_stream_item::SystemStreamItem;
 use crate::lazy::value::LazyValue;
 use crate::read_config::ReadConfig;
 use crate::result::IonFailure;
-use crate::{IonResult, IonType, RawSymbolRef, SymbolTable};
+use crate::{
+    Catalog, Int, IonError, IonResult, IonType, LazyExpandedField, RawSymbolRef, Symbol,
+    SymbolTable,
+};
+use std::ops::Deref;
+use std::sync::Arc;
 
 // Symbol IDs used for processing symbol table structs
 const ION_SYMBOL_TABLE: RawSymbolRef = RawSymbolRef::SymbolId(3);
@@ -74,19 +77,14 @@ pub struct SystemReader<Encoding: Decoder, Input: IonInput> {
     pub(crate) expanding_reader: ExpandingReader<Encoding, Input>,
 }
 
-pub type SystemBinaryReader_1_0<Input> = SystemReader<BinaryEncoding_1_0, Input>;
-pub type SystemTextReader_1_0<Input> = SystemReader<TextEncoding_1_0, Input>;
-pub type SystemTextReader_1_1<Input> = SystemReader<TextEncoding_1_1, Input>;
-
-pub type SystemAnyReader<Input> = SystemReader<AnyEncoding, Input>;
-
 // If the reader encounters a symbol table in the stream, it will store all of the symbols that
 // the table defines in this structure so that they may be applied when the reader next advances.
 #[derive(Default)]
-pub(crate) struct PendingLst {
+pub struct PendingLst {
     pub(crate) has_changes: bool,
     pub(crate) is_lst_append: bool,
-    pub(crate) symbols: Vec<Option<String>>,
+    pub(crate) symbols: Vec<Symbol>,
+    pub(crate) imported_symbols: Vec<Symbol>,
 }
 
 impl PendingLst {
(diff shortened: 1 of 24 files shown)


src/lazy/system_reader.rs
```diff
@@ -228,11 +275,61 @@ impl<Encoding: Decoder, Input: IonInput> SystemReader<Encoding, Input> {
...
+                        ),
+                        // If there's no version, a non-int version, or a version <= 0, we treat it
+                        // as version 1.
+                        _ => Ok(1),
+                    }?;
+
+                    let shared_table = match catalog.get_table_with_version(name.as_ref(), version) {
+                        Some(table) => table,
+                        None => return IonResult::decoding_error(
+                            format!("symbol table import failed, could not find table with name='{name}' and version={version}")
```
zslayton, line 303: 🗺️ `ion-rust` does not implement fallback to older versions of an imported symbol table, choosing instead to treat the unavailability of a specific imported SST version as a hard error.

Ion implementations that offer this capability (e.g. `ion-java`) are forced to track three possible resolved symbol kinds:
1. A symbol with known text
2. A symbol that is known to have no text (`$0`)
3. A symbol that might have text but we don't have the context necessary to look it up

State 3 is defined to allow a reader to "soldier on" in the absence of the necessary symbol definitions, reading the parts of the file it _does_ have. Early versions of `ion-rust` preserved state 3, but it introduced considerable API complexity. We tabled support for it until a concrete use case surfaced.

src/lazy/system_reader.rs
```diff
@@ -333,4 +430,262 @@ mod tests {
         }
         Ok(())
     }
+
```
zslayton, line 433: 🗺️ Unit tests for shared symbol table imports

src/lazy/system_reader.rs
```diff
@@ -333,4 +430,262 @@ mod tests {
...
+        let _symtab = reader.next_item()?.expect_symbol_table()?;
+        // There are 10 symbols in the symbol table, $0 to $9 inclusive. $10 is not defined yet
+        // because we're still parked on the symtab.
+        assert_eq!(reader.symbol_table().len(), 10);
+        // The reader has analyzed the symtab struct and identified what symbols will be added when
+        // it advances beyond it.
+        assert_eq!(
+            reader.pending_symtab_changes().local_symbols()[0].text(),
+            Some("potato salad")
+        );
```
zslayton, line 521: 🗺️ This is a demonstration of the new tooling access to pending symbol table changes.

src/symbol_table.rs
```diff
@@ -32,9 +32,14 @@ impl SymbolTable {
...
-            self.intern_or_add_placeholder(text);
-        }
+        self.add_placeholder(); // $0
+        v1_0::SYSTEM_SYMBOLS[1..]
+            .iter()
+            .copied()
+            .map(Option::unwrap)
+            .for_each(|text| {
+                let _sid = self.add_symbol_for_text(text);
+            });
```
zslayton, line 42: 🗺️ The changes in this file are to offer method versions that accept a `Symbol` instead of a `&str`. This allows the reader to heap-allocate once per symbol definition instead of twice.

zslayton (commented):
🗺️ PR tour 🧭

popematt (approved):
LGTM.

Thanks for keeping the "missing an imported table" case as a hard error. Hopefully, we never need to change that.

src/lazy/system_reader.rs
```diff
@@ -215,9 +250,11 @@ impl<Encoding: Decoder, Input: IonInput> SystemReader<Encoding, Input> {
...
-                    pending_lst.symbols.push(Some(str_ref.text().to_owned()))
+                    pending_lst
+                        .symbols
+                        .push(Symbol::shared(Arc::from(str_ref.deref())))
                 } else {
-                    pending_lst.symbols.push(None)
+                    pending_lst.symbols.push(Symbol::unknown_text())
                 }
             }
         }
```
jobarr-amzn, line 260: Huh, TIL that non-String types in the `symbols` list should be treated as `null` values. I'd never thought about the case before. From https://amazon-ion.github.io/ion-docs/docs/symbols.html :
> The `symbols` field should be a list of strings. If the field is missing or has any other type, it is treated as if it were an empty list.
> 
> Null elements declare undefined symbol IDs (“gaps”) within the sequence; implementations must handle requests for such symbols the same as if the requested ID beyond the end of the list. Any element of the list that is not a string must be interpreted as if it were null.

Is it worth referring back to the spec where we can, or is that all implicit?

jobarr-amzn (approved):

PR #780, opened 2024-05-31
Additional tooling access methods, trait impls, and a bug fix

zslayton:
This PR includes a variety of small fixes and improvements that I found myself needing as I went to upgrade the `ion-cli` to `HEAD`. I'd like to include these in the v1.0-rc4 release. I have completed the `ion-cli` upgrade in my local workspace using this branch, so I'm reasonably confident that these are the last items I need to get in.

* Renames the `RawReaderType` enum to `IonEncoding` because it is generally useful in other contexts too.
* Adds the ability for system and raw readers using `AnyEncoding` to report the encoding they're currently using.
* Adds `value_span` and `annotations_span` methods to `LazyValue`. It already had a `span` method whose bytes included both the annotations and the value.
* The binary 1.0 and 1.1 writers will now write `f64`s as `f32`s to save space when it can be done losslessly.
* Adds `WriteAsIon` implementations for `LazyList`, `LazySExp`, and `LazyStruct`.
* Adds `IntoIterator` impls for `LazyList`, `LazySExp` and `LazyStruct`. Previously, they only existed for borrowed (`&`) references to those types.
* Fixes a bug in the `StreamingRawReader` that could cause a value's annotations span to be overwritten if reading that value consumed all of the data remaining in the buffer.
* Renames the feature-gated `LazyValue::lower` method (which returns a `LazyExpandedValue`) to `LazyValue::expanded()` so I could add a `raw()` method alongside it that returns the underlying `LazyRawValue` when applicable.

### DIFF
--- a/src/lazy/any_encoding.rs
+++ b/src/lazy/any_encoding.rs
@@ -3,9 +3,6 @@
 use std::fmt::Debug;
 use std::ops::Range;
 
-use bumpalo::Bump as BumpAllocator;
-
-use crate::lazy::any_encoding::RawReaderKind::{Binary_1_0, Text_1_0};
 use crate::lazy::binary::raw::annotations_iterator::RawBinaryAnnotationsIterator as RawBinaryAnnotationsIterator_1_0;
 use crate::lazy::binary::raw::r#struct::{
     LazyRawBinaryFieldName_1_0, LazyRawBinaryStruct_1_0, RawBinaryStructIterator_1_0,
@@ -48,8 +45,8 @@ use crate::lazy::text::raw::sequence::{
     LazyRawTextList_1_0, LazyRawTextSExp_1_0, RawTextListIterator_1_0, RawTextSExpIterator_1_0,
 };
 use crate::lazy::text::raw::v1_1::reader::{
-    LazyRawTextFieldName_1_1, LazyRawTextList_1_1, LazyRawTextSExp_1_1, LazyRawTextStruct_1_1,
-    MacroIdRef, RawTextEExpression_1_1, RawTextSequenceCacheIterator_1_1,
+    LazyRawTextFieldName_1_1, LazyRawTextList_1_1, LazyRawTextReader_1_1, LazyRawTextSExp_1_1,
+    LazyRawTextStruct_1_1, MacroIdRef, RawTextEExpression_1_1, RawTextSequenceCacheIterator_1_1,
     RawTextStructCacheIterator_1_1,
 };
 use crate::lazy::text::value::{
@@ -57,6 +54,7 @@ use crate::lazy::text::value::{
     LazyRawTextVersionMarker_1_1, RawTextAnnotationsIterator,
 };
 use crate::{IonResult, IonType, RawSymbolRef};
+use bumpalo::Bump as BumpAllocator;
 
 /// An implementation of the `LazyDecoder` trait that can read any encoding of Ion.
 #[derive(Debug, Clone, Copy)]
@@ -67,7 +65,7 @@ pub struct AnyEncoding;
 // underlying type.
 impl Decoder for AnyEncoding {
     type Reader<'data> = LazyRawAnyReader<'data>;
-    type ReaderSavedState = RawReaderType;
+    type ReaderSavedState = IonEncoding;
     type Value<'top> = LazyRawAnyValue<'top>;
     type SExp<'top> = LazyRawAnySExp<'top>;
     type List<'top> = LazyRawAnyList<'top>;
@@ -262,13 +260,13 @@ pub struct LazyRawAnyReader<'data> {
 }
 
 impl<'data> LazyRawAnyReader<'data> {
-    fn detect_encoding(data: &[u8]) -> RawReaderType {
+    fn detect_encoding(data: &[u8]) -> IonEncoding {
         const BINARY_1_0_IVM: &[u8] = &[0xEA, 0x01, 0x00, 0xE0];
 
         match *data {
-            [0xE0, 0x01, 0x00, 0xEA, ..] => RawReaderType::Binary_1_0,
-            [0xE0, 0x01, 0x01, 0xEA, ..] => RawReaderType::Binary_1_1,
-            _ => RawReaderType::Text_1_0,
+            [0xE0, 0x01, 0x00, 0xEA, ..] => IonEncoding::Binary_1_0,
+            [0xE0, 0x01, 0x01, 0xEA, ..] => IonEncoding::Binary_1_1,
+            _ => IonEncoding::Text_1_0,
         }
     }
 }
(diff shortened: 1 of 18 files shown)


src/lazy/any_encoding.rs
```diff
@@ -67,7 +65,7 @@ pub struct AnyEncoding;
 // underlying type.
 impl Decoder for AnyEncoding {
     type Reader<'data> = LazyRawAnyReader<'data>;
-    type ReaderSavedState = RawReaderType;
+    type ReaderSavedState = IonEncoding;
```
zslayton, line 68: 🗺️ This type (an enum of the four Ion encodings) is now returned by the new (feature gated) `SystemReader<AnyEncoding, _>::detected_encoding()` method, allowing users to determine what encoding is currently being read.

src/lazy/encoder/binary/v1_0/value_writer.rs
```diff
@@ -180,6 +180,13 @@ impl<'value, 'top> BinaryValueWriter_1_0<'value, 'top> {
             return Ok(());
         }
 
+        // See if this value can be losslessly encoded in 4 bytes instead of 8
+        let float32 = value as f32;
+        if float32 as f64 == value {
```
popematt, line 185: What about NaN? It can be written as f32, but won't be considered equal here.

src/lazy/encoder/binary/v1_1/value_writer.rs
```diff
@@ -171,9 +169,14 @@ impl<'value, 'top> BinaryValueWriter_1_1<'value, 'top> {
             self.push_byte(0x5A);
             return Ok(());
         }
+        // See if this value can be losslessly encoded in 4 bytes instead of 8
+        let float32 = value as f32;
+        if float32 as f64 == value {
```
popematt, line 174: Same question about NaN as in the other writer.

src/lazy/any_encoding.rs
```diff
@@ -724,6 +751,24 @@ impl<'top> LazyRawValue<'top, AnyEncoding> for LazyRawAnyValue<'top> {
...
+    }
+
+    fn value_span(&self) -> Span<'top> {
+        match &self.encoding {
+            LazyRawValueKind::Text_1_0(v) => v.value_span(),
+            LazyRawValueKind::Binary_1_0(v) => v.value_span(),
+            LazyRawValueKind::Text_1_1(v) => v.value_span(),
+            LazyRawValueKind::Binary_1_1(v) => v.value_span(),
+        }
+    }
```
zslayton, line 771: 🗺️ The `span()` method returns a byte slice containing the complete value--both annotations _and_ data.

These methods allow you to access the annotations and data separately.

src/lazy/encoder/binary/v1_1/value_writer.rs
```diff
@@ -971,10 +974,20 @@ mod tests {
...
-            expected_encoding.extend_from_slice(&value.to_le_bytes()[..]);
+            let mut expected_encoding = vec![];
+            let float32 = *value as f32;
+            if float32 as f64 == *value {
+                expected_encoding.push(0x5C);
+                expected_encoding.extend_from_slice(&float32.to_le_bytes()[..]);
+            } else {
+                expected_encoding.push(0x5D);
+                expected_encoding.extend_from_slice(&value.to_le_bytes()[..]);
+            }
```
zslayton, line 990: 🗺️ We have a lot of unit tests where we re-invent the wheel for the sake of bootstrapping the encoder/decoder. We should look at migrating some of the more painful ones (like this one) to a roundtrip test using our recently established readers and writers.

src/lazy/encoder/write_as_ion.rs
```diff
@@ -296,29 +296,41 @@ impl<'a, D: Decoder> WriteAsIon for ValueRef<'a, D> {
...
-                let mut struct_ = value_writer.struct_writer()?;
-                for field_result in s {
-                    let field = field_result?;
-                    struct_.write(field.name()?, field.value().read()?)?;
-                }
-                struct_.close()
-            }
+            List(l) => value_writer.write(l),
+            SExp(s) => value_writer.write(s),
+            Struct(s) => value_writer.write(s),
```
zslayton, line 301: 🗺️ Giving these types their own `WriteAsIon` implementations streamlined this code path nicely, but my real aim was to make passing them to a writer in other contexts easier.

src/lazy/streaming_raw_reader.rs
```diff
@@ -55,11 +57,34 @@ impl<Encoding: Decoder, Input: IonInput> StreamingRawReader<Encoding, Input> {
         }
     }
 
+    /// Gets a reference to the data source and tries to fill its buffer.
+    // This is a separate method to better encapsulate its use of `unsafe`
+    #[inline]
+    fn pull_more_data_from_source(&mut self) -> IonResult<usize> {
+        let input = unsafe { &mut *self.input.get() };
```
popematt, line 64: Can you please document this use of unsafe? Specifically, why is the unsafe block required and how are we sure that this is, in fact, safe?

src/lazy/encoding.rs
```diff
@@ -60,6 +60,8 @@ pub trait Encoding: Encoder + Decoder {
     ) -> IonResult<W> {
         Self::default_write_config().encode_all_to(output, values)
     }
+
+    fn encoding(&self) -> IonEncoding;
```
zslayton, line 64: 🗺️ Types implementing the `Encoding` trait now must report which variant of the `IonEncoding` enum they correspond to.

src/lazy/streaming_raw_reader.rs
```diff
@@ -55,11 +57,34 @@ impl<Encoding: Decoder, Input: IonInput> StreamingRawReader<Encoding, Input> {
...
+        input.fill_buffer()
+    }
+
+    /// Returns true if the input buffer is empty.
+    #[inline]
+    fn buffer_is_empty(&self) -> bool {
+        let input = unsafe { &*self.input.get() };
```
popematt, line 71: Can you please add more documentation for this use of unsafe?

src/lazy/sequence.rs
```diff
@@ -282,6 +297,15 @@ impl<'a, 'top, 'data: 'top, D: Decoder> IntoIterator for &'a LazySExp<'top, D> {
     }
 }
 
+impl<'top, 'data: 'top, D: Decoder> IntoIterator for LazySExp<'top, D> {
```
zslayton, line 300: 🗺️ Now that `LazyList`, `LazySExp`, and `LazyStruct` are `Copy`, it makes sense to allow them to be iterated over without having to borrow them first.

src/lazy/streaming_raw_reader.rs
```diff
@@ -83,11 +108,13 @@ impl<Encoding: Decoder, Input: IonInput> StreamingRawReader<Encoding, Input> {
 
             let bytes_read = end_position - starting_position;
             let input = unsafe { &mut *self.input.get() };
-            // If we've exhausted the buffer...
-            if bytes_read >= available_bytes.len() || matches!(result, Err(IonError::Incomplete(_)))
-            {
-                // ...try to pull more data from the data source. If we get more data, try again.
```
zslayton, line 89: 🗺️ This was a subtle bug enabled by my use of `unsafe`. If `result` was `Ok(_)` but _also_ consumed all of the data in `available_bytes`, then this code path would have `input` shift its bytes to the beginning and pull in more data. Because `result` contained an `Ok(value)` referring to a slice of that buffer, its data would get clobbered.

src/lazy/streaming_raw_reader.rs
```diff
@@ -319,6 +352,14 @@ impl<'a> IonInput for StdinLock<'a> {
     }
 }
 
+impl<T: Read, U: Read> IonInput for io::Chain<T, U> {
```
zslayton, line 355: 🗺️ The `ion-cli` uses the `io::Chain` type to link together the first few bytes of an input stream (used to detect gzip, zstd, etc) with the rest of the stream.

zslayton (commented):
🗺️ PR tour

popematt (commented):
I noticed there were some functions where you removed the feature gate, and others where you kept the feature gate. How did you decide for which functions to remove vs retain the feature gate?

zslayton:
> I noticed there were some functions where you removed the feature gate, and others where you kept the feature gate. How did you decide for which functions to remove vs retain the feature gate?

Some methods are on a fuzzy line between being `reader-writer` and `tooling-apis`. For example, `LazyExpandedValue::context` returns the `EncodingContextRef` used to decode that value. It's definitely part of the `reader-writer` API, but whether it's specifically `tooling-access` is debatable.

I decided I was ok with removing the feature gate from methods like that because they're still behind _some_ feature gate, so either way there's no guarantee of stability being made.

src/lazy/encoder/binary/v1_1/value_writer.rs
```diff
@@ -155,29 +156,34 @@ impl<'value, 'top> BinaryValueWriter_1_1<'value, 'top> {
     // TODO: write_f16(...)
 
     pub fn write_f32(mut self, value: f32) -> IonResult<()> {
-        if value == 0f32 && !value.is_sign_negative() {
-            self.push_byte(0x5A);
-            return Ok(());
+        match value.smallest_repr() {
```
popematt, line 159: When we add support for f16, we could also DRY it up like this:
```rust
fn write_float_repr(mut self, value: FloatRepr) -> IonResult<()> {
    match value {
        // ...
    }
    Ok(())
}

pub fn write_f16(mut self, value: f16) -> IonResult<()> {
    self.write_float_repr(value.smallest_repr())
}

pub fn write_f32(mut self, value: f32) -> IonResult<()> {
    self.write_float_repr(value.smallest_repr())
}

pub fn write_f64(mut self, value: f64) -> IonResult<()> {
    self.write_float_repr(value.smallest_repr())
}
```

popematt (approved):

src/lazy/encoder/binary/v1_1/value_writer.rs
zslayton: I was thinking the same thing 👍

PR #781, opened 2024-06-02
Prevent writing empty annotations sequences

zslayton:
Prior to this change, when a `ValueWriter` explicitly called `with_annotations()` and provided an empty array, the writer would write an empty annotations sequence which is not legal. This change detects that case and skips writing any sequence at all. It also adds a `has_annotations()` convenience method to the `LazyRawValue` trait.

### DIFF
--- a/src/lazy/any_encoding.rs
+++ b/src/lazy/any_encoding.rs
@@ -724,6 +724,16 @@ impl<'top> LazyRawValue<'top, AnyEncoding> for LazyRawAnyValue<'top> {
         }
     }
 
+    fn has_annotations(&self) -> bool {
+        use LazyRawValueKind::*;
+        match &self.encoding {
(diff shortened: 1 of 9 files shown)


popematt (commented):
Can you add test cases to verify that the other encodings do not have the same issue? (Or do they already exist?)

zslayton:
> Can you add test cases to verify that the other encodings do not have the same issue? (Or do they already exist?)

The Binary 1.1 writer already had a test for this:
https://github.com/amazon-ion/ion-rust/blob/80088b4dbf9e548d64a273bbe5a7f8350f997a91/src/lazy/encoder/binary/v1_1/value_writer.rs#L2632-L2642

I'll add one for the text writers though.

popematt (approved):

PR #784, opened 2024-06-05
Enables the StreamingRawReader to detect incomplete text

zslayton:
In binary encodings, stream items contain enough data for the reader to tell whether they are complete. In text encodings, it's possible for the buffer to end with data that looks like a complete item but is not. The only way to be certain is to try to read again from the input source to confirm there's no more data. Consider the following examples in which Ion is being pulled from a `File` into a `Vec<u8>`:
```
foo /* comment */   ::bar::baz::1000
└────────┬───────┘ └────────┬───────┘
  buffer contents   remaining in File

              $ion _1_0
└────────┬───────┘ └────────┬───────┘
  buffer contents   remaining in File

               75 1.20
────────┬───────┘ └────────┬───────┘
  buffer contents   remaining in File
```
To avoid misinterpreting the data, the `StreamingRawReader` now performs a final check for text readers who have emptied their buffer: it does not consider the item complete unless the input source is exhausted.

Other changes:
* Fixes an indentation issue that affected annotated values emitted by the text writer.
* Makes `IonStream` a public part of the `experimental-reader-writer` API.
* Fixes a bug in the tooling access API `EncodedBinaryValueData_1_0::body_span` that caused it to fail if the value in question was annotated.

The three commits are each self-contained; reviewing them separately may be helpful.

### DIFF
--- a/src/lazy/any_encoding.rs
+++ b/src/lazy/any_encoding.rs
@@ -384,13 +384,7 @@ impl<'data> LazyRawReader<'data, AnyEncoding> for LazyRawAnyReader<'data> {
 
     #[inline]
     fn save_state(&self) -> <AnyEncoding as Decoder>::ReaderSavedState {
-        use RawReaderKind::*;
-        match &self.encoding {
-            Text_1_0(_) => IonEncoding::Text_1_0,
-            Binary_1_0(_) => IonEncoding::Binary_1_0,
-            Text_1_1(_) => IonEncoding::Text_1_1,
-            Binary_1_1(_) => IonEncoding::Binary_1_1,
-        }
+        self.encoding()
     }
 
     fn position(&self) -> usize {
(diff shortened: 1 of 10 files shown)


src/lazy/any_encoding.rs
```diff
@@ -402,6 +396,16 @@ impl<'data> LazyRawReader<'data, AnyEncoding> for LazyRawAnyReader<'data> {
             Binary_1_1(r) => r.position(),
         }
     }
+
+    fn encoding(&self) -> IonEncoding {
```
zslayton, line 400: 🗺️ All raw readers now provide a `fn encoding(&self) -> IonEncoding`, which allows generic code to tell which Ion encoding it's working with.

src/lazy/binary/raw/value.rs
```diff
@@ -258,7 +258,7 @@ impl<'a, 'top> EncodedBinaryValueData_1_0<'a, 'top> {
     /// value's body (that is: the content of the value that follows its opcode and length).
     pub fn body_range(&self) -> Range<usize> {
         let encoded = &self.value.encoded_value;
-        let body_offset = encoded.header_length();
+        let body_offset = encoded.annotations_header_length as usize + encoded.header_length();
```
zslayton, line 261: 🗺️ This fixes a bug that the `ion-cli` [has to work around](https://github.com/amazon-ion/ion-cli/blob/059a95ce8422121f591a6062db8df6c929f00f53/src/bin/ion/commands/inspect.rs#L928-L936).

zslayton (commented):
🗺️ PR tour 🧭

popematt (approved):

nirosys (approved):

PR #786, opened 2024-06-06
Shift opcodes

zslayton:
Shifts the opcodes used in the binary 1.1 reader/writer to align with [the values in the spec](https://github.com/amazon-ion/ion-docs/blob/ion-11-specification/src/binary-encoding.adoc#opcodes).


nirosys (approved):
thank you!

PR #788, opened 2024-06-07
Implements flexsym annotation reading, adds binary 1.1 roundtrip unit tests

zslayton:
The binary 1.1 reader can now read `FlexSym`-encoded annotations sequences. I have also added some application-level roundtripping unit tests to identify places where the reader and writer may disagree. Because the tests are application-level, the writer generates an LST for several of the tests; quite a bit of the API is being exercised.


src/lazy/binary/raw/v1_1/immutable_buffer.rs
```diff
@@ -173,12 +173,6 @@ impl<'a> ImmutableBuffer<'a> {
         Ok((flex_sym, remaining))
     }
 
-    /// Attempts to decode an annotations wrapper at the beginning of the buffer and returning
-    /// its subfields in an [`AnnotationsWrapper`].
-    pub fn read_annotations_wrapper(&self, _opcode: Opcode) -> ParseResult<'a, AnnotationsWrapper> {
-        todo!();
-    }
-
```
zslayton, line 181: 🗺️ I removed this stub (that I added a while ago) because v1.1 annotations don't involve a 'wrapper' like v1.0 annotations.

src/lazy/binary/raw/v1_1/immutable_buffer.rs
```diff
@@ -340,19 +335,114 @@ impl<'a> ImmutableBuffer<'a> {
...
+        let (annotations_seq, input_after_annotations) = self.read_annotations_sequence(opcode)?;
+        let opcode = input_after_annotations.peek_opcode()?;
+        let mut value = input_after_annotations.read_value_without_annotations(opcode)?;
+        value.encoded_value.annotations_header_length = annotations_seq.header_length;
+        value.encoded_value.annotations_sequence_length = annotations_seq.sequence_length;
+        value.encoded_value.annotations_encoding = annotations_seq.encoding;
+        value.encoded_value.total_length +=
+            annotations_seq.header_length as usize + annotations_seq.sequence_length as usize;
+        // Rewind the input to include the annotations sequence
+        value.input = self;
```
zslayton, line 348: 🗺️ After reading the value, this method goes back and overwrites all of the annotations-related or -affected fields.

src/lazy/binary/raw/v1_1/reader.rs
```diff
@@ -536,17 +536,17 @@ mod tests {
...
+    #[case("2023-10-15T05:04:03Z",                &[0x84, 0x35, 0x7D, 0x85, 0x30, 0x00])]
+    #[case("2023-10-15T05:04:03.123-00:00",       &[0x85, 0x35, 0x7D, 0x85, 0x38, 0xEC, 0x01])]
+    #[case("2023-10-15T05:04:03.000123-00:00",    &[0x86, 0x35, 0x7D, 0x85, 0x38, 0xEC, 0x01, 0x00])]
+    #[case("2023-10-15T05:04:03.000000123-00:00", &[0x87, 0x35, 0x7D, 0x85, 0x38, 0xEC, 0x01, 0x00, 0x00])]
+    #[case("2023-10-15T05:04+01:00",              &[0x88, 0x35, 0x7D, 0x85, 0xE0, 0x01])]
+    #[case("2023-10-15T05:04-01:00",              &[0x88, 0x35, 0x7D, 0x85, 0xA0, 0x01])]
+    #[case("2023-10-15T05:04:03+01:00",           &[0x89, 0x35, 0x7D, 0x85, 0xE0, 0x0D])]
+    #[case("2023-10-15T05:04:03.123+01:00",       &[0x8A, 0x35, 0x7D, 0x85, 0xE0, 0x0D, 0x7B, 0x00])]
+    #[case("2023-10-15T05:04:03.000123+01:00",    &[0x8B, 0x35, 0x7D, 0x85, 0xE0, 0x0D, 0x7B, 0x00, 0x00])]
+    #[case("2023-10-15T05:04:03.000000123+01:00", &[0x8C, 0x35, 0x7D, 0x85, 0xE0, 0x0D, 0x7B, 0x00, 0x00, 0x00])]
```
zslayton, line 549: 🗺️ The reader was treating the `offset` bits in the encoding as 2s complement bits; however, they are an unsigned integer representing the number of quarter hours from `-56` (offset `-14:00`). (Relevant [spec entry](https://github.com/amazon-ion/ion-docs/blob/ion-11-specification/src/binary-encoding.adoc#short_form_timestamp).)

src/lazy/binary/raw/v1_1/value.rs
```diff
@@ -266,7 +255,7 @@ impl<'top> LazyRawBinaryValue_1_1<'top> {
             (OpcodeType::Integer, 0x0) => 0.into(),
             (OpcodeType::Integer, n) => {
                 // We have n bytes following that make up our integer.
-                self.input.consume(1).read_fixed_int(n)?.0.into()
+                self.available_body().read_fixed_int(n)?.0.into()
```
zslayton, line 258: 🗺️ Adding annotations support meant changing a few usages of `self.input` because that buffer contains the annotations sequence (previously empty) and the value itself.

src/lazy/binary/raw/v1_1/struct.rs
```diff
@@ -140,7 +140,7 @@ impl<'top> RawBinaryStructIterator_1_1<'top> {
             bytes_to_skip: 0,
             struct_type: match opcode_type {
                 // TODO: Delimited struct handling
-                OpcodeType::Struct => StructType::FlexSym,
+                OpcodeType::Struct => StructType::SymbolAddress,
```
zslayton, line 143: 🗺️ For now, all structs are treated as `SymbolAddress` structs. We still need to add the switch-to-`FlexSym` functionality for length-prefixed structs.

src/lazy/encoder/binary/v1_1/value_writer.rs
```diff
@@ -313,8 +313,8 @@ impl<'value, 'top> BinaryValueWriter_1_1<'value, 'top> {
         // Compute the offset, its width in bits, and how that will affect the opcode and encoded length.
         let (num_offset_bits, offset_value, opcode_adjustment, length_adjustment) =
             match value.offset() {
-                None => (1, 1, 0, 0), // Unknown offset uses a single bit (1); opcode and length stay the same.
-                Some(0) => (1, 0, 0, 0), // UTC uses a single bit (0); opcode and length stay the same.
+                None => (1, 0, 0, 0), // Unknown offset uses a single bit (0); opcode and length stay the same.
+                Some(0) => (1, 1, 0, 0), // UTC uses a single bit (1); opcode and length stay the same.
```
zslayton, line 317: 🗺️ We only recently defined which `offset` bit value meant `UTC` and which meant `unknown`. I implemented this code long ago and guessed wrong. 🙃

src/lazy/encoder/binary/v1_1/value_writer.rs
```diff
@@ -1156,35 +1159,35 @@ mod tests {
             (
                 "2024-06-01T08:00Z",
                 //        MYYY_YYYY    DDDD_DMMM    mmmH_HHHH    ...._Ummm
-                &[0x83, 0b0011_0110, 0b0000_1011, 0b0000_1000, 0b0000_0000],
+                &[0x83, 0b0011_0110, 0b0000_1011, 0b0000_1000, 0b0000_1000],
```
zslayton, line 1162: 🗺️ Updating all of the timestamp encoding tests to use `UTC=1`, `Unknown=0`.

src/lazy/value.rs
```diff
@@ -310,10 +310,14 @@ impl<'top, D: Decoder> LazyValue<'top, D> {
...
-        let value: Value = value.read()?.try_into()?;
-        Ok(value.with_annotations(annotations))
+    fn try_from(lazy_value: LazyValue<'top, D>) -> Result<Self, Self::Error> {
+        let value: Value = lazy_value.read()?.try_into()?;
+        if lazy_value.has_annotations() {
+            let annotations: Annotations = lazy_value.annotations().try_into()?;
+            Ok(value.with_annotations(annotations))
+        } else {
+            Ok(value.into())
+        }
```
zslayton, line 320: 🗺️ This only calls `annotations()` if `has_annotations()` returns `true`.

zslayton (commented):
🗺️ PR tour 🧭

src/lazy/binary/raw/v1_1/annotations_iterator.rs
```diff
@@ -1,23 +1,55 @@
...
+            },
+            FlexSym => {
+                let (flex_sym, remaining_input) = match self.buffer.read_flex_sym() {
+                    Ok((flex_sym, remaining_input)) => (flex_sym, remaining_input),
+                    Err(error) => return Some(Err(error)),
+                };
+                let raw_symbol = match flex_sym.value() {
+                    FlexSymValue::SymbolRef(raw_symbol) => raw_symbol,
+                    FlexSymValue::Opcode(_) => {
+                        todo!("FlexSym escapes in annotation sequences")
```
popematt, line 46: Are `$0` and `''` going to be handled here or is that handled elsewhere so that it falls under the `SymbolRef` case?

src/lazy/binary/raw/v1_1/immutable_buffer.rs
```diff
@@ -340,19 +335,114 @@ impl<'a> ImmutableBuffer<'a> {
...
+        opcode: Opcode,
+    ) -> ParseResult<'a, EncodedAnnotations> {
+        let input_after_opcode = self.consume(1);
+        // TODO: This implementation actively reads the annotations, which isn't necessary.
+        //       At this phase of parsing we can just identify the buffer slice that contains
+        //       the annotations and remember their encoding; later on, the annotations iterator
+        //       can actually do the reading. That optimization would be impactful for FlexSyms
+        //       that represent inline text.
+        let (sequence, remaining_input) = match opcode.length_code {
+            7 => {
```
popematt, line 371: This is misleading that we're matching `length_code` here, and the length is actually unknown number of bytes or known to be 1 `FlexSym`—but the number we're matching is 7. Can we create an issue to rename `length_code` to `low_nibble` or something else like that? Or, if `length_code` is accurate for everything else, then maybe we figure out a way to set the length code to be something representative of the actual length. Alternately, we could just match the whole opcode byte here. Readability would improve if we were matching against `E7`, `E8`, etc.

src/lazy/binary/raw/v1_1/immutable_buffer.rs
```diff
@@ -340,19 +335,114 @@ impl<'a> ImmutableBuffer<'a> {
...
+                    })?,
+                };
+                (sequence, input_after_sym2)
+            }
+            9 => {
+                let (flex_uint, remaining_input) = input_after_opcode.read_flex_uint()?;
+                let sequence = EncodedAnnotations {
+                    encoding: AnnotationsEncoding::FlexSym,
+                    header_length: u8::try_from(1 + flex_uint.size_in_bytes()).map_err(|_| {
+                        IonError::decoding_error("found a 256+ byte annotations header")
```
popematt, line 404: Can `FlexUInt` even read a number this large? I thought it was limited to 128 bits.

src/lazy/binary/raw/v1_1/annotations_iterator.rs
zslayton: @glitch can comment on his intent, but I believe that's the idea.

src/lazy/binary/raw/v1_1/immutable_buffer.rs
zslayton: Yeah, agreed. It's `length_code` in like ... 80% of cases.

`OpcodeType` is (intentionally) a coarse enumeration:

https://github.com/amazon-ion/ion-rust/blob/8d1178ecbb661ec4f217e2edb44d49d16a83babe/src/lazy/binary/raw/v1_1/type_code.rs#L15-L55

It works well in that questions like "is it a float?" are easily answered without range checks. I'm going to rename it `low_nibble` (which is always accurate).

EDIT: done in 2cef8140c56f2e0042bdba0dbd51b9caf8b296b7

src/lazy/binary/raw/v1_1/immutable_buffer.rs
zslayton: It can't; the current implementation would always raise an error during the `FlexUInt` read. Would you like to see this be an `unreachable!` instead?

zslayton: I'm going to leave this as-is for the moment. If profiling shows that an `unreachable!` or a `cold_code!` would help out here I'll change it.

popematt (approved):

PR #789, opened 2024-06-10
Adds binary 1.1 read support for e-expressions, macro expansion

zslayton:
Introduces a `RawBinaryEExpression_1_1` type (mirroring the existing `RawTextEExpression_1_1` type) and adds methods to the 1.1 `ImmutableBuffer` to parse them. This requires access to the signature of the expression being parsed, which in turn requires access to the macro table. Previously, the buffer types (`ImmutableBuffer`, `TextBufferView`) each held a reference to the allocator in case they needed scratch space for caching child values or sanitizing/decoding input text. I have replaced the `allocator` field with a `context` field--the `EncodingContextRef` type allows access to both the allocator and the macro table.

Finally, this PR also wires the new binary e-expressions up to the macro evaluator for expansion. It adds a handful of unit tests demonstrating this.


src/lazy/binary/raw/v1_1/reader.rs
```diff
@@ -1,139 +1,105 @@
...
             return self.read_ivm(buffer);
         }
-        self.read_value(buffer)
-    }
-
-    /// Runs the provided parsing function on this reader's buffer.
-    /// If it succeeds, marks the reader  as ready to advance by the 'n' bytes
-    /// that were consumed.
-    /// If it does not succeed, the `DataSource` remains unchanged.
-    pub(crate) fn try_parse_next<
```
zslayton, line 119: 🗺️ This method was not being used, so I removed it.

src/lazy/binary/raw/v1_1/type_descriptor.rs
```diff
@@ -9,6 +9,7 @@ pub struct Opcode {
     pub opcode_type: OpcodeType,
     pub ion_type: Option<IonType>,
     pub low_nibble: u8,
+    pub byte: u8,
```
zslayton, line 12: 🗺️ When reading an e-expression in binary, the opcode byte is often the ID itself. Now the `Opcode` captures the original byte so we can inspect it later.

I briefly looked at removing the `low_nibble` field, thinking I could get that value by masking the `byte` field. I realized in the process of putting this PR together that the `low_nibble` field is not _actually_ the low nibble of the opcode (though it often is). It represents either the value length or, if a long encoding is being used, will be `0xFF`.

We'll probably want to change that field name back and document its intent.

src/lazy/expanded/mod.rs
```diff
@@ -90,18 +90,18 @@ pub mod template;
...
-#[derive(Copy, Clone, Debug)]
-pub struct EncodingContext<'top> {
-    pub(crate) macro_table: &'top MacroTable,
-    pub(crate) symbol_table: &'top SymbolTable,
-    pub(crate) allocator: &'top BumpAllocator,
+#[derive(Debug)]
+pub struct EncodingContext {
+    pub(crate) macro_table: MacroTable,
+    pub(crate) symbol_table: SymbolTable,
+    pub(crate) allocator: BumpAllocator,
```
zslayton, line 97: 🗺️ An earlier PR (#763) replaced the three-field `EncodingContext<'top>` in a variety of types with a single pointer. To do this, it put a copy of the three-field `EncodingContext<'top>` in the bump allocator and then gave out a reference to it.

This PR eliminates a layer of conceptual indirection by having the `EncodingContext` type directly own its three fields. Other types hold an `EncodingContextRef<'top>` that points to the `EncodingContext`.

src/lazy/expanded/mod.rs
```diff
@@ -215,12 +238,7 @@ pub struct ExpandingReader<Encoding: Decoder, Input: IonInput> {
     // Holds information found in symbol tables and encoding directives (TODO) that can be applied
     // to the encoding context the next time the reader is between top-level expressions.
     pending_lst: UnsafeCell<PendingLst>,
-    // A bump allocator that is cleared between top-level expressions.
-    allocator: UnsafeCell<BumpAllocator>,
-    // TODO: Make the symbol and macro tables traits on `Encoding` such that they can be configured
-    //       statically. Then 1.0 types can use `Never` for the macro table.
-    symbol_table: UnsafeCell<SymbolTable>,
-    macro_table: UnsafeCell<MacroTable>,
+    encoding_context: UnsafeCell<EncodingContext>,
```
zslayton, line 234: 🗺️ Grouping the `allocator`, `symbol_table`, and `macro_table` into the `encoding_context` field meant that there were a handful of places I could simplify the `UnsafeCell` code in the expanding reader.

src/lazy/expanded/mod.rs
```diff
@@ -232,45 +250,63 @@ impl<Encoding: Decoder, Input: IonInput> ExpandingReader<Encoding, Input> {
...
         // If the user is able to call this method, the PendingLst is not being modified and it's
         // safe to immutably reference.
         unsafe { &*self.pending_lst.get() }
     }
 
+    pub fn pending_lst_mut(&mut self) -> &mut PendingLst {
+        // SAFETY: If the caller has a `&mut` reference to `self`, it is the only mutable reference
+        //         that can modify `self.pending_lst`.
+        unsafe { &mut *self.pending_lst.get() }
+    }
```
zslayton, line 301: 🗺️ These methods where the mutability of the `self` reference (`&self` or `&mut self`) matches the mutability of the returned reference are only unsafe because the type being accessed is an `UnsafeCell`

src/lazy/reader.rs
```diff
@@ -275,4 +277,92 @@ mod tests {
...
+
+    #[test]
+    fn expand_binary_template_macro() -> IonResult<()> {
+        let macro_source = "(macro seventeen () 17)";
+        let encode_macro_fn = |address| vec![0xE0, 0x01, 0x01, 0xEA, address as u8];
+        expand_macro_test(macro_source, encode_macro_fn, |mut reader| {
+            assert_eq!(reader.expect_next()?.read()?.expect_i64()?, 17);
+            Ok(())
+        })
+    }
```
zslayton, line 316: 🗺️ The tests in this file are doing actual expansion of binary encoded e-expressions.

src/lazy/text/buffer.rs
```diff
@@ -776,22 +776,26 @@ impl<'top> TextBufferView<'top> {
...
-                // other parser kinds.
-                Err(e) => {
-                    return {
-                        let error = InvalidInputError::new(self)
-                            .with_label("matching a v1.1 list")
-                            .with_description(format!("{}", e));
-                        Err(nom::Err::Failure(IonParseError::Invalid(error)))
-                    }
+        let (span, child_exprs) = match TextListSpanFinder_1_1::new(
+            self.context.allocator(),
```
zslayton, line 780: 🗺️ The change from `self.allocator` to `self.context.allocator()` caused `rustfmt` to reflow the entire expression.

src/lazy/text/buffer.rs
```diff
@@ -975,22 +979,26 @@ impl<'top> TextBufferView<'top> {
...
-                // other parser kinds.
-                Err(e) => {
-                    return {
-                        let error = InvalidInputError::new(self)
-                            .with_label("matching a v1.1 struct")
-                            .with_description(format!("{}", e));
-                        Err(nom::Err::Failure(IonParseError::Invalid(error)))
-                    }
+        let (span, fields) = match TextStructSpanFinder_1_1::new(
+            self.context.allocator(),
```
zslayton, line 983: 🗺️ Same here; `rustfmt` reflow.

src/lazy/text/buffer.rs
```diff
@@ -1017,26 +1025,27 @@ impl<'top> TextBufferView<'top> {
...
-                return {
-                    let error = InvalidInputError::new(self)
-                        .with_label(format!(
-                            "matching an e-expression invoking macro {}",
-                            macro_name
-                        ))
-                        .with_description(format!("{}", e));
-                    Err(nom::Err::Failure(IonParseError::Invalid(error)))
+        let (span, child_expr_cache) =
+            match TextSExpSpanFinder_1_1::new(self.context.allocator(), sexp_iter)
```
zslayton, line 1029: 🗺️ Reflow here too.

src/lazy/text/raw/v1_1/reader.rs
```diff
@@ -34,6 +34,84 @@ pub struct LazyRawTextReader_1_1<'data> {
     local_offset: usize,
 }
 
+impl<'data> LazyRawReader<'data, TextEncoding_1_1> for LazyRawTextReader_1_1<'data> {
```
zslayton, line 37: 🗺️ I moved this `impl` block so it would be right below the definition of `LazyRawTextReader_1_1`. There are no logic changes.

zslayton (commented):
🗺️ PR tour 🧭

nirosys (approved):

src/lazy/binary/raw/v1_1/immutable_buffer.rs
```diff
@@ -21,7 +29,7 @@ use std::ops::Range;
 /// and a copy of the `ImmutableBuffer` that starts _after_ the bytes that were parsed.
 ///
 /// Methods that `peek` at the input stream do not return a copy of the buffer.
-#[derive(PartialEq, Clone, Copy)]
```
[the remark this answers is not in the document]
zslayton: I think at one time I was using these buffets in unit tests and wanted `assert_eq!` to work. I don't think we're using them like that anymore though. We can/should give it a new `PartialEq` impl like that when we need one.

PR #791, opened 2024-06-19
[serde] Implement serde Serializer/Deserializer::is_human_readable

theli-ua:
Will return true for text encodings, false for binary. is_human_readable is used to determine whether Serialize implementations should serialize in human-readable form.

Some types have a human-readable form that may be somewhat expensive to construct, as well as a binary form that is compact and efficient.

Fixes issue 790

*Issue #, if available:*

*Description of changes:*


desaikd:
As per offline discussion, changes for `is_human_readabale` can use `WriteConfig` on serialization side and `SystemReader` on deserialization side which is now tracked in #792.

desaikd:
As per offline discussion, this PR requires changes on serialization side to use `WriteConfig` and deserialization side to use `SystemReader`. Follow #792 for these changes.

PR #792, opened 2024-07-01
Implement serde Serializer/Deserializer::is_human_readable()

desaikd:
*Related PR #791*

*Description of changes:*
This PR works on adding `is_human_readable()` for `Serializer` and `Deserializer` in serde implementation of Ion.

*List of changes:*
- Added `is_human_readable` property in `ValueSerializer` based on `WriteConfig`.
- Added `is_human_readable` property in `ValueDeserializer` based on `SystemReader`'s `next_item()`.

*Test:*
Added a unit test that converts to/from binary and text as well as compares for roundtrip.

*Note for reviewer:*
_Miri test here is failing due to writer code in lazy writer and is unrelated to current change._


src/lazy/value.rs
```diff
@@ -496,21 +496,21 @@ mod tests {
     #[case::typed_null("null.list", IonType::List)]
     #[case::annotated_typed_null("foo::null.list", IonType::List.with_annotations(["foo"]))]
     #[case::boolean("false", false)]
-    #[case::negative_int("-1", -1)]
+    #[case::negative_int("-1", - 1)]
```
zslayton, line 499: At a glance, the whitespace changes in this file don't seem like something `rustfmt` would require. Did your IDE add these? I'm ok with it if it's a `rustfmt` thing.

src/serde/de.rs
```diff
@@ -19,20 +18,41 @@ where
...
 {
-    let mut reader = Reader::new(AnyEncoding, input)?;
-    let value = reader.expect_next()?;
-    let value_deserializer = ValueDeserializer::new(&value);
-    T::deserialize(value_deserializer)
+    let mut reader = SystemReader::new(AnyEncoding, input);
+    let item = reader.next_item()?;
+    match item {
+        SystemStreamItem::VersionMarker(marker) => {
+            // Note that this uses an older encoding here, but the format (i.e text or binary) stays the same which is why using the older encoding here works fine.
```
zslayton, line 25:
```suggestion
            // Note that this uses the Ion version with which the IVM was encoded rather than
            // the Ion version the stream is switching to. We can do this because the format 
            // (i.e text or binary) stays the same when the version changes.
```

src/serde/ser.rs
```diff
@@ -74,13 +67,15 @@ where
 /// Implements a standard serializer for Ion
 pub struct ValueSerializer<'a, V: ValueWriter> {
     pub(crate) value_writer: V,
+    pub(crate) is_human_readable: bool,
     lifetime: PhantomData<&'a ()>,
 }
 
 impl<'a, V: ValueWriter> ValueSerializer<'a, V> {
-    pub fn new(value_writer: V) -> Self {
+    pub fn new(value_writer: V, is_human_readable: bool) -> Self {
```
zslayton, line 75: Having this as a mandatory parameter on the constructor is a bit unfortunate, but I'm ok with it for now since we'll eventually be able to remove it when `LazyValue` is able to report its encoding.

zslayton (approved):
There are a couple of `clippy` warnings to address, but otherwise this LGTM! At a glance I don't think the Miri warnings are caused by this change--I'll look into those separately. They need to be addressed before we cut another release.

src/lazy/value.rs
desaikd: Ahh. I think yes, it was added by IDE. I will check with `rustfmt` and add change accordingly.

src/serde/ser.rs
desaikd: #793

PR #798, opened 2024-07-11
Fix certain symbols being incorrectly identified as incomplete boolean values

popematt:
**Issue #, if available:**

None

**Description of changes:**

Fixes a small bug in `match_bool` that was incorrectly reporting things like "fal" as incomplete booleans rather than as non-matches.

Also adds test cases, and updates existing test cases with a more concise macro and the ability to specify whether a value is a mismatch vs an incomplete match.

You may want to view this PR in the "split" diff view.

### DIFF
--- a/src/lazy/text/buffer.rs
+++ b/src/lazy/text/buffer.rs
@@ -1062,7 +1062,10 @@ impl<'top> TextBufferView<'top> {
     /// Matches and returns a boolean value.
     pub fn match_bool(self) -> IonParseResult<'top, bool> {
         terminated(
-            alt((value(true, tag("true")), value(false, tag("false")))),
+            alt((
+                value(true, complete_tag("true")),
+                value(false, complete_tag("false")),
+            )),
             Self::peek_stop_character,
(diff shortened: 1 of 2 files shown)


src/lazy/text/buffer.rs
```diff
@@ -1062,7 +1062,10 @@ impl<'top> TextBufferView<'top> {
     /// Matches and returns a boolean value.
     pub fn match_bool(self) -> IonParseResult<'top, bool> {
         terminated(
-            alt((value(true, tag("true")), value(false, tag("false")))),
+            alt((
+                value(true, complete_tag("true")),
+                value(false, complete_tag("false")),
+            )),
```
popematt, line 1068: 🗺️ This is the logic change. The rest of the changes in this file are (a) separating out `mismatch` and `incomplete` in the tests cases, and (b) introducing a macro to eliminate most of the nested functions, boilerplate code, and repeated code in the test cases.

zslayton (approved):

PR #804, opened 2024-08-10
Adds a 1.1 `ValueWriter` setting for writing symbol text inline

zslayton:
This PR builds on the changes in #802.

------

To start, the setting currently governs symbol values and annotations. There is not yet a way to:
* write inline annotations and interned values or vice-versa. 
* write a length-prefixed struct that writes its field names inline.

### DIFF
--- a/src/lazy/binary/raw/v1_1/immutable_buffer.rs
+++ b/src/lazy/binary/raw/v1_1/immutable_buffer.rs
@@ -20,6 +20,7 @@ use crate::lazy::encoder::binary::v1_1::flex_int::FlexInt;
 use crate::lazy::encoder::binary::v1_1::flex_sym::FlexSym;
 use crate::lazy::encoder::binary::v1_1::flex_uint::FlexUInt;
 use crate::lazy::expanded::macro_table::MacroRef;
+use crate::lazy::expanded::template::ParameterEncoding;
 use crate::lazy::expanded::EncodingContextRef;
 use crate::lazy::text::raw::v1_1::arg_group::EExpArgExpr;
 use crate::result::IonFailure;
@@ -221,6 +222,38 @@ impl<'a> ImmutableBuffer<'a> {
         let flex_uint = FlexUInt::read(self.bytes(), self.offset())?;
         let remaining = self.consume(flex_uint.size_in_bytes());
         Ok((flex_uint, remaining))
+    }
+
+    pub fn read_flex_uint_as_lazy_value(self) -> ParseResult<'a, LazyRawBinaryValue_1_1<'a>> {
+        let Some(first_byte) = self.peek_next_byte() else {
+            return IonResult::incomplete("a flex_uint", self.offset());
+        };
+        let size_in_bytes = match first_byte {
+            // If the first byte is zero, this flex_uint is encoded using 9+ bytes. That's pretty
+            // uncommon, so we'll just use the existing logic in the `read` method and discard the
+            // value. If this shows up in profiles, it can be optimized further.
+            0 => FlexUInt::read(self.bytes(), self.offset())?.size_in_bytes(),
+            _ => first_byte.trailing_zeros() as usize + 1,
+        };
+
+        if self.len() < size_in_bytes {
+            return IonResult::incomplete("reading a flex_uint value", self.offset());
+        }
+        // XXX: This *doesn't* slice `self` because FlexUInt::read() is faster if the input
+        //      is at least the size of a u64.
+        let matched_input = self;
+        let remaining_input = self.slice_to_end(size_in_bytes);
+        let value = LazyRawBinaryValue_1_1::for_flex_uint(matched_input);
+        Ok((value, remaining_input))
+    }
+
+    pub fn slice_to_end(&self, offset: usize) -> ImmutableBuffer<'a> {
+        ImmutableBuffer {
+            data: &self.data[offset..],
+            // stream offset + local offset
+            offset: self.offset + offset,
+            context: self.context,
+        }
     }
 
     #[inline]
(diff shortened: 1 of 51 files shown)


src/lazy/binary/raw/v1_1/immutable_buffer.rs
```diff
@@ -462,9 +462,12 @@ impl<'a> ImmutableBuffer<'a> {
...
         let input = self;
-        let header = type_descriptor
-            .to_header()
-            .ok_or_else(|| IonError::decoding_error("found a non-value in value position"))?;
+        let header = type_descriptor.to_header().ok_or_else(|| {
+            IonError::decoding_error(format!(
+                "found a non-value in value position; buffer=<{:X?}>",
+                input.bytes_range(0, 16.min(input.bytes().len()))
+            ))
+        })?;
```
zslayton, line 470: 🗺️ The old error message was maddening. Now you can at least see the unexpected opcode.

src/lazy/encoder/binary/v1_1/container_writers.rs
```diff
@@ -57,20 +58,26 @@ impl<'value, 'top> BinaryContainerWriter_1_1<'value, 'top> {
         start_opcode: u8,
         allocator: &'top BumpAllocator,
         buffer: &'value mut BumpVec<'top, u8>,
+        inline_symbol_text: bool,
```
zslayton, line 61: 🗺️ Most of this patch is just plumbing a new `inline_symbol_text` option through the different binary 1.1 container writers and associated machinery.

src/lazy/encoder/binary/v1_1/flex_sym.rs
```diff
@@ -73,9 +73,14 @@ impl<'top> FlexSym<'top> {
             Ordering::Less => {
                 let flex_int_len = value.size_in_bytes();
                 let len = sym_value.unsigned_abs() as usize;
-                let text = std::str::from_utf8(&input[flex_int_len..flex_int_len + len]).map_err(
-                    |_| IonError::decoding_error("found FlexSym with invalid UTF-8 data"),
-                )?;
+                let flex_sym_end = flex_int_len + len;
+                if input.len() < flex_sym_end {
+                    return IonResult::incomplete("reading a FlexSym", offset);
+                }
```
zslayton, line 79: 🗺️ If the buffer ended in the middle of a `FlexSym`, it would lead to a panic instead of reporting `Incomplete`.

src/lazy/encoder/writer.rs
```diff
@@ -207,9 +228,10 @@ impl<'value, V: ValueWriter> AnnotatableWriter for ApplicationValueWriter<'value
     where
         Self: 'a,
     {
-        if self.encoding.symbol_creation_policy == SymbolCreationPolicy::WriteProvidedToken {
+        if self.has_inline_symbol_text() {
```
zslayton, line 231: 🗺️ `has_inline_symbol_text` is more specific than the initial `SymbolCreationPolicy`. It may be that I can remove that concept, but if so I'll do that in another PR.

src/lazy/encoder/writer.rs
```diff
@@ -207,9 +228,10 @@ impl<'value, V: ValueWriter> AnnotatableWriter for ApplicationValueWriter<'value
...
             // Store the tokens as they are. Text will be written as text, symbol IDs will be written
-            // as symbol IDs. TODO: Lookup SIDs to see if they have text?
+            // as symbol IDs.
+            // TODO: Replace SIDs w/text if they're defined? (Config setting?)
```
zslayton, line 234: 🗺️ If the caller passes a SID, we write it as-is. At a minimum, we should validate that it's in range.

src/serde/de.rs
```diff
@@ -18,26 +21,28 @@ where
...
-            // (i.e text or binary) stays the same when the version changes.
-            // TODO: Use new encoding, once we have APIs to get new/old encodings for the marker.
-            let is_human_readable = marker.encoding().is_text();
-            let value = reader.expect_next_value()?;
-            let value_deserializer = ValueDeserializer::new(&value, is_human_readable);
-            T::deserialize(value_deserializer)
-        }
-        SystemStreamItem::Value(value) => {
-            let value_deserializer = ValueDeserializer::new(&value, true);
-            T::deserialize(value_deserializer)
```
zslayton, line 36: 🗺️ This old implementation would return an error if an LST preceded the first value.

zslayton (commented):
🗺️ PR tour

src/lazy/encoder/binary/v1_1/container_writers.rs
popematt: Is it worth having a `write_options: WriteOptions` (or similar) field instead so that any new options/flags can reuse the same plumbing rather than needing to add a new field in all of these different places?

src/lazy/encoder/binary/v1_1/value_writer.rs
```diff
@@ -50,11 +53,24 @@ impl<'value, 'top> BinaryValueWriter_1_1<'value, 'top> {
...
+    pub fn has_delimited_containers(&self) -> bool {
+        self.delimited_containers
+    }
+
     pub fn with_length_prefixed_containers(mut self) -> Self {
         self.delimited_containers = false;
         self
     }
 
+    pub fn with_inline_symbol_text(mut self) -> Self {
```
popematt, line 65: Should this accept a `bool` so that you could also turn it off?

src/lazy/encoder/binary/v1_1/container_writers.rs
zslayton: Yeah, that is better. I've moved over to a consolidated `ValueWriterConfig` type in https://github.com/amazon-ion/ion-rust/pull/804/commits/ef7c84545db248ee42760aa185fc428e183da56a.

It adds separate settings for:
* Encoding symbol values
* Encoding annotations sequences
* Encoding struct field names
* Delimited vs LP containers

popematt (approved):

PR #813, opened 2024-08-14
Adds text 1.1 read support for eexps with a numeric address

zslayton:
Adds parsing support for numeric macro IDs in an e-expression. Also revises some error messages related to `Int` downcasting--instead of "too big", they now say "out of range".

### DIFF
--- a/src/lazy/text/buffer.rs
+++ b/src/lazy/text/buffer.rs
@@ -1833,7 +1833,7 @@ impl<'top> TextBufferView<'top> {
         terminated(complete_digit1, peek(not(complete_tag("_"))))
             .map(|buffer: TextBufferView| {
                 // The matched buffer is ascii base 10 digits, parsing must succeed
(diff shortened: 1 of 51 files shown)


src/lazy/text/buffer.rs
```diff
@@ -3006,13 +3028,22 @@ mod tests {
...
+            "(:4 1 2 3)",
+            "(:4 (1 2 3))",
+            "(:4 \"foo\")",
+            "(:4 foo)",
         ],
         expect_mismatch: [
             "foo",   // No parens
             "(foo)", // No `:` after opening paren
+            "(4",    // No parens
+            "(4)",   // No `:` after opening paren
```
popematt, line 3042: Here's a couple more test cases that might fail right now because it looks like you're using the same matcher as for integer values.

```suggestion
            "(4)",   // No `:` after opening paren
            "(:0x4)", // Hexadecimal not allowed
            "(:4_000)", // underscores not allowed in number
```

popematt (approved):

PR #815, opened 2024-08-15
Allow `?` and `*` parameters to be omitted in tail position

zslayton:
Previously the text reader only allowed arguments that could use rest syntax to also be omitted altogether.

### DIFF
--- a/src/lazy/text/buffer.rs
+++ b/src/lazy/text/buffer.rs
@@ -1142,7 +1142,7 @@ impl<'top> TextBufferView<'top> {
                 Some(arg) => arg_expr_cache.push(arg),
                 None => {
                     for param in &signature_params[index..] {
-                        if param.rest_syntax_policy() == RestSyntaxPolicy::NotAllowed {
+                        if !param.can_be_omitted() {
                             return fatal_parse_error(
                                 self,
                                 format!(
@@ -1242,8 +1242,16 @@ impl<'top> TextBufferView<'top> {
         }
     }
 
-    pub fn match_empty_arg_group(self) -> IonMatchResult<'top> {
-        recognize(pair(tag("(:"), whitespace_and_then(tag(")"))))(self)
+    pub fn match_empty_arg_group(
+        self,
+        parameter: &'top Parameter,
+    ) -> IonParseResult<'top, EExpArg<'top, TextEncoding_1_1>> {
+        recognize(pair(tag("(:"), whitespace_and_then(tag(")"))))
+            .map(|matched_expr| {
+                let arg_group = TextEExpArgGroup::new(parameter, matched_expr, &[]);
+                EExpArg::new(parameter, EExpArgExpr::ArgGroup(arg_group))
+            })
+            .parse(self)
     }
 
     pub fn match_zero_or_one(
@@ -1251,7 +1259,7 @@ impl<'top> TextBufferView<'top> {
         parameter: &'top Parameter,
     ) -> IonParseResult<'top, Option<EExpArg<'top, TextEncoding_1_1>>> {
         whitespace_and_then(alt((
-            Self::match_empty_arg_group.map(|_| None),
+            Self::parser_with_arg(Self::match_empty_arg_group, parameter).map(Some),
             // TODO: Match a non-empty arg group and turn it into a failure with a helpful error message
             Self::match_sexp_value_1_1.map(|maybe_expr| {
                 maybe_expr.map(|expr| {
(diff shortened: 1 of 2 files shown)


src/lazy/text/buffer.rs
```diff
@@ -1242,16 +1242,24 @@ impl<'top> TextBufferView<'top> {
         }
     }
 
-    pub fn match_empty_arg_group(self) -> IonMatchResult<'top> {
-        recognize(pair(tag("(:"), whitespace_and_then(tag(")"))))(self)
```
zslayton, line 1246: 🗺️ The original impl was just returning the matched slice. Now there's a concrete type that represents an argument group, so we instantiate one of those and return it.

src/lazy/text/buffer.rs
```diff
@@ -2660,8 +2668,8 @@ mod tests {
     }
 
     macro_rules! matcher_tests_with_macro {
-        ($parser:ident $macro_src:literal $($expect:ident: [$($input:literal),+$(,)?]),+$(,)?) => {
-            mod $parser {
+        ($mod_name:ident $parser:ident $macro_src:literal $($expect:ident: [$($input:literal),+$(,)?]),+$(,)?) => {
+            mod $mod_name {
```
zslayton, line 2672: 🗺️ This macro creates a module with a bunch of nested unit tests. However, the module name was also the name of the parser. This meant you couldn't write two sets of unit tests that used the same parser entry point. I added another parameter (`$mod_name`) to this macro to allow the module name to be specified separately.

src/lazy/text/buffer.rs
```diff
@@ -2967,6 +2975,7 @@ mod tests {
     }
 
     matcher_tests_with_macro! {
+        parsing_sexps
```
zslayton, line 2978: 🗺️ In the next few tests, I added a distinct module name.

zslayton (commented):
🗺️ PR Tour 🧭

src/lazy/text/buffer.rs
```diff
@@ -3046,6 +3057,22 @@ mod tests {
...
     }
 
+    matcher_tests_with_macro! {
+        allow_omitting_trailing_optionals
+        match_e_expression
+        "(macro foo (a b+ c? d*) null)"
+        expect_match: [
+            "(:foo 1 2)",
+            "(:foo 1 2 3)",
+            "(:foo 1 2 3 4)",
```
popematt, line 3067:
```suggestion
            "(:foo 1 2 3 4)",
            "(:foo 1 2 (: 3) 4)",
            "(:foo 1 2 3 (: 4))",
```
...unless these are already covered elsewhere.

src/lazy/text/buffer.rs
```diff
@@ -3046,6 +3057,22 @@ mod tests {
...
+            "(:foo 1 2 3 4 5 6)", // implicit rest
```
popematt, line 3068:
```suggestion
            "(:foo 1 2 3 4 5 6)", // implicit rest
            "(:foo 1 2 (: 3) 4 5 6)", // implicit rest
            "(:foo 1 2 3 (: 4 5 6))",
```
...unless these are already covered elsewhere.

src/lazy/text/buffer.rs
```diff
@@ -3046,6 +3057,22 @@ mod tests {
         ]
     }
 
+    matcher_tests_with_macro! {
+        allow_omitting_trailing_optionals
+        match_e_expression
+        "(macro foo (a b+ c? d*) null)"
+        expect_match: [
```
tgregg, line 3064: Worth adding an expectation that the explicit empty argument group `(:)` also works?

src/lazy/text/buffer.rs
```diff
@@ -3046,6 +3057,22 @@ mod tests {
...
+            "(:foo 1 2)",
+            "(:foo 1 2 3)",
+            "(:foo 1 2 3 4)",
+            "(:foo 1 2 3 4 5 6)", // implicit rest
+        ],
+        expect_mismatch: [
+            "(:foo 1)",
+            "(:foo)",
```
popematt, line 3075:
```suggestion
            "(:foo)",
            "(:foo 1 2 3 4 (: 5) 6)", // final argument must be rest or expression group, but not a mix of both
```
...unless this is already covered elsewhere.

src/lazy/text/buffer.rs
zslayton: The text reader doesn't have non-empty arg group support yet.

Issue #794, opened 2024-07-02
Miri tests failure 


desaikd:
#792 and #791 showed miri tests failure on lazy writer `close` code.
Related logs:
https://github.com/amazon-ion/ion-rust/actions/runs/9748900720/job/26904915793?pr=792#step:5:272

PR #821, opened 2024-08-22
Configures Miri to use tree borrows, fixes remaining errors

zslayton:
Related to #794.

Miri's [tree borrows](https://perso.crans.org/vanille/treebor/index.html) aliasing model is more permissive than stacked borrows (the default), which is known to reject some programs that are actually safe.

This PR configures our CI `miri` check to use tree borrows and addresses several of Miri's complaints. This is enough to satisfy the tests currently in CI (the tests for the experimental `serde` feature).

There are still some issues to investigate. When running Miri on the entire test suite, it complains about at least one more problem, though it's unclear whether it's actually a potential problem or another limitation in its aliasing model. This patch is an improvement, however, and I'd rather not block these changes as we look into the others.

### DIFF
--- a/src/lazy/encoder/binary/v1_1/writer.rs
+++ b/src/lazy/encoder/binary/v1_1/writer.rs
@@ -12,7 +12,7 @@ use crate::lazy::encoder::value_writer_config::ValueWriterConfig;
 use crate::lazy::encoder::write_as_ion::WriteAsIon;
 use crate::lazy::encoder::LazyRawWriter;
 use crate::lazy::encoding::Encoding;
-use crate::unsafe_helpers::{mut_ref_to_ptr, ptr_to_mut_ref};
+use crate::unsafe_helpers::{mut_ref_to_ptr, ptr_to_mut_ref, ptr_to_ref};
 use crate::write_config::{WriteConfig, WriteConfigKind};
 use crate::{IonEncoding, IonResult};
 
@@ -74,7 +74,7 @@ impl<W: Write> LazyRawBinaryWriter_1_1<W> {
 
         let encoding_buffer = match encoding_buffer_ptr {
             // If `encoding_buffer_ptr` is set, get the slice of bytes to which it refers.
-            Some(ptr) => unsafe { ptr_to_mut_ref::<'_, BumpVec<'_, u8>>(*ptr).as_slice() },
+            Some(ptr) => unsafe { ptr_to_ref::<'_, BumpVec<'_, u8>>(*ptr).as_slice() },
             // Otherwise, there's nothing in the buffer. Use an empty slice.
             None => &[],
         };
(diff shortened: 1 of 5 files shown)


src/unsafe_helpers.rs
```diff
@@ -19,6 +19,15 @@ pub(crate) unsafe fn ptr_to_mut_ref<'a, T>(ptr: *mut ()) -> &'a mut T {
...
 }
 
+/// Helper function that turns a raw pointer into an immutable reference of the specified type.
+///
+/// The caller is responsible for confirming that `ptr` is a valid reference to some value
+/// of type `T`.
+pub(crate) unsafe fn ptr_to_ref<'a, T>(ptr: *mut ()) -> &'a T {
+    let typed_ptr: *const T = ptr.cast();
+    &*typed_ptr
+}
```
zslayton, line 29: 🗺️ Previously we only had a `ptr_to_mut_ref` which meant we were getting a mutable reference even where a read-only reference would suffice.

src/lazy/encoder/binary/v1_1/writer.rs
```diff
@@ -74,7 +74,7 @@ impl<W: Write> LazyRawBinaryWriter_1_1<W> {
 
         let encoding_buffer = match encoding_buffer_ptr {
             // If `encoding_buffer_ptr` is set, get the slice of bytes to which it refers.
-            Some(ptr) => unsafe { ptr_to_mut_ref::<'_, BumpVec<'_, u8>>(*ptr).as_slice() },
+            Some(ptr) => unsafe { ptr_to_ref::<'_, BumpVec<'_, u8>>(*ptr).as_slice() },
```
zslayton, line 77: 🗺️ In `flush()` we only need to read from the encoding buffer, so we don't need a mut ref.

src/lazy/expanded/mod.rs
```diff
@@ -468,10 +468,12 @@ impl<Encoding: Decoder, Input: IonInput> ExpandingReader<Encoding, Input> {
...
-            let symbol_table: &mut SymbolTable =
-                &mut unsafe { &mut *self.encoding_context.get() }.symbol_table;
-            let macro_table: &mut MacroTable =
-                &mut unsafe { &mut *self.encoding_context.get() }.macro_table;
+            let encoding_context_ref = unsafe { &mut *self.encoding_context.get() };
+            let EncodingContext {
+                macro_table,
+                symbol_table,
+                ..
+            } = encoding_context_ref;
```
zslayton, line 476: 🗺️ Here we get a single mutable reference and then use safe code to destructure it into two mutable references.

src/lazy/expanded/mod.rs
```diff
@@ -530,8 +532,8 @@ impl<Encoding: Decoder, Input: IonInput> ExpandingReader<Encoding, Input> {
         // See if the raw reader can get another expression from the input stream. It's possible
         // to find an expression that yields no values (for example: `(:void)`), so we perform this
         // step in a loop until we get a value or end-of-stream.
-        let allocator: &BumpAllocator = self.context().allocator();
-        let context_ref = EncodingContextRef::new(allocator.alloc_with(|| self.context()));
+        let context_ref = self.context();
+
```
zslayton, line 536: 🗺️ This was copying the `EncodingContextRef` to the bump allocator so it would survive the lifetime `'top`...but it was already going to survive that because the `EncodingContext` lives in the `ExpandingReader`.

zslayton (commented):
🗺️ PR Tour 🧭

tgregg (approved):

PR #818, opened 2024-08-20
Make newly-compiled macros available for use

jobarr-amzn:
Formerly, a macro defined in a macro table definition could not be referenced in a subsequent macro in that same definition. Now it can!

Fixes #817.

*Issue #, if available:* #817

*Description of changes:* 
Removed the `EncodingContext` from `compiler.rs`, replaced with a reference to the macro table under assembly as values are compiled. 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #822, opened 2024-08-23
Enables dependent macros

zslayton:
This builds on PR #821. It also borrows from and supersedes the changes in @jobarr-amzn's PR #818.

Modifies the `TemplateCompiler` to resolve macro IDs in the table of 'pending' macros (defined in the current encoding directive) rather than the active macro table. This allows macros to depend on previously defined macros.

> [!NOTE] 
> `cargo fmt` did some reflows that accounts for ~half of the diff. Fortunately, all of the formatting changes occur in a contiguous block at the end of the PR. I've added a PR tour comment showing where that begins.

### DIFF
--- a/src/lazy/expanded/compiler.rs
+++ b/src/lazy/expanded/compiler.rs
@@ -14,11 +14,12 @@ use crate::lazy::expanded::template::{
 use crate::lazy::expanded::EncodingContextRef;
 use crate::lazy::r#struct::LazyStruct;
 use crate::lazy::sequence::{LazyList, LazySExp};
+use crate::lazy::text::raw::v1_1::reader::MacroIdRef;
 use crate::lazy::value::LazyValue;
 use crate::lazy::value_ref::ValueRef;
 use crate::result::IonFailure;
 use crate::symbol_ref::AsSymbolRef;
-use crate::{v1_1, IonError, IonResult, IonType, Reader, Symbol, SymbolRef};
+use crate::{v1_1, IonError, IonResult, IonType, MacroTable, Reader, Symbol, SymbolRef};
 
 /// Information inferred about a template's expansion at compile time.
 #[derive(Copy, Clone, Debug, PartialEq)]
@@ -94,7 +95,7 @@ impl ExpansionSingleton {
 }
 
 /// Validates a given TDL expression and compiles it into a `TemplateMacro` that can be added
-/// to a [`MacroTable`](crate::lazy::expanded::macro_table::MacroTable).
+/// to a [`MacroTable`].
 pub struct TemplateCompiler {}
 
 impl TemplateCompiler {
@@ -142,12 +143,11 @@ impl TemplateCompiler {
         context: EncodingContextRef,
         expression: &str,
     ) -> IonResult<TemplateMacro> {
-        // TODO: This is a rudimentary implementation that panics instead of performing thorough
-        //       validation. Where it does surface errors, the messages are too terse.
+        // TODO: This is a rudimentary implementation that surfaces terse errors.
         let mut reader = Reader::new(v1_1::Text, expression.as_bytes())?;
         let macro_def_sexp = reader.expect_next()?.read()?.expect_sexp()?;
 
-        Self::compile_from_sexp(context, macro_def_sexp)
+        Self::compile_from_sexp(context, &MacroTable::default(), macro_def_sexp)
     }
 
     /// Pulls the next value from the provided source and confirms that it is a symbol whose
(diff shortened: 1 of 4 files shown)


src/lazy/expanded/compiler.rs
```diff
@@ -94,7 +95,7 @@ impl ExpansionSingleton {
 }
 
 /// Validates a given TDL expression and compiles it into a `TemplateMacro` that can be added
-/// to a [`MacroTable`](crate::lazy::expanded::macro_table::MacroTable).
+/// to a [`MacroTable`].
```
zslayton, line 98: 🗺️ This explicit path was made redundant because this module now imports `MacroTable`.

src/lazy/expanded/compiler.rs
```diff
@@ -142,12 +143,11 @@ impl TemplateCompiler {
         context: EncodingContextRef,
         expression: &str,
     ) -> IonResult<TemplateMacro> {
-        // TODO: This is a rudimentary implementation that panics instead of performing thorough
-        //       validation. Where it does surface errors, the messages are too terse.
+        // TODO: This is a rudimentary implementation that surfaces terse errors.
```
zslayton, line 146: 🗺️ Updated this comment as the impl hasn't used panics in a while.

src/lazy/expanded/compiler.rs
```diff
@@ -450,12 +455,11 @@ impl TemplateCompiler {
     /// [`TemplateBodyExpr`] sequences to the `TemplateBody`.
     ///
     /// If `is_literal` is true, nested symbols and s-expressions will not be interpreted.
-    fn compile_value<'top, D: Decoder>(
-        context: EncodingContextRef<'top>,
-        signature: &MacroSignature,
+    fn compile_value<D: Decoder>(
+        tdl_context: TdlContext,
```
zslayton, line 459: 🗺️ Rather than passing a `context`, `signature`, and (now) `pending_macros` to the compiler for each subexpression, we now pass a single `TdlContext` (defined below on line 840). We can extend this `TdlContext` as needed in the future to avoid similar plumb-the-argument-through PRs.

src/lazy/expanded/compiler.rs
```diff
@@ -631,61 +626,84 @@ impl TemplateCompiler {
...
                     if let Some(name) = s.text() {
-                        let address =
-                            context.macro_table.address_for_name(name).ok_or_else(|| {
-                                IonError::decoding_error(format!("unrecognized macro name: {name}"))
-                            })?;
-                        Ok((Some(name.to_string()), address))
+                        MacroIdRef::LocalName(name)
                     } else {
-                        IonResult::decoding_error("macro names must be an identifier")
+                        return IonResult::decoding_error("macro names must be an identifier");
```
zslayton, line 647: 🗺️ Rather than performing the actual lookup of the ID here, this method now saves the ID in a variable (`macro_id` on line 635 above) and then does the lookup on line 665 below.

I believe that we're going to want to eventually support depending on macros in the active macro table; when we go to add that, we'll just add another `.or_else(|| Self::resolve_macro_id_in(...))` call after line 665.

src/lazy/expanded/template.rs
```diff
@@ -5,19 +5,25 @@ use std::ops::{Deref, Range};
 use bumpalo::collections::Vec as BumpVec;
 use rustc_hash::FxHashMap;
 
-use crate::{Bytes, Decimal, Int, IonResult, IonType, LazyExpandedFieldName, Str, Symbol, SymbolRef, Timestamp, try_or_some_err, Value};
```
zslayton, line 8: 🗺️ 
> [!IMPORTANT]
> ### >> Everything from here on is just a formatting change done by `rustfmt`. << 
> There are no logic changes.

zslayton (commented):
🗺️ PR Tour 🧭

src/lazy/expanded/compiler.rs
```diff
@@ -631,61 +626,84 @@ impl TemplateCompiler {
...
+        // Now that we have our ID (name/address), try to resolve it to a previously defined macro.
+        // We'll look in the pending macros table first...
+        Self::resolve_macro_id_in(macro_id, tdl_context.pending_macros)
+            // ...or we'll raise an error.
+            .ok_or_else(|| {
+                IonError::decoding_error(format!("macro '{macro_id}' has not been defined (yet?)"))
+            })
+        // TODO: Should it be possible for a macro to reference (and therefore "close over") a macro
+        //       in the current context? The referenced macro would not necessarily be accessible
+        //       in the new encoding context, but would still need to be accessible internally.
```
tgregg, line 672: This would be consistent with how the symbol table works, right?

tgregg (approved):

src/lazy/expanded/compiler.rs
zslayton: Yes, in that the active symbol table remains available while reading a new symbol table. However, symbols don't depend on other symbol definitions (you can't put `$10` in the `symbols` list of an LST), so this is new territory in that respect.

PR #825, opened 2024-08-27
Adds support for eexps with 12- and 20-bit addresses

zslayton:
Prior to this PR, e-expressions with an address between `0..64` could be encoded as a single-byte opcode and larger e-expressions could only be encoded as a length-prefixed e-expression.

This patch adds read and write support for binary e-expressions that have a trailing 1- or 2-byte `FixedUInt` to complete the address.

### DIFF
--- a/src/lazy/binary/raw/v1_1/immutable_buffer.rs
+++ b/src/lazy/binary/raw/v1_1/immutable_buffer.rs
@@ -1546,7 +1546,7 @@ mod tests {
         // to its own address.
         let mut macro_definitions = String::from("$ion_encoding::(\n  (macro_table\n");
         for address in MacroTable::FIRST_USER_MACRO_ID..MAX_TEST_MACRO_ADDRESS {
-            write!(macro_definitions, "    (macro m{address} () {address})\n")?;
+            writeln!(macro_definitions, "    (macro m{address} () {address})")?;
         }
         macro_definitions.push_str("  )\n)\n");
         let encoding_directive = Element::read_one(macro_definitions)?;
--- a/src/lazy/binary/raw/v1_1/immutable_buffer.rs
+++ b/src/lazy/binary/raw/v1_1/immutable_buffer.rs
@@ -24,6 +24,7 @@ use crate::lazy::expanded::macro_table::MacroRef;
 use crate::lazy::expanded::template::ParameterEncoding;
 use crate::lazy::expanded::EncodingContextRef;
 use crate::lazy::text::raw::v1_1::arg_group::EExpArgExpr;
+use crate::lazy::text::raw::v1_1::reader::MacroAddress;
 use crate::result::IonFailure;
 use crate::{v1_1, IonError, IonResult};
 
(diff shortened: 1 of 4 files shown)


src/lazy/binary/raw/v1_1/type_code.rs
```diff
@@ -13,18 +13,19 @@ use crate::IonType;
...
-    Decimal,                   // 0x70-0x7F -
-    TimestampShort,            // 0x80-0x8F -
-    String,                    // 0x90-0x9F -
-    InlineSymbol,              // 0xA0-0xAF -
-    List,                      // 0xB0-0xBF -
-    SExpression,               // 0xC0-0xCF -
-    StructEmpty,               // 0xD0      -
+    EExpressionWith4BitAddress,  // 0x00-0x3F -
+    EExpressionWith12BitAddress, // 0x40-0x4F -
+    EExpressionWith20BitAddress, // 0x50-0x5F -
```
zslayton, line 18: 🗺️ Of this block, these three lines (`EExpressionWith*BitAddress`) are the new/interesting ones. The others are highlighted because the comments were automatically realigned.

src/lazy/binary/raw/v1_1/immutable_buffer.rs
```diff
@@ -396,12 +397,11 @@ impl<'a> ImmutableBuffer<'a> {
...
-                ParseValueExprResult::EExp(self.read_eexp_with_address_in_opcode(opcode))
-            }
-            EExpressionAddressFollows => todo!("eexp address follows"),
-            EExpressionWithLengthPrefix => {
-                ParseValueExprResult::EExp(self.read_eexp_with_length_prefix(opcode))
+            EExpressionWith4BitAddress
+            | EExpressionWith12BitAddress
+            | EExpressionWith20BitAddress
+            | EExpressionWithLengthPrefix => {
+                ParseValueExprResult::EExp(self.read_e_expression(opcode))
```
zslayton, line 404: 🗺️ I've been considering adding a less granular `OpcodeKind` or similar alongside `OpcodeType` to capture high-level groups like `EExpression` or `Value` for places like this where we don't actually care about the encoding just yet.

zslayton (commented):
🗺️ PR Tour 🧭

tgregg (approved):

src/lazy/encoder/binary/v1_1/value_writer.rs
```diff
@@ -672,19 +672,58 @@ impl<'value, 'top> BinaryValueWriter_1_1<'value, 'top> {
...
+            MacroIdRef::LocalAddress(address) if address <= MAX_20_BIT_ADDRESS => {
+                // Encode as an opcode with trailing 2-byte FixedUInt
+                const BIAS: usize = MIN_20_BIT_ADDRESS;
+                let biased = address - BIAS;
+                let low_opcode_nibble = biased >> 16;
+                self.encoding_buffer.extend_from_slices_copy(&[
+                    &[
+                        // Opcode with low nibble setting bias
+                        0x50 | low_opcode_nibble as u8,
+                    ],
```
popematt, line 717:
```suggestion
                    // Opcode with low nibble setting bias
                    0x50 | low_opcode_nibble as u8,
```

src/lazy/encoder/binary/v1_1/value_writer.rs
```diff
@@ -672,19 +672,58 @@ impl<'value, 'top> BinaryValueWriter_1_1<'value, 'top> {
...
+                    // Remaining bytes of magnitude in biased address
+                    &(biased as u16).to_le_bytes(),
+                ]);
+            }
             MacroIdRef::LocalAddress(_address) => {
-                todo!("macros with addresses higher than 64");
+                todo!("macros with addresses higher than (2^20)-1");
```
popematt, line 723: Minor nitpick—it's actually macros with addresses higher than `2^20 + 2^12 + 2^6 - 1`.

popematt (approved):
Overall it looks great.

src/lazy/encoder/binary/v1_1/value_writer.rs
zslayton: The [`extend_from_slices_copy`](https://docs.rs/bumpalo/latest/bumpalo/collections/vec/struct.Vec.html#method.extend_from_slices_copy) method takes a slice-of-slices (`&[&[u8]]`). For potentially interesting information about the motivation for this method, check out [the PR that introduced it](https://github.com/fitzgen/bumpalo/pull/240).

Now that I'm re-reading this, though, I could rewrite it to call `(biased as u16).to_le_bytes()` in advance and pack the two bytes into this slice manually. I think I'll do that.

popematt: I see that now. I missed that the call on L701 was `extend_from_slice_copy` whereas the call on L713 was `extend_from_slices_copy`.

PR #826, opened 2024-08-28
Add conformance testing DSL

nirosys:
*Issue #, if available:* n/a

*Description of changes:*
This PR adds an implementation of the conformance DSL, along with a runner for testing conformance tests during CI/CD, and a separate runner that can be used to target files more ad-hoc. The ion-tests submodule has also been updated in order to pull in the conformance tests, which also required some updates to the `test_resource` paths for 1.0.

Most of the code is modeled directly after the grammar. Documents contain fragments and a continuation. Expectations, which are a subset of continuations, act as terminal nodes and trigger the evaluation of the test. Context tracks the fragments, version, and encoding, for the test's evaluation path.

## CI/CD Runner
```
$ cargo test -F experimental --test conformance_tests
   Compiling ion-rs v1.0.0-rc.7 (/Users/glitch/Code/ion-rust)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 3.65s
     Running tests/conformance_tests.rs (target/debug/deps/conformance_tests-6a032b27940772cd)

running 8 tests
test ion_tests::conformance_ion_tests_conformance_null_ion ... ok
test implementation::test_encoding ... ok
test ion_tests::conformance_ion_tests_conformance_core_string_symbol_ion ... ok
test implementation::test_timestamps ... ok
test ion_tests::conformance_ion_tests_conformance_core_empty_document_ion ... ok
test implementation::test_simple_docs ... ok
test ion_tests::conformance_ion_tests_conformance_core_typed_null_ion ... ok
test ion_tests::conformance_ion_tests_conformance_core_toplevel_produces_ion ... ok

test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.02s

$
```

## CLI Runner
```
$ cat test.ion
(ion_1_1 "Test Bytes" (text "{{aGVsbG8=}}") (denotes (Blob "68 65 6c 6c 6f")))
(ion_1_1 "Test Bytes (Fail)" (text "{{aGVsbG8=}}") (denotes (Blob "68 65 6c 6c 5f")))

(ion_1_1 "Test Clob" (text "{{ \"hello\" }}") (denotes (Clob "68 65 6c 6c 6f")))
(ion_1_1 "Test Clob (Fail)" (text "{{ \"hello\" }}") (denotes (Clob "68 65 6c 6c 5f")))

$ cargo test -F experimental --test conformance -- ./test.ion
   Compiling ion-rs v1.0.0-rc.7 (/Users/glitch/Code/ion-rust)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 1.92s
     Running tests/conformance.rs (target/debug/deps/conformance-62318380ee1917b3)
Testing 1 conformance collections.


Running tests: ./test.ion ========================
   Test Bytes         ...  [OK]
   Test Bytes (Fail)  ...  [FAILED]
   Test Clob          ...  [OK]
   Test Clob (Fail)   ...  [FAILED]
-------------------------
File: ./test.ion
Test: Test Bytes (Fail)
Error: ConformanceError(ConformanceErrorImpl { file: "", test_name: "", kind: MismatchedDenotes })
-------------------------
File: ./test.ion
Test: Test Clob (Fail)
Error: ConformanceError(ConformanceErrorImpl { file: "", test_name: "", kind: MismatchedDenotes })
thread 'main' panicked at tests/conformance.rs:51:9:
Conformance test(s) failed
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: test failed, to rerun pass `--test conformance`
$
```

## Notes
* The MacTab fragment is implemented as the spec describes, expanding into a toplevel fragment with named module for defining the macro table. Adding a new named module is not supported in ion-rust yet. Macros can still be added using the encoding fragment.
* Absent symbols are currently not supported. 

I'll follow up after this PR to make errors reporting better. I also have some plans for refactoring and simplifying some of the implementation, post merge.

---

### DIFF
--- a/src/lazy/binary/raw/v1_1/value.rs
+++ b/src/lazy/binary/raw/v1_1/value.rs
@@ -194,7 +194,13 @@ impl<'top> LazyRawValue<'top, BinaryEncoding_1_1> for &'top LazyRawBinaryValue_1
             return Ok(ValueRef::Int(int));
         }
         if self.is_null() {
-            return Ok(ValueRef::Null(self.ion_type()));
+            let ion_type = if self.encoded_value.header.ion_type_code == OpcodeType::TypedNull {
+                let body = self.value_body();
+                ION_1_1_TYPED_NULL_TYPES[body[0] as usize]
+            } else {
+                IonType::Null
+            };
+            return Ok(ValueRef::Null(ion_type));
         }
         // Anecdotally, string and integer values are very common in Ion streams. This `match` creates
         // an inlineable fast path for them while other types go through the general case impl.
--- a/src/lazy/binary/raw/v1_1/value.rs
+++ b/src/lazy/binary/raw/v1_1/value.rs
@@ -119,7 +119,13 @@ impl<'top> HasRange for &'top LazyRawBinaryValue_1_1<'top> {
 
 impl<'top> LazyRawValue<'top, BinaryEncoding_1_1> for &'top LazyRawBinaryValue_1_1<'top> {
     fn ion_type(&self) -> IonType {
-        self.encoded_value.ion_type()
+        // Handle retrieving the type for a typed null.
+        if self.encoded_value.header.type_code() == OpcodeType::TypedNull {
+            let body = self.value_body();
+            ION_1_1_TYPED_NULL_TYPES[body[0] as usize]
+        } else {
+            self.encoded_value.ion_type()
+        }
     }
 
     fn is_null(&self) -> bool {
@@ -145,13 +151,7 @@ impl<'top> LazyRawValue<'top, BinaryEncoding_1_1> for &'top LazyRawBinaryValue_1
         }
 
         if self.is_null() {
-            let ion_type = if self.encoded_value.header.ion_type_code == OpcodeType::TypedNull {
-                let body = self.value_body();
-                ION_1_1_TYPED_NULL_TYPES[body[0] as usize]
-            } else {
-                IonType::Null
-            };
-            return Ok(RawValueRef::Null(ion_type));
+            return Ok(RawValueRef::Null(self.ion_type()));
         }
 
         match self.ion_type() {
@@ -194,13 +194,7 @@ impl<'top> LazyRawValue<'top, BinaryEncoding_1_1> for &'top LazyRawBinaryValue_1
             return Ok(ValueRef::Int(int));
         }
         if self.is_null() {
-            let ion_type = if self.encoded_value.header.ion_type_code == OpcodeType::TypedNull {
-                let body = self.value_body();
-                ION_1_1_TYPED_NULL_TYPES[body[0] as usize]
-            } else {
-                IonType::Null
-            };
-            return Ok(ValueRef::Null(ion_type));
+            return Ok(ValueRef::Null(self.ion_type()));
         }
         // Anecdotally, string and integer values are very common in Ion streams. This `match` creates
         // an inlineable fast path for them while other types go through the general case impl.
--- a/tests/conformance_dsl/context.rs
+++ b/tests/conformance_dsl/context.rs
@@ -3,6 +3,9 @@ use crate::conformance_dsl::*;
 use ion_rs::{Element, ElementReader, Sequence, Reader, IonSlice};
 use ion_rs::{v1_0, v1_1};
 
+/// A Context forms a scope for tracking all of the document fragments and any parent Contexts that
+/// also need to be considered. Through this context the ability to generate the full test
+/// document input, and evaluate any forced encodings or Ion versions, is provided.
 #[derive(Clone, Copy, Debug)]
 pub(crate) struct Context<'a> {
     version: IonVersion,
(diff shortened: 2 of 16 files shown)


src/lazy/binary/raw/v1_1/value.rs
```diff
@@ -194,7 +194,13 @@ impl<'top> LazyRawValue<'top, BinaryEncoding_1_1> for &'top LazyRawBinaryValue_1
...
         }
         if self.is_null() {
-            return Ok(ValueRef::Null(self.ion_type()));
+            let ion_type = if self.encoded_value.header.ion_type_code == OpcodeType::TypedNull {
+                let body = self.value_body();
+                ION_1_1_TYPED_NULL_TYPES[body[0] as usize]
+            } else {
+                IonType::Null
+            };
+            return Ok(ValueRef::Null(ion_type));
```
zslayton, line 203: Why was this needed? `ion_type()` should return the correctly interpreted type from the opcode and optional typed null byte. (`return Ok(ValueRef::Null(self.ion_type()))` also appears below on line 227).

tests/conformance_dsl/context.rs
```diff
@@ -0,0 +1,127 @@
...
+            let empty: Vec<Element> = vec!();
+            return Ok(empty.into());
+        }
+
+        let version = match self.version() {
+            IonVersion::Unspecified => IonVersion::V1_0,
+            v => v,
+        };
+
+        // Ok(Reader::new(AnyEncoding, data_slice)?.read_all_elements()?)
```
zslayton, line 108: Old comment?
```suggestion
        // Ok(Reader::new(AnyEncoding, data_slice)?.read_all_elements()?)
```

tests/conformance_dsl/model.rs
```diff
@@ -0,0 +1,411 @@
+use ion_rs::{Decimal, Element, IonType, Sequence, Timestamp};
+use ion_rs::decimal::coefficient::Coefficient;
+use super::{Clause, ClauseType, ConformanceErrorKind, InnerResult, parse_text_exp, parse_bytes_exp};
+
+use std::collections::HashMap;
+
+#[derive(Debug, Clone, Eq, Hash, PartialEq)]
+pub(crate) enum SymTok {
```
zslayton, line 8: I suggest spelling out the words:
```suggestion
pub(crate) enum SymbolToken {
```

tests/conformance_tests.rs
```diff
@@ -0,0 +1,94 @@
...
+
+mod ion_tests {
+    use super::*;
+
+    #[test_resources("ion-tests/conformance/null.ion")]
+    #[test_resources("ion-tests/conformance/core/typed_null.ion")]
+    #[test_resources("ion-tests/conformance/core/string_symbol.ion")]
+    #[test_resources("ion-tests/conformance/core/empty_document.ion")]
+    #[test_resources("ion-tests/conformance/core/toplevel_produces.ion")]
+    // #[test_resources("ion-tests/conformance/ivm.ion")]
```
zslayton, line 86:
```suggestion
    // #[test_resources("ion-tests/conformance/ivm.ion")]
```
?

zslayton (commented):
This looks good! Please do a documentation pass. It would be helpful to connect implementation sections to the part of the DSL spec they implement. Having an explanation of the motivation for top-level types like `ModelValue` would also be nice.

src/lazy/binary/raw/v1_1/value.rs
nirosys: When parsing the Opcode we don't have the byte containing the true ion type ([here](https://github.com/amazon-ion/ion-rust/blob/main/src/lazy/binary/raw/v1_1/type_descriptor.rs#L79)), so the actual ion type was originally deferred until the value is read ([here](https://github.com/amazon-ion/ion-rust/blob/main/src/lazy/binary/raw/v1_1/value.rs#L147)). Right now `ion_type()` just returns the type from the opcode, but you're right, it's a better place to be handling this logic.

tests/conformance_tests.rs
nirosys: Yea, I had commented this out because it currently doesn't pass due to the symbol syntax that's needed. Removing it for now.

src/lazy/binary/raw/v1_1/value.rs
```diff
@@ -119,7 +119,13 @@ impl<'top> HasRange for &'top LazyRawBinaryValue_1_1<'top> {
...
 impl<'top> LazyRawValue<'top, BinaryEncoding_1_1> for &'top LazyRawBinaryValue_1_1<'top> {
     fn ion_type(&self) -> IonType {
-        self.encoded_value.ion_type()
+        // Handle retrieving the type for a typed null.
+        if self.encoded_value.header.type_code() == OpcodeType::TypedNull {
+            let body = self.value_body();
+            ION_1_1_TYPED_NULL_TYPES[body[0] as usize]
+        } else {
+            self.encoded_value.ion_type()
+        }
```
zslayton, line 128: Could you put this functionality in a `self.read_null()` method instead of having it happen in `ion_type()`?

There are three other raw value types besides `LazyRawBinaryValue_1_1`:
* `LazyRawBinaryValue_1_0`
* `LazyRawTextValue_1_1`
* `LazyRawTextValue_1_0`

Each of them has an `ion_type()` method and I'm pretty sure at least some of them return `IonType::Null` if the value is a null value, even if it's a typed null. The behavior you've implemented is probably more correct, but if we make this change here we'd need to make it in the other three places too.

zslayton (approved):

src/lazy/binary/raw/v1_1/value.rs
zslayton: Added #828 to track this.

nirosys: ack, ty, I will follow up with that today.

Issue #805, opened 2024-08-11
Derive `opcode_length` instead of setting it explicitly


zslayton:
Can this be derived (i.e. `fn is_tagless() -> bool`) rather than storing the opcode length in a field? Tagless encodings _must_ have `value_body_length + length_length == total_length`, and anything else _must_ have an opcode, so `value_body_length + length_length < total_length` is always true for anything that is not a tagless encoded argument. Right?

_Originally posted by @popematt in https://github.com/amazon-ion/ion-rust/pull/801#discussion_r1712903906_

zslayton:
Also look into addressing [this comment](https://github.com/amazon-ion/ion-rust/pull/801/files#r1712904408):

> Perhaps `Header` should actually be an enum so that it could have a `NoHeader` variant in addition to whatever other implementations you have going on?

Issue #829, opened 2024-09-05
`Reader` documentation says it is a `binary reader`


jpschorr:
I believe this comment about being a `binary` reader is out-of-date.

https://github.com/amazon-ion/ion-rust/blob/6f04a8e3e9cbc62d56c7d46eb12563773b8baa08/src/lazy/reader.rs#L13C5-L13C20

zslayton:
Fixed by #830.

PR #830, opened 2024-09-05
Adds support for qualified refs to `$ion`, `$ion_encoding`

zslayton:
* Modifies compiled macros to use `Rc` to store references to dependent macros. This allows the macro to continue referring to its dependent even if the latter is no longer directly addressable.
* Adds support for `module_name::encoding_name::parameter_name` syntax and `(module_name::macro_name ...)` syntax. The only modules currently supported are `$ion` (the system module) and `$ion_encoding` (the active encoding module).

### DIFF
--- a/src/lazy/binary/encoded_value.rs
+++ b/src/lazy/binary/encoded_value.rs
@@ -90,8 +90,6 @@ pub(crate) struct EncodedValue<HeaderType: EncodedHeader> {
     pub annotations_encoding: AnnotationsEncoding,
     // The offset of the type descriptor byte within the overall input stream.
     pub header_offset: usize,
-    // If this value was written with a tagless encoding, this will be 0. Otherwise, it's 1.
-    pub opcode_length: u8,
     // The number of bytes used to encode the optional length VarUInt following the header byte.
     pub length_length: u8,
     // The number of bytes used to encode the value itself, not including the header byte
@@ -128,6 +126,15 @@ impl<HeaderType: EncodedHeader> EncodedValue<HeaderType> {
         let start = self.header_offset;
         let end = start + self.header_length();
         start..end
+    }
+
+    /// Returns the number of bytes used to encode this value's opcode. If this value was serialized
+    /// using a tagless encoding, returns `0`.
+    pub fn opcode_length(&self) -> usize {
+        match self.encoding {
+            ParameterEncoding::Tagged => 1,
+            _ => 0,
+        }
     }
 
     /// Returns the number of bytes used to encode this value's data.
@@ -279,7 +286,6 @@ mod tests {
             annotations_sequence_length: 1,
             annotations_encoding: AnnotationsEncoding::SymbolAddress,
             header_offset: 200,
-            opcode_length: 1,
             length_length: 0,
             value_body_length: 3,
             total_length: 7,
--- a/src/lazy/binary/encoded_value.rs
+++ b/src/lazy/binary/encoded_value.rs
@@ -1,6 +1,6 @@
 use crate::lazy::binary::raw::type_descriptor::Header;
 use crate::lazy::binary::raw::v1_1::immutable_buffer::AnnotationsEncoding;
-use crate::lazy::expanded::template::ParameterEncoding;
+use crate::lazy::binary::raw::v1_1::value::BinaryValueEncoding;
 use crate::IonType;
 use std::ops::Range;
 
@@ -41,7 +41,7 @@ impl EncodedHeader for Header {
 /// without re-parsing its header information each time.
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub(crate) struct EncodedValue<HeaderType: EncodedHeader> {
-    pub(crate) encoding: ParameterEncoding,
+    pub(crate) encoding: BinaryValueEncoding,
     // If the compiler decides that a value is too large to be moved/copied with inline code,
     // it will relocate the value using memcpy instead. This can be quite slow by comparison.
     //
@@ -132,7 +132,7 @@ impl<HeaderType: EncodedHeader> EncodedValue<HeaderType> {
     /// using a tagless encoding, returns `0`.
     pub fn opcode_length(&self) -> usize {
         match self.encoding {
-            ParameterEncoding::Tagged => 1,
+            BinaryValueEncoding::Tagged => 1,
             _ => 0,
         }
     }
@@ -269,14 +269,14 @@ mod tests {
     use crate::lazy::binary::encoded_value::EncodedValue;
     use crate::lazy::binary::raw::type_descriptor::Header;
     use crate::lazy::binary::raw::v1_1::immutable_buffer::AnnotationsEncoding;
-    use crate::lazy::expanded::template::ParameterEncoding;
+    use crate::lazy::binary::raw::v1_1::value::BinaryValueEncoding;
     use crate::{IonResult, IonType};
 
     #[test]
     fn accessors() -> IonResult<()> {
         // 3-byte String with 1-byte annotation
         let value = EncodedValue {
-            encoding: ParameterEncoding::Tagged,
+            encoding: BinaryValueEncoding::Tagged,
             header: Header {
                 ion_type: IonType::String,
                 ion_type_code: IonTypeCode::String,
--- a/src/lazy/binary/raw/v1_1/e_expression.rs
+++ b/src/lazy/binary/raw/v1_1/e_expression.rs
@@ -306,6 +306,9 @@ impl<'top> Iterator for BinaryEExpArgsInputIter<'top> {
                         EExpArg::new(parameter, EExpArgExpr::ValueLiteral(value_ref)),
                         remaining,
                     )
+                }
+                ParameterEncoding::MacroShaped(_macro_ref) => {
+                    todo!("macro-shaped parameter encoding")
                 } // TODO: The other tagless encodings
             },
             // If it's an argument group...
@@ -448,7 +451,7 @@ impl<'top> IntoIterator for BinaryEExpArgGroup<'top> {
 impl<'top> EExpressionArgGroup<'top, BinaryEncoding_1_1> for BinaryEExpArgGroup<'top> {
     type Iterator = BinaryEExpArgGroupIterator<'top>;
 
-    fn encoding(&self) -> ParameterEncoding {
+    fn encoding(&self) -> &ParameterEncoding {
         self.parameter.encoding()
     }
 
--- a/src/lazy/binary/raw/v1_1/immutable_buffer.rs
+++ b/src/lazy/binary/raw/v1_1/immutable_buffer.rs
@@ -11,7 +11,7 @@ use crate::lazy::binary::raw::v1_1::e_expression::{
 };
 use crate::lazy::binary::raw::v1_1::r#struct::LazyRawBinaryFieldName_1_1;
 use crate::lazy::binary::raw::v1_1::value::{
-    DelimitedContents, LazyRawBinaryValue_1_1, LazyRawBinaryVersionMarker_1_1,
+    BinaryValueEncoding, DelimitedContents, LazyRawBinaryValue_1_1, LazyRawBinaryVersionMarker_1_1,
 };
 use crate::lazy::binary::raw::v1_1::{Header, LengthType, Opcode, OpcodeType, ION_1_1_OPCODES};
 use crate::lazy::decoder::{LazyRawFieldExpr, LazyRawValueExpr, RawValueExpr};
@@ -21,7 +21,6 @@ use crate::lazy::encoder::binary::v1_1::flex_int::FlexInt;
 use crate::lazy::encoder::binary::v1_1::flex_sym::{FlexSym, FlexSymValue};
 use crate::lazy::encoder::binary::v1_1::flex_uint::FlexUInt;
 use crate::lazy::expanded::macro_table::MacroRef;
-use crate::lazy::expanded::template::ParameterEncoding;
 use crate::lazy::expanded::EncodingContextRef;
 use crate::lazy::text::raw::v1_1::arg_group::EExpArgExpr;
 use crate::lazy::text::raw::v1_1::reader::MacroAddress;
@@ -663,7 +662,7 @@ impl<'a> ImmutableBuffer<'a> {
         }
 
         let encoded_value = EncodedValue {
-            encoding: ParameterEncoding::Tagged,
+            encoding: BinaryValueEncoding::Tagged,
             header,
             // If applicable, these are populated by the caller: `read_annotated_value()`
             annotations_header_length: 0,
@@ -1542,7 +1541,8 @@ mod tests {
 
         // Construct an encoding directive that defines this number of macros. Each macro will expand
         // to its own address.
-        let mut macro_definitions = String::from("$ion_encoding::(\n  (macro_table\n");
+        let mut macro_definitions =
+            String::from("$ion_encoding::(\n  (macro_table $ion_encoding\n");
         for address in MacroTable::FIRST_USER_MACRO_ID..MAX_TEST_MACRO_ADDRESS {
             writeln!(macro_definitions, "    (macro m{address} () {address})")?;
         }
--- a/src/lazy/binary/raw/v1_1/immutable_buffer.rs
+++ b/src/lazy/binary/raw/v1_1/immutable_buffer.rs
@@ -670,8 +670,6 @@ impl<'a> ImmutableBuffer<'a> {
             annotations_sequence_length: 0,
             annotations_encoding: AnnotationsEncoding::SymbolAddress,
             header_offset,
-            // This is a tagged value, so its opcode length is always 1
-            opcode_length: 1,
             length_length,
             value_body_length,
             total_length,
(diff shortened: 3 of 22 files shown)


src/lazy/binary/encoded_value.rs
```diff
@@ -130,6 +128,15 @@ impl<HeaderType: EncodedHeader> EncodedValue<HeaderType> {
...
 
+    /// Returns the number of bytes used to encode this value's opcode. If this value was serialized
+    /// using a tagless encoding, returns `0`.
+    pub fn opcode_length(&self) -> usize {
+        match self.encoding {
+            BinaryValueEncoding::Tagged => 1,
+            _ => 0,
+        }
+    }
+
```
zslayton, line 139: Rather that setting/storing a number of opcode bytes on every value, we can derive whether an opcode byte was present. Fixes #805.

src/lazy/binary/encoded_value.rs
```diff
@@ -41,7 +41,7 @@ impl EncodedHeader for Header {
 /// without re-parsing its header information each time.
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub(crate) struct EncodedValue<HeaderType: EncodedHeader> {
-    pub(crate) encoding: ParameterEncoding,
+    pub(crate) encoding: BinaryValueEncoding,
```
zslayton, line 44: 🗺️ Previously, `ParameterEncoding` was a `Copy` type:

https://github.com/amazon-ion/ion-rust/blob/6f04a8e3e9cbc62d56c7d46eb12563773b8baa08/src/lazy/expanded/template.rs#L82-L88

Starting with this PR (which stubs out macro-shaped parameter support), parameters needed the ability to hold a reference to a `Macro` that would last as long as necessary. To achieve this, a new variant was added: `MacroShaped(Rc<Macro>)`.

This meant that `ParameterEncoding` was no longer `Copy`. However, the only place it was used that required it to be `Copy` was the `EncodedValue` struct, which was using it as a proxy for "encoding backing this binary value". Since binary values can only be backed by _value_ encodings (not macros), it made sense to introduce a new type for that use case.

src/lazy/binary/raw/v1_1/e_expression.rs
```diff
@@ -448,7 +451,7 @@ impl<'top> IntoIterator for BinaryEExpArgGroup<'top> {
 impl<'top> EExpressionArgGroup<'top, BinaryEncoding_1_1> for BinaryEExpArgGroup<'top> {
     type Iterator = BinaryEExpArgGroupIterator<'top>;
 
-    fn encoding(&self) -> ParameterEncoding {
+    fn encoding(&self) -> &ParameterEncoding {
```
zslayton, line 454: 🗺️ `ParameterEncoding` is no longer `Copy`, so accessors hand out a reference instead of a copy.

src/lazy/binary/raw/v1_1/immutable_buffer.rs
```diff
@@ -1544,7 +1541,8 @@ mod tests {
 
         // Construct an encoding directive that defines this number of macros. Each macro will expand
         // to its own address.
-        let mut macro_definitions = String::from("$ion_encoding::(\n  (macro_table\n");
+        let mut macro_definitions =
+            String::from("$ion_encoding::(\n  (macro_table $ion_encoding\n");
```
zslayton, line 1545: 🗺️ This PR modified the encoding directives to create a new macro table each time while also supporting re-exports of `$ion` and `$ion_encoding`. This meant that several unit tests that used multiple encoding directives saw macro definitions disappear. I've modified those tests to re-export `$ion_encoding` (that is: do a macro table append) as needed.

src/lazy/binary/raw/v1_1/struct.rs
```diff
@@ -337,6 +337,7 @@ mod tests {
                 $ion_encoding::(
                     (symbol_table $ion_encoding)
                     (macro_table
+                        $ion_encoding
```
zslayton, line 340: 🗺️ Macro table append.

src/lazy/encoder/writer.rs
```diff
@@ -69,7 +69,7 @@ impl<E: Encoding, Output: Write> Writer<E, Output> {
         // TODO: LazyEncoder should define a method to construct a new symtab and/or macro table
         let ion_version = E::ion_version();
         let symbol_table = SymbolTable::new(ion_version);
-        let macro_table = MacroTable::new();
+        let macro_table = MacroTable::with_system_macros();
```
zslayton, line 72: 🗺️ `MacroTable` now has a `with_system_macros()` method (that pre-populates it with system macro aliases) and an `empty()` method that constructs a new instance with no macros at all. The latter is used for processing encoding directives.

src/lazy/expanded/compiler.rs
```diff
@@ -272,18 +272,128 @@ impl TemplateCompiler {
...
+                ))
+            }
+        };
+
+        // If it's an unqualified name, look for the associated macro in the local scope.
+        if let (encoding_name, None) = (annotation1, annotation2) {
+            if let Some(macro_ref) =
+                Self::resolve_unqualified_macro_id(context, pending_macros, encoding_name)
+            {
+                Self::validate_macro_shape_for_encoding(&macro_ref)?;
```
zslayton, line 329: 🗺️ This check makes sure the macro isn't a constant. Constants cannot be macro-shaped parameter encodings.

src/lazy/expanded/compiler.rs
```diff
@@ -602,15 +712,11 @@ impl TemplateCompiler {
         let mut expressions = lazy_sexp.iter();
         // Convert the macro ID (name or address) into an address. If this refers to a macro that
         // doesn't exist yet, this will return an error. This prevents recursion.
-        // TODO: Consider storing the name of the invoked target macro in the host's definition
-        //       as debug information. The name cannot be stored directly on the
-        //       TemplateBodyMacroInvocation as that would prevent the type from being `Copy`.
-        let (_maybe_name, macro_address) =
-            Self::name_and_address_from_id_expr(tdl_context, expressions.next())?;
+        let macro_ref = Self::name_and_address_from_id_expr(tdl_context, expressions.next())?;
```
zslayton: 🗺️ Now that we're using `Rc` to hold a reference to the `Macro` itself, the `name` is always available.

src/lazy/expanded/compiler.rs
```diff
@@ -602,15 +712,11 @@ impl TemplateCompiler {
...
         let macro_step_index = definition.expressions.len();
         // Assume the macro contains zero argument expressions to start, we'll update
         // this at the end of the function.
-        definition.push_macro_invocation(macro_address, ExprRange::empty());
+        definition.push_macro_invocation(Rc::clone(&macro_ref), ExprRange::empty());
```
zslayton, line 719: 🗺️ `Rc`s are only _created_ when the `Macro` is compiled. They might get cloned (i.e. shallow-copied) some number of times after that during compilation. After compilation, however, only `&` references are used. This should mean that `Rc` overhead is negligible when amortized across the lifetime of the stream.

src/lazy/expanded/compiler.rs
```diff
@@ -883,16 +988,16 @@ mod tests {
     fn expect_macro(
         definition: &TemplateMacro,
         index: usize,
-        expected_address: usize,
-        expected_num_arguments: usize,
+        expected_macro: Rc<Macro>,
+        expected_num_args: usize,
```
zslayton, line 992: 🗺️ Here and below, we're updating some unit tests that previously relied on templates storing a macro address rather than holding an `Rc<Macro>`.

src/lazy/expanded/encoding_module.rs
```diff
@@ -16,6 +16,13 @@ impl EncodingModule {
             symbol_table,
         }
     }
+    pub fn ion_1_1_system_module() -> Self {
+        Self::new(
+            String::from("$ion"),
+            MacroTable::with_system_macros(),
+            SymbolTable::new(IonVersion::v1_1),
```
zslayton, line 23: Once system symbols are supported, this can be changed to `empty()`.

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -196,12 +186,12 @@ impl<'top, D: Decoder> MacroExpr<'top, D> {
         MacroExprArgsIterator { source: args_kind }
     }
 
-    pub(crate) fn invoked_macro(&self) -> MacroRef<'top> {
+    pub(crate) fn invoked_macro(&self) -> &'top Macro {
```
zslayton, line 189: 🗺️ Previously, the system required that all `MacroExpr`s could tell you their address in the macro table. However, it is now possible for a macro's body to invoke a macro that is not addressable via the active encoding context. This required changing several field types from `MacroRef<'top>` to `&'top Macro`. (`&'top Macro` is only 8 bytes compared to `MacroRef`'s 16 bytes though, so it's actually kind of nice.)

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -1272,6 +1259,81 @@ mod tests {
         )
     }
 
+    #[test]
+    fn macros_close_over_dependencies() -> IonResult<()> {
```
zslayton, line 1263: 🗺️ This unit test demonstrates defining a macro with a dependency on an unaddressable macro.

src/lazy/expanded/macro_table.rs
```diff
@@ -159,14 +158,14 @@ impl<'top> MacroRef<'top> {
 /// its validity and allowing evaluation to begin.
 #[derive(Debug, Clone)]
 pub struct MacroTable {
-    macros_by_address: Vec<Macro>,
+    macros_by_address: Vec<Rc<Macro>>,
```
zslayton, line 161: 🗺️ The `MacroTable` now stores `Rc<Macro>`s, so copying the macro table's contents to a new macro table is substantially cheaper than it was before.

src/lazy/expanded/macro_table.rs
```diff
@@ -183,9 +182,9 @@ impl MacroTable {
     // is expected to change as development continues. It is currently used in several unit tests.
     pub const FIRST_USER_MACRO_ID: usize = Self::NUM_SYSTEM_MACROS;
 
-    pub fn new() -> Self {
-        let macros_by_id = vec![
-            Macro::named(
+    fn system_macros() -> Vec<Rc<Macro>> {
+        vec![
+            Rc::new(Macro::named(
```
zslayton, line 187: 🗺️ I intend to make several thread locals that can be cheaply referenced during encoding directive construction. For now though, I build the system module for each new encoding directive.

src/lazy/expanded/macro_table.rs
```diff
@@ -331,6 +342,25 @@ impl MacroTable {
         Some(MacroRef { address, reference })
     }
 
+    pub(crate) fn clone_macro_with_name(&self, name: &str) -> Option<Rc<Macro>> {
+        let address = *self.macros_by_name.get(name)?;
+        let reference = self.macros_by_address.get(address)?;
+        Some(Rc::clone(reference))
+    }
```
zslayton, line 349: 🗺️ Variants of existing methods that return a cloned `Rc` instead of a `&`.

src/lazy/expanded/macro_table.rs
```diff
@@ -348,7 +378,23 @@ impl MacroTable {
             template.expansion_analysis,
         );
 
-        self.macros_by_address.push(new_macro);
+        self.macros_by_address.push(Rc::new(new_macro));
         Ok(id)
     }
+
+    pub(crate) fn append_all_macros_from(&mut self, other: &MacroTable) -> IonResult<()> {
```
zslayton, line 385: 🗺️ This is used by the system reader to copy macros between encoding modules.

src/lazy/expanded/template.rs
```diff
@@ -434,14 +435,10 @@ impl<'top, D: Decoder> Iterator for TemplateSequenceIterator<'top, D> {
                 TemplateBodyExprKind::MacroInvocation(body_invocation) => {
                     // ...it's a TDL macro invocation. Resolve the invocation to get a reference to the
                     // macro being invoked.
-                    let invoked_macro = self
-                        .context
-                        .macro_table()
-                        .macro_at_address(body_invocation.invoked_macro_address)
-                        .unwrap();
+                    let invoked_macro = body_invocation.invoked_macro.as_ref();
```
zslayton, line 438: 🗺️ There were several places where a macro address would need to be resolved into a macro reference. These are no longer needed because the macro definitions hold an `Rc`.

src/lazy/expanded/template.rs
```diff
@@ -899,46 +892,34 @@ impl TemplateBodyExprKind {
 }
 
 /// A macro invocation found in the body of a template.
-///
-/// Because the template definition lives in the macro table (which may need to grow in the process
-/// of evaluating a stream), this type holds the address of the invoked macro rather than a
-/// reference to it.
```
zslayton, line 905: 🗺️ Now that we're using `Rc`, the template definition's dependent data will live as long as necessary. And because macros cannot have cyclical dependencies, the `Rc`s cannot accidentally form cycles either.

src/lazy/expanded/template.rs
```diff
@@ -994,32 +971,20 @@ impl<'top> TemplateMacroInvocation<'top> {
...
 
     /// Helper method to access the definition of the host template. Useful for debugging,
     /// but not required for macro expansion.
-    pub fn host_macro_ref(&self) -> MacroRef<'top> {
-        self.context()
-            .macro_table()
-            .macro_at_address(self.host_template_address)
-            .unwrap()
+    pub fn host_macro_ref(&self) -> &'top Macro {
+        self.host_template.macro_ref
```
zslayton, line 981: 🗺️ Here and below: more address lookups and unwrapping obviated by holding an actual reference.

src/lazy/reader.rs
```diff
@@ -10,7 +10,7 @@ use crate::read_config::ReadConfig;
 use crate::result::IonFailure;
 use crate::{IonError, IonResult};
 
-/// A binary reader that only reads each value that it visits upon request (that is: lazily).
+/// An Ion reader that only reads each value that it visits upon request (that is: lazily).
```
zslayton, line 13: 🗺️ An incidental doc comment change to address #829.

src/lazy/system_reader.rs
```diff
@@ -378,19 +378,35 @@ impl<Encoding: Decoder, Input: IonInput> SystemReader<Encoding, Input> {
                 "expected a macro table definition operation, but found: {operation:?}"
             ));
         }
-        let mut macro_table = MacroTable::new();
+        let mut macro_table = MacroTable::empty();
```
zslayton, line 381: 🗺️ When reading a `(macro_table ...)` clause, the reader no longer pre-populates the macro table with system macros.

zslayton (commented):
🗺️ PR tour 🧭

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -1272,6 +1259,81 @@ mod tests {
...
+                $ion_encoding::(
+                    (macro_table
+                        $ion_encoding
+                        (macro double ($x) (values $x $x))
+                    )
+                )
+
+                $ion_encoding::(
+                    (macro_table
+                        $ion_encoding // Re-export the active encoding module's macros
```
popematt, line 1314: In this particular case, you could also write `(export double)`, right?

src/lazy/expanded/template.rs
popematt: Also, if the macro `struct` is immutable, that's another guarantee that there will be no `Rc` cycles because you can't create something with a reference to something else that doesn't exist yet.

src/lazy/binary/raw/v1_1/value.rs
```diff
@@ -97,7 +97,8 @@ impl<'top> RawVersionMarker<'top> for LazyRawBinaryVersionMarker_1_1<'top> {
 /// Encodings that can back a lazy binary value. Binary 1.0 values are always backed by
 /// the `Tagged` variant.
 ///
-/// This is a subset of the encodings in the [`ParameterEncoding`] enum.
+/// This is a subset of the encodings in the
+/// [`ParameterEncoding`](crate::lazy::expanded::template::ParameterEncoding) enum.
```
popematt, line 101: This would be a good spot to also add a comment about why this exists and what the subset should be.

popematt (approved):

src/lazy/expanded/macro_evaluator.rs
zslayton: Yes 👍

src/lazy/expanded/compiler.rs
```diff
@@ -272,18 +272,128 @@ impl TemplateCompiler {
...
+                Self::validate_macro_shape_for_encoding(&macro_ref)?;
+                return Ok(ParameterEncoding::MacroShaped(macro_ref));
+            }
+            // If it's not in the local scope, see if it's a built-in.
+            return match encoding_name {
                 "flex_uint" => Ok(ParameterEncoding::FlexUInt),
-                other => IonResult::decoding_error(format!(
-                    "unsupported encoding '{other}' specified for parameter"
+                _ => IonResult::decoding_error(format!(
+                    "unrecognized encoding '{encoding_name}' specified for parameter"
```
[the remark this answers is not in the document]
zslayton: I've opened #832 to track making all of the compiler errors more informative.

Issue #828, opened 2024-09-03
Make `ion_type()` correctly reflect typed nulls


zslayton:
Could you put this functionality in a `self.read_null()` method instead of having it happen in `ion_type()`?

There are three other raw value types besides `LazyRawBinaryValue_1_1`:
* `LazyRawBinaryValue_1_0`
* `LazyRawTextValue_1_1`
* `LazyRawTextValue_1_0`

Each of them has an `ion_type()` method and I'm pretty sure at least some of them return `IonType::Null` if the value is a null value, even if it's a typed null. The behavior you've implemented is probably more correct, but if we make this change here we'd need to make it in the other three places too.

_Originally posted by @zslayton in https://github.com/amazon-ion/ion-rust/pull/826#discussion_r1742481341_

PR #833, opened 2024-09-09
Update conformance testing DSL to support unknown symbols

nirosys:
*Issue #, if available:* n/a

*Description of changes:*
This PR adds missing symbol notation for Toplevel and Produces clauses.

This turned out to be a lot simpler than I had originally thought. My first (few) pass(es) through the spec left my understanding to be that the notation could define symbol mappings, which would have to be created during the evaluation of the fragments, and supported when evaluating expectations. Unfortunately I didn't realize my misunderstanding until after having implemented creating the symbol mappings.

The functionality now allows symbol IDs to be specified as quoted symbols, with the `#$` notation for identifying such symbol IDs.

### Details
When the Toplevel fragment is triggered to encode it's contents into a new document fragment, it uses `ProxyElement` to wrap each of the ion elements that were defined within its body. ProxyElement implements `WriteAsIon` in order to hijack any symbols being written and detect any `#$` prefixed symbols so that it can parse out the symbol table name, and offset. Once parsed, it will write the symbol ID to the document fragment.

When a Produces clause is evaluated, it will walk through its body of ion elements, while reading the concatenated ion document. As each element is evaluated for equality, any symbols found in the Produces' body will also be tested for `#$` prefix, and offsets/symtabs extracted.

In addition to Toplevel and Produces support, the Denotes clause was updated to use a reader rather than the Element API, so that it can access the symbol IDs to complete its data model tests.

### Additionally
I also used this PR to follow up on #828, and rolled the typed null handling into a `read_null` function and reset the `ion_type` back to its original for consistency. I'll follow up with a PR to return the underlying type from `ion_type` for all value implementations.

---

### DIFF
--- a/src/lazy/binary/raw/v1_1/value.rs
+++ b/src/lazy/binary/raw/v1_1/value.rs
@@ -119,13 +119,7 @@ impl<'top> HasRange for &'top LazyRawBinaryValue_1_1<'top> {
 
 impl<'top> LazyRawValue<'top, BinaryEncoding_1_1> for &'top LazyRawBinaryValue_1_1<'top> {
     fn ion_type(&self) -> IonType {
-        // Handle retrieving the type for a typed null.
-        if self.encoded_value.header.type_code() == OpcodeType::TypedNull {
-            let body = self.value_body();
-            ION_1_1_TYPED_NULL_TYPES[body[0] as usize]
-        } else {
-            self.encoded_value.ion_type()
-        }
+        self.encoded_value.ion_type()
     }
 
     fn is_null(&self) -> bool {
@@ -151,7 +145,7 @@ impl<'top> LazyRawValue<'top, BinaryEncoding_1_1> for &'top LazyRawBinaryValue_1
         }
 
         if self.is_null() {
-            return Ok(RawValueRef::Null(self.ion_type()));
+            return self.read_null();
         }
 
         match self.ion_type() {
@@ -170,6 +164,7 @@ impl<'top> LazyRawValue<'top, BinaryEncoding_1_1> for &'top LazyRawBinaryValue_1
             IonType::Struct => Ok(RawValueRef::Struct(self.read_struct()?)),
         }
     }
+
 
     /// This is a fast path for reading values that we know need to be resolved.
     ///
@@ -194,7 +189,7 @@ impl<'top> LazyRawValue<'top, BinaryEncoding_1_1> for &'top LazyRawBinaryValue_1
             return Ok(ValueRef::Int(int));
         }
         if self.is_null() {
-            return Ok(ValueRef::Null(self.ion_type()));
+            return self.read_null()?.resolve(context);
         }
         // Anecdotally, string and integer values are very common in Ion streams. This `match` creates
         // an inlineable fast path for them while other types go through the general case impl.
@@ -218,7 +213,7 @@ impl<'top> LazyRawValue<'top, BinaryEncoding_1_1> for &'top LazyRawBinaryValue_1
             }
 
             if value.is_null() {
-                return Ok(ValueRef::Null(value.ion_type()));
+                return value.read_null()?.resolve(context);
             }
 
             let value_ref =
@@ -352,6 +347,16 @@ impl<'top> LazyRawBinaryValue_1_1<'top> {
     /// contents.
     pub fn read(&'top self) -> ValueParseResult<'top, BinaryEncoding_1_1> {
         <&'top Self as LazyRawValue<'top, BinaryEncoding_1_1>>::read(&self)
+    }
+
+    fn read_null(&self) -> IonResult<RawValueRef<'top, BinaryEncoding_1_1>> {
+        let tpe = if self.encoded_value.header.type_code() == OpcodeType::TypedNull {
+            let body = self.value_body();
+            ION_1_1_TYPED_NULL_TYPES[body[0] as usize]
+        } else {
+            self.encoded_value.ion_type()
+        };
+        Ok(RawValueRef::Null(tpe))
     }
 
     pub fn is_delimited(&self) -> bool {
--- a/src/lazy/binary/raw/v1_1/value.rs
+++ b/src/lazy/binary/raw/v1_1/value.rs
@@ -145,7 +145,7 @@ impl<'top> LazyRawValue<'top, BinaryEncoding_1_1> for &'top LazyRawBinaryValue_1
         }
 
         if self.is_null() {
-            return self.read_null();
+            return Ok(RawValueRef::Null(self.read_null()?));
         }
 
         match self.ion_type() {
@@ -189,7 +189,7 @@ impl<'top> LazyRawValue<'top, BinaryEncoding_1_1> for &'top LazyRawBinaryValue_1
             return Ok(ValueRef::Int(int));
         }
         if self.is_null() {
-            return self.read_null()?.resolve(context);
+            return Ok(ValueRef::Null(self.read_null()?));
         }
         // Anecdotally, string and integer values are very common in Ion streams. This `match` creates
         // an inlineable fast path for them while other types go through the general case impl.
@@ -213,7 +213,7 @@ impl<'top> LazyRawValue<'top, BinaryEncoding_1_1> for &'top LazyRawBinaryValue_1
             }
 
             if value.is_null() {
-                return value.read_null()?.resolve(context);
+                return Ok(ValueRef::Null(value.read_null()?));
             }
 
             let value_ref =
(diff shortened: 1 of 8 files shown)


src/lazy/binary/raw/v1_1/value.rs
```diff
@@ -194,7 +189,7 @@ impl<'top> LazyRawValue<'top, BinaryEncoding_1_1> for &'top LazyRawBinaryValue_1
             return Ok(ValueRef::Int(int));
         }
         if self.is_null() {
-            return Ok(ValueRef::Null(self.ion_type()));
+            return self.read_null()?.resolve(context);
```
zslayton, line 192: The `resolve` method requires a giant `match` over all of the Ion types (see below). Since we already know it's a (possibly typed) `null`, I suggest `return ValueRef::Null(self.read_null())?;`

image

src/lazy/binary/raw/v1_1/value.rs
```diff
@@ -218,7 +213,7 @@ impl<'top> LazyRawValue<'top, BinaryEncoding_1_1> for &'top LazyRawBinaryValue_1
             }
 
             if value.is_null() {
-                return Ok(ValueRef::Null(value.ion_type()));
+                return value.read_null()?.resolve(context);
```
zslayton, line 216: Same thought here RE: `ValueRef::Null(...)`.

src/lazy/binary/raw/v1_1/value.rs
```diff
@@ -354,6 +349,16 @@ impl<'top> LazyRawBinaryValue_1_1<'top> {
         <&'top Self as LazyRawValue<'top, BinaryEncoding_1_1>>::read(&self)
     }
 
+    fn read_null(&self) -> IonResult<RawValueRef<'top, BinaryEncoding_1_1>> {
+        let tpe = if self.encoded_value.header.type_code() == OpcodeType::TypedNull {
```
zslayton, line 353:
```suggestion
        let ion_type = if self.encoded_value.header.type_code() == OpcodeType::TypedNull {
```
?

tests/conformance_dsl/continuation.rs
```diff
@@ -39,18 +39,7 @@ impl Continuation {
...
-                    Err(ConformanceErrorKind::MismatchedProduce)
-                } else {
-                    let zip = expected_elems.iter().zip(elems.iter());
-                    match zip.fold(true, |acc, (x, y)| acc && (*x == *y)) {
-                        true => Ok(()),
-                        false => Err(ConformanceErrorKind::MismatchedProduce),
-                    }
-                }
-            }
+            Continuation::Produces(produces) => produces.evaluate(ctx),
```
zslayton, line 42: Nice, this was helpful.

tests/conformance_dsl/continuation.rs
```diff
@@ -230,6 +209,66 @@ pub(crate) struct EachBranch {
     fragment: Fragment,
 }
 
+#[derive(Clone, Debug, Default)]
```
zslayton, line 212: Could add docs and a reference link to these like you have on the other types?

tests/conformance_dsl/continuation.rs
```diff
@@ -230,6 +209,66 @@ pub(crate) struct EachBranch {
...
+        let mut elem_iter = self.model.iter();
+
+        let mut is_equal = true;
+        while is_equal {
+            let (read_value, expected_element) = (reader.next()?, elem_iter.next());
+            match (read_value, expected_element) {
+                (Some(actual), Some(expected)) => is_equal = is_equal && (expected == actual),
+                (None, None) => break,
+                _ => is_equal = false,
+            }
```
zslayton, line 261: I think this is valid?
```suggestion
            is_equal = match (read_value, expected_element) {
                (Some(actual), Some(expected)) => is_equal && (expected == actual),
                (None, None) => break,
                _ => false,
            }
```

tests/conformance_dsl/fragment.rs
```diff
@@ -140,6 +141,161 @@ impl TryFrom<Sequence> for Fragment {
...
+            }
+        } else {
+            writer.write(self.0)
+        }
+    }
+
+}
+
+impl<T: ion_rs::Decoder> PartialEq<ion_rs::LazyValue<'_, T>> for ProxyElement<'_> {
+    fn eq(&self, other: &ion_rs::LazyValue<'_, T>) -> bool {
```
zslayton, line 184: This method has a lot of `unwrap()`s and `unreachable!`s--would you mind adding either comments/`expect()` text/`unreachable!` messages explaining why they shouldn't happen?

A comments pass would also be helpful.

tests/conformance_dsl/mod.rs
```diff
@@ -306,3 +306,26 @@ pub(crate) fn parse_text_exp<'a, I: IntoIterator<Item=&'a Element>>(elems: I) ->
     let val_string = bytes.iter().map(|v| unsafe { String::from_utf8_unchecked(v.to_vec()) }).collect();
     Ok(val_string)
 }
+
+/// Parses absent symbol notation from a symbol within a Toplevel fragment, or produces
+/// Continuation. The notation is '#$<id>' for an absent symbol id, or '#$<name>#<id>' for a symbol
```
zslayton, line 311: I thought `#$<id>` was the syntax for a symbol ID to be evaluated within the context being created instead of the current one. Is that what you mean by "absent"?

tests/conformance_dsl/fragment.rs
```diff
@@ -140,6 +141,161 @@ impl TryFrom<Sequence> for Fragment {
     }
 }
 
+// newtype to handle writing an Element, after we check to make sure it's not a symbol that has our
+// special absent symbol sauce.
+#[derive(Debug)]
+pub(crate) struct ProxyElement<'a>(pub &'a Element);
```
zslayton, line 147: Could you add a high-level comment to this type/file that explains the general motivation behind and process for creating/reading/writing `ProxyElement` instances?

tests/conformance_dsl/model.rs
```diff
@@ -225,31 +237,14 @@ impl PartialEq<Element> for ModelValue {
...
-                        if let Some(expected_val) = fields.get(&denoted_symtok) {
-                            if expected_val != val {
-                                return false;
-                            }
-                        }
-                    }
-                    true
-                }
-            }
+            ModelValue::Struct(_) => panic!("should not be using element eq for struct"),
```
zslayton, line 240: What is it supposed to be using?

tests/conformance_dsl/model.rs
```diff
@@ -260,6 +255,104 @@ impl PartialEq<Element> for ModelValue {
     }
 }
 
+impl PartialEq<Element> for &ModelValue {
+    fn eq(&self, other: &Element) -> bool {
+        *self == other
+    }
+}
+
+impl<T: ion_rs::Decoder> PartialEq<ion_rs::LazyValue<'_, T>> for &ModelValue {
```
zslayton, line 264: Could you do a comments pass?

tests/conformance_tests.rs
```diff
@@ -11,6 +12,21 @@ use std::str::FromStr;
 mod implementation {
     use super::*;
 
+    #[test]
+    fn test_absent_symbol() {
```
zslayton, line 16: Do we also need tests for the `#table#$id` syntax?

tests/conformance_dsl/mod.rs
nirosys: Yea, that's what I mean, that it's absent from the test document's context.

tests/conformance_dsl/fragment.rs
nirosys: I reworked this impl a bit, factored out the symbol comparison to a new `compare_symbols_eq` function that handles the absent/missing symbol notation on both sides, a little more cleanly. I want to pull the equality tests out of the PartialEq implementations, there's too many options and results to handle.

tests/conformance_dsl/model.rs
nirosys: I transitioned over to using a regular reader instead of the element API in order to access symbol IDs, so the handling of structs, lists, and s-expressions moved over to the LazyValue PartialEq. I cleaned this one up a bit and dropped a comment. Next time I touch this I'll probably get rid of the PartialEqs entirely in favor of something that can return a Result.

nirosys:
As I was reviewing everything I realized I had somehow left out the actual symbol table resolutions. I had stashed some stuff when rolling back some changes and I guess it got lost in the mix. So the latest commit adds in the building of the ion-tests catalog, and validates symbol text between the referenced symbol table, and the current context.

I also tried to document everything mentioned, and added in a couple more tests for 'absent symbols'.

tests/conformance_dsl/fragment.rs
```diff
@@ -1,3 +1,26 @@
...
+//! in the test document. In the case of `Text`, the data is concatenated to an ion text stream
+//! verbatim. The `Toplevel` fragment requires a little more work in order to support the
+//! missing/absent symbol notation. Since all of the Toplevel fragments are provided as ion
+//! literals within the test document, any symbols that is unknown in the context of the test
+//! document, or may otherwise differ in value, needs to be encoded using a notation that
+//! communicates the Symbol ID and an optional Symbol Table, to the DSL parser. When the values are
+//! read by the ion reader they're treated as normal quoted symbols. When the Toplevel fragment is
+//! encoded however, each Element read from the Toplevel clause is wrapped within a `ProxyElement`,
+//! which by implementing `WriteAsIon` is able to recursively look at all values being written to
+//! the test's input and capture any symbols that match the unknown symbol notation.
```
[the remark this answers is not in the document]
nirosys: I'll look into this more and sync up with you when I start working on #839, maybe I can sneak changes in there if it simplifies things.

PR #835, opened 2024-09-14
Adds TDL expr groups, `append_macros` system macro

zslayton:

### DIFF
--- a/src/lazy/expanded/compiler.rs
+++ b/src/lazy/expanded/compiler.rs
@@ -30,6 +30,19 @@ pub struct ExpansionAnalysis {
     pub(crate) expansion_singleton: Option<ExpansionSingleton>,
 }
 
+impl Default for ExpansionAnalysis {
+    /// By default, macros are assumed to be capable of expanding to produce multiple values,
+    /// any of which could be system values.
+    fn default() -> Self {
+        ExpansionAnalysis {
+            could_produce_system_value: true,
+            must_produce_exactly_one_value: false,
+            can_be_lazily_evaluated_at_top_level: false,
+            expansion_singleton: None,
+        }
+    }
+}
+
 impl ExpansionAnalysis {
     pub fn could_produce_system_value(&self) -> bool {
         self.could_produce_system_value
@@ -146,7 +159,7 @@ impl TemplateCompiler {
         let mut reader = Reader::new(v1_1::Text, expression.as_bytes())?;
         let macro_def_sexp = reader.expect_next()?.read()?.expect_sexp()?;
 
-        Self::compile_from_sexp(context, &MacroTable::default(), macro_def_sexp)
+        Self::compile_from_sexp(context, &MacroTable::empty(), macro_def_sexp)
     }
 
     /// Pulls the next value from the provided source and confirms that it is a symbol whose
@@ -493,6 +506,17 @@ impl TemplateCompiler {
             /*target_parameter=*/ None, /*; this is not a macro invocation arg.*/
             body,
         )?;
+        // Make sure there aren't any unexpected expressions following the body.
+        match values.next() {
+            None => {}
+            Some(expr) => {
+                let name = template_name.unwrap_or_else(|| "<anonymous>".into());
+                return IonResult::decoding_error(format!(
+                    "found unexpected expression following the body of macro '{name}': {:?}",
+                    expr?
+                ));
+            }
+        }
         let template_macro = TemplateMacro {
             name: template_name,
             signature,
@@ -565,6 +589,9 @@ impl TemplateCompiler {
     /// [`TemplateBodyExpr`] sequences to the `TemplateBody`.
     ///
     /// If `is_literal` is true, nested symbols and s-expressions will not be interpreted.
+    ///
+    /// If this value is being passed to a macro as an argument, `target_parameter` will be the
+    /// parameter to which the argument corresponds.
     fn compile_value<D: Decoder>(
         tdl_context: TdlContext,
         definition: &mut TemplateBody,
@@ -817,7 +844,28 @@ impl TemplateCompiler {
         // Collect the remaining argument expressions into a (probably) stack-allocated SmallVec.
         let arguments: SmallVec<[LazyValue<'_, D>; 2]> =
             arguments.collect::<IonResult<SmallVec<_>>>()?;
-        debug_assert!(!arguments.is_empty());
+
+        // If there aren't any more arguments, make sure `param` will accept `none`.
+        if arguments.is_empty() {
+            return if param.accepts_none() {
+                // ...and then insert a placeholder `none` invocation.
+                Self::compile_macro(
+                    tdl_context,
+                    definition,
+                    tdl_context.context.none_macro(),
+                    std::iter::empty::<Result<LazyValue<'_, D>, IonError>>(),
+                )
+            } else {
+                IonResult::decoding_error(
+                    format!(
+                        "missing argument in call to macro '{}'; final parameter '{}' has cardinality '{:?}' and cannot be omitted",
+                        invoked_macro_name.unwrap_or("<anonymous>"),
+                        param.name(),
+                        param.cardinality()
+                    )
+                )
+            };
+        }
 
         // If it turns out there was only one expression, we can simply compile it like we would
         // any other argument.
@@ -884,7 +932,7 @@ impl TemplateCompiler {
         Ok(())
     }
 
-    /// Adds a `lazy_sexp` that has been determined to represent a macro invocation to the
+    /// Adds a `lazy_sexp` that has been determined to represent an expression group to the
     /// TemplateBody.
     fn compile_arg_expr_group<'a, D: Decoder>(
         tdl_context: TdlContext,
@@ -893,9 +941,9 @@ impl TemplateCompiler {
         parameter: Parameter,
     ) -> IonResult<()> {
         let expr_group_start_index = definition.expressions.len();
-        // Assume the macro contains zero argument expressions to start, we'll update
-        // this at the end of the function.
-        definition.push_macro_invocation(tdl_context.context.values_macro(), ExprRange::empty());
+        // Put a placeholder expression in the template body. We'll overwrite it at the end of this
+        // method when we have more information.
+        definition.push_placeholder();
         for argument_result in expressions {
             let argument = argument_result?;
             Self::compile_value(
@@ -912,7 +960,7 @@ impl TemplateCompiler {
         let is_none = expr_group_start_index == arguments_end - 1;
 
         // Confirm that this was a legal expression for the corresponding parameter.
-        let macro_ref = if is_none {
+        if is_none {
             if !parameter.accepts_none() {
                 return IonResult::decoding_error(format!(
                     "parameter '{}' has cardinality {:?}; it does not accept empty argument groups: `(;)`",
@@ -920,23 +968,19 @@ impl TemplateCompiler {
                     parameter.cardinality()
                 ));
             }
-            tdl_context.context.none_macro()
-        } else {
-            if !parameter.accepts_multi() {
-                return IonResult::decoding_error(format!(
-                    "parameter '{}' has cardinality {:?}; it does not accept argument groups",
-                    parameter.name(),
-                    parameter.cardinality()
-                ));
-            }
-            tdl_context.context.values_macro()
-        };
+        } else if !parameter.accepts_multi() {
+            return IonResult::decoding_error(format!(
+                "parameter '{}' has cardinality {:?}; it does not accept populated argument groups",
+                parameter.name(),
+                parameter.cardinality()
+            ));
+        }
 
         // update the macro step to reflect the number of child expressions it
         // contains
         let invocation_expr_range = ExprRange::new(expr_group_start_index..arguments_end);
         definition.expressions[expr_group_start_index] =
-            TemplateBodyExpr::arg_expr_group(parameter, macro_ref, invocation_expr_range);
+            TemplateBodyExpr::arg_expr_group(parameter, invocation_expr_range);
         Ok(())
     }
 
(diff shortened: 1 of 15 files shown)


src/lazy/expanded/compiler.rs
```diff
@@ -408,7 +420,7 @@ impl TemplateCompiler {
 
         // TODO: Enforce 'identifier' syntax subset of symbol
         let template_name =
-            Self::expect_nullable_name("a macro name", &mut values)?.map(|name| name.to_owned());
+            Self::expect_nullable_name("a macro name", &mut values)?.map(|name| name.into());
```
zslayton, line 423: 🗺️ Macro names are now `Rc<str>` instead of `String`. This allows them to be cheaply cloned and shrinks the stack size of the name from 24 bytes (`String`) to 16 bytes (`Rc<str>`).

src/lazy/expanded/compiler.rs
```diff
@@ -491,8 +503,20 @@ impl TemplateCompiler {
             tdl_context,
             &mut compiled_body,
             /*is_literal=*/ false,
+            /*target_parameter=*/ None, /*; this is not a macro invocation arg.*/
             body,
         )?;
+        // Make sure there aren't any unexpected expressions following the body.
```
zslayton, line 509: Previously, expressions following the body would be ignored. Now this raises an error.

src/lazy/expanded/e_expression.rs
```diff
@@ -56,35 +44,40 @@ impl<'top, D: Decoder> ArgGroup<'top, D> {
...
-    pub fn invoked_macro(&self) -> MacroRef<'top> {
-        self.invoked_macro
-    }
-    pub fn expressions(&self) -> ArgGroupIterator<'top, D> {
-        ArgGroupIterator::new(self.context, self.raw_arg_group())
+    pub fn expressions(&self) -> EExpArgGroupIterator<'top, D> {
+        EExpArgGroupIterator::new(self.context, self.raw_arg_group())
     }
-    pub fn expand(&self, environment: Environment<'top, D>) -> IonResult<MacroExpansion<'top, D>> {
+    pub fn expand(&self) -> IonResult<MacroExpansion<'top, D>> {
```
zslayton, line 50: 🗺️ Several types that previously required the `Environment` to be passed into their methods now capture the relevant `Environment` at construction time.

src/lazy/expanded/e_expression.rs
```diff
@@ -56,35 +44,40 @@ impl<'top, D: Decoder> ArgGroup<'top, D> {
...
 
-impl<'top, D: Decoder> Debug for ArgGroup<'top, D> {
+impl<'top, D: Decoder> Debug for EExpArgGroup<'top, D> {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        write!(f, "ArgGroup {:?}", self.raw_arg_group)
+        write!(f, "(: {:?}", self.raw_arg_group)?;
+        for expr in self.expressions() {
+            write!(f, " {:?}", expr)?;
+        }
+        write!(f, ")")
```
zslayton, line 80: 🗺️ The derived `Debug` impl gets very noisy if there are several layers of nesting. I rewrote some of these to make the output more compact and human friendly.

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -134,35 +141,53 @@ impl<'top, D: Decoder> MacroExpr<'top, D> {
...
 
-    pub fn expand(&self, environment: Environment<'top, D>) -> IonResult<MacroExpansion<'top, D>> {
+    pub fn expand(&self) -> IonResult<MacroExpansion<'top, D>> {
         match self.kind {
-            MacroExprKind::TemplateMacro(t) => t.expand(environment),
+            MacroExprKind::TemplateMacro(t) => t.expand(),
+            MacroExprKind::TemplateArgGroup(g) => g.expand(),
             MacroExprKind::EExp(e) => e.expand(),
-            MacroExprKind::EExpArgGroup(g) => g.expand(environment),
+            MacroExprKind::EExpArgGroup(g) => g.expand(),
```
zslayton, line 149: 🗺️ As mentioned prior, each of these types now captures a reference to the correct `Environment` and no longer needs one to be passed in.

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -134,35 +141,53 @@ impl<'top, D: Decoder> MacroExpr<'top, D> {
...
 }
 
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone)]
 pub enum MacroExprKind<'top, D: Decoder> {
     /// A macro invocation found in the body of a template.
-    TemplateMacro(TemplateMacroInvocation<'top>),
+    TemplateMacro(TemplateMacroInvocation<'top, D>),
+    /// An argument group found in the body of a template.
+    TemplateArgGroup(TemplateExprGroup<'top, D>),
```
zslayton, line 159: 🗺️ A new variant representing an expression group in a template body.

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -1164,36 +1200,23 @@ impl<'top> TemplateExpansion<'top> {
...
         context: EncodingContextRef<'top>,
         environment: Environment<'top, D>,
     ) -> IonResult<MacroExpansionStep<'top, D>> {
         let expressions = self.template.body().expressions();
-        let value_expr = match expressions.get(self.step_index) {
+        let value_tdl_expr = match expressions.get(self.step_index) {
             None => return Ok(MacroExpansionStep::FinalStep(None)),
             Some(expr) => expr,
         };
+        let value_expr = value_tdl_expr.to_value_expr(context, environment, self.template);
```
zslayton, line 1217: 🗺️ I factored out the `match` logic needed to transform a `TemplateBodyExpr` to a `ValueExpr`, which simplified expansion code in a number of places including this one.

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -1206,30 +1229,20 @@ impl<'top> TemplateExpansion<'top> {
 mod tests {
     use crate::{v1_1, ElementReader, Int, IonResult, MacroTable, Reader};
 
-    /// Reads `input` and `expected` using an expanding reader and asserts that their output
-    /// is the same.
-    fn eval_enc_expr<'data>(input: &'data str, expected: &'data str) -> IonResult<()> {
-        let mut actual_reader = Reader::new(v1_1::Text, input.as_bytes())?;
+    /// Reads `input` and `expected` and asserts that their output is Ion-equivalent.
+    fn stream_eq<'data>(input: &'data str, expected: &'data str) -> IonResult<()> {
```
zslayton, line 1233: 🗺️ Now that the `Reader` understands encoding directives and we have an `:append_macros` system macro, we no longer need to manually register individual macros by source.

src/lazy/expanded/macro_table.rs
```diff
@@ -100,7 +112,7 @@ impl Macro {
 #[derive(Debug, Clone, PartialEq)]
 pub enum MacroKind {
     Void,
-    Values,
+    ExprGroup,
```
zslayton, line 115: As you'll see below, `values` is now a simple `TemplateMacro` that uses `ExprGroup`.

src/lazy/expanded/struct.rs
```diff
@@ -256,48 +255,13 @@ impl<'top, D: Decoder> LazyExpandedStruct<'top, D> {
...
-                            self.context,
-                            element.template(),
-                            first_result_expr.expr_range(),
-                        );
-                        let mut evaluator = MacroEvaluator::new_with_environment(*environment);
-                        let expansion =
-                            MacroExpansion::initialize(*environment, invocation.into())?;
-                        evaluator.push(expansion);
+                let value_expr =
+                    first_result_expr.to_value_expr(self.context, *environment, element.template());
```
zslayton, line 259: 🗺️ Simplified by `to_value_expr`.

src/lazy/expanded/template.rs
```diff
@@ -415,39 +431,15 @@ impl<'top, D: Decoder> Iterator for TemplateSequenceIterator<'top, D> {
...
-                    // macro being invoked.
-                    let invoked_macro = body_invocation.invoked_macro.as_ref();
-                    let invocation = TemplateMacroInvocation::new(
-                        self.context,
-                        self.template,
-                        invoked_macro,
-                        ExprRange::new(current_expr.expr_range().tail()),
-                    );
+
+            let value_expr = current_expr.to_value_expr(self.context, environment, self.template);
```
zslayton, line 435: 🗺️ Simplified by `to_value_expr`.

src/lazy/expanded/template.rs
```diff
@@ -679,6 +644,53 @@ impl TemplateBodyExpr {
     pub fn expr_range(&self) -> ExprRange {
         self.expr_range
     }
+
+    pub fn to_value_expr<'top, D: Decoder>(&'top self,
+                                           context: EncodingContextRef<'top>,
+                                           environment: Environment<'top, D>,
+                                           host_template: TemplateMacroRef<'top>
```
zslayton, line 651: 🗺️ The factored-out logic.

src/lazy/expanded/template.rs
```diff
@@ -1095,37 +1238,8 @@ impl<'top, D: Decoder> Iterator for TemplateMacroInvocationArgsIterator<'top, D>
...
-            TemplateBodyExprKind::MacroInvocation(body_invocation) => {
-                let invocation = body_invocation.resolve(
-                    self.context,
-                    self.host_template,
-                    arg.expr_range(),
-                );
-                ValueExpr::MacroInvocation(invocation.into())
-            }
-        };
+        let arg_expr = arg.to_value_expr(self.context, self.environment, self.host_template);
```
zslayton, line 1242: 🗺️ Simplified by `to_value_expr`.

zslayton (commented):
🗺️ PR tour 🧭

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -1353,6 +1366,94 @@ mod tests {
...
+            r#"
+            "Hello, Gary"
+        "#,
+        )
+    }
+
+    #[test]
+    fn built_in_append_macros() -> IonResult<()> {
+        stream_eq(
+            r#"
```
popematt, line 1423: Can you add a symbol table at the start (and reference it later) to make sure that `append_macros` doesn't clobber the symbol table?

src/lazy/expanded/macro_table.rs
```diff
@@ -158,9 +170,23 @@ impl<'top> MacroRef<'top> {
...
+    /// An instance of the Ion 1.1 system macro table is lazily instantiated once per thread,
+    /// minimizing the number of times macro compilation occurs.
+    ///
+    /// The thread-local instance holds `Rc` references to its macro names and macro definitions,
+    /// making its contents inexpensive to `clone()` and reducing the number of duplicate `String`s
+    /// being allocated over time.
+    ///
+    /// Each time the user constructs a reader, it clones the thread-local copy of the system macro
+    /// table.
+    pub static ION_1_1_SYSTEM_MACROS: MacroTable = MacroTable::construct_system_macro_table();
```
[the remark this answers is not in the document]
zslayton: > Ideally, construct_system_macro_table would be constant, though I understand that might not be possible make it const yet.

Right, constructing a heap-allocated collection is not yet possible in `const`.

> Why does the system macro table need to be cloned rather than borrowed?

(Bit of a journey here with several things you already know, bear with me.)

Each time a `Reader` is created or an `$ion_1_1` is encountered, the macro table needs to be initialized to contain the macros in `$ion`. The macros in `$ion` can also be re-exported at any time in an encoding directive. This means that in any stream, we'll end up building/using `$ion` _at least_ once.

Each macro represents some number of heap-allocated resources--names, values, annotations, etc.

In `ion-rust`, the active macro table is 'flat' in the sense that it doesn't store a tree of other tables to search; it just copies the other tables' macros into its own index/storage. To avoid needlessly performing deep copies, it stores `Rc` references to the already-constructed `Macro` instances.

If we had a static `OnceCell` holding a collection of `Macro`, the `OnceCell` could hand out `&'static Macro`. However, this would make it a different data type than the `MacroTable` we're using everywhere else. This means that the machinery for re-exporting `$ion` would need to be different from the machinery for re-exporting `$ion_encoding` or any user-defined module.

In order for the shared `$ion` macro table to hold `Rc` references, it needs to be tied to the current thread--this brings us to `thread_local`. (Note that `thread_local` is also lazily initialized.)

In aggregate this means that once per thread the system constructs/allocates a copy of the system macro table and then re-uses it across all `Reader` instantiations, `$ion_1_1`s, and re-exports for the rest of that thread and (importantly IMO) `$ion` is treated as just another module.

Thoughts?

src/lazy/expanded/mod.rs
```diff
@@ -182,6 +182,26 @@ impl<'top> EncodingContextRef<'top> {
...
+
+    pub fn system_symbol_table(&self) -> &'top SymbolTable {
+        self.context.system_module.symbol_table()
+    }
+
+    pub fn system_macro_table(&self) -> &'top MacroTable {
+        self.context.system_module.macro_table()
+    }
+
+    pub(crate) fn none_macro(&self) -> Rc<Macro> {
```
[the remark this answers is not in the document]
zslayton: This is an internal API, so we have the freedom to change it in the future if we'd like to.

This gives out an `Rc<Macro>` so a template can refer to the macro in question and guarantee it's around as long as necessary. If this method gave out a `&Macro`, the caller would need to turn it into an `Rc<Macro>` via heap allocation.

src/lazy/expanded/macro_table.rs
```diff
@@ -285,16 +318,89 @@ impl MacroTable {
...
+                body: TemplateBody {
+                    expressions: vec![
+                        // The `$ion_encoding::(...)` s-expression
+                        TemplateBodyExpr::element(
+                            TemplateBodyElement::with_value(TemplateValue::SExp)
+                                // Has the first annotation in annotations storage below; `$ion_encoding`
+                                .with_annotations(0..1),
+                            // Contains expressions 0 (itself) through 5 (exclusive)
+                            ExprRange::new(0..5),
+                        ),
```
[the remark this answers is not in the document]
zslayton: Good catch, thanks.

src/lazy/expanded/macro_table.rs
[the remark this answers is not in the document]
popematt: That was a bit of a journey—but it makes sense.

popematt (approved):

PR #842, opened 2024-10-10
Put Ion 1.1 support behind a feature flag

popematt:
**Issue #, if available:**

**Description of changes:**

I tried to make this as minimally invasive as possible:
* The `experimental-ion-1-1` feature enables`experimental-reader-writer`
* The `experimental` feature flag enables `experimental-ion-1-1`
* When `experimental-ion-1-1` is not enabled
  * the `v1_1` module is only `pub(crate)`
  * the decoder doesn't have a match arm for the `(1, 1)` version.
  * the tests for switching versions have a branch that checks that an `Err` is returned for an Ion 1.1 IVM
  * the conformance dsl tests and a number of Ion 1.1 specific tests are disabled (omitted from compilation, really)
* When `experimental-ion-1-1` is enabled, all APIs and behaviors are as they were prior to this PR

I also updated the build workflow to _not_ cancel all outstanding jobs if one job in the build matrix fails. It's a lot quicker to identify issues if you can see that e.g. the tests pass for certain feature sets but not for others.

### DIFF
--- a/src/lazy/any_encoding.rs
+++ b/src/lazy/any_encoding.rs
@@ -1860,6 +1860,11 @@ mod tests {
 
         expect_int(context_ref, &mut reader, IonEncoding::Text_1_0, 2)?;
 
+        if cfg!(not(feature = "experimental-ion-1-1")) {
+            reader.next(context_ref).expect_err("Ion 1.1 IVM should return an error.");
+            return Ok(())
+        }
+
         // This IVM changes the encoding from 1.0 text to 1.1 text
         expect_version_change(
             context_ref,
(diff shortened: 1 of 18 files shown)


src/lazy/any_encoding.rs
```diff
@@ -1860,6 +1860,11 @@ mod tests {
 
         expect_int(context_ref, &mut reader, IonEncoding::Text_1_0, 2)?;
 
+        if cfg!(not(feature = "experimental-ion-1-1")) {
```
zslayton, line 1863: TIL about the `cfg` macro.

src/lazy/decoder.rs
```diff
@@ -125,6 +125,7 @@ pub trait RawVersionMarker<'top>: Debug + Copy + Clone + HasSpan<'top> {
     fn stream_version_after_marker(&self) -> IonResult<IonVersion> {
         match self.major_minor() {
             (1, 0) => Ok(IonVersion::v1_0),
+            #[cfg(feature = "experimental-ion-1-1")]
```
zslayton, line 128: TIL you could apply `#[cfg(...)]` annotations to a match arm.

zslayton (approved):

PR #858, opened 2024-11-25
API improvements for usage of owned `Element`s

jpschorr:
This PR includes a couple changes intended to make working with owned `Element`s easier are not require cloning for 'projecting' an element to be replaced by something it contains.

- Exposes `OwnedSequenceIterator` and modifies its implementation to clone less
- Adds `try_into_<x>` methods similar to `as_<x>` methods on `Element`
    -  These allow 'partial' conversions of owned `Element`s with the 'failure' case returning the original `Element`

Example simplified usage:
```rust
#[derive(Debug)]
enum ElementIterator {
    Single(Element),
    Sequence(OwnedSequenceIterator),
}

...

match elt.try_into_sequence() {
    Err(elt) => ElementIterator::Single(elt),
    Ok(seq) => ElementIterator::Sequence(seq.into_iter()),
}
```

---

### DIFF
--- a/src/element/mod.rs
+++ b/src/element/mod.rs
@@ -886,13 +886,13 @@ impl Element {
 
 impl IonTypeExpectation for Element {
     fn ion_type(&self) -> IonType {
-        self.ion_type()
+        Element::ion_type(self)
     }
 }
 
 impl IonTypeExpectation for &Element {
     fn ion_type(&self) -> IonType {
-        self.ion_type()
+        Element::ion_type(self)
     }
 }
 
--- a/src/element/mod.rs
+++ b/src/element/mod.rs
@@ -1,5 +1,13 @@
 // Copyright Amazon.com, Inc. or its affiliates.
 
+// Disabled for this file because `Element`'s `try_into_<x>` methods return
+// `ConversionOperationError<Element, <X>>` which embeds an `Element`. `Element` itself is 128 bytes.
+// This trips `clippy::result_large_err`. However, most usages of this type are expected to be
+// converted or destructured immediately.
+//
+// Cf. https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err
+#![allow(clippy::result_large_err)]
+
 //! Provides a dynamically typed, materialized representation of an Ion value.
 //!
 //! An [Element] represents an `(annotations, value)` pair, where a `value` is
--- a/src/element/mod.rs
+++ b/src/element/mod.rs
@@ -12,12 +12,12 @@
 //! [simd-json-value]: https://docs.rs/simd-json/latest/simd_json/value/index.html
 //! [serde-json-value]: https://docs.serde.rs/serde_json/value/enum.Value.html
 
+pub use annotations::{Annotations, IntoAnnotations};
+pub use sequence::{OwnedSequenceIterator, Sequence};
 use std::cmp::Ordering;
 use std::fmt::{Display, Formatter};
 use std::io;
-
-pub use annotations::{Annotations, IntoAnnotations};
-pub use sequence::Sequence;
+use std::ops::ControlFlow;
 
 use crate::{ion_data, Decimal, Int, IonError, IonResult, IonType, Str, Symbol, Timestamp};
 use crate::{Blob, Bytes, Clob, List, SExp, Struct};
@@ -31,6 +31,7 @@ use crate::lazy::reader::Reader;
 use crate::lazy::streaming_raw_reader::{IonInput, IonSlice};
 use crate::result::IonFailure;
 use crate::text::text_formatter::FmtValueFormatter;
+use crate::types::symbol::SymbolText;
 use crate::write_config::WriteConfig;
 
 mod annotations;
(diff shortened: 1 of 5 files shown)


src/element/mod.rs
```diff
@@ -459,6 +504,13 @@ impl Element {
         self.as_int().ok_or_else(|| self.expected(IonType::Int))
     }
 
+    pub fn try_into_int(self) -> ControlFlow<Element, Int> {
```
zslayton, line 507: I think you're using `ControlFlow` everywhere to avoid needing to define another error type to hand back the `Element` in the error case. Is that right? If so, I think there's enough methods like this that we could justify a `CastResult` or `ConversionResult` to make things more self-explanatory.

zslayton (commented):
This looks good, thanks! One question about our error types before we merge this.

src/element/mod.rs
jpschorr: Essentially, yes. I'm not wedded to it's use and open to replace it.

Using something like `Result<Sequence, Element>` for the `try_into_sequence` didn't feel quite right, as using `?` wasn't quite as expected.

`ControlFlow` has `FromResidual` experimental implementation. But I suppose that could always be added for a custom type.

src/element/mod.rs
```diff
@@ -348,6 +349,50 @@ impl std::fmt::Debug for Element {
...
+impl TryFrom<Element> for Sequence {
+    type Error = IonError;
+
+    fn try_from(element: Element) -> Result<Sequence, Self::Error> {
+        match element.try_into_sequence() {
+            ControlFlow::Continue(sequence) => Ok(sequence),
+            ControlFlow::Break(element) => Err(element.expected("sequence value")),
+        }
+    }
+}
```
popematt, line 361: You could do something like this, to DRY some of the boilerplate and also implement `TryFrom<Element>` for more types so that all Ion types are covered. (I think it's a good idea to add it for all of them so that we're not asking later why we didn't do that. The codegen project, in particular, really benefits from having consistent APIs that cover all the types.)
```rust
macro_rules! impl_try_from_element {
    ($to_type:ident, $to_fn:ident, $to_desc:expr) => {
        impl TryFrom<Element> for $to_type {
            type Error = IonError;

            fn try_from(element: Element) -> Result<$to_type, Self::Error> {
                match element.$to_fn() {
                    ControlFlow::Continue(structure) => Ok(structure),
                    ControlFlow::Break(element) => Err(element.expected($to_desc)),
                }
            }
        }
    }
}

impl_try_from_element!(Struct, try_into_struct, "struct value")
impl_try_from_element!(Sequence, try_into_sequence, "sequence value")
impl_try_from_element!(String, try_into_text, "text value")
impl_try_from_element!(Str, try_into_string, "string value")
impl_try_from_element!(Symbol, try_into_symbol, "symbol value")
impl_try_from_element!(bool, try_into_bool, "boolean value")
impl_try_from_element!(f64, try_into_float, "float value")
impl_try_from_element!(Decimal, try_into_decimal, "decimal value")
// etc.
```

src/result/conversion.rs
```diff
@@ -0,0 +1,144 @@
+use crate::{Bytes, Decimal, Int, IonType, Sequence, Str, Struct, Symbol, Timestamp};
+use thiserror::Error;
+
+pub type ConversionResult<T> = Result<T, ConversionError>;
```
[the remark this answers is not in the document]
jpschorr: `ConversionError<FromType>` can't be put directly into `IonError` due to the generic.

```rust
pub enum IonError {
   ...
    #[error("{0}")]
    Conversion(#[from] ConversionError<FromType>),
   ...
}
```

So you end up needing two 'error' structs anyway.

I originally had something like:

```rust
/// Represents a mismatch during conversion between the expected type and the actual value's type.
#[derive(Clone, Debug, Error, PartialEq)]
#[error("Type conversion error; expected a(n) {output_type}, found a(n) {input_type}")]
pub struct ConversionError {
    input_value: String,
    output_type: String,
}

/// Represents a mismatch during conversion between the expected type and the actual value's type.
#[derive(Clone, Debug, Error, PartialEq)]
#[error("Type conversion error; expected a(n) {output_type}, found a(n) {input_type}")]
pub struct ConversionOperationError<FromType>
where
    FromType: ValueTypeExpectation,
{
    input_value: FromType,
    output_type: String,
}
```

So, a pair of `Error`s and a pair of `Result`s. Except the `ConversionOperationError`/`ConversionOperationResult` basically never existed for long and wasn't really used as an error; It was either unwrapped for its success or fail element or converted immediately into a `ConversionError`/`ConversionResult`. On top of this, I kept confusing myself as to which was which between `Conversion<x>` and `ConversionOperation<x>`.

So, I condensed `ConversionOperationError`&`ConversionOperationResult` and into `Conversion`.  However, if you find that more surprising and would prefer the pair of `Error`s/`Result`s or have other structure or naming suggestions, I'm open to change it.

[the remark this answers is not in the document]
jpschorr: I've refactored to a pair of `Error`s & `Result`s.

src/result/conversion.rs
```diff
@@ -0,0 +1,119 @@
...
+/// Holds the original value that was not able to be converted.
+#[derive(Clone, Debug, Error, PartialEq)]
+#[error("Type conversion error; expected a(n) {}, found a(n) {}",
+            ToType::expected_type_name(), .input_value.expected_type_name())]
+pub struct ConversionOperationError<FromType, ToType>
+where
+    FromType: ValueTypeExpectation,
+    ToType: TypeExpectation,
+{
+    input_value: Box<FromType>,
```
zslayton, line 41: Why does this need to be boxed? Are you concerned about stack size?

jpschorr: Without boxing, clippy warns:

```shell
warning: the `Err`-variant returned from this function is very large
   --> src/element/mod.rs:492:34
    |
492 |     pub fn try_into_int(self) -> ConversionOperationResult<Element, Int> {
    |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the `Err`-variant is at least 128 bytes
    |
    = help: try reducing the size of `result::conversion::ConversionOperationError<element::Element, types::integer::Int>`, for example by boxing large elements or replacing it with `Box<result::conversion::ConversionOperationError<element::Element, types::integer::Int>>`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err
    = note: `#[warn(clippy::result_large_err)]` on by default
```

zslayton (approved):

PR #862, opened 2024-11-29
Adds read/write support for system eexps, placeholders for remaining system macros

zslayton:
Builds on PR #861.

* `MacroTable` now implements `Send`, allowing instances to be shared across threads. This made it possible for the system symbol table to be a singleton for all readers. To achieve this:
   * `Rc` references to other `Macro` instances are now `Arc`. This incurs higher overhead for new macro definitions at compile time, but no overhead at usage time. `Reader` and `Writer` instantiation benefit by not having to create a new system macro table instance.
   * `Rc<str>` instances used to store parameter and macro names have been replaced by [`CompactString`](https://github.com/ParkMyCar/compact_str?tab=readme-ov-file#compact_str), which doesn't allocate for strings that are `<=24` bytes long.
* Text read/write support for `$ion::<addr>` and `$ion::<name>` e-expressions.
* Binary read/write support for `0xEF` system e-expressions.
* Stubs out macros for all of the yet-to-be-implemented system macros, so all macros now have the expected address.

### DIFF
--- a/benches/read_many_structs.rs
+++ b/benches/read_many_structs.rs
@@ -49,8 +49,7 @@ fn maximally_compact_1_1_data(num_values: usize) -> TestData_1_1 {
 
     let mut binary_1_1_data = vec![0xE0u8, 0x01, 0x01, 0xEA]; // IVM
     #[rustfmt::skip]
-    let mut binary_1_1_data_body: Vec<u8> = vec![
-        MacroTable::FIRST_USER_MACRO_ID as u8, // Macro ID
+    let mut binary_1_1_data_body: Vec<u8> = [MacroTable::FIRST_USER_MACRO_ID as u8, // Macro ID
         0b10, // [NOTE: `0b`] `parameters*` arg is an arg group
         0x66, // 6-byte integer (`timestamp` param)
         0x75, 0x5D, 0x63, 0xEE, 0x84, 0x01,
@@ -73,8 +72,7 @@ fn maximally_compact_1_1_data(num_values: usize) -> TestData_1_1 {
         0x32, 0x30, 0x3A, 0x35,
         0x39, 0x3A, 0x35, 0x39,
         0x2E, 0x37, 0x34, 0x34,
-        0x30, 0x30, 0x30, 0x5A,
-    ].repeat(num_values);
+        0x30, 0x30, 0x30, 0x5A].repeat(num_values);
     binary_1_1_data.append(&mut binary_1_1_data_body);
     TestData_1_1 {
         name: "maximally compact".to_owned(),
@@ -111,8 +109,7 @@ fn moderately_compact_1_1_data(num_values: usize) -> TestData_1_1 {
     let text_1_1_data = r#"(:event 1670446800245 418 "scheduler-thread-6" "example-client-1" "aws-us-east-5f-abc123" (: "region 4" "2022-12-07T20:59:59.744000Z"))"#.repeat(num_values);
     let mut binary_1_1_data = vec![0xE0u8, 0x01, 0x01, 0xEA]; // IVM
     #[rustfmt::skip]
-    let mut binary_1_1_data_body: Vec<u8> = vec![
-        MacroTable::FIRST_USER_MACRO_ID as u8, // Macro ID
+    let mut binary_1_1_data_body: Vec<u8> = [MacroTable::FIRST_USER_MACRO_ID as u8, // Macro ID
         0b10, // [NOTE: `0b` prefix] `parameters*` arg is an arg group
         0x66, // 6-byte integer (`timestamp` param)
         0x75, 0x5D, 0x63, 0xEE, 0x84, 0x01,
@@ -143,8 +140,7 @@ fn moderately_compact_1_1_data(num_values: usize) -> TestData_1_1 {
         0x32, 0x30, 0x3A, 0x35,
         0x39, 0x3A, 0x35, 0x39,
         0x2E, 0x37, 0x34, 0x34,
-        0x30, 0x30, 0x30, 0x5A,
-    ].repeat(num_values);
+        0x30, 0x30, 0x30, 0x5A].repeat(num_values);
 
     binary_1_1_data.append(&mut binary_1_1_data_body);
     TestData_1_1 {
@@ -182,8 +178,7 @@ fn length_prefixed_moderately_compact_1_1_data(num_values: usize) -> TestData_1_
     let text_1_1_data = r#"(:event 1670446800245 418 "scheduler-thread-6" "example-client-1" "aws-us-east-5f-abc123" (: "region 4" "2022-12-07T20:59:59.744000Z"))"#.repeat(num_values);
     let mut binary_1_1_data = vec![0xE0u8, 0x01, 0x01, 0xEA]; // IVM
     #[rustfmt::skip]
-    let mut binary_1_1_data_body: Vec<u8> = vec![
-        0xF5, // LP invocation
+    let mut binary_1_1_data_body: Vec<u8> = [0xF5, // LP invocation
         ((MacroTable::FIRST_USER_MACRO_ID * 2) + 1) as u8, // Macro ID
         0xDF, // Length prefix: FlexUInt 111
         0b10, // [NOTE: `0b` prefix] `parameters*` arg is an arg group
@@ -216,8 +211,7 @@ fn length_prefixed_moderately_compact_1_1_data(num_values: usize) -> TestData_1_
         0x32, 0x30, 0x3A, 0x35,
         0x39, 0x3A, 0x35, 0x39,
         0x2E, 0x37, 0x34, 0x34,
-        0x30, 0x30, 0x30, 0x5A,
-    ].repeat(num_values);
+        0x30, 0x30, 0x30, 0x5A].repeat(num_values);
 
     binary_1_1_data.append(&mut binary_1_1_data_body);
     TestData_1_1 {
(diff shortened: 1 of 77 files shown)


src/lazy/expanded/compiler.rs
```diff
@@ -379,7 +413,7 @@ impl TemplateCompiler {
         context: EncodingContextRef<'_>,
         pending_macros: &'a MacroTable,
         macro_id: impl Into<MacroIdRef<'a>>,
-    ) -> Option<Rc<Macro>> {
+    ) -> Option<Arc<Macro>> {
```
zslayton, line 416: 🗺️ Macros now reference each other with `Arc`, allowing the system `MacroTable` to be a lazily initialized singleton shared across all readers. The speedup offered by not having to allocate/copy the system table should outweigh the cost of creating `Arc`s.

There is no overhead when dereferencing an `Arc` (it's as fast as `&`), only when creating or destroying one.

src/lazy/expanded/e_expression.rs
```diff
@@ -127,6 +127,7 @@ impl<'top, D: Decoder> EExpression<'top, D> {
                 environment = self.new_evaluation_environment()?;
                 MacroExpansionKind::Template(TemplateExpansion::new(template_ref))
             }
+            MacroKind::ToDo => todo!("system macro {}", invoked_macro.name().unwrap()),
```
zslayton, line 130: 🗺️ All of the not-yet-implemented system macros now have the `MacroKind` `ToDo` and will panic if used. This allowed me to insert placeholders for the remaining macro definitions, which aligned the existing macro definitions to the expected addresses.

src/lazy/expanded/macro_table.rs
```diff
@@ -171,23 +174,17 @@ impl<'top> MacroRef<'top> {
...
-    /// Each time the user constructs a reader, it clones the thread-local copy of the system macro
-    /// table.
-    pub static ION_1_1_SYSTEM_MACROS: MacroTable = MacroTable::construct_system_macro_table();
-}
+/// A lazily initialized singleton instance of the Ion 1.1 system macro table.
+///
+/// This instance is shared by all readers, minimizing the number of times that macro compilation
+/// and heap allocation occurs.
+pub static ION_1_1_SYSTEM_MACROS: LazyLock<MacroTable> =
+    LazyLock::new(MacroTable::construct_system_macro_table);
```
zslayton, line 187: 🗺️ Originally, each reader or writer would instantiate its own copy of the system macro table. This involved several heap allocations and some compilation logic that could feel quite heavyweight for a short-lived stream.

To avoid paying this cost for every reader and writer, I made the system macro table a thread-local value. Only the first reader or writer on a thread would pay the initialization cost, and `Macro` and `Parameter` instances could continue using lightweight `Rc`s to share data within the thread.

However, this introduced a problem: unlike `static` values that live for the duration of the program, thread locals only live for the duration of their thread; they do not have a named lifetime.  You can only access a thread local value [via a closure](https://doc.rust-lang.org/std/thread/struct.LocalKey.html#method.with), and you cannot retain a reference to it. As a result, it wasn't possible to get a reference (`&`) to a system macro from the thread local system macro table—only an `Rc`.

As a compromise, I used a thread local to build the first system macro table—performing the necessary macro compilation steps—and then had subsequent readers and writers `clone()` that instance of the table. This allowed most readers and writers to avoid the cost of compiling system macros, but they still had to pay for the heap allocations. Not terribly satisfying.

Recently, in Rust v1.80, [the `LazyLock` type](https://doc.rust-lang.org/beta/std/sync/struct.LazyLock.html) was stabilized. It uses synchronization primitives to manage initializing the value, but all accesses thereafter can get a `&` reference for 'free'.

By switching the internals of the `Macro` type from `Rc<str>` to `CompactString` and `Rc<Macro>` to `Arc<Macro>`, I was able to create an immutable, lazily-initialized global singleton system macro table. References to the system macros are now trivial to get and can be reused across all readers and writers.

src/lazy/expanded/macro_table.rs
```diff
@@ -257,65 +280,137 @@ impl MacroTable {
...
-                    could_produce_system_value: false,
-                    must_produce_exactly_one_value: true,
-                    can_be_lazily_evaluated_at_top_level: true,
-                    expansion_singleton: Some(ExpansionSingleton {
-                        is_null: false,
-                        ion_type: IonType::String,
-                        num_annotations: 0,
-                    }),
-                },
+                ExpansionAnalysis::single_application_value(IonType::String),
```
zslayton, line 283: 🗺️ Note the use of named `ExpansionAnalysis` constructors for clarity.

src/lazy/expanded/macro_table.rs
```diff
@@ -707,8 +938,9 @@ impl MacroTable {
         }
     }
 
+    // TODO: Remove this
```
zslayton, line 941: 🗺️ Once we implement the encoding module sequence, we will no longer initialize the local macro table by cloning the system macro table.

src/lazy/text/buffer.rs
```diff
@@ -1177,8 +1177,49 @@ impl<'top> TextBuffer<'top> {
         Ok((exp_body_after_id, id))
     }
 
+    pub fn match_system_eexp_id(self) -> IonParseResult<'top, MacroIdRef<'top>> {
```
zslayton, line 1180: 🗺️ Text parsing logic for `$ion::<name>` and `$ion::<addr>`. Later on we can generalize this for any qualified macro reference.

zslayton (commented):
🗺️ PR Tour 🧭

src/lazy/expanded/macro_table.rs
```diff
@@ -257,65 +280,137 @@ impl MacroTable {
...
+                        ParameterEncoding::Tagged,
+                        ParameterCardinality::ZeroOrOne,
+                        RestSyntaxPolicy::NotAllowed,
+                    ),
+                    Parameter::new(
+                        "offset_minutes",
+                        ParameterEncoding::Tagged,
+                        ParameterCardinality::ZeroOrOne,
+                        RestSyntaxPolicy::NotAllowed,
+                    ),
```
jobarr-amzn, line 373: This stretch makes me wonder if named `Parameter` constructors would also be helpful. Is the required cardinality too high? I would think that at least `tagged`, `tagged_optional`, and `tagged_rest` would be helpful (roughly covering `x`, `x?`, and `x*`).

This also makes me think that the system macro table might be a good place to point people at for usage examples of parameters, we run a gamut here. Not as much as we used to with tagged encodings for the system macros, but still there's a fair bit of diversity of use here.

jobarr-amzn (approved):

benches/read_many_structs.rs
```diff
@@ -73,8 +72,7 @@ fn maximally_compact_1_1_data(num_values: usize) -> TestData_1_1 {
         0x32, 0x30, 0x3A, 0x35,
         0x39, 0x3A, 0x35, 0x39,
         0x2E, 0x37, 0x34, 0x34,
-        0x30, 0x30, 0x30, 0x5A,
-    ].repeat(num_values);
+        0x30, 0x30, 0x30, 0x5A].repeat(num_values);
```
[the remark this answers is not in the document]
zslayton: Yeah, `rustfmt` chose this layout.

benches/read_many_structs.rs
```diff
@@ -92,27 +90,26 @@ fn moderately_compact_1_1_data(num_values: usize) -> TestData_1_1 {
...
-                'threadId': thread_id,
-                'threadName': thread_name,
+                'timestamp': (%timestamp),
+                'threadId': (%thread_id),
+                'threadName': (%thread_name),
                 'loggerName': "com.example.organization.product.component.ClassName",
-                'logLevel': (literal INFO),
+                'logLevel': (.literal INFO),
                 'format': "Request status: {} Client ID: {} Client Host: {} Client Region: {} Timestamp: {}",
                 'parameters': [
```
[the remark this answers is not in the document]
zslayton: Good point! Both true. I'll update it.

benches/read_many_structs.rs
```diff
@@ -418,7 +412,9 @@ mod benchmark {
         let seq_1_1 = reader_1_1.read_all_elements().unwrap();
         assert!(
             IonData::eq(seq_1_0, &seq_1_1),
-            "{name} binary Ion 1.1 sequence was not equal to the original Ion 1.0 sequence"
+            "{name} binary Ion 1.1 sequence was not equal to the original Ion 1.0 sequence: \n{:#?}\n  !=\n{:#?}",
+            seq_1_0.elements().take(10).collect::<Vec<_>>(),
+            seq_1_1.elements().take(100).collect::<Vec<_>>(),
```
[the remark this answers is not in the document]
zslayton: 🤦 This was a debugging holdover, will remove.

src/lazy/encoder/binary/v1_1/value_writer.rs
```diff
@@ -2914,6 +2919,28 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn write_system_macro_invocation() -> IonResult<()> {
+        encoding_test(
+            |writer: &mut LazyRawBinaryWriter_1_1<&mut Vec<u8>>| {
+                let mut args = writer.eexp_writer(MacroIdRef::SystemAddress(
+                    SystemMacroAddress::new(3).unwrap(), /* make_string */
```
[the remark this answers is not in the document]
zslayton: I wrote this before I wrote `SystemMacroAddress`, then never went back to update it.

jobarr-amzn (approved):

PR #870, opened 2024-12-04
Adds binary read support for delimited arg groups

zslayton:

### DIFF
--- a/src/lazy/binary/raw/v1_1/e_expression.rs
+++ b/src/lazy/binary/raw/v1_1/e_expression.rs
@@ -473,7 +473,7 @@ pub struct BinaryEExpArgGroupIterator<'top> {
 
 #[derive(Debug, Copy, Clone)]
 enum BinaryEExpArgGroupIteratorSource<'top> {
-    /// The iterator is parsing expressions from a slice of the input buffer
+    /// The iterator is parsing expressions from a slice of the input buffer.
     Input(BinaryBuffer<'top>),
     /// The iterator is iterating over the bump-allocated cache of previously parsed expressions.
     Cache(&'top [LazyRawValueExpr<'top, BinaryEncoding_1_1>]),
--- a/src/lazy/binary/raw/v1_1/e_expression.rs
+++ b/src/lazy/binary/raw/v1_1/e_expression.rs
@@ -1,11 +1,9 @@
 #![allow(non_camel_case_types)]
 
-use std::fmt::{Debug, Formatter};
-use std::ops::Range;
-
 use crate::lazy::binary::raw::v1_1::immutable_buffer::{
     ArgGrouping, ArgGroupingBitmapIterator, BinaryBuffer,
 };
+use crate::lazy::binary::raw::v1_1::value::DelimitedContents;
 use crate::lazy::decoder::LazyRawValueExpr;
 use crate::lazy::encoding::BinaryEncoding_1_1;
 use crate::lazy::expanded::e_expression::EExpArgGroup;
@@ -17,6 +15,8 @@ use crate::lazy::expanded::EncodingContextRef;
 use crate::lazy::text::raw::v1_1::arg_group::{EExpArg, EExpArgExpr};
 use crate::lazy::text::raw::v1_1::reader::MacroIdRef;
 use crate::{try_or_some_err, v1_1, Environment, HasRange, HasSpan, IonResult, Span};
+use std::fmt::{Debug, Formatter};
+use std::ops::Range;
 
 /// An e-expression which has been parsed from a binary Ion 1.1 stream.
 #[derive(Copy, Clone)]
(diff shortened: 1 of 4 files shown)


src/lazy/binary/raw/v1_1/e_expression.rs
```diff
@@ -332,24 +332,53 @@ impl<'top> Iterator for BinaryEExpArgsInputIter<'top> {
...
-                    group_header_flex_uint.size_in_bytes() as u8,
-                );
-                (
-                    EExpArg::new(parameter, EExpArgExpr::ArgGroup(arg_group)),
-                    self.remaining_args_buffer.consume(arg_group_length),
-                )
+                match group_header_flex_uint.value() {
+                    // A FlexUInt of `0` indicates that the arg group is delimited.
+                    // We need to step through it to determine its span.
+                    0 if *parameter.encoding() == ParameterEncoding::Tagged => {
```
zslayton, line 340: 🗺️ This arm contains the new logic.

src/lazy/binary/raw/v1_1/e_expression.rs
```diff
@@ -332,24 +332,53 @@ impl<'top> Iterator for BinaryEExpArgsInputIter<'top> {
...
+                            group_header_flex_uint.size_in_bytes() + bytes_to_read;
+                        let arg_group = BinaryEExpArgGroup::new(
+                            parameter,
+                            self.remaining_args_buffer.slice(0, arg_group_length),
+                            group_header_flex_uint.size_in_bytes() as u8,
+                        );
+                        (
+                            EExpArg::new(parameter, EExpArgExpr::ArgGroup(arg_group)),
+                            self.remaining_args_buffer.consume(arg_group_length),
+                        )
```
zslayton, line 379: 🗺️ Relocated here from above.

zslayton (commented):
🗺️ PR tour 🧭

src/lazy/binary/raw/v1_1/e_expression.rs
```diff
@@ -428,28 +468,48 @@ impl<'top> HasSpan<'top> for BinaryEExpArgGroup<'top> {
...
-    remaining_args_buffer: BinaryBuffer<'top>,
+    source: BinaryEExpArgGroupIteratorSource<'top>,
+}
+
+#[derive(Debug, Copy, Clone)]
+enum BinaryEExpArgGroupIteratorSource<'top> {
+    /// The iterator is parsing expressions from a slice of the input buffer
+    Input(BinaryBuffer<'top>),
+    /// The iterator is iterating over the bump-allocated cache of previously parsed expressions.
+    Cache(&'top [LazyRawValueExpr<'top, BinaryEncoding_1_1>]),
```
popematt, line 479: So, the name and doc comments suggest to me that these expressions could be shared somehow. (Maybe there's a variable that's used more than once...?) Yet, further down, I see that this is being mutated. Is this actually a slice that points to a cache that is owned somewhere else? And it's effectively just the start and length of the slice that is updated?

Are any of those cached expressions stateful? Are we sure that there's no possibility of interference between different use-sites of the same cached expressions?

src/lazy/binary/raw/v1_1/value.rs
```diff
@@ -282,6 +282,7 @@ impl<'top> LazyRawValue<'top, BinaryEncoding_1_1> for &'top LazyRawBinaryValue_1
 
 #[derive(Debug, Copy, Clone)]
 pub enum DelimitedContents<'top> {
+    // I believe this is intended to be a flat version of Option<DelimitedContents>
```
popematt, line 285: Note to self? Or the start of a doc comment that you forgot to finish?

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -1302,6 +1302,36 @@ mod tests {
...
+        // Empty delimited argument group
+        bin_stream_eq(
+            &[
+                0xEF, // System macro, address follows as 1-byte FixedUInt
+                0x03, // make_string
+                0x02, // Argument group
+                0x01, // FlexUInt 0: delimited group
+                0xF0, // Delimited END
+            ],
+            r#""" // <-- empty string "#,
```
popematt, line 1316: It took me a minute to figure out what "empty string" was referring to. This might make it easier for other readers.
```suggestion
            r#" "" // <-- empty string "#,
```

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -1302,6 +1302,36 @@ mod tests {
...
+                0x03, // make_string
+                0x02, // Argument group
+                0x01, // FlexUInt 0: delimited group
+                0x93, // Opcode: 3-byte string follows
+                0x66, 0x6F, 0x6F, // "foo"
+                0x93, // Opcode: 3-byte string follows
+                0x62, 0x61, 0x72, // "bar"
+                0xF0, // Delimited END
+            ],
+            r#""foobar""#,
```
popematt, line 1331: Consider adding some whitespace to separate the Rust quotation marks from the Ion quotation marks.

popematt (commented):
There's one part where I am a bit confused, and it would help to have some more explanation, but otherwise it looks good. I've commented on a few other minor things that are not blockers.

jobarr-amzn (approved):

src/lazy/binary/raw/v1_1/value.rs
zslayton: Oops, this was a note to self. I wasn't sure if `None` meant "empty contents" or "not applicable". It's the latter. I'll remove the comment, thanks.

src/lazy/binary/raw/v1_1/e_expression.rs
zslayton: Both of `BinaryEExpArgGroupIteratorSource`'s variants refer to data stored elsewhere with a lifetime of `'top`:
* In the `Input` case, the `BinaryBuffer` holds a slice of the input buffer, which is frozen for the duration of `'top`.
* In the `Cache` case, the reader bump-allocated a `&'top [...]` in which to store raw lexed expressions it found as it traversed a delimited group. Data stored in the bump is guaranteed to outlive `'top`.

`BinaryEExpArgGroupIteratorSource` instances are immutable, but they are each wrapped by a `BinaryEExpArgGroupIterator` that _is_ mutable. Each time the iterator's `next()` method is called, it either advances the slice of the source's `Input` variant that it holds or it pops the head off of the slice of the `Cache` variant that it holds.

Both the iterator and its source are trivially copy-able, so you could easily/safely make several iterators traversing the expression group at once. Each iterator would mutate its own state as it traversed the immutable source data.

Is that helpful?

popematt (approved):

PR #871, opened 2024-12-05
Raw 1.1 writer now clears encoding buffer on `flush()`

### LINKED ISSUE #865: v1_1 `Writer::flush()` seems to not empty the buffer, resulting in ever-growing repeated preambles
Every `flush()` results in copies of all "entries" logged so far, followed by the most recent "entry". Is this PEBKAC[1], or a bug?

```rust
type IonWriter = Writer<v1_1::Binary, File>;

pub struct Log {
    ion_writer: IonWriter,
}

impl Log {
    pub fn write_entry(
        &mut self,
        module_path:   &str,
        event_kind:    EventKind,
        write_content: impl FnOnce(&mut EntryWriter<'_>),
    ) {
        {
            let mut entry_writer = EntryWriter::new(self, module_path, event_kind);
            write_content(&mut entry_writer);
            let _ = entry_writer.ion_writer.close();
        }
        let _ = self.ion_writer.flush();
    }
}

#[rustfmt::skip]
type IonContentWriter<'a> =
    // Viewing the top-level writer as a context where values can be written...
    <<IonWriter as MakeValueWriter>
    // ...this will be the list writer type.
        ::ValueWriter<'a> as ValueWriter>::ListWriter;
        // This is indeed an unfortunately complicated way to have to refer to the type. x_x
        // This type (a list writer) implements `SequenceWriter`.

pub struct EntryWriter<'f> {
    ion_writer: IonContentWriter<'f>,
}

impl<'f> EntryWriter<'f> {
    pub fn new(
        log:         &'f mut Log,
        module_path: &str,
        event_kind:  EventKind,
    ) -> Self {
        let mut ion_writer = log.ion_writer.list_writer().expect("logger ListWriter creation failed");

        use chrono::Utc;

        let now = Utc::now().naive_utc();
        let event_timestamp = ion_rs::Timestamp::from_datetime(now, ion_rs::TimestampPrecision::Second);
        let _ = ion_writer.write(event_timestamp);
        let event_source = format!("{:?}::{module_path}", EventSource::iotaSOFT_UI);
        let _ = ion_writer.write(event_source);
        let _ = ion_writer.write(format!("{:?}", event_kind));

        Self {
            ion_writer: ion_writer,
        }
    }

    pub fn write_name(&mut self, s: &str) {
        let _ = self.ion_writer.write_symbol(s);
    }

    pub fn write_str(&mut self, s: &str) {
        let _ = self.ion_writer.write(s);
    }
    ...
}
```

[1] PEBKAC: Problem Exists Between Keyboard and Chair.

zslayton:
Fixes #865.

The raw writer has an encoding buffer stored in its bump allocator.
As values are written, their serialized bytes are written to the encoding buffer.
When `flush()` is called, the encoding buffer's bytes are copied to output and the bump allocator is reset.
However, the writer was _not_ clearing the `encoding_buffer_ptr`, which continued to refer to a (now invalid) memory region in the bump.
That memory still contained the old encoding buffer and so would result in the old data being written out again.

The 1.0 writer had the correct behavior and a cleaner implementation in general, so I copied it over.

### DIFF
--- a/src/lazy/encoder/binary/v1_1/writer.rs
+++ b/src/lazy/encoder/binary/v1_1/writer.rs
@@ -72,16 +72,15 @@ impl<W: Write> LazyRawBinaryWriter_1_1<W> {
             encoding_buffer_ptr,
         } = self;
 
-        let encoding_buffer = match encoding_buffer_ptr {
-            // If `encoding_buffer_ptr` is set, get the slice of bytes to which it refers.
-            Some(ptr) => unsafe { ptr_to_ref::<'_, BumpVec<'_, u8>>(*ptr).as_slice() },
-            // Otherwise, there's nothing in the buffer. Use an empty slice.
-            None => &[],
-        };
-        // Write our top level encoding buffer's contents to the output sink.
-        output.write_all(encoding_buffer)?;
-        // Flush the output sink, which may have its own buffers.
-        output.flush()?;
+        if let Some(ptr) = encoding_buffer_ptr {
+            let encoding_buffer = unsafe { ptr_to_ref::<'_, BumpVec<'_, u8>>(*ptr).as_slice() };
+            // Write our top level encoding buffer's contents to the output sink.
+            output.write_all(encoding_buffer)?;
+            // Flush the output sink, which may have its own buffers.
+            output.flush()?;
+        }
+        // Now that we've written the encoding buffer's contents to output, clear it.
+        self.encoding_buffer_ptr = None;
         // Clear the allocator. A new encoding buffer will be allocated on the next write.
         allocator.reset();
         Ok(())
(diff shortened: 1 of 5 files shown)


src/lazy/encoder/binary/v1_1/writer.rs
```diff
@@ -72,16 +72,15 @@ impl<W: Write> LazyRawBinaryWriter_1_1<W> {
...
-        output.flush()?;
+        if let Some(ptr) = encoding_buffer_ptr {
+            let encoding_buffer = unsafe { ptr_to_ref::<'_, BumpVec<'_, u8>>(*ptr).as_slice() };
+            // Write our top level encoding buffer's contents to the output sink.
+            output.write_all(encoding_buffer)?;
+            // Flush the output sink, which may have its own buffers.
+            output.flush()?;
+        }
+        // Now that we've written the encoding buffer's contents to output, clear it.
+        self.encoding_buffer_ptr = None;
```
zslayton, line 83: This line was missing in the original impl, resulting in a dangling pointer. The second time `flush()` was called, the `match` arm on the old line 77 would incorrectly be taken.

jobarr-amzn (approved):

PR #872, opened 2024-12-06
Overhauls system macro table bootstrapping

zslayton:
This patch modifies `EncodingContext` to have an empty macro table when constructed for use in Ion 1.0 readers.

This allows the Ion 1.0 reader to be used in the process of bootstrapping the Ion 1.1 system macro table, which was previously not possible due to the (now eliminated) circular dependency between macros and readers.

System macros that can be expressed in TDL are now constructed from a template definition encoded in Ion 1.0. As a result, the system macro table initialization logic is now substantially easier to read and more straightforward to modify.

### DIFF
--- a/benches/read_many_structs.rs
+++ b/benches/read_many_structs.rs
@@ -396,7 +396,7 @@ mod benchmark {
         let name = test_data_1_1.name.as_str();
 
         let empty_context = EncodingContext::for_ion_version(IonVersion::v1_1);
-        let compiled_macro = TemplateCompiler::compile_from_text(
+        let compiled_macro = TemplateCompiler::compile_from_source(
             empty_context.get_ref(),
             &test_data_1_1.template_definition_text,
         )
--- a/src/lazy/binary/raw/v1_1/immutable_buffer.rs
+++ b/src/lazy/binary/raw/v1_1/immutable_buffer.rs
@@ -1317,8 +1317,9 @@ mod tests {
         encode_macro_fn: impl FnOnce(MacroAddress) -> Vec<u8>,
         test_fn: impl FnOnce(BinaryEExpArgsIterator_1_1<'_>) -> IonResult<()>,
     ) -> IonResult<()> {
-        let mut context = EncodingContext::empty();
-        let template_macro = TemplateCompiler::compile_from_text(context.get_ref(), macro_source)?;
+        let mut context = EncodingContext::for_ion_version(IonVersion::v1_1);
+        let template_macro =
+            TemplateCompiler::compile_from_source(context.get_ref(), macro_source)?;
         let macro_address = context.macro_table.add_macro(template_macro)?;
         let opcode_byte = u8::try_from(macro_address).unwrap();
         let binary_ion = encode_macro_fn(opcode_byte as usize);
--- a/src/lazy/expanded/compiler.rs
+++ b/src/lazy/expanded/compiler.rs
@@ -2,6 +2,7 @@
 //! evaluation.
 use crate::element::iterators::SymbolsIterator;
 use crate::lazy::decoder::Decoder;
+use crate::lazy::expanded::macro_table::ION_1_1_SYSTEM_MACROS;
 use crate::lazy::expanded::template::{
     ExprRange, MacroSignature, Parameter, ParameterCardinality, ParameterEncoding,
     RestSyntaxPolicy, TemplateBody, TemplateBodyElement, TemplateBodyExpr, TemplateMacro,
@@ -14,7 +15,10 @@ use crate::lazy::text::raw::v1_1::reader::MacroIdRef;
 use crate::lazy::value::LazyValue;
 use crate::lazy::value_ref::ValueRef;
 use crate::result::IonFailure;
-use crate::{v1_1, IonError, IonResult, IonType, Macro, MacroTable, Reader, Symbol, SymbolRef};
+use crate::{
+    AnyEncoding, IonError, IonInput, IonResult, IonType, Macro, MacroTable, Reader, Symbol,
+    SymbolRef,
+};
 use rustc_hash::FxHashMap;
 use smallvec::SmallVec;
 use std::ops::Range;
@@ -50,6 +54,17 @@ impl ExpansionAnalysis {
         }
     }
 
+    /// Returns an expansion analysis for a macro that will produce a single value that may or may
+    /// not be a system value.
+    pub const fn possible_system_value() -> Self {
+        ExpansionAnalysis {
+            could_produce_system_value: true,
+            must_produce_exactly_one_value: true,
+            can_be_lazily_evaluated_at_top_level: false,
+            expansion_singleton: None,
+        }
+    }
+
     /// Returns an expansion analysis for a macro that always produces exactly one non-system value.
     pub const fn single_application_value(ion_type: IonType) -> Self {
         ExpansionAnalysis {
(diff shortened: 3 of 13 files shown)


benches/read_many_structs.rs
```diff
@@ -396,7 +396,7 @@ mod benchmark {
         let name = test_data_1_1.name.as_str();
 
         let empty_context = EncodingContext::for_ion_version(IonVersion::v1_1);
-        let compiled_macro = TemplateCompiler::compile_from_text(
+        let compiled_macro = TemplateCompiler::compile_from_source(
```
zslayton, line 399: 🪧 The `compile_from_text` method previously required Ion 1.1. It now works with any Ion encoding; when used during bootstrapping we only pass it Ion 1.0.

src/lazy/binary/raw/v1_1/immutable_buffer.rs
```diff
@@ -1317,8 +1317,9 @@ mod tests {
         encode_macro_fn: impl FnOnce(MacroAddress) -> Vec<u8>,
         test_fn: impl FnOnce(BinaryEExpArgsIterator_1_1<'_>) -> IonResult<()>,
     ) -> IonResult<()> {
-        let mut context = EncodingContext::empty();
-        let template_macro = TemplateCompiler::compile_from_text(context.get_ref(), macro_source)?;
+        let mut context = EncodingContext::for_ion_version(IonVersion::v1_1);
```
zslayton, line 1320: 🪧 At first, Ion 1.0 and 1.1 readers would use the same `EncodingContext`. Later, `EncodingContext` was modified to be able to take an Ion version, allowing it to initialize the symbol table differently to accommodate Ion 1.1's new symtab. This PR continues that differentiation, causing 1.0 encoding contexts to be initialized with an empty macro table.

src/lazy/expanded/compiler.rs
```diff
@@ -50,6 +54,17 @@ impl ExpansionAnalysis {
         }
     }
 
+    /// Returns an expansion analysis for a macro that will produce a single value that may or may
+    /// not be a system value.
+    pub const fn possible_system_value() -> Self {
```
zslayton, line 59: 🪧 I added a few more named `ExpansionAnalysis` constructors so their intent can be expressed at a high level.

src/lazy/expanded/compiler.rs
```diff
@@ -146,7 +174,7 @@ pub struct TemplateCompiler {}
 
 impl TemplateCompiler {
     /// Takes a TDL expression in the form:
-    /// ```ion_1_1
+    /// ```ion
```
zslayton, line 177: The `rustdoc` renderer doesn't differentiate between these two identifiers, but now that the template compiler is Ion-version-agnostic, I wanted this fence to be too.

src/lazy/expanded/compiler.rs
```diff
@@ -431,10 +459,7 @@ impl TemplateCompiler {
         let macro_id = macro_id.into();
         match module_name {
             // If the module is `$ion`, this refers to the system module.
-            "$ion" => context
-                .system_module
-                .macro_table()
-                .clone_macro_with_id(macro_id),
+            "$ion" => ION_1_1_SYSTEM_MACROS.clone_macro_with_id(macro_id),
```
zslayton, line 462: 🪧 System macro resolution now happens in the singleton instance.

src/lazy/expanded/compiler.rs
```diff
@@ -459,7 +484,61 @@ impl TemplateCompiler {
         // The `params` clause of the macro definition is an s-expression enumerating the parameters
         // that the macro accepts. For example: `(flex_uint::x, y*, z?)`.
         let params_clause = Self::expect_sexp("an s-expression defining parameters", &mut values)?;
+        let signature = Self::compile_signature_from_sexp(context, pending_macros, params_clause)?;
```
zslayton, line 487: 🪧 This diff rendering is unfortunate. I extracted the logic needed to compile a sexp containing parameters (for example: `(foo bar? baz*)`) into a `MacroSignature`, making a new helper method. If you skim ahead to the next large deletion, you'll see that this was mostly just a code block being moved.

The new helper method is now also used when bootstrapping the 1.1 system macro table.

src/lazy/expanded/compiler.rs
```diff
@@ -520,44 +599,7 @@ impl TemplateCompiler {
                 Parameter::new(name, parameter_encoding, cardinality, rest_syntax_policy);
             compiled_params.push(compiled_param);
         }
-        let signature = MacroSignature::new(compiled_params)?;
```
zslayton, line 523: 🪧 The aforementioned block of moved code.

src/lazy/expanded/encoding_module.rs
```diff
@@ -16,13 +16,7 @@ impl EncodingModule {
...
         }
     }
-    pub fn ion_1_1_system_module() -> Self {
-        Self::new(
-            String::from("$ion"),
-            MacroTable::with_system_macros(),
-            SymbolTable::new(IonVersion::v1_1),
-        )
-    }
+
```
zslayton, line 19: 🪧 This was unused and, in light of the singleton instance, no longer makes sense.

src/lazy/expanded/macro_table.rs
```diff
@@ -204,504 +202,195 @@ impl MacroTable {
     pub const FIRST_USER_MACRO_ID: usize = Self::NUM_SYSTEM_MACROS;
 
     fn compile_system_macros() -> Vec<Arc<Macro>> {
```
zslayton, line 204: 🪧 This method is where the system macro table is bootstrapped, making it the primary benefactor of our refactoring.

src/lazy/expanded/macro_table.rs
```diff
@@ -204,504 +202,195 @@ impl MacroTable {
...
+        };
+
+        // Creates a `Macro` whose implementation is provided by the system
+        let builtin = |name: &str,
+                       signature: &str,
+                       kind: MacroKind,
+                       expansion_analysis: ExpansionAnalysis| {
             Arc::new(Macro::named(
+                name,
+                TemplateCompiler::compile_signature(context, signature).unwrap(),
```
zslayton, line 222: 🪧 Using the previously defined helper method, we can now compile the signature from source instead of manually instantiating a `Vec` of `Parameter`s.

src/lazy/expanded/macro_table.rs
```diff
@@ -204,504 +202,195 @@ impl MacroTable {
...
+                        (macro set_symbols (symbols*)
+                             $ion_encoding::(
+                               // Set a new symbol table
+                               (symbol_table [(%symbols)])
+                               // Include the active encoding module macros
+                               (macro_table $ion_encoding)
+                             )
+                       )
+                    "#,
+            ),
```
zslayton, line 303: 🪧 My next task is to switch over to the default module syntax, which means redefining these system template macros. That job will now be much easier--I can just revise this Ion 1.0 text.

src/lazy/expanded/macro_table.rs
```diff
@@ -827,4 +522,10 @@ impl MacroTable {
         }
         Ok(())
     }
+
+    pub(crate) fn reset_to_system_macros(&mut self) {
```
zslayton, line 526: 🪧 Used (for now) after encountering an Ion 1.1 IVM.

src/lazy/expanded/mod.rs
```diff
@@ -182,22 +176,14 @@ impl<'top> EncodingContextRef<'top> {
         &self.context.macro_table
     }
 
-    pub fn system_symbol_table(&self) -> &'top SymbolTable {
```
zslayton, line 185: 🪧 These are no longer needed; we have the singleton.

src/lazy/reader.rs
```diff
@@ -287,20 +286,18 @@ mod tests {
     fn expand_macro_test(
         macro_source: &str,
         encode_macro_fn: impl FnOnce(MacroAddress) -> Vec<u8>,
-        test_fn: impl FnOnce(Reader<AnyEncoding, &[u8]>) -> IonResult<()>,
+        test_fn: impl FnOnce(Reader<v1_1::Binary, &[u8]>) -> IonResult<()>,
```
zslayton, line 289: 🪧 This test was written before encoding directives were implemented. It manually registers a macro definition with the reader instead. Now that 1.0 readers don't have macros, it needs to be limited to working with 1.1 data.

zslayton (commented):
🗺️ PR Tour 🧭

popematt (approved):

src/lazy/expanded/macro_table.rs
jobarr-amzn: Wonderful, this is such a nice change.

PR #873, opened 2024-12-06
Switches from "encoding module" to "default module"

zslayton:
Builds on PR #872.

Switches encoding directive syntax from:
```ion
$ion_encoding::(/*...*/)
```
to:
```ion
$ion::(module _ /*...*/)
```
in anticipation of multi-module support.

### DIFF
--- a/src/lazy/encoder/writer.rs
+++ b/src/lazy/encoder/writer.rs
@@ -165,8 +165,10 @@ impl<E: Encoding, Output: Write> Writer<E, Output> {
 
         let mut directive = directive_writer
             .value_writer()
-            .with_annotations(v1_1::system_symbols::ION_ENCODING)?
+            .with_annotations(v1_1::system_symbols::ION)?
             .sexp_writer()?;
+
+        directive.write_symbol("module")?.write_symbol("_")?;
 
         let pending_symbols = context
             .symbol_table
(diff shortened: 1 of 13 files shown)


src/lazy/encoder/writer.rs
```diff
@@ -165,9 +165,11 @@ impl<E: Encoding, Output: Write> Writer<E, Output> {
 
         let mut directive = directive_writer
             .value_writer()
-            .with_annotations(v1_1::system_symbols::ION_ENCODING)?
+            .with_annotations(v1_1::system_symbols::ION)?
             .sexp_writer()?;
 
+        directive.write_symbol("module")?.write_symbol("_")?;
```
popematt, line 171: There should be a system symbol `module`, right?

src/lazy/expanded/compiler.rs
```diff
@@ -460,11 +460,9 @@ impl TemplateCompiler {
         match module_name {
             // If the module is `$ion`, this refers to the system module.
             "$ion" => ION_1_1_SYSTEM_MACROS.clone_macro_with_id(macro_id),
-            // If the module is `$ion_encoding`, this refers to the active encoding module.
-            "$ion_encoding" => context.macro_table().clone_macro_with_id(macro_id),
-            _ => todo!(
-                "qualified references to modules other than $ion_encoding (found {module_name}"
-            ),
+            // If the module is `_`, this refers to the active encoding module.
+            "_" => context.macro_table().clone_macro_with_id(macro_id),
```
popematt, line 464: It would be good to have a `pub const DEFAULT_MODULE_NAME` for this so that we can use a named constant instead of a string.

popematt (approved):
Looks good. Only 2 minor comments about the use of hard coded strings.

PR #874, opened 2024-12-09
Implements `flatten`, `make_list`, `meta`

zslayton:
Changes:
* Modifies the system macro table to allow macros to easily depend on earlier macros
* Adds an implementation of the `flatten` macro
* Defines `make_list` as a template that leverages `flatten`
* Re-defines `make_sexp` as a template that leverages `flatten`
* Removes old bespoke impl of `make_sexp`
* Defines `meta` as a template that leverages `none`

### DIFF
--- a/benches/read_many_structs.rs
+++ b/benches/read_many_structs.rs
@@ -438,7 +438,7 @@ mod benchmark {
             let mut context = EncodingContext::for_ion_version(IonVersion::v1_1);
             context
                 .macro_table_mut()
-                .add_macro(compiled_macro.clone())
+                .add_template_macro(compiled_macro.clone())
                 .unwrap();
             let context_ref = context.get_ref();
             b.iter(|| {
--- a/src/lazy/expanded/e_expression.rs
+++ b/src/lazy/expanded/e_expression.rs
@@ -9,9 +9,9 @@ use crate::lazy::decoder::{Decoder, RawValueExpr};
 use crate::lazy::encoding::TextEncoding_1_1;
 use crate::lazy::expanded::compiler::{ExpansionAnalysis, ExpansionSingleton};
 use crate::lazy::expanded::macro_evaluator::{
-    AnnotateExpansion, EExpressionArgGroup, ExprGroupExpansion, IsExhaustedIterator,
-    MacroExpansion, MacroExpansionKind, MacroExpr, MacroExprArgsIterator, MakeSExpExpansion,
-    MakeTextExpansion, RawEExpression, TemplateExpansion, ValueExpr,
+    AnnotateExpansion, EExpressionArgGroup, ExprGroupExpansion, FlattenExpansion,
+    IsExhaustedIterator, MacroExpansion, MacroExpansionKind, MacroExpr, MacroExprArgsIterator,
+    MakeSExpExpansion, MakeTextExpansion, RawEExpression, TemplateExpansion, ValueExpr,
 };
 use crate::lazy::expanded::macro_table::{MacroKind, MacroRef};
 use crate::lazy::expanded::template::TemplateMacroRef;
@@ -125,6 +125,11 @@ impl<'top, D: Decoder> EExpression<'top, D> {
             }
             MacroKind::MakeSExp => MacroExpansionKind::MakeSExp(MakeSExpExpansion::new(arguments)),
             MacroKind::Annotate => MacroExpansionKind::Annotate(AnnotateExpansion::new(arguments)),
+            MacroKind::Flatten => MacroExpansionKind::Flatten(FlattenExpansion::new(
+                self.context,
+                environment,
+                arguments,
+            )),
             MacroKind::Template(template_body) => {
                 let template_ref = TemplateMacroRef::new(invoked_macro.reference(), template_body);
                 environment = self.new_evaluation_environment()?;
--- a/src/lazy/expanded/e_expression.rs
+++ b/src/lazy/expanded/e_expression.rs
@@ -11,7 +11,7 @@ use crate::lazy::expanded::compiler::{ExpansionAnalysis, ExpansionSingleton};
 use crate::lazy::expanded::macro_evaluator::{
     AnnotateExpansion, EExpressionArgGroup, ExprGroupExpansion, FlattenExpansion,
     IsExhaustedIterator, MacroExpansion, MacroExpansionKind, MacroExpr, MacroExprArgsIterator,
-    MakeSExpExpansion, MakeTextExpansion, RawEExpression, TemplateExpansion, ValueExpr,
+    MakeTextExpansion, RawEExpression, TemplateExpansion, ValueExpr,
 };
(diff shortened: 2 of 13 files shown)


benches/read_many_structs.rs
```diff
@@ -438,7 +438,7 @@ mod benchmark {
             let mut context = EncodingContext::for_ion_version(IonVersion::v1_1);
             context
                 .macro_table_mut()
-                .add_macro(compiled_macro.clone())
+                .add_template_macro(compiled_macro.clone())
```
zslayton, line 441: 🪧 This method (which is specific to template macros) was renamed to accommodate adding a more general purpose `add_macro` method alongside it.

src/lazy/expanded/e_expression.rs
```diff
@@ -123,8 +123,12 @@ impl<'top, D: Decoder> EExpression<'top, D> {
             MacroKind::MakeSymbol => {
                 MacroExpansionKind::MakeSymbol(MakeTextExpansion::symbol_maker(arguments))
             }
-            MacroKind::MakeSExp => MacroExpansionKind::MakeSExp(MakeSExpExpansion::new(arguments)),
             MacroKind::Annotate => MacroExpansionKind::Annotate(AnnotateExpansion::new(arguments)),
+            MacroKind::Flatten => MacroExpansionKind::Flatten(FlattenExpansion::new(
+                self.context,
+                environment,
+                arguments,
+            )),
```
zslayton, line 131: 🪧 Now that `make_sexp` is a trivial template that uses `flatten`, we no longer need a special `MacroKind` and custom impl for it. While this PR adds 3 new system macros, `MacroKind`'s number of variants is unchanged.

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -419,20 +420,19 @@ impl<'top, D: Decoder> ValueExpr<'top, D> {
 
 /// Indicates which of the supported macros this represents and stores the state necessary to
 /// continue evaluating that macro.
-#[derive(Copy, Clone, Debug)]
+#[derive(Debug)]
```
zslayton, line 423: 🪧 `flatten` is the first system macro to be implemented which requires a macro evaluator to persist across calls to `evaluator.next()`. As a result, the `MacroExpansion` family of types can no longer implement `Copy`, as each one might own a bump-allocated evaluator of its own. The `Copy` impl was only being used directly in one speculatively defined function that had no callers. I'll point it out later in the diff.

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -499,18 +499,12 @@ impl<'top, D: Decoder> MacroExpansion<'top, D> {
...
             Annotate(annotate_expansion) => annotate_expansion.next(context, environment),
+            Flatten(flatten_expansion) => flatten_expansion.next(),
             // `none` is trivial and requires no delegation
             None => Ok(MacroExpansionStep::FinalStep(Option::None)),
         }
     }
-
-    // Calculate the next step in this macro expansion without advancing the expansion.
-    pub fn peek_next_step(&self) -> IonResult<MacroExpansionStep<'top, D>> {
-        let mut expansion_copy = *self;
```
zslayton, line 511: 🪧 The aforementioned use of `Copy`, which would have enabled `peek()`ing at the next step of a macro. This is still possible without `Copy`, but will take a bit more code.

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -718,7 +712,7 @@ impl<'top, D: Decoder> MacroEvaluator<'top, D> {
 
     #[inline(never)]
     pub fn push_general_case(&mut self, new_expansion: MacroExpansion<'top, D>) {
-        match self.state {
+        match mem::take(&mut self.state) {
```
zslayton, line 715: 🪧 An indirect use of `Copy`: because `self.state` isn't `Copy` any more, we need to set its value to the default (`Empty`) so we can update its state using the (owned) contents.

src/lazy/expanded/macro_table.rs
```diff
@@ -188,43 +189,69 @@ impl Default for MacroTable {
 }
 
 impl MacroTable {
-    pub const SYSTEM_MACRO_KINDS: &'static [MacroKind] = &[
-        MacroKind::None,
-        MacroKind::ExprGroup,
-        MacroKind::MakeString,
-        MacroKind::MakeSExp,
-        MacroKind::Annotate,
-    ];
```
zslayton, line 197: 🪧 This was a holdover from before we had a proper system macro table. It's no longer in use.

src/lazy/expanded/macro_table.rs
```diff
@@ -511,18 +528,21 @@ impl MacroTable {
         Ok(id)
     }
 
+    pub(crate) fn append_macro(&mut self, macro_ref: &Arc<Macro>) -> IonResult<()> {
```
zslayton, line 531: 🪧 This method was extracted from the `append_all_macros_from` function below. It is now also be used in `compile_system_macros()`.

src/lazy/expanded/macro_table.rs
```diff
@@ -490,7 +507,7 @@ impl MacroTable {
         }
     }
 
-    pub fn add_macro(&mut self, template: TemplateMacro) -> IonResult<usize> {
+    pub fn add_template_macro(&mut self, template: TemplateMacro) -> IonResult<usize> {
```
zslayton, line 510: 🪧 As mentioned prior, this was renamed to accommodate the introduction of a more general `append_macro` method (see below).

src/lazy/expanded/sequence.rs
```diff
@@ -212,8 +211,6 @@ pub enum ExpandedSExpSource<'top, D: Decoder> {
     /// Because the TDL treats s-expressions as literals, this variant only applies when the
     /// s-expression appeared within a `literal` operation.
     Template(Environment<'top, D>, TemplateElement<'top>),
-    /// The SExp was produced by a call to `make_sexp`.
-    Constructed(Environment<'top, D>, MacroExprArgsIterator<'top, D>),
```
zslayton, line 216: Defining `make_sexp` in terms of `flatten` allowed me to delete lots of special-case code for `make_sexp`. 🎉

src/lazy/text/buffer.rs
```diff
@@ -2822,7 +2822,10 @@ mod tests {
         fn register_macro(&mut self, text: &str) -> &mut Self {
             let new_macro =
                 TemplateCompiler::compile_from_source(self.context.get_ref(), text).unwrap();
-            self.context.macro_table.add_macro(new_macro).unwrap();
+            self.context
+                .macro_table
+                .add_template_macro(new_macro)
+                .unwrap();
```
zslayton, line 2828: 🪧 `rustfmt` change

zslayton (commented):
🗺️ PR Tour 🧭

popematt (approved):

PR #878, opened 2024-12-11
Implements the `make_field` system macro

zslayton:

### DIFF
--- a/src/lazy/expanded/struct.rs
+++ b/src/lazy/expanded/struct.rs
@@ -112,7 +112,7 @@ pub enum ExpandedStructSource<'top, D: Decoder> {
         &'top TemplateStructIndex,
     ),
     // The struct was produced by the `make_struct` macro.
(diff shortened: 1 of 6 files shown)


src/lazy/expanded/struct.rs
```diff
@@ -283,6 +282,18 @@ impl<'top, D: Decoder> LazyExpandedStruct<'top, D> {
                     }
                 }
             }
+            // For any other kind of struct, do a linear scan over its fields until we encounter
+            // one with the requested name.
```
jobarr-amzn, line 286:
```suggestion
            // For any other kind of struct (e.g. struct literal or make_struct call),
            // do a linear scan over its fields until we encounter one with the requested name.
```

I like a reminder of what the other cases are. It doesn't have to come this way.

jobarr-amzn (approved):

src/lazy/expanded/struct.rs
zslayton: I'm going to incorporate this in another PR I'm about to open.

PR #885, opened 2024-12-19
Container tooling

zslayton:
This PR adds lots of new tooling methods and bookkeeping:
* Tags each value with the variable that produced it, if any.
* Renames `UnexpandedField` to `FieldExpr`, which is much clearer.
* Adds container iterators that yield both macro invocations _and_ their expansions, allowing tooling to see both and work with whatever they need.
* Adds `Span` accessor for the delimited `END` byte.

### DIFF
--- a/src/lazy/any_encoding.rs
+++ b/src/lazy/any_encoding.rs
@@ -65,6 +65,7 @@ use crate::lazy::text::value::{
     LazyRawTextVersionMarker_1_1, RawTextAnnotationsIterator,
 };
 use crate::symbol_table::{SystemSymbolTable, SYSTEM_SYMBOLS_1_0, SYSTEM_SYMBOLS_1_1};
+use crate::LazyRawValueKind::{Binary_1_0, Binary_1_1, Text_1_0, Text_1_1};
 use crate::{try_next, Encoding, IonResult, IonType, RawStreamItem, RawSymbolRef};
 
 /// An implementation of the `LazyDecoder` trait that can read any encoding of Ion.
@@ -1044,6 +1045,15 @@ impl<'top> LazyRawValue<'top, AnyEncoding> for LazyRawAnyValue<'top> {
             Binary_1_0(v) => v.is_null(),
             Text_1_1(v) => v.is_null(),
             Binary_1_1(v) => v.is_null(),
+        }
+    }
+
+    fn is_delimited(&self) -> bool {
+        match &self.encoding {
+            Text_1_0(v) => v.is_delimited(),
+            Binary_1_0(v) => v.is_delimited(),
+            Text_1_1(v) => v.is_delimited(),
+            Binary_1_1(v) => v.is_delimited(),
         }
     }
 
--- a/src/lazy/binary/raw/value.rs
+++ b/src/lazy/binary/raw/value.rs
@@ -115,6 +115,10 @@ impl<'top> LazyRawValue<'top, BinaryEncoding_1_0> for LazyRawBinaryValue_1_0<'to
         self.is_null()
     }
 
+    fn is_delimited(&self) -> bool {
+        false
+    }
+
     fn has_annotations(&self) -> bool {
         self.has_annotations()
     }
@@ -220,6 +224,19 @@ pub trait EncodedBinaryValue<'top, D: Decoder>: LazyRawValue<'top, D> {
         let body_length = self.body_length();
         let body_bytes = &value_span.bytes()[value_span.len() - body_length..];
         Span::with_offset(value_span.range().end - body_length, body_bytes)
+    }
+
+    fn delimited_end_span(&self) -> Span<'top> {
+        let bytes = self.span().bytes();
+        let end = bytes.len();
+        let range = if !self.is_delimited() {
+            end..end
+        } else {
+            debug_assert!(bytes[end - 1] == 0xF0);
+            end - 1..end
+        };
+        let end_bytes = bytes.get(range).unwrap();
+        Span::with_offset(self.range().end - end_bytes.len(), end_bytes)
     }
 }
 
--- a/src/lazy/decoder.rs
+++ b/src/lazy/decoder.rs
@@ -382,7 +382,7 @@ impl<D: Decoder> HasRange for LazyRawFieldExpr<'_, D> {
 // function while also preventing users from seeing or depending on it.
 pub(crate) mod private {
     use crate::lazy::expanded::macro_evaluator::{MacroExpr, RawEExpression};
-    use crate::lazy::expanded::r#struct::UnexpandedField;
+    use crate::lazy::expanded::r#struct::FieldSource;
     use crate::lazy::expanded::EncodingContextRef;
     use crate::{try_next, try_or_some_err, IonResult, LazyExpandedValue, LazyRawFieldName};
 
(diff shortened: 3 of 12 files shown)


src/lazy/any_encoding.rs
```diff
@@ -1047,6 +1048,15 @@ impl<'top> LazyRawValue<'top, AnyEncoding> for LazyRawAnyValue<'top> {
         }
     }
 
+    fn is_delimited(&self) -> bool {
```
zslayton, line 1051: 🪧 Each encoded value kind can now report whether it is delimited. For text 1.0 and binary 1.0, this method just returns `false`.

src/lazy/binary/raw/value.rs
```diff
@@ -221,6 +225,19 @@ pub trait EncodedBinaryValue<'top, D: Decoder>: LazyRawValue<'top, D> {
...
+        let end = bytes.len();
+        let range = if !self.is_delimited() {
+            end..end
+        } else {
+            debug_assert!(bytes[end - 1] == 0xF0);
+            end - 1..end
+        };
+        let end_bytes = bytes.get(range).unwrap();
+        Span::with_offset(self.range().end - end_bytes.len(), end_bytes)
+    }
```
zslayton, line 240: 🪧 For scalars and length-prefixed containers, this returns an empty slice.

src/lazy/decoder.rs
```diff
@@ -257,6 +258,21 @@ pub enum LazyRawFieldExpr<'top, D: Decoder> {
...
+                name.resolve(context),
+                LazyExpandedValue::from_literal(context, value),
+            ),
+            NameEExp(name, eexp) => {
+                FieldExpr::NameMacro(name.resolve(context), eexp.resolve(context)?.into())
+            }
+            EExp(eexp) => FieldExpr::EExp(eexp.resolve(context)?),
+        };
+        Ok(field)
+    }
```
zslayton, line 274: 🪧 A convenience method pattern used widely throughout the crate; `resolve` turns a raw value (or in this case, field) into its higher-level, context-ful equivalent.

src/lazy/decoder.rs
```diff
@@ -395,47 +411,44 @@ pub(crate) mod private {
     }
 
     pub trait LazyRawStructPrivate<'top, D: Decoder> {
-        /// Creates an iterator that converts each raw struct field into an `UnexpandedField`, a
+        /// Creates an iterator that converts each raw struct field into an `FieldExpr`, a
         /// common representation for both raw fields and template fields that is used in the
         /// expansion process.
-        fn unexpanded_fields(
+        fn field_exprs(
```
zslayton, line 417: 🪧 `UnexpandedField` was a clunky name, hopefully this is better. The new relationship between `LazyExpandedField` and `FieldExpr` mirrors the existing relationship between `LazyExpandedValue` and `ValueExpr`.

src/lazy/encoder/binary/v1_1/flex_sym.rs
```diff
@@ -95,7 +95,7 @@ impl<'top> FlexSym<'top> {
             Ordering::Equal => {
                 // Get the first byte following the leading FlexInt
                 let flex_int_len = value.size_in_bytes();
-                if input.len() < flex_int_len {
+                if input.len() <= flex_int_len {
```
zslayton, line 98: 🪧 This bug surfaced while I was working on an example file from @jobarr-amzn. This fixed it.

src/lazy/expanded/macro_evaluator.rs
```diff
@@ -114,7 +114,7 @@ where
 #[derive(Copy, Clone)]
 pub struct MacroExpr<'top, D: Decoder> {
     kind: MacroExprKind<'top, D>,
-    variable: Option<TemplateVariableReference<'top>>,
+    pub(crate) variable: Option<TemplateVariableReference<'top>>,
```
zslayton, line 117: 🪧 Macros now remember the variable (if any) that produced them, allowing them in turn to tag the values they produce.

src/lazy/expanded/sequence.rs
```diff
@@ -202,6 +207,50 @@ impl<'top, D: Decoder> Iterator for ExpandedListIterator<'top, D> {
     }
 }
 
+/// Like an [`ExpandedListIterator`], but will yield macro invocations before also yielding
+/// that macro's expansion.
+#[derive(Debug)]
+pub struct ListValueExprIterator<'top, D: Decoder> {
```
zslayton, line 213: 🪧 This section introduces iterators that return both the expressions _and_ the values they expand to.

zslayton (commented):
🗺️ PR Tour 🧭

src/lazy/expanded/struct.rs
```diff
@@ -1,31 +1,76 @@
...
+}
+
+impl<'top, D: Decoder> FieldExpr<'top, D> {
+    pub fn name(&self) -> Option<&LazyExpandedFieldName<'top, D>> {
+        use FieldExpr::*;
+        let name = match self {
+            NameValue(name, _) | NameMacro(name, _) => name,
+            EExp(_) => return None,
+        };
+        Some(name)
```
jobarr-amzn, line 42:
```suggestion
        match self {
            NameValue(name, _) | NameMacro(name, _) => Some(name),
            EExp(_) => None,
        }
```

src/lazy/expanded/struct.rs
```diff
@@ -580,34 +654,240 @@ impl<'top, D: Decoder> ExpandedStructIterator<'top, D> {
...
+    /// ```ion
+    /// {
+    ///   bar: (:values 1 2 3),
+    /// }
+    /// ```
+    /// An `ExpandedStructIterator` would yield a `LazyExpandedField` representing each
+    /// of the name/values pairs in the expansion: `(bar, 1)`, `(bar, 2)`, and `(bar, 3)`.
+    ///
+    /// In contrast, the `FieldExprIterator` would yield a `FieldExpr` for the name/macro
+    /// field expression (`NameMacro("foo", MacroExpr)`) followed by a `FieldExpr` for each of
```
jobarr-amzn, line 711:
```suggestion
    /// field expression (`NameMacro("bar", MacroExpr)`) followed by a `FieldExpr` for each of
```

src/lazy/expanded/struct.rs
```diff
@@ -580,34 +654,240 @@ impl<'top, D: Decoder> ExpandedStructIterator<'top, D> {
...
-        let iter: &'top mut ExpandedStructIterator<'top, D> = struct_.bump_iter();
-        *state = ExpandedStructIteratorState::InliningAStruct(iter);
-        Ok(())
+
+            expect_name_value(fields, "foo", 0)?;
+            assert!(matches!(
+                fields.next().unwrap()?,
+                FieldExpr::NameMacro(name, invocation)
+                    if name.read()?.text() == Some("bar") && matches!(invocation.kind(), MacroExprKind::EExp(eexp) if eexp.invoked_macro.name() == Some("three_values"))
+            ));
```
jobarr-amzn, line 877: On the one hand, it's cool that we can do this in Rust. On the other hand... 😨

I wish there was a better syntax for expressing a set of constraints like this.

jobarr-amzn (approved):

src/lazy/expanded/struct.rs
zslayton: A lot of the lower-level types are missing convenience methods like `expect_next()`, `expect_name_macro()`, etc that would make this prettier. :disappointed: The unit tests are the only place that would really benefit, but that might be reason enough to implement them.

Issue #888, opened 2024-12-25
Getting, then later writing, a symbol's ID with v1.1 BinaryWriter?


barries:
In the telemetry framework I'm writing, it would be nice (tm) to be able to acquire symbol IDs for `&str`s, then emit as needed. My motivation is to be able to hold onto a symbol like the currently executing function name without having to do a heap allocation to hold that name between function entry and exit.

Is that a likely feature one day?

Two possible API concepts:

```rust
let symbol_id = ion_writer.write_symbol("foo")?;
... time passes ...
ion_writer.write(symbol_id)
```

```rust
let symbol_id = ion_writer.symbol_table().get("foo");
ion_writer.write(symbol_id)
... time passes ...
ion_writer.write(symbol_id)
```

zslayton:
The latter is meant to be supported, but it looks like the `symbol_table()` method on the value writer isn't `pub`.

zslayton:
This issue (along with #887) was fixed by PR #889.

barries:
@zslayton, thanks for making SymbolTable public; given that `add_symbol_for_text()` is not `pub`, is there a way to add symbols directly (like the old `intern()` method, before writing them?

Or do I need to write one, then look it up using the same string (which is a bit of extra complexity on the call side and seems inefficient)?

PR #889, opened 2024-12-26
Makes `AnnotatableWriter` and writer's symbol table experimentally pub

### LINKED ISSUE #887: Writing annotated values with v1.1 BinaryWriter?
I'd like to do something like the following `.with_annotations(...)` call--is this the sort of thing that's planned, or is there some other way I can achieve it without trying to dig into the `ion-rs` internals or copy `value` to, say, an `Element` tree before serializing?

Thanks!

```rust
#[rustfmt::skip]
type IonListWriter<'a> =
    // Viewing the top-level writer as a context where values can be written...
    <<IonWriter as ContextWriter>
    // ...this will be the list writer type.
        ::NestedValueWriter<'a> as ValueWriter>::ListWriter;
        // This is indeed an unfortunately complicated way to have to refer to the type. x_x
        // This type (a list writer) implements `SequenceWriter`.

pub struct EntryWriter<'f> {
    ion_list_writer: IonListWriter<'f>,
}

impl<'f> EntryWriter<'f> {
    ...
    pub fn write_value(&mut self, pretty_printer_fmt: &str, value: &impl serde::Serialize) {
        use ion_rs::serde::ser::ValueSerializer;

        let is_human_readable = false;
        let serializer = ValueSerializer::new(
            self.ion_list_writer.value_writer().with_annotations([pretty_printer_fmt]),
            is_human_readable
        );
        let _ = value.serialize(serializer);
    }
}
```

zslayton:
Fixes #887 and #888.

### DIFF
--- a/src/element/mod.rs
+++ b/src/element/mod.rs
@@ -26,7 +26,7 @@ use std::cmp::Ordering;
 use std::fmt::{Display, Formatter};
 use std::io;
 
-use crate::{ion_data, Decimal, Int, IonResult, IonType, Str, Symbol, Timestamp};
+use crate::{ion_data, Decimal, Int, IonResult, IonType, Str, Symbol, SymbolRef, Timestamp};
 use crate::{Blob, Bytes, Clob, List, SExp, Struct};
 // Re-export the Value variant types and traits so they can be accessed directly from this module.
 use crate::element::builders::{SequenceBuilder, StructBuilder};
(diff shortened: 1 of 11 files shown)


tgregg (approved):

barries:
Thanks, @zslayton! I wanted to attempt a PR for this, but seeing all that you did, I wouldn't have been familiar enough to make all the necessary tweaks. Very much appreciated!

Issue #904, opened 2025-02-05
Implement the support for location metadata in the text buffer


desaikd:
Location metadata here means row and column values for the current text input of the buffer. It will be based on the `offset` value currently stored in the buffer and newline character counts (i.e. characters `\r` and `\n`). 

Text buffer would need to do some book keeping based on the parameters specified above for properties row and previous newline byte offset. It needs to add access methods `row()` and `column()` where the row will just render the property row, whereas column will be based on current offset value and previous newline byte offset.

To implement this feature here are the list of changes required:

- [x] Modify winnow parser methods for getting newline count 
  - [x] match_whitespace0
  - [x] match_whitespace1 
  - [x] match_text_until_unescaped
- [x] Write unit tests to verify buffer properties row and column results into correct numbers

Reference WIP: https://github.com/amazon-ion/ion-rust/commit/2af4d3c21a083d62d8102d024dbc29be12098d2e

desaikd:
Closed with https://github.com/amazon-ion/ion-rust/pull/941

PR #905, opened 2025-02-06
Adds location information to text buffer

desaikd:
*Issue https://github.com/amazon-ion/ion-rust/issues/904*

*Description of changes:*
This PR adds location information to the underlying text buffer. The location information includes row and column value for the buffer input data.

*Row Calculation*
Row numbers start at 1 and increment with each newline character encountered in the input. This maintains consistency with common text editor line numbering.
Row calculations will be based on occurrence of newline characters `\r` and `\n`. These characters have different implications in different OS systems. In order to correctly count the row position:
1. Consider `\r\n` as a single count if they appear together in the input.
2. Count the total occurrences of `\r` and `\n` characters individually.
3. Subtract count in step 1 from count in step 2. This will be the final count of newline characters.

*Column Calculation*
Column numbers also start at 1 and represent the number of characters(here the character count to consider is byte count/offset) consumed within the current row, including whitespaces. The column count will be based on the previous newline offset position. At any given point the column count represent `CURRENT_OFFSET_VALUE - PREVIOUS_NEWLINE_OFFSET_VALUE`. 

*List of changes:*
- Modifies winnow parser methods for row and column calculation. This includes modifying following parser methods:
  - `match_whitespace0`
  - `match_whitespace1` 
  - `match_text_until_unescaped_str`
- Adds `update_location_metadata`  which helps all the parser methods to update location metadata appropriately on newline parsing.
  - This method does the book keeping for row count and previous newline offset value.
- Requires the checkpoint reset to reset row and previous newline offset values. (This is similar to how `offset` property is reset) 
- Adds unit tests to verify buffer properties row and column results into correct numbers.

### DIFF
--- a/src/lazy/text/buffer.rs
+++ b/src/lazy/text/buffer.rs
@@ -362,6 +362,7 @@ impl<'top> TextBuffer<'top> {
         )
             .take()
             .parse_next(self)?;
+        self.update_location_metadata(result.data);
         Ok(result)
     }
 
@@ -2940,6 +2941,8 @@ mod tests {
     #[case::comment("/*comment*/", (1,12))]
     #[case::newline_before_comment("\n/*comment*/", (2,12))]
     #[case::newline_after_comment("/*comment*/\n", (2,1))]
+    #[case::newline_inside_comment("/*multiline \n comment*/", (2,11))]
+    #[case::newlines_inside_comment("/*this is a \n multiline \n comment*/", (3,11))]
     fn expect_whitespace_with_comment(#[case] input: &str, #[case] expected_location: (usize, usize)) {
         MatchTest::new_1_0(input).expect_match_location(match_length(TextBuffer::match_optional_comments_and_whitespace), expected_location);
     }
--- a/src/lazy/text/buffer.rs
+++ b/src/lazy/text/buffer.rs
@@ -215,7 +215,7 @@ impl<'top> TextBuffer<'top> {
     }
 
     /// Returns the row position for this buffer.
-    /// _Note: Row positions are calculated based on newline characters `\n` and `\r`._
+    /// _Note: Row positions are calculated based on newline characters `\n` and `\r`. `\r\n` together in this order is considered a single newline._
     pub fn row(&self) -> usize {
         self.row
     }
@@ -271,7 +271,7 @@ impl<'top> TextBuffer<'top> {
     pub fn match_whitespace1(&mut self) -> IonMatchResult<'top> {
         let result = take_while(1.., WHITESPACE_BYTES).parse_next(self)?;
         self.update_location_metadata(result.data);
-       Ok(*self)
+       Ok(result)
     }
 
     /// Updates the location metadata based on the matched whitespace bytes in the consumed buffer
@@ -289,13 +289,12 @@ impl<'top> TextBuffer<'top> {
             self.row += newline_match_count;
 
             // Stores this newline offset as previous newline offset for calculating column position since this has already been matched/parsed
-            if self.offset < non_newline_match_length {
+            self.prev_newline_offset = if self.offset < non_newline_match_length {
                 // this means that the input is not yet consumed hence get the input length + the current offset
-                self.prev_newline_offset = self.offset + data.len() - non_newline_match_length;
+                self.offset + data.len() - non_newline_match_length
             } else {
-                self.prev_newline_offset = self.offset - non_newline_match_length;
-
-            }
+                self.offset - non_newline_match_length
+            };
         }
     }
 
@@ -303,10 +302,11 @@ impl<'top> TextBuffer<'top> {
     pub fn match_whitespace0(&mut self) -> IonMatchResult<'top> {
         let result = take_while(0.., WHITESPACE_BYTES).parse_next(self)?;
         self.update_location_metadata(result.data);
-        Ok(*self)
+        Ok(result)
     }
 
     /// Matches any amount of contiguous comments and whitespace, including none.
+    #[inline]
     pub fn match_optional_comments_and_whitespace(&mut self) -> IonMatchResult<'top> {
         pub fn full_match_optional_comments_and_whitespace<'t>(
             input: &mut TextBuffer<'t>,
@@ -498,7 +498,6 @@ impl<'top> TextBuffer<'top> {
         )
             .map(|(maybe_annotations, value)| input.apply_annotations(maybe_annotations, value))
             .parse_next(self)
-        // let length = result.with_span().map(|(_output, consumed)| consumed.len());
     }
 
     /// Matches a struct field name. That is:
@@ -1641,8 +1640,6 @@ impl<'top> TextBuffer<'top> {
         let delimiter_head = delimiter.as_bytes()[0];
         // Whether we've encountered any escapes while looking for the delimiter
         let mut contained_escapes = false;
-        // This input may contain newline characters hence update the location metadata.
-        self.update_location_metadata(self.bytes());
         // The input left to search
         let mut remaining = *self;
         loop {
@@ -1650,7 +1647,7 @@ impl<'top> TextBuffer<'top> {
             // If the input doesn't contain one, this will return an `Incomplete`.
             // `match_text_until_escaped` does NOT include the delimiter byte in the match,
             // so `remaining_after_match` starts at the delimiter byte.
-            let (_matched_input, segment_contained_escapes) =
+            let (matched_input_buffer, segment_contained_escapes) =
                 remaining.match_text_until_unescaped(delimiter_head, true)?;
             contained_escapes |= segment_contained_escapes;
 
@@ -1659,6 +1656,8 @@ impl<'top> TextBuffer<'top> {
                 let relative_match_end = remaining.offset() - self.offset();
                 let matched_input = self.slice(0, relative_match_end);
                 self.consume(relative_match_end);
+                // This input may contain newline characters hence update the location metadata.
+                self.update_location_metadata(matched_input_buffer.bytes());
                 return Ok((matched_input, contained_escapes));
             } else {
                 // Otherwise, advance by one and try again.
@@ -2042,20 +2041,11 @@ impl<'data> Stream for TextBuffer<'data> {
     }
 
     fn checkpoint(&self) -> Self::Checkpoint {
-        let mut checkpoint = *self;
-        // Reset row, column at checkpoint
-        checkpoint.row = 0;
-        checkpoint.prev_newline_offset = 0;
-        checkpoint
+        *self
     }
 
     fn reset(&mut self, checkpoint: &Self::Checkpoint) {
-        let current_row = self.row;
-        let prev_column_value = self.prev_newline_offset;
-
         *self = *checkpoint;
-        self.row = current_row + checkpoint.row;
-        self.prev_newline_offset = prev_column_value + checkpoint.prev_newline_offset;
     }
 
     fn raw(&self) -> &dyn Debug {
--- a/src/lazy/text/buffer.rs
+++ b/src/lazy/text/buffer.rs
@@ -283,9 +283,11 @@ impl<'top> TextBuffer<'top> {
             // Subtract the crlf_count from total count of all newline characters to get the correct number of newline match count.
             let newline_match_count = data.iter().filter(|b| NEWLINE_BYTES.contains(b)).count() - crlf_count;
 
-            // Gets index for the last occurrence of the newline byte and subtracts from the result length to get non newline bytes length
-            let last_index_newline_byte = data.iter().rposition(|b| NEWLINE_BYTES.contains(b)).unwrap_or(0);
-            let non_newline_match_length = data.len() - last_index_newline_byte - 1;
+            // Gets index for the last occurrence of the newline byte and subtracts from the result length to get non newline bytes length.
+            // Adding 1 to the index because we want to include everything after the newline -
+            // if newline is at index 5, we want to start counting from index 6 (5 + 1).
+            let last_index_newline_byte = data.iter().rposition(|b| NEWLINE_BYTES.contains(b)).map(|i| i + 1).unwrap_or(0);
+            let non_newline_match_length = data.len() - last_index_newline_byte;
             self.row += newline_match_count;
 
             // Stores this newline offset as previous newline offset for calculating column position since this has already been matched/parsed
@@ -2045,7 +2047,12 @@ impl<'data> Stream for TextBuffer<'data> {
     }
 
     fn reset(&mut self, checkpoint: &Self::Checkpoint) {
+        let current_row = self.row;
+        let prev_column_value = self.prev_newline_offset;
+
         *self = *checkpoint;
+        self.row = current_row;
+        self.prev_newline_offset = prev_column_value;
     }
 
     fn raw(&self) -> &dyn Debug {
@@ -2893,6 +2900,8 @@ mod tests {
     }
 
     #[rstest]
+    #[case::no_whitespace("", (1,1))]
+    #[case::whitespace_without_newlines("\t\t", (1,3))]
     #[case::newlines("\n\r", (3,1))]
     #[case::crlf("\r\n\r\n", (3,1))]
     #[case::mixed("\r\n\n\r\n", (4,1))]
@@ -2900,6 +2909,14 @@ mod tests {
     #[case::mix_tabs_and_newlines("\n\t\n", (3,1))]
     fn expect_whitespace(#[case] input: &str, #[case] expected_location: (usize, usize)) {
         MatchTest::new_1_0(input).expect_match_location(match_length(TextBuffer::match_whitespace0), expected_location);
+    }
+
+    #[rstest]
+    #[case::no_whitespace("", (1,1))]
+    #[case::whitespace_with_inline_comment("//comment ", (1,11))]
+    #[case::whitespace_with_comment("/*comment*/ \n", (2,1))]
+    fn expect_whitespace_with_comment(#[case] input: &str, #[case] expected_location: (usize, usize)) {
+        MatchTest::new_1_0(input).expect_match_location(match_length(TextBuffer::match_optional_comments_and_whitespace), expected_location);
     }
 
     #[test]
--- a/src/lazy/text/buffer.rs
+++ b/src/lazy/text/buffer.rs
@@ -88,8 +88,6 @@ impl Debug for TextBuffer<'_> {
 /// '\x0C', Form feed
 pub(crate) const WHITESPACE_BYTES: &[u8] = b" \t\r\n\x09\x0B\x0C";
 
-pub(crate) const NON_NEWLINE_BYTES: &[u8] = b"\t\x09\x0B\x0C";
-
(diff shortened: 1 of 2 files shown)


src/lazy/text/buffer.rs
```diff
@@ -245,16 +269,44 @@ impl<'top> TextBuffer<'top> {
...
     /// Matches zero or more whitespace characters.
     pub fn match_whitespace0(&mut self) -> IonMatchResult<'top> {
-        take_while(0.., WHITESPACE_BYTES).parse_next(self)
+        let result = take_while(0.., WHITESPACE_BYTES).parse_next(self)?;
+        self.update_location_metadata(result.data);
+        Ok(*self)
     }
 
     /// Matches any amount of contiguous comments and whitespace, including none.
-    #[inline]
```
zslayton, line 257: What was the motivation for removing this?

src/lazy/text/buffer.rs
```diff
@@ -1987,11 +2042,20 @@ impl<'data> Stream for TextBuffer<'data> {
...
+        checkpoint
     }
 
     fn reset(&mut self, checkpoint: &Self::Checkpoint) {
+        let current_row = self.row;
+        let prev_column_value = self.prev_newline_offset;
+
         *self = *checkpoint;
+        self.row = current_row + checkpoint.row;
+        self.prev_newline_offset = prev_column_value + checkpoint.prev_newline_offset;
```
zslayton, line 2058: In `TextBuffer`'s case, `Self::Checkpoint` is `TextBuffer`. Saving with `checkpoint` copies `self`, and restoring with `reset` overwrites `self` with `checkpoint`. As a result, I believe that this method should just be `*self = *checkpoint`.

src/lazy/text/buffer.rs
```diff
@@ -2183,6 +2248,26 @@ mod tests {
...
 
+        fn expect_match_location<'data, P, O>(&'data self, parser: P, expected_location: (usize, usize))
+        where
+            P: Parser<TextBuffer<'data>, O, IonParseError<'data>>,
+        {
+            let result = self.try_match(parser).unwrap_or_else(|e| {
+                panic!("Unexpected parse fail for input <{}>\n{e}", self.input)
+            });
+            let match_length = result.1;
+            // Inputs have a trailing newline and `0` that should _not_ be part of the match
```
zslayton, line 2259: This was a hack from an old implementation of the text reader that couldn't properly detect incompleteness. It's no longer needed and can also be removed from where you copied it.

src/lazy/text/buffer.rs
```diff
@@ -2157,22 +2221,23 @@ mod tests {
             self
         }
 
-        fn try_match<'data, P, O>(&'data self, parser: P) -> IonParseResult<'data, usize>
+        fn try_match<'data, P, O>(&'data self, parser: P) -> IonParseResult<'data, (TextBuffer<'data>, usize)>
```
zslayton, line 2224: Why was this needed?

src/lazy/text/buffer.rs
```diff
@@ -2157,22 +2221,23 @@ mod tests {
...
         fn expect_match<'data, P, O>(&'data self, parser: P)
         where
             P: Parser<TextBuffer<'data>, O, IonParseError<'data>>,
         {
-            let result = self.try_match(parser);
-            let match_length = result.unwrap_or_else(|e| {
+            let result = self.try_match(parser).unwrap_or_else(|e| {
                 panic!("Unexpected parse fail for input <{}>\n{e}", self.input)
             });
+            let match_length = result.1;
```
zslayton, line 2240: It seems like `try_match` was modified to also return the buffer, but this just discards it?

src/lazy/text/buffer.rs
```diff
@@ -2817,6 +2902,21 @@ mod tests {
         assert!(contains_escapes);
     }
 
+    #[rstest]
+    #[case::newlines("\n\r", (3,1))]
+    #[case::crlf("\r\n\r\n", (3,1))]
+    #[case::mixed("\r\n\n\r\n", (4,1))]
+    #[case::tabs("\n\t\t\t", (2,4))]
+    #[case::mix_tabs_and_newlines("\n\t\n", (3,1))]
```
zslayton, line 2910: Could you please add test cases for:
* No whitespace
* Whitespace without newlines
* Whitespace that includes a comment
* Whitespace that includes a multi-line comment.

src/lazy/text/buffer.rs
desaikd: I changed try_match to return buffer and length so that I can just use the same method for both usecases, 1. checking length and 2. checking row-column values are right.

src/lazy/text/buffer.rs
desaikd: Yes, I added buffer as return for using it in expect_match_location. And I changed try_match to return buffer and length so that I can just use the same method for both usecases, 1. checking length and 2. checking row-column values are right.

src/lazy/text/buffer.rs
```diff
@@ -1987,11 +2042,20 @@ impl<'data> Stream for TextBuffer<'data> {
     }
 
     fn checkpoint(&self) -> Self::Checkpoint {
-        *self
+        let mut checkpoint = *self;
+        // Reset row, column at checkpoint
+        checkpoint.row = 0;
+        checkpoint.prev_newline_offset = 0;
```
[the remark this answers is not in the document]
desaikd: I thought the checkpoint was not saving states for row and column values properly. But upon removing my changes for `checkpoint` and `reset` like you said it correctly saves the state of self at checkpoint and reset copies the state point checkpoint correctly.

src/lazy/text/buffer.rs
desaikd: I thought the checkpoint was not saving states for row and column values properly. But upon removing my changes for checkpoint and reset like you said it correctly saves the state of self at checkpoint and reset copies the state point checkpoint correctly.

src/lazy/text/buffer.rs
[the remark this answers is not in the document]
desaikd: I will remove the reset I am doing here.

src/lazy/text/buffer.rs
```diff
@@ -1588,6 +1641,8 @@ impl<'top> TextBuffer<'top> {
         let delimiter_head = delimiter.as_bytes()[0];
         // Whether we've encountered any escapes while looking for the delimiter
         let mut contained_escapes = false;
+        // This input may contain newline characters hence update the location metadata.
+        self.update_location_metadata(self.bytes());
```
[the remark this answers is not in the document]
desaikd: What did you mean by match? Are you saying we should perform a match first and then based on that count newline char counts?

src/lazy/text/buffer.rs
```diff
@@ -446,6 +498,7 @@ impl<'top> TextBuffer<'top> {
         )
             .map(|(maybe_annotations, value)| input.apply_annotations(maybe_annotations, value))
             .parse_next(self)
+        // let length = result.with_span().map(|(_output, consumed)| consumed.len());
```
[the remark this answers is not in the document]
desaikd: Added that while debugging, will remove that

src/lazy/text/buffer.rs
desaikd: I didn't intend to remove this. while implementing I had added some helper methods there and then had refactored back to the original methods. Will add it back.

src/lazy/text/buffer.rs
```diff
@@ -245,16 +269,44 @@ impl<'top> TextBuffer<'top> {
...
+    fn update_location_metadata(&mut self, data: &'top [u8]) {
+        if !data.is_empty() {
+            // If the bytes contain '\r\n' in this order then this must be coming from windows line ending pattern and hence should be counted as 1.
+            let crlf_count = data.windows(2).filter(|window| window == b"\r\n").count();
+
+            // Subtract the crlf_count from total count of all newline characters to get the correct number of newline match count.
+            let newline_match_count = data.iter().filter(|b| NEWLINE_BYTES.contains(b)).count() - crlf_count;
+
+            // Gets index for the last occurrence of the newline byte and subtracts from the result length to get non newline bytes length
+            let last_index_newline_byte = data.iter().rposition(|b| NEWLINE_BYTES.contains(b)).unwrap_or(0);
```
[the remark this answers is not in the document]
desaikd: Hmm. Sure, can take a look at that suggestion and benchmark.

src/lazy/text/buffer.rs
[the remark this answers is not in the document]
zslayton: The `loop` below is scanning ahead looking for the closing delimiter. Imagine that `self` is 100 bytes, and the closing delimiter is found 8 bytes in. You want to update the newline count for newlines found in the first 8 bytes, not the full 100 bytes.

src/lazy/text/buffer.rs
desaikd: I think I found out why I had identified the row and column values were incorrect with checkpoint reset. But if you think I am missing something here do let me know.
Whenever I think the input is taken, I followed the path and seemed to find [this place](https://docs.rs/winnow/latest/src/winnow/combinator/impls.rs.html#399) where it saves the checkpoint before parsing and then updates with reset based on that saved checkpoint. Which means nothing gets updated (offset, row or column). But the offset is explicitly updated with a call to `next_slice` passing the updated offset value after parsing. However, row and column stays the same as prior to parsing which means it doesn't get updated properly unless we add the saved state to the current state like how I do in the `reset` here.

src/lazy/text/buffer.rs
```diff
@@ -245,12 +269,42 @@ impl<'top> TextBuffer<'top> {
...
+            let non_newline_match_length = data.len() - last_index_newline_byte;
+            self.row += newline_match_count;
+
+            // Stores this newline offset as previous newline offset for calculating column position since this has already been matched/parsed
+            self.prev_newline_offset = if self.offset < non_newline_match_length {
+                // this means that the input is not yet consumed hence get the input length + the current offset
+                self.offset + data.len() - non_newline_match_length
+            } else {
+                self.offset - non_newline_match_length
+            };
```
popematt, line 299: This is performing 3 different iterations over the data when only one is really needed. We should be able to do something like this where we have essentially a mini state machine that would allow us to make only one pass over the data.
```rust
// The accumulator for fold is:
// (is this byte following a carriage return, num of rows found, position of last newline)
let (_, rows, prev_newline_offset) = data.iter().enumerate().fold((0, 0, false), |(follows_cr, rows, offset), (i, b) | {
    match (b, follows_cr) {
        // When there's a "\r", add a row and update the offset
        (b"\r", _)     -> (true,  rows + 1, i),
        // When there's a "\n" not after "\r", add a row and update the offset
        (b"\n", false) -> (false, rows + 1, i),
        // When there's "\n" immediately following "\r", update the offset without adding a row
        (b"\n", true)  -> (false, rows,     i),
        _              -> (false, rows,     offset), 
    }
});
if rows > 0 {
  self.prev_newline_offset = self.offset - data.len() + prev_newline_offset;
  self.row += rows;
}
```

src/lazy/text/buffer.rs
```diff
@@ -206,6 +214,22 @@ impl<'top> TextBuffer<'top> {
...
+        self.row
+    }
+
+    /// Returns the column position for this buffer.
+    /// _Note: Column positions are calculated based on current offset and previous newline byte offset._
+    pub fn column(&self) -> usize {
+        self.offset - self.prev_newline_offset + 1
+    }
+
+    pub fn prev_newline_offset(&self) -> usize {
```
popematt, line 229: I think we should probably hide the implementation details and only expose the row/col functions. Is there a specific reason this function needs to exist?

src/lazy/text/buffer.rs
```diff
@@ -1991,7 +2047,12 @@ impl<'data> Stream for TextBuffer<'data> {
     }
 
     fn reset(&mut self, checkpoint: &Self::Checkpoint) {
+        let current_row = self.row;
+        let prev_column_value = self.prev_newline_offset;
+
         *self = *checkpoint;
+        self.row = current_row;
+        self.prev_newline_offset = prev_column_value;
```
popematt, line 2055: Can you explain what's going on here? If we're reseting to a specific checkpoint, why would we _not_ want to reset the row/col accordingly?

src/lazy/text/buffer.rs
```diff
@@ -2817,6 +2899,31 @@ mod tests {
...
     }
 
+    #[rstest]
+    #[case::no_whitespace("", (1,1))]
+    #[case::whitespace_without_newlines("\t\t", (1,3))]
+    #[case::newlines("\n\r", (3,1))]
+    #[case::crlf("\r\n\r\n", (3,1))]
+    #[case::mixed("\r\n\n\r\n", (4,1))]
+    #[case::tabs("\n\t\t\t", (2,4))]
+    #[case::mix_tabs_and_newlines("\n\t\n", (3,1))]
```
popematt, line 2909: I think we can be both more targeted with our test cases (i.e. let's specify the purpose/goal of the test cases) and a little more exhaustive

Here are the cases I would check. What do you think?
```    
// No whitespace
"", (1, 1)

// No CR or LF
"\t", (1, 2) 
"\t\t", (1, 3) 

// One LF
"\n", (2, 1)

// One CR
"\r", (2, 1)

// One CRLF
"\r\n", (2, 1)

// All possible combinations of 3 CR/LF; check correct number of rows
"\n\n\n", (4, 1)
"\n\n\r", (4, 1)
"\n\r\n", (4, 1)
"\n\r\r", (4, 1)
"\r\n\n", (3, 1)
"\r\n\r", (3, 1)
"\r\r\n", (3, 1)
"\r\r\r", (4, 1)

// Whitespace after newline does affect column count
"\n\t", (2, 2)
"\n\t\t", (2, 3)

// Whitespace before newline does not affect column count
"\t\n", (2, 1)
"\t\n\t", (2, 2)
"\t\t\n", (2, 1)
"\t\t\n\t", (2, 2)

// Whitespace between CR and LF is two separate row endings
"\r\t\n", (3, 1)
```

src/lazy/text/buffer.rs
```diff
@@ -2817,6 +2899,31 @@ mod tests {
...
+    fn expect_whitespace(#[case] input: &str, #[case] expected_location: (usize, usize)) {
+        MatchTest::new_1_0(input).expect_match_location(match_length(TextBuffer::match_whitespace0), expected_location);
+    }
+
+    #[rstest]
+    #[case::no_whitespace("", (1,1))]
+    #[case::whitespace_with_inline_comment("//comment ", (1,11))]
+    #[case::whitespace_with_comment("/*comment*/ \n", (2,1))]
```
popematt, line 2917: Suggested test cases:
* Empty input
* `//` comment by itself
* `//` comment preceded by a newline
* `//` comment followed by a newline
* `//` comment followed by another `//` comment on a new line
* `/**/` comment on a single line
* `/**/`  comment preceded by a newline
* `/**/` comment followed by a newline
* `/**/` comment containing one newline
* `/**/` comment containing two newlines

src/lazy/text/buffer.rs
```diff
@@ -2817,6 +2899,31 @@ mod tests {
...
+    fn expect_whitespace_with_comment(#[case] input: &str, #[case] expected_location: (usize, usize)) {
+        MatchTest::new_1_0(input).expect_match_location(match_length(TextBuffer::match_optional_comments_and_whitespace), expected_location);
+    }
+
+    #[test]
+    fn expect_newline_long_text() {
```
popematt, line 2923: Suggested test cases
* Empty single-segment long string.
* Single segment long string containing only `\n`
* Single segment long string containing other stuff
* Two-segment long string on a single line (i.e. `'''a''' '''b'''`)
* Two-segment long string on separate lines (i.e. `'''a'''\n'''b'''`)
* Two-segment long string containing new lines (i.e. `'''a\n''' '''b\n'''`)
* Two-segment long string containing new lines and on separate lines (i.e. `'''a\n'''\n'''b\n'''`)

Also, I think it's probably worth adding (or generalizing this to include) some basic test cases for blobs, clobs, list, sexp, and structs that contain newlines; newlines between annotations; and newlines between field names and field values.

src/lazy/text/buffer.rs
popematt: Also, is it feasible to add/update a unit test for this function to include the row/col reset behavior?

src/lazy/text/buffer.rs
desaikd: I intended to add this for wherever buffer gets created some states are being saved in the reader implementation. so similar to how the offset is being saved we would have to do it for `row` and `prev_column_value` eventually. 
However this can be `pub(crate)`. 
Here's the saved state usage which stores offset right now but should have row and column later: https://github.com/amazon-ion/ion-rust/blob/main/src/lazy/any_encoding.rs#L479
When we will add implementation for adding this information into the `LazyValue` we will need this info and which is why I added it here. 
I am happy to remove this method entirely if thats confusing and will add it when we will require while implementing support for location with `LazyValue`.

src/lazy/text/buffer.rs
desaikd: My understanding of why this is required: 
https://github.com/amazon-ion/ion-rust/pull/905#discussion_r1949866069

src/lazy/text/buffer.rs
desaikd: For all your suggestions on testcases. I have only included the ones that were basic and I understand there could be more test cases here. We already have test suite which runs for all read operations for all these cases which is why for just row and column counts I carved out only a few test cases. 
I am happy to add these cases though!

popematt: The conformance test suite we have doesn't do any verification of row/col of the values, which is why we need to cover a variety of test cases here. We don't need to cover all possible cases, but we should try to cover the basic permutations.

> And I thought as we also will be adding test cases at the LazyValue level where we can be more exhaustive.

A unit test should (as much as possible) test a unit of code in isolation. We can make sure that the logic is working correctly here, and then we have more confidence to build upon this. If we have good tests here, then the tests for `LazyValue` can assume that this functionality works correctly, and so they only need to make sure that the transfer of information from `TextBuffer` to `LazyValue` is happening correctly.

desaikd: Sure. I will add the test cases.

src/lazy/text/buffer.rs
desaikd: I think this discussion is going beyond the scope of this PR. I will just remove this method for now. And we can decide on keeping or not when it is actually required by other places.

popematt (approved):

src/lazy/text/buffer.rs
desaikd: `next_slice` only takes the `offset` as input. And this definition is based on [Stream](https://docs.rs/winnow/latest/winnow/stream/trait.Stream.html#tymethod.next_slice) trait. We need to pass the row and column values and thats not possible with this method.

popematt: Let's create an issue to follow up on this, and move on for now, then.

zslayton (approved):

PR #910, opened 2025-02-18
Adds experimental-location feature flag

desaikd:
*Issue #, if available:*

*Description of changes:*
This PR works on adding an `experimental-location` feature flag for the location-metadata implementation. This allows users to opt-in based on their performance requirements.

*List of changes:*
- define `experimnetal-location` feature flag in the `Cargo.toml`.
- add configuration wherever row and column properties are used or updated in the `TextBuffer`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ default = []
 experimental-ion-hash = ["digest", "experimental-reader-writer"]
 
 # Access location information of the input Ion from underlying buffer.
-experimental-location = []
+source-location = []
 
 # Feature for indicating particularly bleeding edge APIs or functionality in the library.
 # These are not guaranteed any sort of API stability and may also have non-standard
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,9 @@ rust-version = "1.80"
 default = []
 experimental-ion-hash = ["digest", "experimental-reader-writer"]
 
+# Access location information of the input Ion from underlying buffer.
+experimental-location = []
+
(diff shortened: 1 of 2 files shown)


src/lazy/text/buffer.rs
```diff
@@ -2264,6 +2281,7 @@ mod tests {
                 self.input,
                 &self.input[..match_length]
             );
+            #[cfg(feature = "experimental-location")]
```
popematt, line 2284: Just a suggestion—you could get rid of the warnings on this function by removing this and putting it at the start of this function. I.e.:

```rust
#[cfg(not(feature = "experimental-location"))]
fn expect_match_location<'data, P, O>(&'data self, _parser: P, _expected_location: (usize, usize)) { }

#[cfg(feature = "experimental-location")]
fn expect_match_location<'data, P, O>(&'data self, parser: P, expected_location: (usize, usize)) {
    // ...
}
```

Cargo.toml
```diff
@@ -24,6 +24,9 @@ rust-version = "1.80"
 default = []
 experimental-ion-hash = ["digest", "experimental-reader-writer"]
 
+# Access location information of the input Ion from underlying buffer.
+experimental-location = []
```
popematt, line 28: Do we want to release this as "experimental"? Can we call it `element-location` instead?

desaikd: Sure. I think it should be okay to move this out of experimental. I don't want to attach it with element though. What if we decide to keep it with `LazyValue` and add support for `Element` later. 
Maybe we can just say `source-location`?

popematt (approved):

Issue #902, opened 2025-01-30
Update ion-tests integration to latest version of ion-tests


zslayton:
Our `ion-tests` submodule is currently several commits behind the latest upstream version. `HEAD` includes tests which exercise features that were modified or renamed after the testing DSL was initially implemented.

We need to:
1. Update to the latest version of `ion-tests`.
2. Put together a skip list that documents which tests are failing and why.
3. Create child issues that track the gaps identified.
4. Whittle down the skip list and associated sub-issues until we're caught up.

nirosys:
I've updated `ion-tests`, and added a file oriented skip-list over in #911. I'm putting together a table of all of the conformance test failures for tracking.

PR #911, opened 2025-02-19
Update ion-tests and start cataloging conformance issues.

nirosys:
*Issue #, if available:* #902

*Description of changes:*
This PR adds a skip list to the conformance tests run during CI/CD, while also adding all of the current conformance tests to the generated unit tests.

The skip list is implemented to skip test files, which may contain multiple tests. This was done because tests are not required to have a name, and many of them do not. This means that a test file cannot be removed from the skip list until each test within passes. I'll be tracking all of the individual tests, and can post a more detailed list, which I will also be sharing with the team shortly. The comments captured in the skip list are mostly the first error to trigger the test failure. I'll keep them updated as I work through them, but the list I'll share will be the most accurate summary of issues.

In addition the conformance CLI has been updated to catch panics and recover, so that we can run all tests within a file even if one of the tests hits a `todo!`, or fatal error.

The CLI has also been updated to allow a `-i` option, which can be specified before any files in order to output the test results in an ion format for easier processing.

---

### DIFF
--- a/tests/conformance_tests.rs
+++ b/tests/conformance_tests.rs
@@ -155,7 +155,8 @@ mod ion_tests {
     #[test_resources("ion-tests/conformance/system_macros/*.ion")]
     #[test_resources("ion-tests/conformance/tdl/*")]
     fn conformance(file_name: &str) {
-        if !GLOBAL_CONFORMANCE_SKIPLIST.iter().any(|f| *f == file_name) {
+        // Test for skip list. Convert windows '\\' separators into '/' to match skiplist.
+        if !GLOBAL_CONFORMANCE_SKIPLIST.iter().any(|f| *f == file_name.replace("\\", "/")) {
             println!("TESTING: {}", file_name);
             let collection = TestCollection::load(file_name).expect("unable to load test file");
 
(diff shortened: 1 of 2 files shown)


nirosys:
Unit test failures on windows due to skip-list not allowing windows separators. As much as I try to forget windows exists.. it's still there.

tests/conformance_tests.rs
```diff
@@ -79,17 +79,90 @@ mod implementation {
...
+    #[test_resources("ion-tests/conformance/data_model/*.ion")]
+    #[test_resources("ion-tests/conformance/eexp/*.ion")]
+    #[test_resources("ion-tests/conformance/eexp/binary/*.ion")]
+    #[test_resources("ion-tests/conformance/ion_encoding/*.ion")]
+    #[test_resources("ion-tests/conformance/ion_encoding/module/*.ion")]
+    #[test_resources("ion-tests/conformance/ion_encoding/module/macro/cardinality/*.ion")]
+    #[test_resources("ion-tests/conformance/ion_encoding/module/macro/template/*.ion")]
+    #[test_resources("ion-tests/conformance/ion_encoding/module/macro/trivial/*.ion")]
+    #[test_resources("ion-tests/conformance/system_macros/*.ion")]
+    #[test_resources("ion-tests/conformance/tdl/*")]
```
zslayton, line 156: Does this macro support `**` syntax for nesting at any depth? (Or maybe you're enumerating the subdirectories to leave something out?)

zslayton (approved):

Issue #914, opened 2025-02-27
Conformance: Text fragment generation of macro invocations merge macro id and first argument.


nirosys:
Using a toplevel clause such as:
```
(toplevel ('#$:make_list' 1 2))
```
Using `make_list` like this is wrong based on the docs, but it highlights the issue.

```
ConformanceError(ConformanceErrorImpl {
   file: "",
   test_name: "",
   kind: IonError(Decoding(DecodingError { 
      description: "could not find macro with id LocalName(\"make_list1\")\n    while reading a v1.1 top-level expression\n        offset=30\n        input=  2  )  \n        buffer len=8\n        ",
      position: Some(Position { byte_offset: 30, byte_length: Some(8), line_column: None })
   }))
})
```

nirosys:
Fixed with #919.

Issue #912, opened 2025-02-19
Implement E-Expression Transcription


nirosys:
There are quite a few conformance test failures due to the e-expression invocation notation not being supported yet.

We have a total of 323 tests. Of those, 147 are passing, and 51 are currently failing due to e-expression transcription.

nirosys:
## List of Failing Tests
|Test File|Test Name|Test Status|Error|
|-|-|-|-|
|tdl/if_single.ion|when invoked in TDL,|Panic|not yet implemented: e-expression transcription
|tdl/if_single.ion|when the first argument is|Panic|not yet implemented: e-expression transcription
|tdl/if_single.ion|when the first argument is a single value|Panic|not yet implemented: e-expression transcription
|tdl/if_single.ion|when the first argument is none|Panic|not yet implemented: e-expression transcription
|ion_encoding/module/macro/cardinality/invoke_cardinality_ee.ion|! parameters|Panic|not yet implemented: e-expression transcription
|ion_encoding/module/macro/cardinality/invoke_cardinality_ee.ion|? parameters|Panic|not yet implemented: e-expression transcription
|ion_encoding/module/macro/cardinality/invoke_cardinality_ee.ion|+ parameters|Panic|not yet implemented: e-expression transcription
|ion_encoding/module/macro/cardinality/invoke_cardinality_ee.ion|* parameters|Panic|not yet implemented: e-expression transcription
|ion_encoding/mactab.ion|top-level macro_table|Panic|not yet implemented: e-expression transcription
|ion_encoding/module/macro/trivial/invoke_tl.ion|Local macros shadow system macros|Panic|not yet implemented: e-expression transcription
|ion_encoding/module/macro/trivial/invoke_tl.ion|Qualified references|Panic|not yet implemented: e-expression transcription
|ion_encoding/module/macro/trivial/invoke_tl.ion|Local references|Panic|not yet implemented: e-expression transcription
|ion_encoding/module/macro/trivial/invoke_tl.ion|Local names shadow `use`d names|Panic|not yet implemented: e-expression transcription
|tdl/if_none.ion|when invoked in TDL,|Panic|not yet implemented: e-expression transcription
|tdl/if_none.ion|when the first argument is|Panic|not yet implemented: e-expression transcription
|tdl/if_none.ion|when the first argument is none|Panic|not yet implemented: e-expression transcription
|tdl/if_none.ion|when the first argument is some|Panic|not yet implemented: e-expression transcription
[...]

PR #915, opened 2025-02-27
Implement macro invocation transcription

nirosys:
*Issue #, if available:* #912

*Description of changes:*
This PR updates the conformance DSL to allow macro invocation transcription. This allows a `toplevel` clause to contain s-expressions of the form `('#$:foo' 1 2)`, where the `#$` prefix denotes the document being generated by the fragment should insert a macro invocation for the macro ID `foo`, using the remaining values as macro arguments.

This first PR provides support for numeric symbol IDs and text IDs when generating a document using text encoding. Due to the use of raw writers, while generating a binary document text IDs need an additional lookup. This PR does not implement that but I intend to implement it after this PR. In order to lessen the impact, I've also changed the default format to text with this PR, as many tests seem to not define a format. 

The conformance DSL describes `(mactab ..)` as equal to defining a module named `M` to contain the macros defined within the clause. A lot of the tests that use this clause reference the macros via `('#$:M:name' ..)`, since we currently only support the default module, special handling has not been implemented for this. For text fragments it should pass through fine, but support for binary formats will need to be added. The tests themselves, that use the `M` module, can either be left as-is or updated to use the default module; will open this for discussion.

One other minor change in this PR is the move of the conformance DSL unit tests into the `conformance_dsl` module. This makes testing the dsl itself a little easier by filtering on `conformance_dsl::tests`.

### Conformance Update
There are 323 conformance tests. Prior to this PR 176 tests were failing either due to panics or conformance errors. After this PR, 40 individual tests now pass, leaving 139 tests in a failure state.

While working through this, I've identified an issue with some macro invocations where the text encoder appears to be joining the first argument and the macro ID. I've created #914 to track this.

---


tests/conformance_dsl/fragment.rs
```diff
@@ -358,7 +358,19 @@ impl WriteAsIon for ProxyElement<'_> {
                     Some(Symbol(symbol)) if symbol.text().is_some() => {
                         let text = symbol.text().unwrap();
                         if text.starts_with("#$:") {
-                            todo!("e-expression transcription")
+                            let macro_id = text.strip_prefix("#$:").unwrap(); // SAFETY: Tested above.
```
zslayton, line 361: Personally, I associate `// SAFETY: ...` with uses of `unsafe`.

zslayton (approved):
🚀

tests/conformance_dsl/fragment.rs
nirosys: Yea, I can see that; wondering if using `expect` would be a better self-documenting way to highlight that it is a justified unwrap.

PR #918, opened 2025-02-28
Introduces a `LazyElement` API

zslayton:
This PR introduces a new API: `LazyElement`. 

Like `LazyValue<'_, Encoding>`, `LazyElement` does not read or materialize any of its contained data after identifying the value in input. Unlike `LazyValue`, `LazyElement` uses reference counting to share ownership of its backing resources with the `Reader`. This requires a small amount of opt-in bookkeeping overhead, but allows `LazyElement`s to be read in any order, stored indefinitely, and used beyond the lifetime of the originating `Reader`. It also enables the `Reader` to implement the `Iterator` trait, which makes writing lots of application code simpler.

For example usages, see the associated unit tests:
https://github.com/amazon-ion/ion-rust/blob/261bf6c4044c33221b2a871f1d39a67eb9e241f5/src/lazy/expanded/lazy_element.rs#L273-L324

We may wish to rename `LazyValue` for improved consistency with other naming conventions. `Element`/`LazyElement`/`LazyValue` feels a bit off to me—perhaps `Element`/`LazyElement`/`LazyElementRef` would be better.


src/lazy/binary/binary_buffer.rs
```diff
@@ -7,10 +7,10 @@ use crate::binary::constants::v1_0::{length_codes, IVM};
 use crate::binary::int::DecodedInt;
 use crate::binary::var_int::VarInt;
 use crate::binary::var_uint::VarUInt;
-use crate::lazy::binary::encoded_value::EncodedValue;
+use crate::lazy::binary::encoded_value::EncodedBinaryValue;
 use crate::lazy::binary::raw::r#struct::LazyRawBinaryFieldName_1_0;
 use crate::lazy::binary::raw::type_descriptor::{Header, TypeDescriptor, ION_1_0_TYPE_DESCRIPTORS};
-use crate::lazy::binary::raw::v1_1::immutable_buffer::AnnotationsEncoding;
+use crate::lazy::binary::raw::v1_1::binary_buffer::AnnotationsEncoding;
```
zslayton, line 13: 🪧 The first commit in this PR did a mechanical rename of the file `immutable_buffer` to `binary_buffer` to align with the name of the primary type therein. I suggest skipping the first commit.

src/lazy/binary/encoded_value.rs
```diff
@@ -40,7 +40,7 @@ impl EncodedHeader for Header {
 /// allowing a user to re-read (that is: parse) the body of the value as many times as necessary
 /// without re-parsing its header information each time.
 #[derive(Clone, Copy, Debug, PartialEq)]
-pub(crate) struct EncodedValue<HeaderType: EncodedHeader> {
+pub struct EncodedBinaryValue<HeaderType: EncodedHeader> {
```
zslayton, line 43: 🪧 Prior to this change there was an `EncodedTextValue` and an `EncodedValue`. I renamed the latter for symmetry/clarity.

src/lazy/encoding.rs
```diff
@@ -118,8 +119,8 @@ pub struct BinaryEncoding_1_0;
 #[derive(Copy, Clone, Debug, Default)]
 pub struct BinaryEncoding_1_1;
 
-impl BinaryEncoding<'_> for BinaryEncoding_1_0 {}
-impl BinaryEncoding<'_> for BinaryEncoding_1_1 {}
+impl BinaryEncoding for BinaryEncoding_1_0 {}
+impl BinaryEncoding for BinaryEncoding_1_1 {}
```
zslayton, line 123: 🪧 The lifetime in the `BinaryEncoding<'_>` and `TextEncoding<'_>` traits turned out to be unnecessary. 

This required using higher-ranked generics in the trait implementations, but keeping the lifetime out of the trait name simplified a lot of other changes.

src/lazy/encoding.rs
```diff
@@ -241,51 +242,51 @@ impl Encoding for TextEncoding_1_1 {
 }
 
 /// Marker trait for binary encodings of any version.
-pub trait BinaryEncoding<'top>: Encoding<Output = Vec<u8>> + Decoder {}
+pub trait BinaryEncoding: Encoding<Output = Vec<u8>> + Decoder {}
 
 /// Marker trait for text encodings.
-pub trait TextEncoding<'top>:
+pub trait TextEncoding:
```
zslayton, line 248: 🪧 The rest of the changes in this file move the lifetime to the method signatures.

src/lazy/reader.rs
```diff
@@ -139,17 +131,117 @@ impl<Encoding: Decoder, Input: IonInput> Reader<Encoding, Input> {
...
 
     pub fn register_template(&mut self, template_macro: TemplateMacro) -> IonResult<MacroAddress> {
-        self.system_reader.expanding_reader.register_template(template_macro)
+        self.system_reader
+            .expanding_reader
+            .register_template(template_macro)
+    }
+}
+
+impl<Encoding: Decoder, Input: IonInput> Iterator for Reader<Encoding, Input> {
```
zslayton, line 140: 🪧 `Reader` now implements `Iterator`, which makes a LOT of user code simpler.

src/lazy/span.rs
```diff
@@ -63,6 +68,26 @@ impl<'a> Span<'a> {
     pub fn is_empty(&self) -> bool {
         self.bytes.is_empty()
     }
+
+    pub fn slice(&self, offset: usize, length: usize) -> Span<'a> {
```
zslayton, line 72: 🪧 This PR modified the `TextBuffer` type to use a `Span` internally. A future PR will do the same for `BinaryBuffer`.

src/lazy/streaming_raw_reader.rs
```diff
@@ -241,6 +247,11 @@ impl<Encoding: Decoder, Input: IonInput> StreamingRawReader<Encoding, Input> {
                 }
             }
 
+            // At this point, `self.input` is no longer being modified.
+            // We can hand out a reference to it that's good for the duration of 'top.
+            let io_buffer_handle: &dyn IoBufferHandle = self.input.get_mut();
+            unsafe { context.set_io_buffer_handle(io_buffer_handle) };
```
zslayton, line 253: 🪧 When the streaming reader has finished modifying the input buffer, it gets a fat pointer to it and stores it in the `EncodingContext`. If the user asks to make a `LazyElement`, it will use that fat pointer to get a reference-counted copy of the input buffer.

src/lazy/streaming_raw_reader.rs
```diff
@@ -325,19 +374,159 @@ impl<SliceType: AsRef<[u8]>> IonDataSource for IonSlice<SliceType> {
...
         );
     }
+
+    fn position(&self) -> usize {
+        self.position
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct IoBuffer {
```
zslayton, line 384: 🪧 In order for fixed slice inputs and streaming inputs to have a common language for saving input buffers, I reworked things to use this type consistently. `IonSlice` (a fixed input) will avoid constructing an `IoBuffer` as an optimization, but once it's been created it can be re-used by every value in the stream.

src/lazy/text/matched.rs
```diff
@@ -1232,7 +1232,7 @@ mod tests {
             let expected: Int = expected.into();
             let encoding_context = EncodingContext::empty();
             let context = encoding_context.get_ref();
-            let mut buffer = TextBuffer::new(context, data.as_bytes(), true);
+            let mut buffer = TextBuffer::new(context, data.as_bytes());
```
zslayton, line 1235: 🪧 `TextBuffer::new` is only used in the unit tests where `is_final_data` is always `true`, so I removed that argument.

src/lazy/text/raw/sequence.rs
```diff
@@ -21,11 +21,11 @@ use crate::{IonResult, IonType};
 // ===== Lists =====
 
 #[derive(Copy, Clone)]
-pub struct RawTextList<'data, E: TextEncoding<'data>> {
+pub struct RawTextList<'data, E: TextEncoding> {
```
zslayton, line 24: 🪧 Throughout this file you can see the benefit of removing the lifetime from the `TextEncoding` trait.

src/lazy/text/value.rs
```diff
@@ -21,21 +21,24 @@ use std::ops::Range;
 /// includes a text definition for these items whenever one exists, see
 /// [`crate::lazy::value::LazyValue`].
 #[derive(Copy, Clone)]
-pub struct LazyRawTextValue<'top, E: TextEncoding<'top>> {
-    pub(crate) encoded_value: EncodedTextValue<'top, E>,
+pub struct LazyRawTextValue<'top, Encoding: TextEncoding> {
```
zslayton, line 24: 🪧 I tried to replace a lot of one-letter generics with a more meaningful name.

src/lazy/value.rs
```diff
@@ -260,6 +263,52 @@ impl<'top, D: Decoder> LazyValue<'top, D> {
     pub fn read(&self) -> IonResult<ValueRef<'top, D>> {
         self.expanded_value.read_resolved()
     }
+
+    pub(crate) fn context(&self) -> EncodingContextRef<'top> {
+        self.expanded_value.context()
+    }
+
+    pub fn to_owned(&self) -> LazyElement<D> {
```
zslayton, line 271: 🪧 This method is key to the whole API. By realizing it boiled down to this, I avoided needing to make a big family of `LazyElement` types to mirror the `LazyValue` family of types.

zslayton (commented):
🗺️ PR Tour 🧭

src/lazy/text/encoded_value.rs
```diff
@@ -40,14 +40,24 @@ pub struct EncodedTextValue<'top, E: TextEncoding<'top>> {
...
+impl<'top, E: TextEncoding> EncodedTextValue<'top, E> {
     pub(crate) fn new(matched_value: MatchedValue<'top, E>) -> EncodedTextValue<'top, E> {
         EncodedTextValue {
             data_offset: 0,
             matched_value,
         }
     }
 
+    pub(crate) fn with_offset(
+        offset: u16,
```
popematt, line 52: Just curious—why `u16` specifically (instead of `usize`, for example)?

src/unsafe_helpers.rs
```diff
@@ -37,3 +37,9 @@ pub(crate) fn mut_ref_to_ptr<T>(reference: &mut T) -> *mut () {
     let untyped_ptr: *mut () = ptr.cast();
     untyped_ptr
 }
+
+pub fn ref_as_ptr<T>(reference: &mut T) -> *mut () {
```
popematt, line 41: Should this be `pub(crate)`?
How is this different from `mut_ref_to_ptr()`? They appear to have the same signature and implementation.

popematt (commented):
It would be nice (although certainly not required for this PR) to update some of the benchmarks to include `LazyElement`. It might also be worth trying to update the ion-tests runner to exercise `LazyElement` as well.

src/lazy/binary/raw/v1_1/value.rs
```diff
@@ -282,6 +282,16 @@ impl<'top> LazyRawValue<'top, BinaryEncoding_1_1> for &'top LazyRawBinaryValue_1
...
     }
+
+    fn with_backing_data(&self, span: Span<'top>) -> Self {
+        let buffer =
+            BinaryBuffer::new_with_offset(self.input.context(), span.bytes(), span.offset());
+        let allocator = self.input.context().allocator();
+        allocator.alloc_with(move || LazyRawBinaryValue_1_1 {
+            input: buffer,
+            ..**self
+        })
```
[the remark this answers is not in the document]
zslayton: Good question!

`BinaryEncoding_1_1` is the only encoding (for the moment!) that uses a `&'top` reference as its `Decoder::Value` associated type:

https://github.com/amazon-ion/ion-rust/blob/a80cb0b91ee81a2b69965d19d8ea6a0557acd011/src/lazy/encoding.rs#L466

The `LazyRaw______Value` types can be pretty large, and passing around copies of them has a noticeable performance impact. Storing the `LazyRawBinaryValue_1_1` in the bump allocator and passing around a reference to it instead of copying the whole thing was a nice performance boost. I need to do the same thing for the other three encodings, but haven't had the opportunity yet.

src/lazy/expanded/mod.rs
```diff
@@ -148,13 +243,22 @@ impl EncodingContext {
...
+            ..
+        } = self;
+        (Rc::make_mut(macro_table), Rc::make_mut(symbol_table))
     }
 }
 
 #[derive(Debug, Copy, Clone)]
 pub struct EncodingContextRef<'top> {
-    context: &'top EncodingContext,
+    pub(crate) context: &'top EncodingContext,
```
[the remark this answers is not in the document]
zslayton: At the moment, that would probably be ok. However, I expect that `EncodingContextRef` will eventually be visible to end-users in some form, so I don't want `EncodingContext` itself to be accessible outside the crate.

src/lazy/expanded/mod.rs
```diff
@@ -508,10 +610,14 @@ impl<Encoding: Decoder, Input: IonInput> ExpandingReader<Encoding, Input> {
         unsafe { &*self.raw_reader.get() }.encoding()
     }
 
+    pub fn save_buffer(&self) -> IoBuffer {
+        unsafe { &*self.raw_reader.get() }.save_buffer()
+    }
```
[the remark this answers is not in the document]
zslayton: This method was unused—it was supplanted by the `save_io_buffer` trait method. I've removed it.

src/lazy/text/encoded_value.rs
zslayton: `EncodedTextValue` is rather large, and a `u16` could be added in the existing padding bytes without increasing its size. The `data_offset` field is just the offset from the beginning of the encoded value (that is, bypassing annotations and interstitial whitespace), so 65k bytes seemed like a healthy limit.

src/unsafe_helpers.rs
zslayton: This was also unused, thanks. Removed.

Years ago when I started this crate I added an `#![allow(dead_code)]` at the crate root because the lint was overly conservative and would complain about unused-but-public methods. I'm sure that's been fixed by now. I'll do another PR after this to re-enable it so we can prune stuff like this.

Issue #909, opened 2025-02-17
Implement the support for location metadata in `LazyValue`


desaikd:
Location metadata can be pulled up to `LazyValue` from the underlying buffer. For binary implementations this will be `None` and for text implementations it will come from `TextBuffer::row()` and `TextBuffer::column()`.

List of changes required for this implementation:
- [ ] Modify `LazyValue` to add a property for location metadata
  - [ ] For binary Ion it would always be None
  - [ ] For text Ion get it from the buffer
  - [ ] Would require having a `HasLocation` trait similar to `HasSpan` and `HasRange` for different encodings (Ion 1.1 text, Ion 1.0 text, etc.) or adding an associated type to the Decoder
- [ ] Write unit tests on `LazyValue` for location metadata

desaikd:
Completed with https://github.com/amazon-ion/ion-rust/pull/941

PR #922, opened 2025-03-03
Adds location support with `LazyValue` API

desaikd:
*Issue #909*

*Description of changes:*
This PR works on adding support for location metadata in `LazyValue` API.

*List of changes:*
- Modifies top level matching logic in buffer to preserve `row` and `prev_newline_offset` count.
- Adds a way to store row and `prev_newline_offset` value in the `RawReaderState`. This helps in preserving saved state whenever resume is used on the streaming reader. 
- Since streaming reader uses readers for each encoding get these values for encoding `Reader` types.
- Add location (usize, usize) to `LazyValue`
- Adds unit tests for location support

### DIFF
--- a/src/lazy/binary/raw/v1_1/value.rs
+++ b/src/lazy/binary/raw/v1_1/value.rs
@@ -292,6 +292,11 @@ impl<'top> LazyRawValue<'top, BinaryEncoding_1_1> for &'top LazyRawBinaryValue_1
             ..**self
         })
     }
+
+    #[cfg(feature = "source-location")]
+    fn location(&self) -> (usize, usize) {
+        (0, 0)
+    }
 }
 
 /// Nested expressions parsed and cached while reading (e.g.) a [`LazyRawBinaryValue_1_1`].
--- a/src/lazy/expanded/mod.rs
+++ b/src/lazy/expanded/mod.rs
@@ -1117,6 +1117,14 @@ impl<'top, Encoding: Decoder> LazyExpandedValue<'top, Encoding> {
         }
         None
     }
+
+    #[cfg(feature = "source-location")]
+    pub fn location(&self) -> Option<(usize, usize)> {
+        if let ExpandedValueSource::ValueLiteral(value) = &self.source {
+            return Some(value.location());
+        }
+        None
+    }
 }
 
 impl<'top, Encoding: Decoder> From<LazyExpandedValue<'top, Encoding>>
--- a/src/lazy/any_encoding.rs
+++ b/src/lazy/any_encoding.rs
@@ -599,7 +599,9 @@ impl<'data> From<LazyRawBinaryReader_1_1<'data>> for LazyRawAnyReader<'data> {
 impl<'data> LazyRawReader<'data, AnyEncoding> for LazyRawAnyReader<'data> {
     fn new(context: EncodingContextRef<'data>, data: &'data [u8], is_final_data: bool) -> Self {
         let encoding = Self::detect_encoding(data);
-        let state = RawReaderState::new(data, 0, is_final_data, encoding);
+        let state = RawReaderState::new(data, 0,         #[cfg(feature = "source-location")]
+        0,         #[cfg(feature = "source-location")]
+        0, is_final_data, encoding);
         LazyRawAnyReader {
             new_encoding: None,
             encoding_reader: RawReaderKind::resume_at_offset(context, state),
@@ -636,6 +638,10 @@ impl<'data> LazyRawReader<'data, AnyEncoding> for LazyRawAnyReader<'data> {
             return RawReaderState::new(
                 reader_state.data(),
                 reader_state.offset(),
+                #[cfg(feature = "source-location")]
+                reader_state.row(),
+                #[cfg(feature = "source-location")]
+                reader_state.prev_newline_offset(),
                 reader_state.is_final_data(),
                 new_encoding,
             );
@@ -684,6 +690,28 @@ impl<'data> LazyRawReader<'data, AnyEncoding> for LazyRawAnyReader<'data> {
             Binary_1_0(r) => r.position(),
             Text_1_1(r) => r.position(),
             Binary_1_1(r) => r.position(),
+        }
+    }
+
+    #[cfg(feature = "source-location")]
+    fn row(&self) -> usize {
+        use RawReaderKind::*;
+        match &self.encoding_reader {
+            Text_1_0(r) => r.row(),
+            Binary_1_0(r) => r.row(),
+            Text_1_1(r) => r.row(),
+            Binary_1_1(r) => r.row(),
+        }
+    }
+
+    #[cfg(feature = "source-location")]
+    fn prev_newline_offset(&self) -> usize {
+        use RawReaderKind::*;
+        match &self.encoding_reader {
+            Text_1_0(r) => r.prev_newline_offset(),
+            Binary_1_0(r) => r.prev_newline_offset(),
+            Text_1_1(r) => r.prev_newline_offset(),
+            Binary_1_1(r) => r.prev_newline_offset(),
         }
     }
 
@@ -1118,6 +1146,16 @@ impl<'top> LazyRawValue<'top, AnyEncoding> for LazyRawAnyValue<'top> {
                     LazyRawValueKind::Binary_1_1(v.with_backing_data(span))
                 }
             },
+        }
+    }
+
+    #[cfg(feature = "source-location")]
+    fn location(&self) -> (usize, usize) {
+        match &self.encoding {
+            LazyRawValueKind::Text_1_0(v) => v.location(),
+            LazyRawValueKind::Binary_1_0(v) => v.location(),
+            LazyRawValueKind::Text_1_1(v) => v.location(),
+            LazyRawValueKind::Binary_1_1(v) => v.location(),
         }
     }
 }
--- a/src/lazy/binary/raw/reader.rs
+++ b/src/lazy/binary/raw/reader.rs
@@ -118,7 +118,9 @@ impl<'data> LazyRawReader<'data, BinaryEncoding_1_0> for LazyRawBinaryReader_1_0
     fn new(context: EncodingContextRef<'data>, data: &'data [u8], is_final_data: bool) -> Self {
         Self::resume(
             context,
-            RawReaderState::new(data, 0, is_final_data, IonEncoding::Binary_1_0),
+            RawReaderState::new(data, 0,         #[cfg(feature = "source-location")]
+            0,         #[cfg(feature = "source-location")]
+            0, is_final_data, IonEncoding::Binary_1_0),
         )
     }
 
@@ -137,6 +139,10 @@ impl<'data> LazyRawReader<'data, BinaryEncoding_1_0> for LazyRawBinaryReader_1_0
         RawReaderState::new(
             &self.data.buffer.bytes()[self.data.bytes_to_skip..],
             stream_offset,
+            #[cfg(feature = "source-location")]
+            0,
+            #[cfg(feature = "source-location")]
+            0,
             // The binary readers do not care whether the data is final because they can detect
             // incomplete values in any case. They always report `false` for simplicity.
             false,
@@ -150,6 +156,16 @@ impl<'data> LazyRawReader<'data, BinaryEncoding_1_0> for LazyRawBinaryReader_1_0
 
     fn position(&self) -> usize {
         self.data.buffer.offset() + self.data.bytes_to_skip
+    }
+
+    #[cfg(feature = "source-location")]
+    fn row(&self) -> usize {
+        0
+    }
+
+    #[cfg(feature = "source-location")]
+    fn prev_newline_offset(&self) -> usize {
+        0
     }
 
     fn encoding(&self) -> IonEncoding {
--- a/src/lazy/binary/raw/v1_1/reader.rs
+++ b/src/lazy/binary/raw/v1_1/reader.rs
@@ -83,7 +83,9 @@ impl<'data> LazyRawReader<'data, BinaryEncoding_1_1> for LazyRawBinaryReader_1_1
     fn new(context: EncodingContextRef<'data>, data: &'data [u8], is_final_data: bool) -> Self {
         Self::resume(
             context,
-            RawReaderState::new(data, 0, is_final_data, IonEncoding::Binary_1_1),
+            RawReaderState::new(data, 0,         #[cfg(feature = "source-location")]
+            0,         #[cfg(feature = "source-location")]
+            0, is_final_data, IonEncoding::Binary_1_1),
         )
     }
 
@@ -95,6 +97,10 @@ impl<'data> LazyRawReader<'data, BinaryEncoding_1_1> for LazyRawBinaryReader_1_1
         RawReaderState::new(
             self.input.bytes(),
             self.position(),
+            #[cfg(feature = "source-location")]
+            0,
+            #[cfg(feature = "source-location")]
+            0,
             // The binary reader doesn't care about `is_final`, so we just use `false`.
             false,
             self.encoding(),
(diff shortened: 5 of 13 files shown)


desaikd:
Benchmarks by disabling and enabling location metadata for `LazyValue`

`LazyValu`e support and with `source-locatio`n feature disabled:
```
Gnuplot not found, using plotters backend
Text Ion 1.0 data size: 3690000 bytes
Bin  Ion 1.0 data size: 2670089 bytes
text 1.0/scan all       time:   [28.825 ms 28.988 ms 29.147 ms]
                        change: [+58.126% +58.957% +59.878%] (p = 0.00 < 0.05)
                        Performance has regressed.
text 1.0/read all       time:   [26.999 ms 27.078 ms 27.163 ms]
                        change: [+2.0400% +3.2512% +4.1874%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 19 outliers among 100 measurements (19.00%)
  14 (14.00%) high mild
  5 (5.00%) high severe
text 1.0/read 'format' field
                        time:   [22.133 ms 22.194 ms 22.261 ms]
                        change: [+2.9679% +3.3273% +3.6755%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild

=== v1.1: maximally compact ===
Binary data size: 620000 bytes
Text   data size: 890000 bytes
maximally compact text 1.1/scan all
                        time:   [8.1125 ms 8.1636 ms 8.2191 ms]
                        change: [-12.529% -11.686% -10.873%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  4 (4.00%) high mild
  1 (1.00%) high severe
maximally compact text 1.1/read all
                        time:   [19.497 ms 19.558 ms 19.620 ms]
                        change: [-7.2783% -6.8479% -6.4477%] (p = 0.00 < 0.05)
                        Performance has improved.
maximally compact text 1.1/read 'format' field
                        time:   [9.4405 ms 9.4808 ms 9.5267 ms]
                        change: [-11.395% -10.733% -10.121%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 12 outliers among 100 measurements (12.00%)
  5 (5.00%) high mild
  7 (7.00%) high severe

=== v1.1: moderately compact ===
Binary data size: 1120000 bytes
Text   data size: 1360000 bytes
moderately compact text 1.1/scan all
                        time:   [8.0763 ms 8.0853 ms 8.0948 ms]
                        change: [-13.695% -13.311% -12.976%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
moderately compact text 1.1/read all
                        time:   [16.860 ms 17.047 ms 17.357 ms]
                        change: [-10.334% -9.0061% -7.2801%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high severe
moderately compact text 1.1/read 'format' field
                        time:   [9.7043 ms 9.7375 ms 9.7724 ms]
                        change: [-9.3995% -8.8850% -8.4146%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

=== v1.1: moderately compact w/length-prefixed top level ===
Binary data size: 1140000 bytes
Text   data size: 1360000 bytes
moderately compact w/length-prefixed top level text 1.1/scan all
                        time:   [8.1667 ms 8.1939 ms 8.2284 ms]
                        change: [-14.549% -13.700% -12.941%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 16 outliers among 100 measurements (16.00%)
  7 (7.00%) high mild
  9 (9.00%) high severe
moderately compact w/length-prefixed top level text 1.1/read all
                        time:   [16.795 ms 16.833 ms 16.872 ms]
                        change: [-9.2602% -8.7515% -8.3216%] (p = 0.00 < 0.05)
                        Performance has improved.
moderately compact w/length-prefixed top level text 1.1/read 'format' field
                        time:   [9.5801 ms 9.5935 ms 9.6092 ms]
                        change: [-11.235% -10.807% -10.464%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
  3 (3.00%) high severe
```
`LazyValue` support and with `source-location` feature enabled:

```
Gnuplot not found, using plotters backend
Text Ion 1.0 data size: 3690000 bytes
Bin  Ion 1.0 data size: 2670089 bytes
text 1.0/scan all       time:   [24.931 ms 25.045 ms 25.183 ms]
                        change: [-14.226% -13.603% -12.917%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  3 (3.00%) high mild
  6 (6.00%) high severe
text 1.0/read all       time:   [31.497 ms 31.548 ms 31.598 ms]
                        change: [+16.092% +16.505% +16.896%] (p = 0.00 < 0.05)
                        Performance has regressed.
text 1.0/read 'format' field
                        time:   [26.873 ms 26.964 ms 27.076 ms]
                        change: [+20.944% +21.496% +22.129%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 16 outliers among 100 measurements (16.00%)
  1 (1.00%) low mild
  7 (7.00%) high mild
  8 (8.00%) high severe

=== v1.1: maximally compact ===
Binary data size: 620000 bytes
Text   data size: 890000 bytes
maximally compact text 1.1/scan all
                        time:   [9.1104 ms 9.1191 ms 9.1289 ms]
                        change: [+10.946% +11.705% +12.418%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 10 outliers among 100 measurements (10.00%)
  4 (4.00%) high mild
  6 (6.00%) high severe
maximally compact text 1.1/read all
                        time:   [20.958 ms 20.993 ms 21.028 ms]
                        change: [+6.9559% +7.3383% +7.7246%] (p = 0.00 < 0.05)
                        Performance has regressed.
maximally compact text 1.1/read 'format' field
                        time:   [10.495 ms 10.529 ms 10.573 ms]
                        change: [+10.389% +11.061% +11.729%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 12 outliers among 100 measurements (12.00%)
  3 (3.00%) high mild
  9 (9.00%) high severe

=== v1.1: moderately compact ===
Binary data size: 1120000 bytes
Text   data size: 1360000 bytes
moderately compact text 1.1/scan all
                        time:   [9.2199 ms 9.2506 ms 9.2918 ms]
                        change: [+13.998% +14.412% +14.860%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 13 outliers among 100 measurements (13.00%)
  3 (3.00%) high mild
  10 (10.00%) high severe
moderately compact text 1.1/read all
                        time:   [18.236 ms 18.269 ms 18.305 ms]
                        change: [+5.2620% +7.1710% +8.3892%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 7 outliers among 100 measurements (7.00%)
  2 (2.00%) low mild
  3 (3.00%) high mild
  2 (2.00%) high severe
moderately compact text 1.1/read 'format' field
                        time:   [10.620 ms 10.632 ms 10.645 ms]
                        change: [+8.7749% +9.1862% +9.5837%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

=== v1.1: moderately compact w/length-prefixed top level ===
Binary data size: 1140000 bytes
Text   data size: 1360000 bytes
moderately compact w/length-prefixed top level text 1.1/scan all
                        time:   [9.2430 ms 9.2693 ms 9.2992 ms]
                        change: [+12.563% +13.125% +13.659%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 12 outliers among 100 measurements (12.00%)
  5 (5.00%) high mild
  7 (7.00%) high severe
moderately compact w/length-prefixed top level text 1.1/read all
                        time:   [18.280 ms 18.322 ms 18.366 ms]
                        change: [+8.4830% +8.8496% +9.1932%] (p = 0.00 < 0.05)
                        Performance has regressed.
moderately compact w/length-prefixed top level text 1.1/read 'format' field
                        time:   [10.673 ms 10.699 ms 10.725 ms]
                        change: [+11.189% +11.524% +11.860%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

Above results show 16% regression for this implementation to support location in `LazyValue`.

src/lazy/binary/raw/v1_1/value.rs
```diff
@@ -292,6 +292,11 @@ impl<'top> LazyRawValue<'top, BinaryEncoding_1_1> for &'top LazyRawBinaryValue_1
             ..**self
         })
     }
+
+    #[cfg(feature = "source-location")]
+    fn location(&self) -> (usize, usize) {
```
popematt, line 297: Do we have a `RowColumn` struct? (IIRC, it came up in one of the previous PRs.) If we do, we should use it here.

src/lazy/expanded/mod.rs
```diff
@@ -1117,6 +1117,14 @@ impl<'top, Encoding: Decoder> LazyExpandedValue<'top, Encoding> {
...
         None
     }
+
+    #[cfg(feature = "source-location")]
+    pub fn location(&self) -> Option<(usize, usize)> {
+        if let ExpandedValueSource::ValueLiteral(value) = &self.source {
+            return Some(value.location());
+        }
+        None
+    }
```
popematt, line 1127: What would you think of something like this? We can have the function always be part of the public API (so that it's easier to turn on/off), but when the feature is not enabled, it always returns `None`.
```suggestion
    pub fn location(&self) -> Option<(usize, usize)> {
        #[cfg(feature = "source-location")]
        if let ExpandedValueSource::ValueLiteral(value) = &self.source {
            return Some(value.location());
        }
        None
    }
```

popematt (commented):
As I've been reading the code, I think that we could simplify things by storing the data as something like this:
```rust
#[cfg(not(feature = "source-location"))]
pub type LocationInternalState = PhantomData<()>;
#[cfg(feature = "source-location")]
pub type LocationInternalState = Option<(NonZero<usize>, usize)>;
```
Then, we don't need to have `#[cfg(feature = "source-location")]` all over the codebase. We can have these fields always present with no performance penalty because when the feature is not enabled, it's a zero-sized type.

I am not certain, but I suspect that the compiler is smart enough to optimize a [NonZero](https://doc.rust-lang.org/std/num/struct.NonZero.html) that's inside another type, like `Option<(NonZero<usize>, usize)>`, so that using `Option` costs no extra space, and the binary reader could just use `None` instead of `(0, 0)`.

src/lazy/binary/raw/v1_1/value.rs
desaikd: No. I think it came in the design discussion though. I will add something for storing this.

src/lazy/expanded/mod.rs
desaikd: I meant to hide the location information when the flag is not enabled. I think having it with `None` will work in the case though.

src/lazy/binary/raw/v1_1/value.rs
popematt: It's fine if it's literally just one of these:
```
pub struct RowColumn(pub usize, pub usize);
// or
pub type RowColumn = (usize, usize);
```
I know we don't normally make fields public, but in this case we're not really trying to define a new struct, we're just giving a descriptive name to `(usize, usize)`. For that same reason, I don't think it really needs any `impl` blocks.

I don't know which one of these two is better in this case. @zslayton, do you have any thoughts on the matter?

src/lazy/expanded/mod.rs
desaikd: I think this is particular to the `ion-schema-rust` usecase and I think the point of having the feature flag is that the API is only available under the feature flag. Also, we tried several benchmarks previously to decide on whether to keep `Option` vs just hide the properties under the feature flag and found that the feature flag was closest to zero-cost solution, which is why we decided to hide it under the feature flag in the last PR. 
I am just going by those results and going with feature-flag everywhere. But that being said, I can modify this to be an `Option` if that is something that is needed within `ion-schema-rust`.

popematt: > I think this is particular to the ion-schema-rust usecase

Even if it is particular to `ion-schema-rust`, it would still make `ion-schema-rust` a nicer library for users if `ion-schema-rust` can automatically work both with and without the `source-location` feature flag.

> I think the point of having the feature flag is that the API is only available under the feature flag

That's not quite correct. The goal isn't to gate the feature—the reason we have a feature flag is to mitigate the performance impact caused by the feature.

Maybe I didn't explain my suggestion well enough. I am suggesting that we _keep_ the feature flag. When the feature flag is enabled, it looks pretty much like it does in this PR. However, when the feature flag is _not_ enabled:
* the internal bookkeeping code is not included in the compilation unit
  * as in this (and the prior) PR
* any internal fields used for tracking the locations can still exist, but they are switched to zero-sized types
  * I suggested this elsewhere in this PR; that's for readability, but it's a two-way decision, so I won't fight for it if you disagree
* the public API remains unchanged—that is to say `pub fn location(&self)` still exists, but it has a trivial implementation that always returns `None`.
  * This should not have any performance impact because it's not adding to the memory footprint of any of the types. It's also not adding additional computation anywhere—it's returning a constant, so even the function call is easily inlined by the compiler.

Does it make more sense if I re-phrase the suggested change like this?
```suggestion
    #[cfg(feature = "source-location")]
    pub fn location(&self) -> Option<(usize, usize)> {
        if let ExpandedValueSource::ValueLiteral(value) = &self.source {
            return Some(value.location());
        }
        None
    }
    
    #[cfg(not(feature = "source-location"))]
    pub fn location(&self) -> Option<(usize, usize)> { None }
```

desaikd: Yes, that makes sense. I previously did not get the part where you said this was for when the feature isn't enabled as you mentioned something about the performance measurement. Bu this makes sense and I am happy to add this change. 

Were you referring to your comment [here](https://github.com/amazon-ion/ion-rust/pull/922#pullrequestreview-2659472670) for the zero-sized types? I do think its useful for readability(Which is why I agree that should be changed) but I do not think its gonna change performance as it would still require the same two proprieties `prev_newline_offset` and `row` for storing location metadata.
Also, the reason I am saying this is because I am trying to to use the new `LazyElement` changes where I can count newlines based on `IonInput` whenever more iput is read into buffer instead of calculating that at `TextBuffer`. which is why I am only focusing on changes that could impact the current implementation performance so that I can compare it with the new suggested change once its ready.(which might likely have better performance)

popematt: > Were you referring to your comment https://github.com/amazon-ion/ion-rust/pull/922#pullrequestreview-2659472670 for the zero-sized types?

Yes. I don't think it's going to impact performance, which is why it should be okay. I can explain in more detail:

Instead of passing around two fields for the `row` and `previous_newline_offset`, we can encapsulate that into a type such as `pub(crate) LocationTrackingState { row: usize, previous_newline_offset: usize }`. That reduces some of the noise in the code simply by going from two fields/arguments to one. However, the more interesting thing is that it allows us to use a trick involving _zero-sized types_.

Instead of conditionally compiling fields and parameters that store `LocationTrackingState` based on the feature flag, we change the _declaration_ of `LocationTrackingState` based on the feature flag:
```rust
#[cfg(feature="source-location")]
#[derive(Default, Copy)]
pub(crate) struct LocationTrackingState {
    row: usize,
    previous_newline_offset: usize
}

#[cfg(not(feature="source-location"))]
#[derive(Default, Copy)]
pub(crate) struct LocationTrackingState;
```

Now, we don't need to have conditionally compiled function arguments or struct fields anywhere that `LocationTrackingState` is used because when the feature flag is _not_ enabled, `LocationTrackingState` is a [zero-sized type](https://doc.rust-lang.org/nomicon/exotic-sizes.html#zero-sized-types-zsts), and "Rust largely understands that any operation that produces or stores a ZST can be reduced to a no-op", so the cost of having those fields should be nothing.

This also makes it easy to identify which code needs to have the feature flag on it—it's only code that needs to read or write the fields of `LocationTrackingState` (so, for example, the conditional code in that checkpoint function no longer needs to be conditional).

desaikd: Sorry, If I was not clear earlier I do understand that this makes sense. I was trying to explain why I am holding off on changes that do not affect performance.

popematt:
@desaikd This PR was replaced by #941, right? Can we close this?

desaikd:
Yes, this has been replaced. Closing it with completion of #941

PR #924, opened 2025-03-03
Re-enables dead code lint

zslayton:
Back when `ion-rust` was getting started, the `rustc`'s lints for detecting dead code had lots of rough edges. We would get false positives enough that we set `#![allow(dead_code)]` on the crate root. Since then, the lints have gotten much better. This PR re-enables the lint and eliminates unused code throughout the library.


Cargo.toml
```diff
@@ -73,6 +73,7 @@ phf = { version = "0.11.2", features = ["macros"] }
 sha2 = { version = "0.9", optional = true }
 serde = { version = "1.0", features = ["derive"], optional = true }
 serde_with = { version = "3.7.0", optional = true }
+visibility = "0.1.1"
```
zslayton, line 76: 🪧 This is a very light dependency that makes it possible to use `#[cfg_attr(...)]` to modify the visibility of a type or `impl` block based on the enabled features.

src/lazy/binary/raw/v1_1/binary_buffer.rs
```diff
@@ -958,7 +953,7 @@ impl<'a> BinaryBuffer<'a> {
         // the batch size because it is a power of two and ~400 bytes seemed like a reasonable
         // chunk of stack space. This can be changed as needed.
         const ARG_BATCH_SIZE: usize = 4;
-        let mut args_array: ArrayVec<ValueExpr<'_, v1_1::Binary>, 4> = ArrayVec::new();
+        let mut args_array: ArrayVec<ValueExpr<'_, v1_1::Binary>, ARG_BATCH_SIZE> = ArrayVec::new();
```
zslayton, line 956: 🪧 The `const` above was defined and then never used.

zslayton (commented):
🗺️ PR Tour 🧭 

This diff annotates several things with `#[allow(dead_code)]` because they are a conflicting combination of used, not-used, visible, and not-visible depending on which set of experimental feature flags is enabled. Refactoring all of the APIs to avoid needing to do this is possible, but IMO not worth the time investment required. As the features stabilize, the feature flags will go away anyway.

popematt (approved):

src/lazy/streaming_raw_reader.rs
```diff
@@ -119,6 +113,10 @@ impl<Encoding: Decoder, Input: IonInput> StreamingRawReader<Encoding, Input> {
         self.read_next(context, /*is_peek=*/ false)
     }
 
+    #[allow(dead_code)]
+    // TODO: The lower-level readers support a 'peek' operation; it isn't currently exposed at higher
+    //      levels because that would require storing ephemeral values that are peeked.
+    //      We should either remove this feature or finish it.
```
jobarr-amzn, line 119:
```suggestion
    // TODO: The lower-level readers support a 'peek' operation; it isn't currently exposed at higher
    //      levels because that would require storing ephemeral values that are peeked.
    //      We should either remove this feature or finish it.
    //      See: https://github.com/amazon-ion/ion-rust/issues/925
```

jobarr-amzn (approved):

jobarr-amzn (approved):

PR #927, opened 2025-03-06
Adds IonDataHash to enable Hash for IonData

popematt:
**Issue #, if available:**

None, but this is needed for Ion Schema Rust.

**Description of changes:**

* Adds `IonDataHash` trait with implementations for `Element`, `Value`, etc.
* Implements `Hash` for `IonData` using `IonDataHash`.
* Adds regular `Hash` for some other types where it makes sense, such as the `IonType` and `TimestampPrecision` enums.
* Adds integer values to `coefficient::Sign` so that it can be cast to an integer value. (I could have just added `derive(Hash)`, but this seemed like a useful little improvement.)

### DIFF
--- a/src/types/decimal/coefficient.rs
+++ b/src/types/decimal/coefficient.rs
@@ -14,8 +14,8 @@ use crate::{Int, UInt};
 /// When the magnitude is zero, the `Sign` can be used to distinguish between -0 and 0.
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug)]
 pub enum Sign {
-    Negative,
-    Positive,
+    Negative = -1,
+    Positive = 1,
 }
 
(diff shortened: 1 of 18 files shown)


src/types/decimal/coefficient.rs
```diff
@@ -293,4 +293,10 @@ mod coefficient_tests {
         assert_eq!(Int::try_from(Coefficient::new(0)), Ok(Int::from(0)));
         assert!(Int::try_from(Coefficient::negative_zero()).is_err());
     }
+
+    #[test]
+    fn test_casting_sign() {
+        assert_eq!(-1, Sign::Negative as i8);
+        assert_eq!(1, Sign::Positive as i8);
```
zslayton, line 300: Interesting, I didn't know this was a thing.

zslayton (approved):
I wonder if we could also rename `IonOrd` to `IonDataOrd` to suggest that it is also practical machinery rather than a formally defined algorithm.

Issue #937, opened 2025-03-12
Reading with `AnyEncoding` not resetting encoding context when IVM of same version encountered.


nirosys:
While working through some conformance dsl issues I noticed that using `AnyEncoding`, like the first reader in the test below, would read the provided input without issue, while the `v1_1::Text` reader would fail when trying to find the `m` macro at the invocation.

Discussing with zslayton highlighted the redundant IVMs in the DSL's fragment rendering (called out in #936), and located the core issue.

Fixing this will cause a decent amount of failures in the conformance tests, so I'm going to bundle a fix for this with a fix for #936 to address the extraneous IVMs.

```rust
    #[test]
    fn test_macro() {
        use ion_rs::*;
        let test = "$ion_1_1 $ion_1_1 $ion::(module _ (macro_table (macro m (v '!' ) ('%' v ) ) ) )  $ion_1_1 (:m 1 )";

        let mut reader = Reader::new(AnyEncoding, test).expect("Unable to create reader");
        let elements = reader.read_all_elements().expect("Unable to get elements"); // <-- Reads: 1
        println!("{:?}", elements);

        let mut reader = Reader::new(v1_1::Text, test).expect("Unable to create reader");
        let elements = reader.read_all_elements().expect("Unable to get elements"); // <-- Fails
        println!("{:?}", elements);
    }

```

nirosys:
Fixed with #940.

Issue #936, opened 2025-03-12
Rendering of Conformance fragments includes IVM


nirosys:
Currently, when rendering fragments with the conformance DSL new IVMs are inserted into the start of every fragment. This causes previously configured encoding context to be lost such as macros.

nirosys:
Fixed with #940.

PR #940, opened 2025-03-19
Reset encoding context whenever IVM encountered & update conformance DSL fragment rendering.

nirosys:
*Issue #, if available:* #937, #936.

*Description of changes:*
This PR addresses 2 issues found during work on conformance testing. The first, is that when reading with AnyEncoding the encoding context would not be reset unless a new ion version was used. The second issue, was that the conformance DSL was rendering each fragment using a new writer, which lead to each fragment having its own IVM.

The redundant IVMs caused issues with the conformance DSL's reader which uses the required version and encoding. When testing some failures (due to context reset) in the CLI we found that the AnyEncoding was not resetting the encoding context as expected.

This PR addresses each by:
* Removing a check that would only reset the context at the start of a stream, or when a new IVM with a different ion version was encountered.
* Refactored the conformance DSL's fragment rendering to render all fragments using the same writer in order to maintain encoding context.

---


zslayton (approved):

nirosys:
*sigh* new laptop.. and I didn't setup my fmt and clippy pre-commit handler. >_<

PR #941, opened 2025-03-25
Adds location metadata to `LazyValue` and unit tests

desaikd:
*Related PR #922*

*Description of changes:*
This PR works on adding lazy calculation of location metadata to `LazyValue` API.
 
 *List of changes:*
 * Adds `lazzy-soucre-location` feature flag
* Adds `row` and `prev_newline_offset`  for `IonSlice` and `IonStream`
* Adds unit tests for `IonSlice` and `IonStream`

### DIFF
--- a/src/lazy/expanded/mod.rs
+++ b/src/lazy/expanded/mod.rs
@@ -178,7 +178,7 @@ impl EncodingContext {
     }
 
     pub fn get_ref(&self) -> EncodingContextRef<'_> {
-        EncodingContextRef { context: self }
+        EncodingContextRef { context: self, value_start: 0, value_end: 0 }
     }
 
     pub fn save_io_buffer(&self) -> IoBuffer {
@@ -259,11 +259,13 @@ impl EncodingContext {
 #[derive(Debug, Copy, Clone)]
 pub struct EncodingContextRef<'top> {
     pub(crate) context: &'top EncodingContext,
+    pub(crate) value_start: usize,
+    pub(crate) value_end: usize,
 }
 
 impl<'top> EncodingContextRef<'top> {
     pub fn new(context: &'top EncodingContext) -> Self {
-        Self { context }
+        Self { context, value_start: 0, value_end: 0 }
     }
 
     pub fn allocator(&self) -> &'top BumpAllocator {
@@ -289,6 +291,58 @@ impl<'top> EncodingContextRef<'top> {
             .clone_macro_with_name("values")
             .expect("`values` macro in system macro table")
     }
+
+
+    #[cfg(feature = "lazy-source-location")]
+    pub fn set_value_start(&mut self, start: usize) {
+        self.value_start = start;
+    }
+
+    #[cfg(feature = "lazy-source-location")]
+    pub fn set_value_end(&mut self, end: usize) {
+        self.value_end = end;
+    }
+
+    #[cfg(feature = "lazy-source-location")]
+    pub fn location(&self) -> Option<(usize, usize)> {
+        match unsafe { &*self.io_buffer_source.get() } {
+            IoBufferSource::IoBuffer(ref buffer) => Some((buffer.row(), buffer.column())),
+            IoBufferSource::Reader(handle) => {
+                let prev_newline = handle.prev_newline_offset();
+                let current_row = handle.row();
+
+                if prev_newline == 0 {
+                    // If no newlines have been encountered yet, we're on the first row
+                    // The column is the start position of the value
+                    Some((1, self.value_start))
+                } else {
+                    // If newlines have been encountered
+                    if self.value_start > prev_newline {
+                        // If the value starts after the last newline,
+                        // we're on the current row and can calculate the column
+                        Some((current_row, self.value_start - prev_newline))
+                    } else {
+                        // If the value starts before or at the last newline,
+                        // we need to find the correct row and column
+                        if let Some(last_newline_pos) = handle.newlines().iter().enumerate()
+                            .take_while(|&(_index, pos)| pos <= &self.value_start)
+                            .last() {
+                            // Found the last newline before or at the value start
+                            // Row is the index of this newline + 2 (1-indexed and we're on the next line)
+                            // Column is the distance from this newline to the value start
+                            Some((last_newline_pos.0 + 2, self.value_start - *last_newline_pos.1))
+                        } else {
+                            // If we couldn't find a newline, we're on the first row
+                            Some((1, self.value_start))
+                        }
+                    }
+                }
+            },
+            IoBufferSource::None => None
+        }
+    }
+
+
 }
 
 impl Deref for EncodingContextRef<'_> {
--- a/src/lazy/streaming_raw_reader.rs
+++ b/src/lazy/streaming_raw_reader.rs
@@ -265,6 +265,18 @@ impl<Encoding: Decoder, Input: IonInput> StreamingRawReader<Encoding, Input> {
 pub trait IoBufferHandle {
     /// Creates an inexpensive copy of the I/O buffer with no lifetime.
     fn save_io_buffer(&self) -> IoBuffer;
+
+    /// Represents the row position for the I/O buffer.
+    #[cfg(feature = "lazy-source-location")]
+    fn row(&self) -> usize;
+
+    /// Represents the previous newline offset for the I/O buffer.
+    #[cfg(feature = "lazy-source-location")]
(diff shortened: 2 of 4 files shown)


src/lazy/streaming_raw_reader.rs
```diff
@@ -345,6 +362,64 @@ impl<SliceType: AsRef<[u8]>> IoBufferHandle for IonSlice<SliceType> {
...
+            1
+        }
+    }
+
+    #[cfg(feature = "lazy-source-location")]
+    fn newlines(&self) -> &[usize] {
+        // This will act as a no-op, as for `IonSlice` we don't need to rely on newlines.
+        // It is only used by `IonStream`, to help keep track of all past newlines offsets.
+        &[]
+    }
```
popematt, line 422: This source location implementation for `IonSlice` seems like it's going to be pretty inefficient. Both `row()` and `prev_newline_offset()` iterate through the entire source data every time they are called.

Instead, consider adding a `newlines: OnceCell<Vec<usize>>` (or `source_location_state: OnceCell<SourceLocationState>`) to `IonSlice` and then storing all of the newline locations there. Yes, it requires some auxiliary storage, but then all of the newline positions can be found once, and reused for every call to `row()` and `prev_newline_offset()`.

src/lazy/expanded/mod.rs
```diff
@@ -178,7 +178,7 @@ impl EncodingContext {
     }
 
     pub fn get_ref(&self) -> EncodingContextRef<'_> {
-        EncodingContextRef { context: self }
+        EncodingContextRef { context: self, value_start: 0, value_end: 0 }
```
popematt, line 181: This looks like `value_start` and `value_end` do not belong in `EncodingContextRef`.

If `value_start` and `value_end` are part of an `EncodingContextRef`, then you should be able to set them based on some value in `EncodingContext` instead of defaulting to `0`, but if they are part of `EncodingContext`, then these fields are unnecessary because they are already included via the `context` field.

src/lazy/value.rs
```diff
@@ -268,6 +268,17 @@ impl<'top, D: Decoder> LazyValue<'top, D> {
         self.expanded_value.context()
     }
 
+    #[cfg(feature = "lazy-source-location")]
+    pub fn location(&self) -> Option<(usize, usize)> {
+        let mut context = self.expanded_value.context();
+        // set the value start and end positions, this help in location calculation
+        if let Some(span) = self.expanded_value.span() {
+            context.set_value_start(span.offset() + 1);
+            context.set_value_end(span.offset() + span.len());
```
popematt, line 277: It looks like `context.set_value_start()` and `context.set_value_end()` are being used only as optional arguments to `context.location()`. This is generally an antipattern because it creates an undocumented (an unenforceable) relationship between these methods—i.e., the compiler can't check to make sure you're passing the necessary arguments. (Also, in other languages there is the additional possibility of `value_start` and `value_end` being modified concurrently, causing unexpected results.)

Can you refactor `location()` to be something like `location_for_span(&self, span: Option<Span<'_>>)`?

src/lazy/expanded/mod.rs
```diff
@@ -289,6 +291,58 @@ impl<'top> EncodingContextRef<'top> {
...
+                    if self.value_start > prev_newline {
+                        // If the value starts after the last newline,
+                        // we're on the current row and can calculate the column
+                        Some((current_row, self.value_start - prev_newline))
+                    } else {
+                        // If the value starts before or at the last newline,
+                        // we need to find the correct row and column
+                        if let Some(last_newline_pos) = handle.newlines().iter().enumerate()
+                            .take_while(|&(_index, pos)| pos <= &self.value_start)
+                            .last() {
```
popematt, line 329:
```suggestion
                            .rfind(|&(_index, pos)| pos <= &self.value_start) {
```

src/lazy/expanded/mod.rs
```diff
@@ -289,6 +291,58 @@ impl<'top> EncodingContextRef<'top> {
...
+                    // If no newlines have been encountered yet, we're on the first row
+                    // The column is the start position of the value
+                    Some((1, self.value_start))
+                } else {
+                    // If newlines have been encountered
+                    if self.value_start > prev_newline {
+                        // If the value starts after the last newline,
+                        // we're on the current row and can calculate the column
+                        Some((current_row, self.value_start - prev_newline))
+                    } else {
```
popematt, line 324: I believe all of this is actually unnecessary for correctness. The method you have for calculating the row/column on lines 325-337 should work for these cases too. Is there a noticeable performance benefit from having these special cases up front? (If not, then can you simplify it by removing the special cases?)

src/lazy/expanded/mod.rs
desaikd: Thats a good point. I will change that to just make `location()` take in `value_start` and `value_end` instead of setting it up this way.

src/lazy/expanded/mod.rs
desaikd: Actually this is necessary for correctness. We have two implementations of `IobufferHanlde` one with `IonSlice` and one with `IonStream`. the `IonStream` one reuses its buffer and hence we have to keep track of the value_start and value_end in order to decide the correct location info for it.
If we remove these lines that you mentioned are unnecessary the tests we have in `LazyValue` will fail, due to incorrect count of location.

e.g.
```
    #[case::lf_lf_cr("{foo: 1, bar: 2}\n\n\r\"hello\"", (4,1))]
```
 fails with
 ```
   left: (1, 20)
   right: (4, 1)
```

File: src/lazy/streaming_raw_reader.rs [hunk 6]
```diff
@@ -300,6 +312,9 @@ pub struct IonSlice<SliceType> {
     source: SliceType,
     // The offset of the first byte that hasn't yet been consumed.
     position: usize,
+    #[cfg(feature = "lazy-source-location")]
+    // The offset of the last byte that has been consumed previously.
+    prev_position: usize,
```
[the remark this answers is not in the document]
desaikd: I think the comment here might misleading. I meant the last byte of the consumed value. but it just stores the previous position value before last `consume` operation.

src/lazy/streaming_raw_reader.rs
desaikd: I think this implementation will still be faster then storing location information. Since this `row` and `prev_newline_offset` are only called if user asks for location metadata i.e when `LazyValue#location()` is called. And I think since now it is only called when user asks for location its not that much of an overhead, unlike storing newlines and keeping it updated with each consumption might be expensive.

desaikd: Is it okay to iterate over this later, I can create an issue for it. As I am not sure If this is really going to improve performance I would need to run benchmarks for this particular suggestion and also because I think the current solution/lazy evaluation is faster.

src/lazy/value.rs
```diff
@@ -268,6 +268,17 @@ impl<'top, D: Decoder> LazyValue<'top, D> {
...
+        }
+        context.location()
```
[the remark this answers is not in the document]
desaikd: This is now handled with `value_start` changes.

src/lazy/streaming_raw_reader.rs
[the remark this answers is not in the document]
desaikd: Changed the comment here. Let me know if anything else needs change.

src/lazy/expanded/mod.rs
```diff
@@ -289,6 +289,44 @@ impl<'top> EncodingContextRef<'top> {
...
+                } else {
+                    // If newlines have been encountered
+                    if value_start > prev_newline {
+                        // If the value starts after the last newline,
+                        // we're on the current row and can calculate the column
+                        Some((current_row, value_start - prev_newline))
+                    } else {
+                        // If the value starts before or at the last newline,
+                        // we need to find the correct row and column
+                        if let Some(last_newline_pos) = handle.newlines().iter().enumerate()
```
popematt, line 314: This branch is not used for `IonSlice`, right? That suggests to me that things aren't encapsulated the right way. It doesn't necessarily need to be addressed now, but this could easily become a friction point if we're trying to refactor things at some point in the future (either related to source location or related to `IoBufferHandle`, etc).

popematt (approved):

src/lazy/expanded/mod.rs
desaikd: https://github.com/amazon-ion/ion-rust/issues/944

PR #946, opened 2025-03-28
Implement AsRef<Element> for Element

popematt:
**Issue #, if available:**

None.

**Description of changes:**

The main substance of this commit is the first commit, in [src/element/mod.rs](https://github.com/amazon-ion/ion-rust/compare/main...popematt:ion-rust:main?expand=1#diff-91c614c02808b6b0727c79965075d97b2882d5c7c06267d31273a99596bb6f17). I literally just added a trivial implementation of `AsRef<Element>`. I was surprised there isn't already a blanket implementation like `impl AsRef<T> for T`, but it looks like it's due to [a restriction in Rust's type system](https://doc.rust-lang.org/std/convert/trait.AsRef.html#reflexivity).

Why is this change needed? This will allow us, in `ion-schema-rust`, to define things that can be generic over `Element` and `&Element` because we can just accept `AsRef<Element>`.

In the second commit, it's all `cargo fmt` changes related to the recent release of Rust 1.85.0.

### DIFF
--- a/src/lazy/streaming_raw_reader.rs
+++ b/src/lazy/streaming_raw_reader.rs
@@ -368,13 +368,13 @@ impl<SliceType: AsRef<[u8]>> IoBufferHandle for IonSlice<SliceType> {
     fn row(&self) -> usize {
         // Update row for location metadata
         let available_range = ..self.prev_position;
-        let data  = &self.stream_bytes()[available_range];
+        let data = &self.stream_bytes()[available_range];
         if !data.is_empty() {
             // Calculate `rows` based on occurrence of newline bytes. If we encounter:
             // 1. b'\r' then increment row count by 1.
             // 2.a. If there was no b'\r' encountered before b'\n' then increment row count by 1.
             // 2.b. If there was no b'\r' encountered before b'\n' then don't increment as b'\r\n' should be counted as 1 based on windows line ending pattern.
-           let (_, rows) = data.iter().fold((false, 0), |(follows_cr, rows), b| {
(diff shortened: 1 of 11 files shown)


src/lazy/streaming_raw_reader.rs
```diff
@@ -558,25 +558,38 @@ impl IoBuffer {
...
+                        // When there's '\n' immediately following '\r', update the offset without adding a row
+                        (b'\n', true) => {
+                            offset.pop();
+                            offset.push(self.stream_offset + range.0 + i + 1);
+                            (false, offset)
+                        }
+                        _ => (false, offset),
+                    }
+                },
+            );
```
jobarr-amzn, line 592: I don't like what rustfmt is doing in much of this, but this one makes me particularly irritated.

tests/conformance_dsl/mod.rs
```diff
@@ -157,7 +159,11 @@ pub(crate) fn parse_document_like<T: DocumentLike>(clause: &Clause) -> InnerResu
     let mut continuation = continuation::Continuation::None;
 
     // We have an optional name as the second argument..
-    if let Some(elem) = clause.body.first().filter(|e| e.ion_type() == IonType::String) {
+    if let Some(elem) = clause
+        .body
+        .first()
+        .filter(|e| e.ion_type() == IonType::String)
+    {
```
jobarr-amzn, line 166: 😢

jobarr-amzn (approved):

PR #947, opened 2025-04-01
Adds location metadata support for `Element` API

desaikd:
*Issue #, if available:*

*Description of changes:*
This PR adds location metadata support for `Element` API.

### DIFF
--- a/src/element/mod.rs
+++ b/src/element/mod.rs
@@ -466,9 +466,22 @@ impl Element {
         &self.value
     }
 
-    /// Returns the source location metadata (row, column).
-    /// If element was constructed from a `LazyValue` then it returns `Some(_)`.
-    /// Otherwise, returns `None` for programmatically constructed element.
+    /// Returns the source location (row, column) of this element in the original Ion text.
+    ///
+    /// The location metadata is primarily intended for error reporting and debugging purposes,
+    /// helping applications provide meaningful feedback to users about the source of issues.
+    ///
+    /// # Returns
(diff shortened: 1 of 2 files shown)


src/element/mod.rs
```diff
@@ -453,6 +466,13 @@ impl Element {
         &self.value
     }
 
+    /// Returns the source location metadata (row, column).
+    /// If element was constructed from a `LazyValue` then it returns `Some(_)`.
```
popematt, line 470: A couple of doc suggestions:
* I don't think library users necessarily care about whether it was constructed from LazyValue. What's more meaningful to say is that if it was read from Ion _text_, then it will _probably_ have a location.
* Maybe also mention that this is intended to help applications surface errors to users, and so it is not guaranteed to be present in all cases, and should not be relied upon for programmatic manipulation of the source of the Ion data.

src/element/mod.rs
```diff
@@ -453,6 +466,13 @@ impl Element {
...
+    /// Otherwise, returns `None` for programmatically constructed element.
+    pub fn location(&self) -> Option<(usize, usize)> {
```
popematt, line 472: I don't want to blow up the scope of this PR, so can you create an issue to re-evaluate the public API for this before we release this feature? I still think that we should consider having a (possibly opaque) `SourceLocation` type.

popematt (approved):

src/element/mod.rs
desaikd: #948

Issue #948, opened 2025-04-02
Add an abstract wrapper for storing location metadata


desaikd:
Reference: https://github.com/amazon-ion/ion-rust/pull/947#discussion_r2025364935
We can use a struct called `SourceLocation` with row and column properties.

PR #949, opened 2025-04-02
Adds implementation of SourceLocation

desaikd:
*Issue #948*

*Description of changes:*
This PR works on adding a `SourceLocation` wrapper for location metadata.

### DIFF
--- a/src/location.rs
+++ b/src/location.rs
@@ -1,0 +1,43 @@
+
+/// Represents the source location (row, column) of this element in the original Ion text.
+///
+/// The source location metadata is primarily intended for error reporting and debugging purposes,
+/// helping applications provide meaningful feedback to users about the source of issues.
+///
+/// # Returns
+/// * `Some((row, column))` - Position where this element was found in the source text
+/// * `None` - Location information is not available
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct SourceLocation {
+    location: Option<(usize, usize)>
+}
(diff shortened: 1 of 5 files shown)


src/location.rs
```diff
@@ -0,0 +1,43 @@
...
+/// The source location metadata is primarily intended for error reporting and debugging purposes,
+/// helping applications provide meaningful feedback to users about the source of issues.
+///
+/// # Returns
+/// * `Some((row, column))` - Position where this element was found in the source text
+/// * `None` - Location information is not available
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct SourceLocation {
+    location: Option<(usize, usize)>
+}
```
popematt, line 13: Any particular reason you chose to implement `SourceLocation` with an `Option` inside of it instead of using `Option<SourceLocation>` for method return types? Do you anticipate having other source location information in this struct at some point (perhaps `offset`) that would make it possible to have source location without necessarily having a row/column?

desaikd: I think I just meant to keep the multiple usage of `Option<(usize, usize)>` concise with `SourceLocation`. It would keep API cleaner than wrapping everything in `Option<SourceLocation>`. But I havent particularly thought about adding an extension with `offset`. Also, I think if we do we have an offset then we should be able to calculate `row` and `column` based on that.

popematt (approved):

Issue #951, opened 2025-04-10
A row/column location should not be available when reading Ion binary


popematt:
Current implementation does not distinguish between Ion text and Ion binary, and so it will report row/column locations for binary Ion, which could be misleading to users. Additionally, we should ensure that the source location feature has not introduced any performance regression for reading Ion binary, since it seems that the newline counter is counting newlines even for binary input.

PR #952, opened 2025-04-10
Fixes location metadata to support nested values

### LINKED ISSUE #944: Improve encapsulation for source location calculation in `IonSlice` and `IonStream`
> This branch is not used for `IonSlice`, right? That suggests to me that things aren't encapsulated the right way. It doesn't necessarily need to be addressed now, but this could easily become a friction point if we're trying to refactor things at some point in the future (either related to source location or related to `IoBufferHandle`, etc).

_Originally posted by @popematt in https://github.com/amazon-ion/ion-rust/pull/941#discussion_r2017268891_

popematt:
**Issue #, if available:**

None.

**Description of changes:**

I discovered that the source location was being reported correctly for top-level values, but that any nested value simply reported the location of its top-level ancestor. This fixes that issue, and in the process, it also fixes #944.

I benchmarked this change, and it is performance-neutral compared to the current tip of `main`.

In the process of working on this, I also realized that row/column location is incorrectly available for Ion binary. I have added a test case to reproduce the problem in `value.rs`, but annotated it as `#[ignore]` because I don't yet know what is the right solution. See #951.

### DIFF
--- a/src/location.rs
+++ b/src/location.rs
@@ -1,40 +1,129 @@
+use crate::Span;
+use std::cell::RefCell;
+use std::rc::Rc;
+
 /// Represents the source location (row, column) of this element in the original Ion text.
 ///
 /// The source location metadata is primarily intended for error reporting and debugging purposes,
 /// helping applications provide meaningful feedback to users about the source of issues.
-///
-/// # Returns
-/// * `Some((row, column))` - Position where this element was found in the source text
-/// * `None` - Location information is not available
-#[derive(Debug, Clone, Eq, PartialEq)]
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Default)]
 pub struct SourceLocation {
-    location: Option<(usize, usize)>,
+    location: (usize, usize),
 }
 
 impl SourceLocation {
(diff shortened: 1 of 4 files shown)


src/location.rs
```diff
@@ -1,40 +1,129 @@
...
 /// helping applications provide meaningful feedback to users about the source of issues.
-///
-/// # Returns
-/// * `Some((row, column))` - Position where this element was found in the source text
-/// * `None` - Location information is not available
-#[derive(Debug, Clone, Eq, PartialEq)]
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Default)]
 pub struct SourceLocation {
-    location: Option<(usize, usize)>,
+    location: (usize, usize),
```
popematt, line 11: 🪧 Since `(0, 0)` is not a valid row/column, we don't need to have the `Option`. Removing the option reduces the size of this type from 24 bytes to 16 bytes (on 64-bit systems).

desaikd (approved):
Looks good! Thanks for adding calculation at one place with `SourecLocationState`.

Issue #935, opened 2025-03-10
Conformance DSL implementation needs to properly encode expression groups.


nirosys:
Currently when transcribing e-expressions the conformance DSL does not catch expression groups, and instead will serialize an expression group as an s-expression with a `::` symbol.

nirosys:
Completed with #975

PR #953, opened 2025-04-10
Implement expression group transcription for conformance DSL.

nirosys:
*Issue #, if available:* #935 

*Description of changes:*
This PR adds support for transcribing expression groups in the `toplevel` clause of the DSL, similar to expression invocations.

Additionally, a small change was also made to the `mactab` clause to automatically include the current unnamed module when re-defining it in order to keep `$ion` imported.

Example usage:
```
(ion_1_1
   (toplevel ('#$:values' ('#$:' 0 1 2)))
   (produces 0 1 2)
)
```

---

### DIFF
--- a/tests/conformance_dsl/clause.rs
+++ b/tests/conformance_dsl/clause.rs
@@ -31,10 +31,10 @@ pub(crate) enum ClauseType {
     Ivm,
     /// Specify a ion data to be inserted into the document, using inline ion syntax.
     TopLevel,
-    /// Provide ion data defining the contents of an '$ion_encoding' directive.
-    Encoding,
     /// Provide ion data defining the contents of a macro table wrapped by a module within an encoding directive.
     MacTab,
+    /// Provide a set of strings to be inserted into the symbol table.
+    SymTab,
     /// Define data that is expected to be produced by the test's document, using inline ion
     /// syntax.
     Produces,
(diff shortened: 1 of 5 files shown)


popematt (approved):

nirosys:
Coming back to this. After merging with main, to resolve conflicts, some of the conformance dsl tests are failing.

Post-merge the toplevel fragment now handles pre-populating macros that are defined in the toplevel's sub-elements. Unfortunately those elements are then serialized to the input doc which re-defines the macros, resulting in decoding errors.

For example one of the tests uses the following 2 fragments:
```
   (mactab (macro m (v!) (%v)))
   (:m 1)
```
The test is to ensure that we don't add new IVMs between the fragments, which was an issue previously.

When these fragments are serialized, it produces (formatting added):
```
$ion_1_1
$ion::(module _ (symbol_table _ ) (macro_table _ (macro m (v ) ('%' v ) ) ) )
$ion::(module _ (macro_table _ (macro m (v '!' ) ('%' v ) ) ) )
(:m 1)
```
It appears the change made to the top-level fragment is adding the macros, but they're being written out in addition to the toplevel elements that are defining them as well. This results in an error when deserializing:
```
Decoding(DecodingError { description: "macro named 'm' already exists", position: None })
```

It might be better to walk the toplevel elements looking for encoding directives and register them as we find them so that we can skip serializing them with the rest of the elements. 

Looking to see how involved that will be.

nirosys:
okie dokie.. f me then, closing since this pr isn't needed anymore..

Issue #955, opened 2025-04-16
Reader does not correctly count newlines when stream buffer needs to be filled between `\r` and `\n`


popematt:
Reader does not correctly count newlines when stream buffer needs to be filled between `\r` and `\n`.

Minimal reproduction:

```rust
#[test]
fn minimal_repro() {
    let mut input: Box<dyn Read> = Box::new(io::empty());
    for input_byte in "1\r\n2".as_bytes() {
        input = Box::new(input.chain(Cursor::new([*input_byte])));
    }
    let values: Vec<_> = Reader::new(AnyEncoding, input)
        .expect("expected a reader")
        .collect::<IonResult<_>>()
        .expect("expected values");
    assert_eq!(values.get(0).and_then(|it| it.as_lazy_value().location().row_column()), Some((1, 1)));
    assert_eq!(values.get(1).and_then(|it| it.as_lazy_value().location().row_column()), Some((2, 1)));
}
```
Fails with:
```
assertion `left == right` failed
  left: Some((3, 1))
 right: Some((2, 1))
```

popematt:
As per [this conversation](https://github.com/amazon-ion/ion-rust/pull/956#discussion_r2049240059), the optimal solution is probably to drop support for tracking `\r`.

Issue #954, opened 2025-04-16
Reader does not detect encoding correctly when stream requires loading in IVM


popematt:
Here is a minimal-ish reproduction:
```rust
#[test]
fn test_detect_encoding_from_stream() {
    let data = [
        0xE0u8, 0x01, 0x00, 0xEA, // IVM
        0x83, 65, 66, 67, // String: "ABC"
    ];

    let mut input: Box<dyn Read> = Box::new(io::empty());
    for input_byte in data {
        input = Box::new(input.chain(Cursor::new([input_byte])));
    }
    let _values: Vec<_> = Reader::new(AnyEncoding, input)
        .expect("a reader")
        .collect::<IonResult<_>>()
        .expect("values should be read successfully");
}
```
It fails with:
```
values should be read successfully: Decoding(DecodingError { description: "found unrecognized syntax\n    while reading a value\n        offset=0\n        input=[E0]\n        buffer len=1\n        ", position: Some(Position { byte_offset: 0, byte_length: Some(1), line_column: None }) })
```

zslayton:
This can be addressed by having the detection method return `IonError::Incomplete(_)` when the input has fewer than 4 bytes and the bytes that are present are an IVM prefix.

PR #956, opened 2025-04-16
Stop providing misleading row/column locations for binary Ion

### LINKED ISSUE #951: A row/column location should not be available when reading Ion binary

popematt:
**Issue #, if available:**

fixes #951 

**Description of changes:**

Adds an `encoding()` method to lazy raw values so that a lazy value can call `self.raw().encoding()` and determine what `IonEncoding` (if any) is backing the current lazy value, and then avoid returning a row/column location for values that are backed by binary Ion.

I also significantly refactored the value location tests, and in doing so, I discovered a few more bugs. The affected test cases are commented out, and I opened #954 and #955.

### DIFF
--- a/src/lazy/value.rs
+++ b/src/lazy/value.rs
@@ -242,11 +242,19 @@ impl<'top, D: Decoder> LazyValue<'top, D> {
     }
 
     pub fn location(&self) -> SourceLocation {
-        let context = self.expanded_value.context();
-        // set the value start and end positions, this help in location calculation
-        context
-            .location_for_span(self.expanded_value.span())
-            .unwrap_or_default()
+        if let Some(raw) = self.raw() {
+            if raw.encoding().is_text() {
+                self.expanded_value
+                    .context()
+                    .location_for_span(self.expanded_value.span())
+                    .unwrap_or_default()
+            } else {
+                // No row/column for binary Ion
+                SourceLocation::empty()
+            }
+        } else {
+            SourceLocation::empty()
+        }
     }
 
     pub fn to_owned(&self) -> LazyElement<D> {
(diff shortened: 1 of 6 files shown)


src/lazy/value.rs
```diff
@@ -729,27 +782,109 @@ mod tests {
...
+    #[case::no_crlf( "{}\"hello\"",       [(1, 1), (1, 3)])]
+    #[case::lf_lf_lf("{}\n\n\n\"hello\"", [(1, 1), (4, 1)] )]
+    #[case::lf_lf_cr("{}\n\n\r\"hello\"", [(1, 1), (4, 1)] )]
+    #[case::lf_cr_cr("{}\n\r\r\"hello\"", [(1, 1), (4, 1)] )]
+    // FIXME: Currently failing because of https://github.com/amazon-ion/ion-rust/issues/955
+    // #[case::lf_cr_lf("{}\n\r\n\"hello\"", [(1, 1), (3, 1)] )]
+    // #[case::cr_lf_lf("{}\r\n\n\"hello\"", [(1, 1), (3, 1)] )]
+    // #[case::cr_lf_cr("{}\r\n\r\"hello\"", [(1, 1), (3, 1)] )]
+    // #[case::cr_cr_lf("{}\r\r\n\"hello\"", [(1, 1), (3, 1)] )]
+    #[case::cr_cr_cr("{}\r\r\r\"hello\"", [(1, 1), (4, 1)] )]
```
zslayton, line 795: Making a note of our offline conversation: I was surprised to see that we support standalone `\r`s in our row/column calculations as these are only (?) used in pre-2000-era mac systems. Counting `\n`s (which inherently includes `\r\n`s) should be a `memchr` SIMD operation while counting `\r`, `\n`, and `\r\n`s separately requires a state machine.

zslayton (approved):

desaikd (approved):

PR #959, opened 2025-04-25
Fixes clippy warnings

zslayton:

### DIFF
--- a/src/lazy/decoder.rs
+++ b/src/lazy/decoder.rs
@@ -125,6 +125,7 @@ pub trait RawVersionMarker<'top>: Debug + Copy + Clone + HasSpan<'top> {
     }
 
     /// The `IonVersion` that was used to encode this IVM.
+    #[allow(dead_code)]
     fn stream_version_before_marker(&self) -> IonVersion {
         self.stream_encoding_before_marker().version()
     }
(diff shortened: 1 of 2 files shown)


src/lazy/decoder.rs
```diff
@@ -125,6 +125,7 @@ pub trait RawVersionMarker<'top>: Debug + Copy + Clone + HasSpan<'top> {
     }
 
     /// The `IonVersion` that was used to encode this IVM.
+    #[allow(dead_code)]
     fn stream_version_before_marker(&self) -> IonVersion {
```
zslayton, line 129: 🪧 This trait method is not yet used, but I want it to be available.

src/lazy/value.rs
```diff
@@ -257,7 +257,7 @@ impl<'top, D: Decoder> LazyValue<'top, D> {
         }
     }
 
-    pub fn to_owned(&self) -> LazyElement<D> {
```
zslayton, line 260: 🪧 Clippy complains that methods named `to_owned` usually take `self` by value. `LazyValue` implements `Copy`, so we can take `self` by value instead of by reference.

zslayton (commented):
🗺️ PR Tour 🧭

jobarr-amzn (approved):

PR #961, opened 2025-04-28
Optimize the layout of `Decimal`

zslayton:
Cuts `Decimal` from 48 bytes (23 bytes padding) to 32 bytes (7 bytes padding). See the layout comment in the PR for details.

Many types contain a `Decimal`, including `Timestamp` and `ValueRef`, so this removes 16 bytes from many types. Binary 1.0 benchmarks showed that scanning is ~5% faster.

### DIFF
--- a/src/types/decimal/mod.rs
+++ b/src/types/decimal/mod.rs
@@ -39,19 +39,56 @@ pub mod coefficient;
 /// ```
 #[derive(Copy, Clone, Debug)]
 pub struct Decimal {
-    // A Coefficient is a `(Sign, UInt)` pair supporting integers of arbitrary size
-    pub(crate) coefficient: Coefficient,
+    // ===== A note on layout =====
+    // A `Coefficient` is a `(Sign, Int)` pair. The `Sign` is a one-byte enum that allows the
+    // `Coefficient` to distinguish between positive and negative zero. If the value is not
+    // negative zero, then the `Sign` will agree with the sign of the `Int` value.
+    //
+    // `Decimal` previously contained a `Coefficient`, but this resulted in large amounts of wasted
+    // memory. Because of alignment requirements, the `Coefficient` was stored like this:
+    //
+    //              Int      Sign    Padding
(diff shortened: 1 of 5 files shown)


src/types/decimal/mod.rs
```diff
@@ -39,19 +39,56 @@ pub mod coefficient;
...
+    //                   Decimal
+    // Of the `Decimal`'s 48 bytes, 23 were padding—a huge waste. Because many types contain a
+    // Decimal (however indirectly), this meant that lots of data types had 23 bytes of dead space.
+    //
+    // `Decimal` now stores the value and sign fields itself and uses them to construct a `Coefficient`
+    // on demand. As a result, `Decimal` is now only 32 bytes, 7 of which are padding.
+    //
+    //              Int        Exponent  Padding
+    //       ┌───────┴───────┐ ┌───┴──┐  ┌──┴──┐
+    //       IIIIIIII IIIIIIII EEEEEEEE S_______
```
popematt, line 71: Since we only need one _bit_ for the sign, could we do some bit-trickery to reduce either the exponent or the coefficient by one bit and store the exponent in that space? Would it make much difference, do you think, to get this down to only 24 bytes?

jobarr-amzn: I'm also curious about this.

jobarr-amzn (approved):

src/types/decimal/mod.rs
zslayton: > Since we only need one bit for the sign, could we do some bit-trickery to reduce either the exponent or the coefficient by one bit and store the exponent in that space?

I'm open to suggestions! The one that springs to mind is reserving `int128::MIN` to mean `-0`. It does make the representable coefficient range slightly less tidy ("`i128` minus one"), but maybe it's worth trying.

> Would it make much difference, do you think, to get this down to only 24 bytes?

It's possible! Hard to say without benchmarking, though.

I'm going to go ahead and merge this. We can experiment more in another PR.

PR #962, opened 2025-04-29
Make special case parsing logic for ints 1-3 bytes in length

zslayton:
When decoding a binary Ion 1.0 encoding primitive `UInt`, the reader uses a general-purpose helper method that copies the input bytes into a fixed-size buffer and converts the whole thing into a `u128`. Because it works on a slice of any length, the compiler will choose not to inline the resulting `memcpy`. This causes integer-reading to show up prominently in profiles.

This change special cases 1-, 2-, and 3-byte integers to avoid the `memcpy` and inline all the logic.

### DIFF
--- a/src/binary/uint.rs
+++ b/src/binary/uint.rs
@@ -1,5 +1,4 @@
 use std::io::Write;
-use std::mem;
 
 use crate::result::{IonFailure, IonResult};
 use crate::{Int, IonError, UInt};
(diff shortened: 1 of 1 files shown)


src/binary/uint.rs
```diff
@@ -18,20 +17,30 @@ impl DecodedUInt {
...
     #[inline]
     pub(crate) fn uint_from_slice(uint_bytes: &[u8]) -> IonResult<u128> {
-        if uint_bytes.len() > mem::size_of::<u128>() {
-            return IonResult::decoding_error(
+        const MAX_BYTES: usize = size_of::<u128>();
+        // The `uint_from_slice_unchecked` method will work for a uint of any size up to and including
+        // 16 bytes. However, because the slice is of an unknown length, the `memcpy` performed by
+        // that method does not get inlined by the compiler. Here we check for some common int sizes
+        // and construct the int manually to allow the hot path to be inlined.
+        match uint_bytes.len() {
```
popematt, line 25: No case needed for `0`?

popematt (approved):

src/binary/uint.rs
jobarr-amzn: I'd assume that's handled in a different layer and would skip calling this method entirely, but if so that would be a good thing to comment at least.

jobarr-amzn (approved):
Why specialize for 1-3 bytes, only, when this method handles up to 16 bytes? 
The PR description says that the `memcpy` "causes integer-reading to show up prominently in profiles" - does the benefit diminish after special-casing 1-3? What makes the cutoff?

zslayton:
> Why specialize for 1-3 bytes, only, when this method handles up to 16 bytes?

In many (most?) data sets, the integers are not random; they represent counts or quantities. As a result, smaller integers are more common. This is true in the couple of data sets that I'm benchmarking.

> The PR description says that the `memcpy` "causes integer-reading to show up prominently in profiles" - does the benefit diminish after special-casing 1-3? What makes the cutoff?

The aim is to allow the most common cases to be inlined by the caller. Without this change, `uint_from_slice` gets inlined but its child call to `memcpy` does not. With this change, processing for smaller integers gets inlined while processing for larger integers falls back to `memcpy`. If we try to special case too many integer sizes, then code generated for `uint_from_slice` itself gets too large, and neither `uint_from_slice` nor `memcpy` get inlined.

src/binary/uint.rs
zslayton: > No case needed for 0?

As it was written, the empty slice would use the general-purpose path. I didn't add a special case for it because I was concerned about the code for `uint_from_slice` itself getting to big. However, I just ran my benchmarks with the special case for `0` added and it doesn't seem to have pushed it over that size threshold, so let's do it!

PR #964, opened 2025-05-06
Simplify `FlexUInt::read`

zslayton:
When `FlexUInt::read` was first written, Rust did not yet have a `u128` data type. In order to support decoding 9- and 10-byte `FlexUInt`s, we made a buffer on the stack and manipulated it manually. Now that Rust has a `u128` type, we can use integer bitshifting operations to simplify our decoding in the general case, resulting in a deletion of ~150 LOC.

### DIFF
--- a/src/lazy/encoder/binary/v1_1/flex_uint.rs
+++ b/src/lazy/encoder/binary/v1_1/flex_uint.rs
@@ -3,7 +3,6 @@ use crate::{IonResult, UInt};
 use bumpalo::collections::Vec as BumpVec;
 use ice_code::ice as cold_path;
 use std::io::Write;
-use std::mem;
 
 const BITS_PER_U128: usize = 128;
 const BITS_PER_ENCODED_BYTE: usize = 7;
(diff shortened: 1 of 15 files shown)


src/lazy/encoder/binary/v1_1/flex_uint.rs
```diff
@@ -46,79 +45,49 @@ impl FlexUInt {
     ///          an appropriate error message if reading fails.
     #[inline(always)]
     pub fn read(input: &[u8], offset: usize) -> IonResult<FlexUInt> {
-        const COMMON_CASE_INPUT_BYTES_NEEDED: usize = 8;
-
+        const COMMON_CASE_INPUT_BYTES_NEEDED: usize = 4;
```
zslayton, line 48: 🪧 The hot path below now uses per-size `const` method invocations. This was lowered from `8` to `4` to strike a balance between compiled method size (for inlining) and performance.

src/lazy/encoder/binary/v1_1/flex_uint.rs
```diff
@@ -46,79 +45,49 @@ impl FlexUInt {
...
+                let num_encoded_bytes = input[0].trailing_zeros() as usize + 1;
+                // By branching on particular values, we make the value of `num_encoded_bytes` in their
+                // corresponding arm `const`. This allows us to use `read_n_bytes` to optimize for those
+                // sizes.
+                let mut buffer = [0u8; size_of::<u64>()];
+                match num_encoded_bytes {
+                    1 => Self::read_n_bytes::<1>(input, &mut buffer),
+                    2 => Self::read_n_bytes::<2>(input, &mut buffer),
+                    3 => Self::read_n_bytes::<3>(input, &mut buffer),
+                    4 => Self::read_n_bytes::<4>(input, &mut buffer),
```
zslayton, line 66: 🪧 For the most common sizes, we use a `const` helper function to do the reading. It's very slightly faster in Rust v1.86. However, the `nightly` version of the compiler is able to do magical things with it (presumably because it's using a newer LLVM version). Nightly can be ~10% faster than stable in some cases.

src/lazy/encoder/binary/v1_1/flex_uint.rs
```diff
@@ -46,79 +45,49 @@ impl FlexUInt {
...
-        Ok(flex_uint)
+        // General-purpose FlexUInt reading logic. Checks for empty input and supports FlexUInts
+        // up to U64::MAX.
+        Self::read_flex_primitive_as_uint(input, offset, "reading a FlexUInt")
     }
 
-    /// Helper method that reads a [`FlexUInt`] with 7 or fewer bytes of magnitude from the buffer.
-    // Caller must confirm that `bytes` has at least 8 bytes.
     #[inline]
-    fn read_small_flex_uint(bytes: &[u8]) -> FlexUInt {
```
zslayton, line 74: 🪧 This (removed) method was the decoder for `FlexUInt`s up to 7 bytes in size. If the magnitude was only 7 bytes, then the complete serialized form of the `FlexUInt` could fit in a `u64`, simplifying decoding.

src/lazy/encoder/binary/v1_1/flex_uint.rs
```diff
@@ -46,79 +45,49 @@ impl FlexUInt {
...
-    ///
-    /// The current implementation supports flex primitives with up to 64 bits of representation
-    /// beyond the leading header bits. Flex primitives requiring 10 bytes to encode have 70 magnitude
-    /// bits. If this value is unsigned (`support_sign_extension=false`), the six bits beyond the
-    /// supported 64 must all be `0`. If this value will later be re-interpreted as a signed value,
-    /// (`support_sign_extension=true`), then the six bits beyond the supported 64 must all be the
-    /// same as the 64th (highest supported) bit. This will allow encodings of up to 70 bits
-    /// to be correctly interpreted as positive, negative, or beyond the bounds of the 64 bit
-    /// limitation.
     pub(crate) fn read_flex_primitive_as_uint(
```
zslayton, line 87: 🪧 This was the general-purpose decoder, handling incomplete data and magnitudes greater than 7 bytes. Because the serialized form couldn't fit in a `u64` (and Rust didn't yet have a `u128`), we had to manipulate a byte array on the stack to achieve the desired effect.

src/lazy/encoder/binary/v1_1/flex_uint.rs
```diff
@@ -129,147 +98,30 @@ impl FlexUInt {
...
-        // Shift the magnitude from the last 8 bytes over and combine it with the six bits we
-        // carried over from the second byte.
-        let value = (remaining_magnitude << 6) | magnitude_low_six as u64;
-        let flex_uint = FlexUInt::new(10, value);
-        Ok(flex_uint)
+        let mut buffer = [0u8; size_of::<u128>()];
+        buffer[..num_encoded_bytes].copy_from_slice(&input[..num_encoded_bytes]);
+        let big_value = u128::from_le_bytes(buffer).wrapping_shr(num_encoded_bytes as u32);
+        let value = big_value as u64;
+        Ok(FlexUInt::new(num_encoded_bytes, value))
```
zslayton, line 124: 🪧 In the new version, the serialized form fits entirely in a `u128` so we can use standard bitshifting to eliminate the continuation flag bits.

zslayton (commented):
🗺️ PR Tour 🧭

popematt (approved):

PR #965, opened 2025-05-12
Adds initial guard rails to the `ApplicationEExpWriter`

zslayton:
The `ApplicationEExpWriter` is intended to be the common validation layer for user interactions with the underlying `Encoding::RawEExpWriter` type. It resolves the provided ID to a macro, then verifies that each argument aligns with that macro's signature. This PR begins the process of implementing that. 

Specifically, it implements the initial macro ID resolution, raising an error if/when the user attempts to write an e-expression with an unrecognized ID. This caused several conformance tests to break because the `Writer` is asked to emit macro definitions that it does not have in its own encoding context. I have added a workaround for this that I'll highlight in the PR tour. We can implement a more robust fix down the road.

### DIFF
--- a/src/lazy/encoder/value_writer.rs
+++ b/src/lazy/encoder/value_writer.rs
@@ -93,7 +93,9 @@ pub trait ValueWriter: AnnotatableWriter + Sized {
     fn list_writer(self) -> IonResult<Self::ListWriter>;
     fn sexp_writer(self) -> IonResult<Self::SExpWriter>;
     fn struct_writer(self) -> IonResult<Self::StructWriter>;
-    fn eexp_writer<'a>(self, macro_id: impl MacroIdLike<'a>) -> IonResult<Self::EExpWriter>;
+    fn eexp_writer<'a>(self, macro_id: impl MacroIdLike<'a>) -> IonResult<Self::EExpWriter>
+    where
+        Self: 'a;
 
     fn write(self, value: impl WriteAsIon) -> IonResult<()> {
         value.write_as_ion(self)
@@ -198,7 +200,7 @@ macro_rules! delegate_value_writer_to {
                 fn eexp_writer<'a>(
                     self,
                     macro_id: impl MacroIdLike<'a>,
-                 ) -> IonResult<Self::EExpWriter>;
+                 ) -> IonResult<Self::EExpWriter> where Self: 'a;
 
             }
         }
@@ -460,9 +462,9 @@ pub trait SequenceWriter: MakeValueWriter {
     }
 
     fn eexp_writer<'a>(
-        &mut self,
+        &'a mut self,
         macro_id: impl MacroIdLike<'a>,
-    ) -> IonResult<<Self::NestedValueWriter<'_> as ValueWriter>::EExpWriter> {
+    ) -> IonResult<<Self::NestedValueWriter<'a> as ValueWriter>::EExpWriter> {
         self.value_writer().eexp_writer(macro_id)
     }
 
--- a/src/lazy/encoder/writer.rs
+++ b/src/lazy/encoder/writer.rs
@@ -32,6 +32,8 @@ use crate::{
     TemplateCompiler, TemplateMacro, Timestamp, UInt, Value,
 };
 
+/// A thin wrapper around a `SymbolTable` that tracks the number of symbols whose definition has
+/// not yet been written to output.
 pub(crate) struct WriterSymbolTable {
     symbols: SymbolTable,
     num_pending: usize,
@@ -72,6 +74,8 @@ impl Deref for WriterSymbolTable {
     }
 }
 
+/// A thin wrapper around a `MacroTable` that tracks the number of macros whose definition has
+/// not yet been written to output.
 pub(crate) struct WriterMacroTable {
     macros: MacroTable,
     num_pending: usize,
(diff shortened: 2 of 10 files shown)


src/lazy/encoder/value_writer.rs
```diff
@@ -93,7 +93,9 @@ pub trait ValueWriter: AnnotatableWriter + Sized {
     fn list_writer(self) -> IonResult<Self::ListWriter>;
     fn sexp_writer(self) -> IonResult<Self::SExpWriter>;
     fn struct_writer(self) -> IonResult<Self::StructWriter>;
-    fn eexp_writer<'a>(self, macro_id: impl MacroIdLike<'a>) -> IonResult<Self::EExpWriter>;
+    fn eexp_writer<'a>(self, macro_id: impl MacroIdLike<'a>) -> IonResult<Self::EExpWriter>
+    where
+        Self: 'a;
```
zslayton, line 98: 🪧 I had to tweak the signature of this method to communicate that the returned value will outlive the provided identifier.

src/lazy/encoder/writer.rs
```diff
@@ -27,34 +28,107 @@ use crate::result::IonFailure;
 use crate::write_config::WriteConfig;
 use crate::{
     AnyEncoding, ContextWriter, Decimal, Element, ElementWriter, Int, IonInput, IonResult, IonType,
-    IonVersion, MacroTable, RawSymbolRef, Reader, Symbol, SymbolTable, TemplateCompiler, Timestamp,
-    UInt, Value,
+    IonVersion, MacroDef, MacroTable, RawSymbolRef, Reader, Symbol, SymbolId, SymbolTable,
+    TemplateCompiler, TemplateMacro, Timestamp, UInt, Value,
 };
 
-pub(crate) struct WriterContext {
```
zslayton, line 34: 🪧 The symbol table can be modified by the user at any depth by writing a new symbol, annotation, or field name. This required value writer types to pass around a `&mut WriterContext` to make updates as needed.

In contrast, the macro table can only be modified at the top level, between values. When the `ApplicationEExpWriter` goes to store a shared reference to the invoked macro, it causes a conflict. There can't simultaneously be a `&mut WriterContext` _and_ a `&MacroDef` that points inside the `WriterContext`. Therefore, I've split the `WriterContext` into a `WriterSymbolTable` and `WriterMacroTable` that can be managed separately. Each new type is a thin wrapper around `SymbolTable`/`MacroTable` and does some minor bookkeeping each time the underlying table is modified.

src/lazy/encoder/writer.rs
```diff
@@ -96,36 +170,42 @@ impl<E: Encoding, Output: Write> Writer<E, Output> {
...
+            self.macro_table(),
             &MacroTable::empty(),
             macro_def_sexp,
         )?;
 
-        let address = self
-            .context
-            .macro_table
-            .add_template_macro(template_macro)?;
-        self.context.num_pending_macros += 1;
```
zslayton, line 115: 🪧 The new `Writer_____Table` types make the code a bit less verbose.

src/lazy/encoder/writer.rs
```diff
@@ -332,38 +404,37 @@ impl<E: Encoding, Output: Write> SequenceWriter for Writer<E, Output> {
 }
 
 pub struct ApplicationValueWriter<'a, V: ValueWriter> {
-    encoding: &'a mut WriterContext,
+    symbols: &'a mut WriterSymbolTable,
+    macros: &'a WriterMacroTable,
```
zslayton, line 408: 🪧 Notice that one of these is `&mut` and the other is simply `&`.

src/lazy/encoder/writer.rs
```diff
@@ -628,35 +699,49 @@ impl<'value, V: ValueWriter> ValueWriter for ApplicationValueWriter<'value, V> {
...
-    fn eexp_writer<'a>(self, macro_id: impl MacroIdLike<'a>) -> IonResult<Self::EExpWriter> {
+    fn eexp_writer<'a>(self, macro_id: impl MacroIdLike<'a>) -> IonResult<Self::EExpWriter>
+    where
+        Self: 'a,
+    {
+        let resolved_id = macro_id.resolve(self.macros)?;
+        let macro_ref = self
+            .macros
+            .macro_at_address(resolved_id.address())
+            .expect("just resolved");
```
zslayton, line 717: 🪧 This method now resolves the provided macro ID up front.

src/lazy/encoder/writer.rs
```diff
@@ -846,21 +939,30 @@ impl<V: ValueWriter> SequenceWriter for ApplicationSExpWriter<'_, V> {
...
 
 pub struct ApplicationEExpWriter<'value, V: ValueWriter> {
-    encoding: &'value mut WriterContext,
+    symbols: &'value mut WriterSymbolTable,
+    macros: &'value WriterMacroTable,
     raw_eexp_writer: V::EExpWriter,
     value_writer_config: ValueWriterConfig,
+    // TODO: these are now available but not yet used.
+    _invoked_macro: &'value MacroDef,
+    _param_index: usize,
```
zslayton, line 948: 🪧 The next PR will add validation for each argument-writing method call that happens next.

src/lazy/expanded/compiler.rs
```diff
@@ -843,7 +843,7 @@ impl TemplateCompiler {
 
     /// Adds a `lazy_sexp` that has been determined to represent a macro invocation to the
     /// TemplateBody.
-    fn compile_macro<'top, D: Decoder>(
+    fn compile_macro_invocation<'top, D: Decoder>(
```
zslayton, line 846: 🪧 I renamed this method for clarity. The entire module is dedicated to compiling macros. This particular method is compiling a TDL macro invocation within a template.

tests/conformance_dsl/context.rs
```diff
@@ -274,11 +275,11 @@ mod tests {
             .input(IonEncoding::Text)
             .expect("failed to render fragments");
 
-        let fragments_str = String::from_utf8(bytes).expect("Invalid input string generated");
-        assert_eq!(
-            fragments_str,
-            "$ion_1_1 $ion::(module _ (macro_table (macro m (v '!' ) ('%' v ) ) ) ) (:m 1)"
-                .to_string(),
-        );
```
zslayton, line 282: 🪧 This test was doing a string-equality check that started failing because of the workaround I added later. I changed the test to use Ion stream equality instead.

tests/conformance_dsl/fragment.rs
```diff
@@ -408,18 +408,51 @@ impl FragmentImpl for TopLevel {
...
-            writer.write(ProxyElement(elem, ctx))?;
-        }
+        self.write(ctx, &mut writer)?;
+
         buffer = writer.close()?;
         Ok(buffer)
     }
 
-    fn write<E: ion_rs::Encoding, O: std::io::Write>(
+    fn write<E: Encoding, O: std::io::Write>(
```
zslayton, line 417: 🪧 The comment below explains the problem I encountered and the workaround that's now in place.

zslayton (commented):
🗺️ PR Tour 🧭

PR #968, opened 2025-05-29
Adds guard rails to ApplicationEExpWriter

zslayton:
This PR adds validation to method calls on the `ApplicationEExpWriter`, the user-facing API for invoking a macro.


src/lazy/binary/raw/v1_1/binary_buffer.rs
```diff
@@ -997,16 +994,22 @@ impl<'a> BinaryBuffer<'a> {
...
             .macro_at_address(macro_address)
             .ok_or_else(|| {
                 IonError::decoding_error(format!(
                     "invocation of macro at unknown address '{macro_address:?}'"
                 ))
-            })?
-            .reference();
+            })?;
+        let macro_ref = MacroRef::new(
+            QualifiedAddress::new(ModuleKind::Default, macro_address),
```
zslayton, line 1009: 🪧 `MacroRef`s now hold a `QualifiedAddress`, which is a `(module, address)` pair rather than just an `address`.

src/lazy/encoder/binary/v1_0/writer.rs
```diff
@@ -156,6 +158,14 @@ impl<W: Write> LazyRawWriter<W> for LazyRawBinaryWriter_1_0<W> {
...
         Ok(())
     }
+
+    fn macro_table(&self) -> &WriterMacroTable {
+        &EMPTY_MACRO_TABLE
+    }
+
+    fn macro_table_mut(&mut self) -> Option<&mut WriterMacroTable> {
+        None
+    }
```
zslayton, line 168: 🪧 `LazyRawWriter` implementations are now required to own their own `WriterMacroTable`. 

As the trait method doc comments explain (later in the diff), for read-only access there is no `Option` type because that's a very hot path. Ion 1.0 writers can return a reference to the global static `EMPTY_MACRO_TABLE` to satisfy the API.

For mutable access (only required to modify the encoding context, which is comparatively rare), implementations can return an `Option<&mut WriterMacroTable>`. This forces the managed writer to confirm on each access that there's a macro table to modify.

src/lazy/encoder/binary/v1_1/container_writers.rs
```diff
@@ -24,6 +28,7 @@ pub(crate) struct BinaryContainerWriter_1_1<'value, 'top> {
     allocator: &'top BumpAllocator,
     encoder: ContainerEncodingKind<'value, 'top>,
     value_writer_config: ValueWriterConfig,
+    macros: &'value MacroTable,
```
zslayton, line 31: 🪧 There are lots of new function parameters simply plumbing through a reference to the macro table so nested value writers can invoke macros as needed.

src/lazy/encoder/binary/v1_1/container_writers.rs
```diff
@@ -407,6 +432,8 @@ pub struct BinaryEExpWriter_1_1<'value, 'top> {
     allocator: &'top BumpAllocator,
     buffer: &'value mut BumpVec<'top, u8>,
     value_writer_config: ValueWriterConfig,
+    macros: &'value MacroTable,
+    signature_iter: SignatureIterator<'value>,
```
zslayton, line 436: 🪧 Common logic for iterating over the parameters in a macro signature are encapsulated in the `SignatureIterator` type, which is now nested in both the binary and text e-expression writers.

src/lazy/encoder/binary/v1_1/value_writer.rs
```diff
@@ -681,58 +701,61 @@ impl BinaryValueWriter_1_1<'_, '_> {
...
-                let biased = address - BIAS;
-                let low_opcode_nibble = biased >> 8;
-                self.encoding_buffer.extend_from_slice_copy(&[
-                    // Opcode with low nibble setting bias
-                    0x40 | low_opcode_nibble as u8,
-                    // Remaining byte of magnitude in biased address
-                    biased as u8,
-                ]);
+        let macro_ref = macro_id.resolve(self.macros)?;
+        let address = macro_ref.address();
```
zslayton, line 705: 🪧 Now that the raw writers have access to the macro table, they can do their own ID resolution. This means that the binary writer no longer has to handle the case where the caller provided a text macro ID that may or may not be valid. The ID can either be mapped to an address or it's an error.

src/lazy/encoder/text/v1_1/value_writer.rs
```diff
@@ -271,31 +291,47 @@ pub struct TextEExpWriter_1_1<'value, W: Write> {
...
         writer: &'value mut LazyRawTextWriter_1_0<W>,
         depth: usize,
         parent_type: ParentType,
-        container_type: ContainerType,
-        opening_delimiter: &str,
-        value_delimiter: &'static str,
-        trailing_delimiter: &'static str,
+        opening_text: &str,
+        macros: &'value MacroTable,
+        invoked_macro: MacroRef<'value>,
```
zslayton, line 311: 🪧 Adding parameters to this function (to plumb through the macro information) caused it to trip `clippy`'s alarm for excessive parameters. This is only called in one place, so I refactored it to eliminate some parameters.

src/lazy/expanded/macro_table.rs
```diff
@@ -369,38 +373,55 @@ pub enum MacroKind {
 /// A Macro reference that is guaranteed to be valid for its lifespan.
 #[derive(Debug, Copy, Clone, PartialEq)]
 pub struct MacroRef<'top> {
-    id: MacroIdRef<'top>,
-    reference: &'top MacroDef,
+    qualified_address: QualifiedAddress,
+    def: &'top MacroDef,
 }
```
zslayton, line 378: 🪧 As mentioned prior, `MacroRef` now tracks which module the macro was found in.

src/lazy/expanded/template.rs
```diff
@@ -141,6 +141,63 @@ impl Parameter {
             ParameterCardinality::ZeroOrOne | ParameterCardinality::ZeroOrMore
         )
     }
+
+    /// If the parameter's encoding matches the provided argument, returns `Ok`.
+    /// Otherwise, returns an `Err`.
+    #[inline]
+    pub fn expect_encoding(&self, arg_encoding: &ParameterEncoding) -> IonResult<&Self> {
```
zslayton, line 148: 🪧 Helper methods for validating that the current macro signature parameter has the expected encoding/cardinality.

src/lazy/text/raw/v1_1/reader.rs
```diff
@@ -203,50 +214,64 @@ pub trait MacroIdLike<'a>: Sized + Copy {
...
-            LocalAddress(address) => {
-                if let Some(name) = macro_table
-                    .macro_at_address(address)
-                    .and_then(|macro_ref| macro_ref.name())
-                {
-                    return Ok(ResolvedId::new(Some(name), address));
-                }
-            }
-            SystemAddress(_system_address) => {
-                todo!("resolving qualified macro IDs")
```
zslayton, line 224: 🪧 Because macro handles now store a `QualifiedAddress`, it's possible to handle this case during resolution.

src/element/builders.rs
```diff
@@ -73,6 +73,12 @@ impl SequenceBuilder {
         self
     }
 
+    /// Adds all of the provided elements to the end of the [`Sequence`] being constructed.
+    pub fn push_all<E: Into<Element>, I: IntoIterator<Item = E>>(mut self, elements: I) -> Self {
+        self.values.extend(elements.into_iter().map(|e| e.into()));
+        self
+    }
```
zslayton, line 80: 🪧 Helper method for building a sequence when you have an iterator.

zslayton (commented):
🗺️ PR Tour 🧭

PR #975, opened 2025-06-09
Add expression group transcription to conformance DSL

nirosys:
*Issue #, if available:* #935

*Description of changes:*
Prior to this PR using an expression group within a `toplevel` clause would fail. The existing code captured the `::` for a normal expression group, but due to the DSL's transcription syntax we need to capture `#$::` and construct an expression group through the eexp writer.

This PR captures the `#$::` notation, and constructs the expression group via the writer.

Two tests were also added to the conformance DSL unit tests to ensure the behavior as well.

---

### DIFF
--- a/tests/conformance_dsl/fragment.rs
+++ b/tests/conformance_dsl/fragment.rs
@@ -360,6 +360,7 @@ impl WriteAsIon for ProxyElement<'_> {
                     Some(Symbol(symbol)) if symbol.text().is_some() => {
                         let text = symbol.text().unwrap();
                         const EEXP_PREFIX: &str = "#$:";
+                        const EXP_GROUP_PLACEHOLDER: &str = "#$::";
                         if text.starts_with(EEXP_PREFIX) {
                             // It's an e-expression. Start by isolating the macro ID.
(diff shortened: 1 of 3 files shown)


tests/conformance_dsl/fragment.rs
```diff
@@ -391,11 +392,24 @@ impl WriteAsIon for ProxyElement<'_> {
...
+                                } else if arg_is_expr_group {
+                                    // ...if it is an expression group, we need to encode it in an
+                                    // expression group built via our expression writer
+                                    // ProxyElement cannot do this alone.
+                                    let mut group_writer = eexp_writer.expr_group_writer()?;
+                                    let group_args = arg
+                                        .as_sexp()
+                                        .unwrap() // Verified above
+                                        .iter()
+                                        .skip(1);
```
jobarr-amzn, line 409: This is skipping the first element, the `#$::`, right? If I am correct could you add a comment for posterity, and if I am wrong could you add a comment telling me what is the deal here?

jobarr-amzn (approved):

tests/conformance_dsl/fragment.rs
nirosys: Yep that's exactly why. I'll add a comment, good call.

Issue #942, opened 2025-03-26
Conformance tests failing due to unimplemented system macros


nirosys:
We currently have 31 conformance tests failing due to missing system macros:
```
PanicHookInfo { payload: Any { .. }, location: Location { file: \"src/lazy/expanded/e_expression.rs\", line: 143, col: 32 }, can_unwind: true, force_no_backtrace: false }
```

```bash
% cat conformance_test_failures_update_3.json | jq -r '.[] | [to_entries[] | .key as $key | .value[] | select(.result // "" | . != "Ok") | {path: $key, error: .error}] | [.[] | select(.error | contains("e_expression.rs\", line: 143"))][].path' | sort | uniq -c
   3 ./ion-tests/conformance/system_macros/delta.ion
   5 ./ion-tests/conformance/system_macros/make_decimal.ion
  11 ./ion-tests/conformance/system_macros/make_timestamp.ion
   4 ./ion-tests/conformance/system_macros/parse_ion.ion
   4 ./ion-tests/conformance/system_macros/repeat.ion
   4 ./ion-tests/conformance/system_macros/sum.ion
```

> NOTE: I'll update with a list of the macros in use.

PR #981, opened 2025-06-18
Implement delta system macro.

nirosys:
*Issue #, if available:* #942

*Description of changes:*
This PR implements the `delta` system macro.

```
Running tests: ion-tests/conformance/system_macros/delta.ion
  delta can be invoked                                     ...  [Ok]
  delta produces a stream of values that is the delta of   ...  [Ok]
  arguments may not be                                     ...  [Ok]
  delta and repeat can be combined to generate             ...  [Failed]
  it is possible to create a delta of deltas encoding      ...  [Failed]
============================================================================
Source: ion-tests/conformance/system_macros/delta.ion
Test  : delta and repeat can be combined to generate
Error : ConformanceError(ConformanceErrorImpl { file: "", test_name: "", kind: IonError(Decoding(DecodingError { description: "macro named 'from_x_count_n_by_step' already exists", position: None })) })
----------------------------------------------
Source: ion-tests/conformance/system_macros/delta.ion
Test  : it is possible to create a delta of deltas encoding
Error : ConformanceError(ConformanceErrorImpl { file: "", test_name: "", kind: IonError(Decoding(DecodingError { description: "unrecognized encoding 'flex_int' specified for parameter", position: None })) })
----------------------------------------------
```
---


nirosys:
Patch % is down in part because I test the args for ValueLiteral. There is only one argument to delta, and it gets collapsed into an arg group. So the handling of ValueLiteral's when walkinging the arguments iterator isn't really needed. I'll probably re-work the implementation a little more to lean on the single argument.

nirosys:
Converted back to draft because I found an issue. Tracking it down to fix.

PR #991, opened 2025-07-21
Implement uint8 tagless encoding

nirosys:
*Issue #, if available:* #943

*Description of changes:*
This PR implements the "backend" support for tagless encoding of uint8s.

By "backend" I mean that the ability to read tagless uint8s is provided, however writing them leans on a function exposed by the EExpr writer for writing `uint8`s specifically. This mirrors the support for `FlexUInt` and will require adding some functionality to existing writers to allow plumbing parameter info down so that we can verify values being written to ensure they fit the range of numeric primitives, etc., as well as handle expression groups.

### CLI Reading
```
 % ./target/release/ion inspect ~/Code/sandbox/tagless_output/output.10n
┌──────────────┬──────────────┬─────────────────────────┬──────────────────────┐
│    Offset    │    Length    │       Binary Ion        │       Text Ion       │
├──────────────┼──────────────┼─────────────────────────┼──────────────────────┘
│            0 │            4 │ e0 01 01 ea             │ $ion_1_1 // Version marker
├──────────────┼──────────────┼─────────────────────────┤
│            4 │            6 │ e7 f9 24 69 6f 6e       │ $ion:: // <text>
│           10 │           44 │ fc 55                   │ (
│           12 │            2 │ ee 10                   │ · module
│           14 │            2 │ a1 5f                   │ · _
│           16 │            5 │ c4                      │ · (
│           17 │            2 │ ee 0f                   │ · · symbol_table
│           19 │            2 │ a1 5f                   │ · · _
│              │              │                         │ · )
│           21 │           33 │ fc 3f                   │ · (
│           23 │            2 │ ee 0e                   │ · · macro_table
│           25 │            2 │ a1 5f                   │ · · _
│           27 │           27 │ fc 33                   │ · · (
│           29 │            6 │ a5 6d 61 63 72 6f       │ · · · macro
│           35 │            4 │ a3 66 6f 6f             │ · · · foo
│           39 │           10 │ c9                      │ · · · (
│           40 │            7 │ e7 f7 75 69 6e 74 38    │ · · · · uint8:: // <text>
│           47 │            2 │ a1 78                   │ · · · · x
│              │              │                         │ · · · )
│           49 │            5 │ c4                      │ · · · (
│           50 │            2 │ a1 25                   │ · · · · '%'
│           52 │            2 │ a1 78                   │ · · · · x
│              │              │                         │ · · · )
│              │              │                         │ · · )
│              │              │                         │ · )
│              │              │                         │ )
[...]

PR #992, opened 2025-08-01
Add support for tagless types in binary writer through standard API

nirosys:
*Issue #, if available:* n/a

*Description of changes:*
> NOTE: This PR includes the changes in #991.

This PR extends the binary writer API by adding a `BinaryEExpParameterValueWriter_1_1` which becomes the `ValueWriter` implementation returned by the `BinaryEExpWriter_1_1`. The implementation wires up `uint8` encoding, and continues supporting tagged encodings for all of the existing types that will ultimately support taggless encodings.

This extends the writer API to provide seamless handling of tagless encoding for code like:
```rust
//  (macro foo (uint8::x) (%x))
let mut eexp_writer = writer.eexp_writer("foo")?;
eexp_writer.write_int(&5.into())?;
eexp_writer.close()?;
```
To generate the expected:
```
$ ./target/release/ion inspect ~/Code/sandbox/tagless_output/output.10n
┌──────────────┬──────────────┬─────────────────────────┬──────────────────────┐
│    Offset    │    Length    │       Binary Ion        │       Text Ion       │
├──────────────┼──────────────┼─────────────────────────┼──────────────────────┘
│            0 │            4 │ e0 01 01 ea             │ $ion_1_1 // Version marker
├──────────────┼──────────────┼─────────────────────────┤
│            4 │            6 │ e7 f9 24 69 6f 6e       │ $ion:: // <text>
│           10 │           44 │ fc 55                   │ (
│           12 │            2 │ ee 10                   │ · module
│           14 │            2 │ a1 5f                   │ · _
│           16 │            5 │ c4                      │ · (
│           17 │            2 │ ee 0f                   │ · · symbol_table
│           19 │            2 │ a1 5f                   │ · · _
│              │              │                         │ · )
│           21 │           33 │ fc 3f                   │ · (
│           23 │            2 │ ee 0e                   │ · · macro_table
│           25 │            2 │ a1 5f                   │ · · _
│           27 │           27 │ fc 33                   │ · · (
│           29 │            6 │ a5 6d 61 63 72 6f       │ · · · macro
│           35 │            4 │ a3 66 6f 6f             │ · · · foo
│           39 │           10 │ c9                      │ · · · (
│           40 │            7 │ e7 f7 75 69 6e 74 38    │ · · · · uint8:: // <text>
│           47 │            2 │ a1 78                   │ · · · · x
│              │              │                         │ · · · )
│           49 │            5 │ c4                      │ · · · (
│           50 │            2 │ a1 25                   │ · · · · '%'
│           52 │            2 │ a1 78                   │ · · · · x
│              │              │                         │ · · · )
│              │              │                         │ · · )
│              │              │                         │ · )
│              │              │                         │ )
├──────────────┼──────────────┼─────────────────────────┤
│           54 │            2 │ 18                      │ (:foo
│           55 │            1 │ 05                      │ · 5 // x
│              │              │                         │ )
├──────────────┼──────────────┼─────────────────────────┤
│           55 │            1 │ 05                      │ 5
├──────────────┼──────────────┼─────────────────────────┤
│           56 │              │                         │  // End of stream
└──────────────┴──────────────┴─────────────────────────┘
```

When trying to write a value greater than 255, we get an IonError:
```
unable to write int: Encoding(EncodingError { description: "error with value provided for 'x': provided unsigned integer value does not fit within 1 byte" })
```

With this change, type checking and parameter cardinality is checked in a way that broke a couple unit tests. Originally two unit tests existed that wrote invocations for `(:none foo bar baz)`, and `(:make_string foo bar baz)`. The first unit test no longer works due to `none` not taking arguments. The second unit test fails because the eexp invocation requires building an expression group which currently is not implemented for binary ion. These issues now get caught, rather than writing the equivalent of `(:none) foo bar baz`.

---

### DIFF
--- a/src/lazy/encoder/binary/v1_1/fixed_uint.rs
+++ b/src/lazy/encoder/binary/v1_1/fixed_uint.rs
@@ -65,6 +65,20 @@ impl FixedUInt {
     pub fn size_in_bytes(&self) -> usize {
         self.size_in_bytes
     }
+
+    /// Write the provided UInt-like as a uint8 ensuring that the value fits in 8bits.
+    #[inline]
+    pub(crate) fn write_as_uint8<W: Write>(output: &mut W, value: impl Into<UInt>) -> IonResult<()> {
+        let value = value.into().data;
+        let encoded_bytes = value.to_le_bytes();
+
+        if !(0..256u128).contains(&value) {
+            return IonResult::encoding_error("provided unsigned intger value does not fit within 1 byte");
+        }
+
+        output.write_all(&encoded_bytes[..1])?;
+        Ok(())
+    }
 }
 
 impl TryFrom<FixedUInt> for Coefficient {
--- a/src/lazy/encoder/binary/v1_1/value_writer.rs
+++ b/src/lazy/encoder/binary/v1_1/value_writer.rs
@@ -1075,18 +1075,13 @@ impl<'value, 'top> ValueWriter for BinaryEExpParameterValueWriter_1_1<'value, 't
             .and_then(|p| p.expect_single_expression())?;
 
         let result = match param.encoding() {
-            PE::UInt8 => {
-                value
-                    .try_into()
-                    .and_then(|uint: UInt| FixedUInt::write_as_uint8(self.buffer, uint))
-                    .map(|_| ())
-            }
-            PE::FlexUInt => {
-                value
-                    .try_into()
-                    .and_then(|uint: UInt| FlexUInt::write(self.buffer, uint))
-                    .map(|_| ())
-            }
+            PE::UInt8 => value
+                .try_into()
+                .and_then(|uint: UInt| FixedUInt::write_as_uint8(self.buffer, uint)),
+            PE::FlexUInt => value
+                .try_into()
+                .and_then(|uint: UInt| FlexUInt::write(self.buffer, uint))
+                .map(|_| ()),
             PE::Tagged => {
                 let value_writer = BinaryValueWriter_1_1::new(
                     self.allocator,
@@ -1096,8 +1091,9 @@ impl<'value, 'top> ValueWriter for BinaryEExpParameterValueWriter_1_1<'value, 't
                 );
                 value_writer.write_i64(value)
             }
-            encoding => return IonResult::encoding_error(format!("value does not satisfy encoding type {encoding}")),
-
+            encoding => IonResult::encoding_error(
+                format!("value does not satisfy encoding type {encoding}")
+            ),
         };
 
         result.map_err(|err| error_context(param.name(), err))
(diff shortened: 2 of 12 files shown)


src/lazy/encoder/binary/v1_1/fixed_uint.rs
```diff
@@ -65,6 +65,20 @@ impl FixedUInt {
     pub fn size_in_bytes(&self) -> usize {
         self.size_in_bytes
     }
+
+    /// Write the provided UInt-like as a uint8 ensuring that the value fits in 8bits.
+    #[inline]
+    pub(crate) fn write_as_uint8<W: Write>(output: &mut W, value: impl Into<UInt>) -> IonResult<()> {
```
popematt, line 71: Why can't we use `value: impl Into<u8>`? I'm sure you have a good reason, but I don't know what that is and my gut reaction is that it seems like accepting `impl Into<UInt>` is just concealing a potential problem from the users.

src/lazy/encoder/binary/v1_1/value_writer.rs
```diff
@@ -967,6 +968,323 @@ impl<'value, 'top> BinaryAnnotatedValueWriter_1_1<'value, 'top> {
...
+        bool => write_bool,
+        &Decimal => write_decimal,
+        &Timestamp => write_timestamp,
+        impl AsRef<str> => write_string,
+        impl AsRef<[u8]> => write_clob,
+        impl AsRef<[u8]> => write_blob,
+    );
+
+    // These types need to have ranges validated to ensure we can write the value to a tagless
+    // parameter if the parameter is tagless.
```
popematt, line 1039: Which types?

src/lazy/encoder/binary/v1_1/value_writer.rs
```diff
@@ -967,6 +968,323 @@ impl<'value, 'top> BinaryAnnotatedValueWriter_1_1<'value, 'top> {
...
+            .and_then(|p| p.expect_single_expression())?;
+
+        let result = match param.encoding() {
+            PE::UInt8 => value
+                .try_into()
+                .and_then(|uint: UInt| FixedUInt::write_as_uint8(self.buffer, uint)),
+            PE::FlexUInt => value
+                .try_into()
+                .and_then(|uint: UInt| FlexUInt::write(self.buffer, uint))
+                .map(|_| ()),
```
popematt, line 1083: It's been a while... does the `buffer` track its own position so that it's safe to ignore the returned "number of bytes written"?

src/lazy/encoder/binary/v1_1/value_writer.rs
```diff
@@ -2920,21 +3238,23 @@ mod tests {
         Ok(())
     }
 
+    // TODO: Revisit this unit test, we need to change the system macro being called.. originally
+    // it was calling (:none), but with the parameter tracking it will fail since :none does not
```
popematt, line 3242: `meta` is like `none`, but it takes arguments. You could use `meta` and leave the rest of this test largely unchanged, I think.

src/lazy/encoder/binary/v1_1/value_writer.rs
nirosys: Yep, `buffer` is a `BumpVec` which tracks writes by extending its length.

src/lazy/encoder/binary/v1_1/value_writer.rs
nirosys: Oh nice, good call. I'll still have to circle back around to it though after writing e-groups in binary is implemented.

src/lazy/encoder/binary/v1_1/fixed_uint.rs
popematt: > I could switch this over to use a TryInto<u8>, and use the implementation to validate the range of the underlying value. 

No, don't worry about it. I see also that this is a `pub(crate)` function, so even if we were to change it, it would need to be changed elsewhere (public signatures) in order for the user to actually get any benefit.

popematt (approved):

Issue #844, opened 2024-10-21
Trait `WriteAsIon` is not implemented for `Element`


popematt:
```
error[E0277]: the trait bound `ion_rs::Element: WriteAsIon` is not satisfied
   --> src/bin/ion/commands/schema/validate.rs:234:26
    |
234 |             writer.write(ion_list![].with_annotations(["valid"]))
    |                    ----- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `WriteAsIon` is not implemented for `ion_rs::Element`
    |                    |
    |                    required by a bound introduced by this call
    |
note: required by a bound in `ion_rs::ValueWriter::write`
   --> /Users/popematt/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ion-rs-1.0.0-rc.8/src/lazy/encoder/value_writer.rs:78:32
    |
78  |     fn write(self, value: impl WriteAsIon) -> IonResult<()> {
    |                                ^^^^^^^^^^ required by this bound in `ValueWriter::write`
help: consider borrowing here
    |
234 |             writer.write(&ion_list![].with_annotations(["valid"]))
    |                          +
```

More generally, it seems like `WriteAsIon` is implemented for references rather than owned values for a number of things in the data model. This is going to be a bit of a surprise to everyone who is starting to use `ion-rust` for the first time if we leave things the way they are.

PR #1003, opened 2025-09-30
Both `Element` and `&Element` implement `WriteAsIon` now

Shaurya0108:
*Issue #, if available:*
#844 

*Description of changes:*
Both `Element` and `&Element` implement `WriteAsIon` now
- `Element` implements it directly
- `&Element` gets it via the implementation `impl<T: WriteAsIon> WriteAsIon for &T` in `write_as_ion.rs`

This fixes the issue of `Element` not being able to implement the trait, new users can now use the API without needing to remember to add `&`.

*Example:*
```
// using this to invoke the api
// this creates an empty List using the macro and adds the annotation "valid" to the List
// then writes the annotated element to the ion stream
writer.write(ion_list![].with_annotations(["valid"]))?;
```

Before
```
// gave this error because Element did not implement WriteAsIon
error[E0277]: the trait bound `ion_rs::Element: WriteAsIon` is not satisfied
```

After
```
// works with or without the '&' because both implement WriteAsIon now
// produces an empty List, annotated with 'valid::'
```

### DIFF
--- a/src/lazy/encoder/write_as_ion.rs
+++ b/src/lazy/encoder/write_as_ion.rs
@@ -77,7 +77,7 @@ pub trait WriteAsIon {
     }
 }
 
-impl WriteAsIon for &Element {
+impl WriteAsIon for Element {
     fn write_as_ion<V: ValueWriter>(&self, writer: V) -> IonResult<()> {
         if self.annotations().is_empty() {
             self.value().write_as_ion(writer)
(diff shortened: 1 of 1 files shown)


popematt:
Can you please add and/or update the tests to cover this change?

jobarr-amzn (approved):

PR #1004, opened 2025-10-01
Move typed null handling from ion_type() to read_null() methods

Shaurya0108:
*Issue #, if available:*
#828 

*Description of changes:*

Typed nulls in Ion 1.1 use `OpcodeType::TypedNull` encoding where the actual type information is stored in the value body. The `ion_type()` method was  only reading the header type while missing the logic to decode the actual type from the body.

Applying the typed null handling logic from `read_null()` to `ion_type()` determines if this is a typed null or regular value.


Shaurya0108:
Updated to not use the read null function specifically in ion 1.1 binary reader, instead added a check to map null types to binary values

nirosys (approved):

PR #1012, opened 2025-10-24
Update opcodes and opcode creation; update or ignore failing tests

nirosys:
*Issue #, if available:* n/a

*Description of changes:*
This PR takes the first steps towards updating to the latest ion 1.1 spec by updating the Opcode values, creation, and kinds, to match the current spec.

Tests that were identified as failing due to manually supplied binary ion have been updated though I have not verified that every test is passing correctly. Next efforts will be to validate the primitive values, and begin working on changes to containers and templates.

Throughout the unit tests I've left `TODO`s to revisit failing tests (that are now ignored, or commented out) once 1.1 features are updated or implemented. These comments will serve as a way to identify tests that should be working as the efforts continue.

I've also disabled the conformance tests for now, until we have read and write support since the conformance dsl relies heavily on both.

---


nirosys:
Wow that was dumb of me. I updated the typed null unit tests, but not the actual parsing..

PR #1013, opened 2025-10-29
Evaluate scalar value parsing to ensure they match the current 1.1 spec

Shaurya0108:
*Description of changes:*

Test Coverage Improvements:
 - Added test cases for lists and s-expressions
 - Added timestamp test coverage with short-form case
 - Fixed string encoding in variable-length list tests (0x99 -> 0xF8)
 - Cleaned up symbol test cases

Opcode Compliance Fix by adding a missing 0xFD opcode mapping for FlexSym mode structs in type descriptor, more information [here](https://amazon-ion.github.io/ion-docs/books/ion-1-1/binary/opcodes.html).

Following the Ion 1.1 specification which states FlexSym should only handle symbol IDs and inline text:
 - FlexInt value 0 now correctly represents symbol ID 0
 - FlexInt positive values represent symbol IDs 
 - FlexInt negative values represent inline text length
 - No special opcode escape handling for FlexInt value 0

Implementing Ion 1.1 spec formula for symbols with a symbol address: symbolId = (flexUInt << 3) | (opcode & 0b111) as well as extracting the low 3 bits from opcode byte as LSB and reading FlexUInt from value body as MSB

### DIFF
--- a/src/lazy/binary/raw/v1_1/binary_buffer.rs
+++ b/src/lazy/binary/raw/v1_1/binary_buffer.rs
@@ -638,8 +638,8 @@ impl<'a> BinaryBuffer<'a> {
                 let (flex_uint, _) = input_after_opcode.read_flex_uint()?;
                 let flex_uint_size = flex_uint.size_in_bytes();
                 let total_length = 1 + flex_uint_size; // opcode + FlexUInt
-                (total_length, 0, flex_uint_size, DelimitedContents::None) }
-            else {
+                (total_length, 0, flex_uint_size, DelimitedContents::None)
+            } else {
                 let length = match header.length_type() {
                     LengthType::InOpcode(n) => FlexUInt::new(0, n as u64),
                     LengthType::Unknown => FlexUInt::new(0, 0), // Delimited value, we do not know the size.


src/lazy/binary/raw/v1_1/reader.rs
```diff
@@ -283,15 +282,6 @@ mod tests {
...
             0x67, 0x74, 0x68, 0x20, 0x65, 0x6E, 0x63, 0x6f, 0x64, 0x69, 0x6E, 0x67,
-
-            // Symbol ID: 1
-            // 0x51, 0x01,
-
-            // // Symbol ID: 257
-            // 0x52, 0x01, 0x00,
-
-            // // Symbol ID: 65,793
-            // 0x50, 0x00, 0x00, 0x00,
```
nirosys: We should add these back in (with the appropriate opcodes) now that you've updated the symbol ID handling.

src/lazy/binary/raw/v1_1/binary_buffer.rs
```diff
@@ -633,6 +633,12 @@ impl<'a> BinaryBuffer<'a> {
                 let total_length = after.offset() - self.offset();
                 let value_body_length = total_length - 1; // Total length - sizeof(opcode)
                 (total_length, 0, value_body_length, contents)
+            } else if opcode.opcode_type == OpcodeType::SymbolAddress {
+                let input_after_opcode = input.consume(1);
+                let (flex_uint, _) = input_after_opcode.read_flex_uint()?;
+                let flex_uint_size = flex_uint.size_in_bytes();
+                let total_length = 1 + flex_uint_size; // opcode + FlexUInt
+                (total_length, 0, flex_uint_size, DelimitedContents::None)
```
nirosys: This shouldn't be needed since it should fall through with the `Unknown` case. With the changes we discussed see about removing this and lmk if there's an issue still.

nirosys:
CI build errors are unrelated, new rust brings new clippy lints.
```
Error: error: this `impl` can be derived
   --> src/lazy/expanded/macro_evaluator.rs:687:1
    |
687 | / impl<D: Decoder> Default for EvaluatorState<'_, D> {
688 | |     fn default() -> Self {
689 | |         Self::Empty
690 | |     }
691 | | }
    | |_^
```

Happy rust release day!

src/lazy/binary/raw/v1_1/binary_buffer.rs
Shaurya0108: Removing this causes the `incompleteError` we noticed: 
```
Error: Incomplete(IncompleteError { label: "reading a FlexUInt", position: Position { byte_offset: 47, byte_length: None, line_column: None } })
```

Without this the stream parsing gets out of sync and fails.
With this the stream parsing stays synchronized.

Edit: looking into ways to change the FlexUInt reading for the tests without having to add this

src/lazy/binary/raw/v1_1/binary_buffer.rs
```diff
@@ -633,7 +633,13 @@ impl<'a> BinaryBuffer<'a> {
                 let total_length = after.offset() - self.offset();
                 let value_body_length = total_length - 1; // Total length - sizeof(opcode)
                 (total_length, 0, value_body_length, contents)
-            } else {
+            } else if opcode.opcode_type == OpcodeType::SymbolAddress {
```
Shaurya0108: Without this, the parser is interpreting the FlexUInt as a length field @nirosys. Symbol ID extraction in read_symbol_id() is failing.

nirosys (approved):

PR #1014, opened 2025-11-03
Handle the field-value-based delimited-container-end opcode

Shaurya0108:
*Issue #, if available:*

*Description of changes:*
Updated FlexSym's opcode handler to handle 0xEF as a direct field-value-based delimited-container-end opcode, separate from FlexSym parsing. Previously, FlexSym would parse and return an opcode when it encountered 0xEF. Now 0xEF is detected directly at the container parsing level before attempting FlexSym parsing.

**FlexSymValue Enum:**
```
// Before:
#[derive(Debug, Clone, Copy)]
pub enum FlexSymValue<'top> {
    SymbolRef(RawSymbolRef<'top>),
    Opcode(Opcode),
}

// After:
#[derive(Debug, Clone, Copy)]
pub enum FlexSymValue<'top> {
    SymbolRef(RawSymbolRef<'top>),
}
```

0xEF detection happens only in delimited struct parsing and 0xEF is handled as direct opcode in the value position,  isolating it to delimited structs.

Catching 0xEF:
1. peek_delimited_field_flexsym - Catches 0xEF when looking for delimited struct field names
2. peek_delimited_struct_value - Catches 0xEF when looking for delimited struct field values
3. peek_value_expr - Catches 0xEF when looking for any struct field values

### DIFF
--- a/src/lazy/binary/raw/v1_1/binary_buffer.rs
+++ b/src/lazy/binary/raw/v1_1/binary_buffer.rs
@@ -565,9 +565,7 @@ impl<'a> BinaryBuffer<'a> {
             let (field, after_value) = match after_name.peek_delimited_struct_value()? {
                 (None, after) => {
                     // Check if this is the throwaway field name + container end pattern
-                    if let Some(opcode) = after.peek_opcode()
-                        && opcode.is_delimited_end()
-                    {
+                    if after.peek_opcode().map_or(false, |opcode| opcode.is_delimited_end()) {
                         return Ok((None, after.consume(1)));
                     }
                     if after.is_empty() {


src/lazy/binary/raw/v1_1/binary_buffer.rs
```diff
@@ -535,12 +541,14 @@ impl<'a> BinaryBuffer<'a> {
             return Ok((None, *self));
         }
 
+        // Check for delimited container end opcode (0xEF) before trying to read FlexSym
+        if let Some(0xEF) = self.peek_next_byte() {
+            let after = self.consume(1);
+            return Ok((None, after));
+        }
+
```
nirosys: This should be removed. The name position is interpreted as a FlexSym and `0xEF` is a valid start for a 4byte flexsym value.

src/lazy/binary/raw/v1_1/struct.rs
```diff
@@ -190,12 +190,12 @@ impl<'top> RawBinaryStructIterator_1_1<'top> {
             return Ok(None);
         }
 
+        if let Some(0xEF) = buffer.peek_next_byte() {
+            return Ok(None);
+        }
+
```
nirosys: Same as above, potentially valid flexsym.

src/lazy/binary/raw/v1_1/struct.rs
```diff
@@ -237,6 +237,11 @@ impl<'top> RawBinaryStructIterator_1_1<'top> {
         BinaryBuffer<'top>,
     )> {
         let opcode = buffer.expect_opcode()?;
+
+        if opcode.is_delimited_end() {
+            return Ok((None, buffer));
+        }
+
```
nirosys: nit: this could be worked into the if-else below. It's not really any more special than the `NOP` handling.

Should this return:
```rust
Ok((None, buffer.consume(1)))
```
If the original ByteBuffer is setup correctly, then the end of container marker should be the last byte in the buffer. So moving beyond it, should leave us with an empty buffer, which will be a quick check if someone tries to read again.

nirosys (commented):
Had a double-take. Looks like there's duplicate code between the ByteBuffer, and `RawBinaryStructIterator_1_1`. We could probably circle back and clean that up later, but all of the value peeking and stuff looks to be duplicated.

PR looks good, just have a couple issues with testing for the delimiter in the name position. Left a couple other notes to think about, but they're not blocking.

src/lazy/binary/raw/v1_1/binary_buffer.rs
```diff
@@ -564,6 +564,13 @@ impl<'a> BinaryBuffer<'a> {
 
             let (field, after_value) = match after_name.peek_delimited_struct_value()? {
                 (None, after) => {
+                    // Check if this is the throwaway field name + container end pattern
+                    if let Some(opcode) = after.peek_opcode() {
+                        if opcode.is_delimited_end() {
+                            // Throwaway field name followed by container end - consume 0xEF and end
+                            return Ok((None, after.consume(1)));
+                        }
+                    }
```
nirosys: nit: could be merged with `if` below.

src/lazy/encoder/binary/v1_1/flex_sym.rs
```diff
@@ -12,7 +11,6 @@ use crate::RawSymbolRef;
 #[derive(Debug, Clone, Copy)]
 pub enum FlexSymValue<'top> {
```
nirosys: Definitely not something for this PR, just making a note. We can probably convert this to a struct, now that it's just wrapping a symbol ref. Something to come back to later.

src/lazy/binary/raw/v1_1/binary_buffer.rs
Shaurya0108: whoops this slipped

PR #1019, opened 2026-05-05
Add support for arbitrary precision integers, decimals, and timestamps

popematt:
### Issue #, if available:

#7 

### Description of changes:

This PR removes the `i128`/`u128` size limit on Ion integer values by introducing a tagged-union representation that stores small values inline and heap-allocates only when needed.

My design goals were:
* Optimize for the common case. Smaller numbers are prioritized at the expense of very large numbers. I have some inefficiencies for the large numbers (a lot of intermediate allocation is the main thing), and I've chosen not to bother optimizing them. If it's a problem, we can revisit that later.
* Don't re-implement big integer arithmetic. I could have used `Box<[u8]>`, but it was simpler to just use `BigInt` and `BigUInt`.
* Avoid/minimize changes to the public API.
  * Aside from the inevitable loss of `Copy`, I tried to make purely additive changes to the public API.
  * I intentionally do not expose any APIs with BigInt and BigUInt. We've been through the hassle of re-exports before, and I don't want to reintroduce that again. The public APIs for getting large values of out Int and UInt are intentionally sparse (`to_le_bytes()`, `to_le_signed_bytes()`) so that there are no API dependencies.

Note that this PR does not necessarily add support for arbitrarily large values in Ion 1.1. (It did get added in some cases as a consequence of ensuring that the code will still compile, but it is not assured to be universally supported.)

I've grouped the changes into logical commits to make it easier to review.

   **Commit 1: Remove `Copy` from `Int`, `UInt`, `Timestamp`, and `Decimal`**

   These types can no longer be `Copy` because the new representation may heap-allocate. This commit makes the mechanical change across the codebase — replacing implicit copies with explicit `.clone()` calls and adjusting move semantics.

   **Commit 2: Add `IntData` and `UIntData`**

   Introduces `src/types/int_data.rs`, a pair of 16-byte tagged unions that serve as the backing storage for `Int` and `UInt`:

   - Values that fit in 127 bits (unsigned) or 126 bits (signed) are stored inline with no allocation.
   - Larger values are stored as heap-allocated `BigUint`/`BigInt`.
   - A tag bit in byte[0] distinguishes the two representations.
   - Includes arithmetic operations (`Add`, `Sub`, `Mul`, `Div`, `Rem` and their assign variants), comparison, hashing, display, and serialization to/from LE and BE byte representations.
   - The `from_le_bytes` / `from_le_signed_bytes` fast paths avoid `BigInt` allocation entirely when the input is ≤16 bytes.

   **Commit 3: Wire up arbitrary-precision support across the codebase**

   - `Int` and `UInt` now wrap `IntData`/`UIntData` instead of raw `i128`/`u128`.
   - Binary encoding (`DecodedInt::write`, `read_int`) handles integers of any size using sign-and-magnitude BE byte representation.
   - Text parsing (`MatchedInt`, `MatchedDecimal`, `MatchedTimestamp`) falls back to `BigUint::parse_bytes` for values exceeding `i128`.
   - `Decimal` comparison uses `BigUint` for the slow path when coefficients exceed `u128`.
   - `Timestamp` nanosecond conversion uses `UIntData` division for coefficients exceeding `u128`.
   - Ion hash representation handles big integer magnitudes via BE byte encoding.
   - `FixedUInt::write` and `FixedInt::encoded_size` delegate to `UIntData`/`IntData` methods directly.
   - Adds `from_le_bytes`, `to_le_bytes`, `from_le_signed_bytes`, and `to_le_signed_bytes` as public methods on `UInt`/`Int`.


src/types/int_data.rs
almann, line 1: This encoding seems really clever and a half, are we sure it is justified?  Why not something more simple unless we have the data to do this with unsafe `union`?

```rust
pub(crate) enum UIntData {
    Inline(u128),
    Heap(Box<BigUint>),
}
```

Now the above is roughly 24-bytes from the 16-bytes we had before, but what are we trying to optimize for here?

The other issue I see with this design is that we have two indirections to get to the heap allocated value, so I am wondering if `Heap(BigUint)` is good enough too.  Granted the latter gets you to 32-bytes, so again, I think the extra indirection _could_ be warranted, but I would like to understand the rationale for this code better.

almann: As per offline discussion, you are planning to make this `enum`, so I will hold off on the other comments.

almann (changes requested):

src/types/decimal/mod.rs
```diff
@@ -241,21 +243,31 @@ impl Decimal {
...
-        let mut scaled_coefficient: u128 = d1.coefficient().magnitude().data;
-        scaled_coefficient *= 10u128.pow(exponent_delta as u32);
-        UInt::from(scaled_coefficient).cmp(&d2.coefficient().magnitude())
+        let d1_mag = d1.coefficient().magnitude();
+        let d2_mag = d2.coefficient().magnitude();
+        // Fast path: both fit in u128
+        if let (Some(m1), Some(m2)) = (d1_mag.as_u128(), d2_mag.as_u128()) {
+            if let Some(scaled) = m1.checked_mul(10u128.pow(exponent_delta as u32)) {
+                return scaled.cmp(&m2);
+            }
```
almann, line 254: I believe pow can panic here, this probably should be checked_pow; it might be good to make sure we have tests with large exponents and whatnot to test/prove this (>39) in case this is not really an issue.

almann: Some example tests that I believe should prove the panic (Claude also found/demonstrated some other panic cases that are similar when I asked it to verify this with unit tests):

```rust
    #[test]
    fn convert_large_coefficient_decimal_to_bigdecimal() {
        use crate::decimal::coefficient::Coefficient;
        use crate::Int;
        // 2^128 + 1 is a valid Ion decimal coefficient that exceeds u128::MAX.
        let mut bytes = vec![1u8];
        bytes.extend(vec![0u8; 16]);
        bytes[16] = 1;
        let big_int = Int::from_le_signed_bytes(&bytes);
        let d = Decimal::new(Coefficient::new(big_int), 0i64);
        let result: Result<BigDecimal, _> = d.try_into();
        assert!(result.is_ok());
    }

    #[test]
    fn compare_decimals_with_large_exponent_difference() {
        // 1e40 and 1e0 are both valid Ion decimals. Comparing them requires scaling
        // one coefficient by 10^40, which exceeds i128::MAX.
        let d1 = Decimal::new(1i64, 40i64);
        let d2 = Decimal::new(1i64, 0i64);
        assert_eq!(d1.cmp(&d2), Ordering::Greater);
    }

    #[test]
    fn trunc_decimal_with_large_negative_exponent() {
        // 1e-40 is a valid Ion decimal. trunc() should return 0.
        let d = Decimal::new(1i64, -40i64);
        assert_eq!(d.trunc(), Decimal::new(0i64, 0i64));
    }

    #[test]
    fn fract_decimal_with_large_negative_exponent() {
        // 1e-39 is a valid Ion decimal. fract() should return the value itself
        // since the integer part is zero.
        let d = Decimal::new(1i64, -39i64);
        assert_eq!(d.fract(), Decimal::new(1i64, -39i64));
    }

    #[test]
    fn subtract_decimals_with_large_exponent_difference() {
        // 1e40 - 1e0 is valid Ion decimal arithmetic.
        let lhs = Decimal::new(1i64, 40i64);
        let rhs = Decimal::new(1i64, 0i64);
        let result = lhs - rhs;
        assert!(result > Decimal::new(0i64, 0i64));
    }
```

src/lazy/streaming_raw_reader.rs
```diff
@@ -262,7 +262,7 @@ impl<Encoding: Decoder, Input: IonInput> StreamingRawReader<Encoding, Input> {
 }
 
 // This is a separate trait so it can be `dyn`-compatible.
-pub trait IoBufferHandle {
+pub(crate) trait IoBufferHandle {
```
almann, line 265: Do we have to worry about public API concerns on narrowing this?

src/result/mod.rs
```diff
@@ -66,6 +66,12 @@ pub enum IonError {
     Conversion(#[from] ConversionError),
 }
 
+impl From<Infallible> for IonError {
+    fn from(_: Infallible) -> Self {
+        unreachable!()
+    }
+}
+
```
almann: Why is this needed?  Probably warrants a comment.

src/types/integer/int_data.rs
```diff
@@ -0,0 +1,712 @@
...
+/// Generates `TryFrom<Data> for primitive` impls that extract the inline value and narrow it.
+macro_rules! impl_try_from_data_to_primitive {
+    ($data:ty => $($t:ty),*) => ($(
+        impl std::convert::TryFrom<$data> for $t {
+            type Error = IonError;
+            fn try_from(value: $data) -> Result<Self, IonError> {
+                value.as_small_value()
+                    .and_then(|v| <$t>::try_from(v).ok())
+                    .ok_or_else(|| IonError::decoding_error(
+                        concat!("{} value exceeds range of ", stringify!($data), stringify!($t))
```
almann: I think this is not right, or rather it looks like it should be `format!`.

src/types/integer/int_data.rs
```diff
@@ -0,0 +1,712 @@
...
+
+    pub(crate) fn is_negative(&self) -> bool {
+        match &self.0 {
+            SmallValue(value) => *value < 0,
+            BigValue(big) => cold_path! { big.is_negative() },
+        }
+    }
+
+    pub(crate) fn byte_len(&self) -> usize {
+        self.to_le_bytes().len()
```
almann: This forces allocation, while this is probably fine for now, I think we should document this fact.

src/types/timestamp.rs
```diff
@@ -220,22 +218,19 @@ impl Timestamp {
...
+                } else {
+                    let divisor = 10u128.saturating_pow(nano_exp.unsigned_abs() as u32);
+                    let divided = mag.data / divisor;
+                    // Default value can only be reached if the fractional seconds are greater than 1.
+                    u128::try_from(divided).unwrap_or(0)
                 };
-                // Adjust for the exponent
-                let nanoseconds = (magnitude as f64 * 10f64.powi(exponent_delta as i32)) as u32;
-                Some(nanoseconds)
+                Some(nanos as u32)
```
almann, line 236: This will truncate precision--is that intended? I don't quite think this is correct.

almann (changes requested):

src/types/integer/int_data.rs
popematt: 🤦‍♂️

src/lazy/streaming_raw_reader.rs
popematt: No. This is already effectively private. The change is to fix a Clippy complaint.

src/types/timestamp.rs
popematt: I think it's okay.
* It was already truncating to u32 before.
* This method returns only the nanoseconds part, not an absolute number of nanos, so the max possible number of nanoseconds is 999,999,999 compared to `u32::MAX` is 4,294,967,296. If this ended up actually truncating anything, then we're not properly validating/normalizing the data when the Timestamp is constructed.

I can add a comment to that effect.

src/result/mod.rs
popematt: It was needed for a prior revision, but it's not needed anymore. I'll remove it.

src/types/integer/int_data.rs
popematt: I'll just fix it to not use allocation. It's easy enough to calculate the minimal number of bytes to represent a number.

popematt: I spoke too soon. It's not so simple for the BigInt negative number case.

src/types/integer/int_data.rs
```diff
@@ -288,14 +311,14 @@ impl From<UIntData> for IntData {
...
+        impl std::convert::TryFrom<$from_type> for $to_type {
             type Error = IonError;
-            fn try_from(value: $data) -> Result<Self, IonError> {
+            fn try_from(value: $from_type) -> Result<Self, IonError> {
                 value.as_small_value()
-                    .and_then(|v| <$t>::try_from(v).ok())
+                    .and_then(|v| <$to_type>::try_from(v).ok())
                     .ok_or_else(|| IonError::decoding_error(
-                        concat!("{} value exceeds range of ", stringify!($data), stringify!($t))
+                        concat!(stringify!($from_type), " value exceeds range of ", stringify!($to_type))
```
popematt, line 321: This line is the fix. The rest of the changes in this macro are just renaming the macro arguments to be more clear.

almann (approved):
LGTM--I do notice some warning nags about unused imports, worth considering a fixup if not spurious.

Issue #970, opened 2025-06-03
Mathematical and convenience operations on the numeric types (Int, Decimal, Float)


jobarr-amzn:
We need more mathematical and convenience operations on the numeric types, e.g. `+`, `-`, `*`, `/`, `%` and various type conversions like `to_f64` even if lossy. 

For e.g. `Int` these are quite easy to implement, for `Decimal` it's more complicated. These conveniences make the Ion values much easier to work with.

Issue #838, opened 2024-09-30
Consider switching chrono to jiff


zbraniecki:
[Jiff](https://docs.rs/jiff/latest/jiff/) is a new Rust crate for time management, aligned with [Temporal proposal](https://tc39.es/proposal-temporal/), and [better alignment](https://docs.rs/jiff/latest/jiff/_documentation/comparison/index.html) with the direction of the industry.

zslayton:
Thanks for opening an issue! At least for the time being, `chrono` is an implementation detail in `ion-rust`, so replacing it isn't a priority. Would `jiff` enable any new functionality, or are you just considering the library's general maintenance?

zbraniecki:
I'm working on a config/storage system which exposes different file stores (JSON, Postcard, TOML, Yaml etc.).
I'm looking to support Timestamp type and based on my research settled on Jiff.
I started looking into Ion today and would like to consider supporting it, but bringing jiff and chrono into my deps is suboptimal, esp that jiff seems like a better choice for Ion.

Totally understand this is not a prio

Issue #810, opened 2024-08-12
Difficulty accessing `Span` from `Lazy*` types.


jpschorr:
Using
```
ion-rs = { version = "1.0.0-rc.6", features = ["experimental"] }
```

It's somewhat difficult (and sometimes not possible unless I'm missing something) to access the `Span` for a given `Lazy*`

Here's what I'm trying:
```
pub(crate) trait IonSpan<'a> {
    fn get_span(&self) -> Option<Span<'a>>;
}

impl<'a> IonSpan<'a> for LazyValue<'a, AnyEncoding> {
    fn get_span(&self) -> Option<Span<'a>> {
        Some(self.raw()?.span())
    }
}

impl<'a> IonSpan<'a> for LazyStruct<'a, AnyEncoding> {
    fn get_span(&self) -> Option<Span<'a>> {
        self.as_value().get_span()
    }
}

impl<'a> IonSpan<'a> for LazyList<'a, AnyEncoding> {
    fn get_span(&self) -> Option<Span<'a>> {
        todo!("ion-rs doesn't seem to support span or as_value for LazyList")
    }
}

impl<'a> IonSpan<'a> for LazyField<'a, AnyEncoding> {
    fn get_span(&self) -> Option<Span<'a>> {
        todo!("ion-rs doesn't seem to support span or as_value for LazyList")
    }
}

impl<'a> IonSpan<'a> for ValueRef<'a, AnyEncoding> {
    fn get_span(&self) -> Option<Span<'a>> {
        todo!("ion-rs doesn't seem to support span or as_value for LazyList")
    }
} 
```

I'm not sure if the implemented methods are ideal, and for several (see the `todo`s), I could find no way to get to the span.

Issue #809, opened 2024-08-12
It seems `Position` is not publicly nameable


jpschorr:
It seems `Position` is not publicly accessible: https://github.com/amazon-ion/ion-rust/blob/main/src/lib.rs#L181

Using
```
ion-rs = { version = "1.0.0-rc.6", features = ["experimental"] }
```

Given an `IonError`, `err`:

This works:
```
        let position: Option<_> = match err {
            IonError::Incomplete(e) => Some(e.position()),
            IonError::Decoding(e) => e.position(),
            _ => None,
        };
```

But there is no way to explicitly name the type for the position variable, because I can’t actually import the `Position` type
```
        let position: Option<&Position> = ...
```

```
286 |         let position: Option<&Position> = match &self.0 {
    |                               ^^^^^^^^ not found in this scope
```

PR #1024, opened 2026-06-05
Clean up public API for v1.0.0 release

popematt:
Reduces the public API surface in preparation for v1.0.0. The goal is a minimal stable surface that we can grow from.
Items can always be added in minor releases but cannot be removed after 1.0.

This PR contains three commits:

1. `79484b5` — Gate, remove, or reorganize public API items
2. `f9858cb` — Remove arithmetic trait impls from numeric types
3. `8bf799c` — Gate chrono conversions behind `experimental-chrono`

If you're interested, the full output of `cargo public-api` is [available here](https://gist.github.com/popematt/022a994afe65de6cba2f2f92d6dfa804), but what follows is a summary of the changes and the reasoning for each.

----

#### Gate reader/writer-only types behind `experimental-reader-writer` (79484b5)

Several types were unconditionally public despite being useless without the streaming Reader/Writer. Consistent with the approach established in #576:

- `AnyEncoding` — only meaningful when constructing a `Reader`
- `ElementReader`, `ElementWriter` — only have implementations when
  the Reader/Writer is available
- `SymbolId` — no stable API method accepts or returns it
- `Catalog`, `EmptyCatalog`, `MapCatalog` — only consumed by `Reader`
- `SharedSymbolTable` — only useful with catalogs

#### Gate tooling types behind `experimental-tooling-apis` (79484b5)

`HasRange`, `HasSpan`, and `Span` expose encoding-level byte details.
Every implementor is already behind this feature flag, and users who need position info have `SourceLocation` in the stable API (#810).

#### Remove vestigial items (79484b5)

- `Format` enum — discussed in #246 but never wired up; zero usages
- `IonResultIterExt` — no known internal or external users

#### Restrict visibility of implementation details (79484b5)

- `OwnedSequenceIterator` — users receive it via `IntoIterator`, never need to name it
- `WriteConfig<*Encoding_1_1>::new()` — Ion 1.1 types not accessible without their feature flag

#### Fix un-nameable types in public signatures (79484b5)

- `SourceLocation` — re-exported at `ion_rs::SourceLocation` (#809)
- `IonInput` — re-exported at `ion_rs::IonInput`; enables generic functions over Ion input sources. Documented as sealed for now. (Long term intention is #783, so making it public but sealed is fine for now. )

#### Clean up decimal module exports (79484b5)

- `Coefficient` and `Sign` moved to `ion_rs::decimal::Coefficient`
  and `ion_rs::decimal::Sign`
- `coefficient` submodule now `pub(crate)`

#### Remove leaked trait bounds from `ConversionOperationError` (79484b5)

Internal traits (`ValueTypeExpectation`, `TypeExpectation`) removed
from the struct definition; moved to impl blocks only.

#### Prevent implicit feature flags (79484b5)

`digest` and `sha2` changed to `dep:digest`, `dep:sha2` so they can
only be enabled via `experimental-ion-hash`.

#### Remove arithmetic traits from `Int`, `UInt`, `Decimal` (f9858cb)

`Add`, `Sub`, `Neg`, and `num_traits::Zero` were unused outside of
tests and exposed `num_traits` as a public dependency. Issues #970
and #286 ask for arithmetic but have no design resolution —
committing now would lock in a premature decision. Inherent
`is_zero()` methods replace `Zero::is_zero()`.

#### Gate chrono behind `experimental-chrono` (8bf799c)

Chrono is an implementation detail (#838) that coupled ion-rs's
SemVer to chrono's. All `From`/`TryInto` impls between `Timestamp`
and chrono types now require the feature flag. Internal code uses
`pub(crate)` helper methods.

**Potential correctness concern:** The `From<NaiveDateTime>` conversion
is semantically suspect — Ion timestamps with unknown offset still
represent UTC instants, but `NaiveDateTime` carries no UTC guarantee.
This should be investigated before promoting the feature out of
experimental.

----

### DIFF
--- a/src/lazy/expanded/lazy_element.rs
+++ b/src/lazy/expanded/lazy_element.rs
@@ -169,8 +169,7 @@ impl<Encoding: Decoder> TryFrom<&LazyElement<Encoding>> for Element {
 #[cfg(test)]
 mod tests {
     use crate::lazy::expanded::lazy_element::LazyElement;
-    use crate::lazy::reader::IonResultIterExt;
-    use crate::{AnyEncoding, Element, IonResult, IonType, Reader, Sequence};
+    use crate::{AnyEncoding, Element, IonResult, Reader, Sequence};
 
     #[cfg(feature = "experimental-ion-1-1")]
     fn test_data() -> String {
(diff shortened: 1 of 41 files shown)


src/lazy/expanded/lazy_element.rs
```diff
@@ -268,58 +267,4 @@ mod tests {
...
-        let mut reader = Reader::new(AnyEncoding, r#"1 2 3 4 5 6 7 8 9 10"#)?;
-        let evens: IonResult<Vec<i64>> = reader
-            .try_filter(|element| Ok(element.ion_type() == IonType::Int))
-            .try_map(|element| element.read()?.clone().expect_i64())
-            .try_filter(|i| Ok(i % 2 == 0))
-            .collect();
-        drop(reader);
-        assert_eq!(vec![2, 4, 6, 8, 10], evens?);
-        Ok(())
-    }
```
almann, line 283: Are these tests redundant and that is why they are removed?

src/lazy/reader.rs
```diff
@@ -181,79 +181,6 @@ impl<Encoding: Decoder, Input: IonInput> Iterator for LazyElementIterator<'_, En
     }
 }
 
-/// Extension methods for iterators that return an `IonResult`.
-///
-/// These methods are analogous to methods that already exist on `Iterator`,
-/// but automatically handle situations where the input `IonResult` is an `Err`
-/// sparing the user from writing more boilerplate.
-// This code is not used within the library but is intentionally available to tests and applications.
-pub trait IonResultIterExt<Item>: Iterator<Item = IonResult<Item>> {
```
almann, line 190: Are we at all concerned about breaking any downstream with this?

src/types/timestamp.rs
almann: Import seems unused (according to CI)

src/types/timestamp.rs
```diff
@@ -1211,61 +1257,35 @@ fn downconvert_to_naive_datetime_with_nanoseconds(timestamp: &Timestamp) -> Naiv
     }
 }
 
-// Allows a Timestamp with an unknown offset to be converted to a NaiveDateTime.
+#[cfg(feature = "experimental-chrono")]
 impl TryInto<NaiveDateTime> for Timestamp {
```
almann, line 1260: I see the import is conditional--we probably need to annotate that import if we want to make it not whine.

src/types/timestamp.rs
almann: Do these imports need conditional compilation?

almann: Same for the above.

almann (approved):
Overall looks good--though I see some unused imports that might need conditional compilation to make them not complain.

src/lazy/expanded/lazy_element.rs
popematt: They were demonstrating the `IonResultIterExt` trait, which is being removed.

src/lazy/reader.rs
popematt: The trait has a pretty distinctive name, so I checked GitHub and Amazon internal code, and no one names or `use`s this trait.

If someone is using it in private code, they can either let us know (and we can re-add it) or recreate it themselves. If there is any breakage, it will be at compile time and easy enough to fix (unlike other languages where this sort of change might cause a runtime failure).

PR #1028, opened 2026-06-11
fix: remove `Eq` from `Element` and fix `Struct` equality

### LINKED ISSUE #1025: `Element`'s `Eq` impl violates reflexivity requirement
`Element`'s `Eq` impl violates the reflexivity requirement. Consider removing it in favor of having `Element` only implement `PartialEq`, or make `NaN == NaN` for `Element`.

### LINKED ISSUE #1026: Inconsistent handling of float values in `Value`'s impl of `PartialEq`
`Struct`'s impl of `PartialEq` internally calls `ion_eq` on field values, which means `NaN == NaN` and `0.0 != -0.0` inside structs. But `Value`'s impl of `PartialEq` for `Float` uses standard `f64::eq`, where `NaN != NaN` and `0.0 == -0.0`. So the same float value compares differently depending on whether it's bare or inside a struct:

```rust
let nan = Element::float(f64::NAN);
let s1 = ion_struct! { "x": f64::NAN };
let s2 = ion_struct! { "x": f64::NAN };

nan == nan   // false (f64 semantics)
s1 == s2     // true  (Ion semantics inside struct)
```

popematt:
**Issue #, if available:**

fixes #1025
fixes #1026 

**Description of changes:**

* `Element`'s `Eq` impl violated reflexivity because `Value::Float` uses standard `f64` semantics where `NaN != NaN`. Remove `Eq` from `Element`, `Sequence`, `List`, `SExp`, and `Struct`.

* Fix `Struct`'s `PartialEq` to use `==` instead of `ion_eq` for field comparison, and fix a multiset matching bug where duplicate field values could all match the same target value. `IonEq for Struct` now independently uses `ion_eq` for field comparison.

### DIFF
--- a/src/element/mod.rs
+++ b/src/element/mod.rs
@@ -1029,8 +1029,6 @@ impl PartialEq for Element {
     }
 }
 
-impl Eq for Element {}
-
 // This implementation allows APIs that require an Into<Element> to accept references to an existing
 // Element.
 impl<'a> From<&'a Element> for Element {
--- a/src/element/sequence.rs
+++ b/src/element/sequence.rs
@@ -15,7 +15,7 @@ use std::io;
 ///
 /// A `Sequence` is not itself an Ion value type, but can represent a series of Ion values appearing
 /// in a [`List`](crate::List), a [`SExp`](crate::SExp), or at the top level.
-#[derive(Clone, PartialEq, Eq)]
+#[derive(Clone, PartialEq)]
 pub struct Sequence {
     elements: Vec<Element>,
 }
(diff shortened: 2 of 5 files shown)


deirdresama (approved):
LGTM. Correct fix for two real bugs, with a clean separation of PartialEq and IonEq semantics.

Minor (non-blocking): the greedy find(...) match relies on eq being a true equivalence relation; holds for both == and ion_eq, but a short comment would guard against a future non-transitive comparator.

NovemberZulu (approved):

PR #1037, opened 2026-08-13
Substitute padded tables for unresolvable imports with max_id

### LINKED ISSUE #1036: Reader errors on unresolvable shared symbol table import even when max_id is defined
## Summary

When a local symbol table's `imports` field references a shared symbol table that is not present in the catalog, the reader unconditionally returns `IonError::Decoding` even when the import declares `max_id`.

Per the Ion 1.0 spec's import-resolution algorithm, an error is prescribed only when `max_id` is undefined. When `max_id` is defined, the spec instead directs the implementation to fall back to the greatest available version of the table (allocating `max_id` SIDs "regardless of the actual size of the referenced table") or, failing that, to substitute a dummy table of `max_id` unknown-text symbols and continue processing. This padding behavior exists precisely so that documents can be read (with unknown symbol text) without prior knowledge of every shared symbol table they import.

This behavior doesn't match Ion Java.

## Versions

- `ion-rs`: 1.0.1 (crates.io, default features)
- `rustc`: 1.96.0
- Affects both text and binary Ion 1.0 input (identical error)
- Also reproduces on `main` @ `fb36103` (repro compiles unchanged; identical error)

## Minimal reproduction

Fails via the stable `Element` API with default features — no experimental flags needed:

```rust
use ion_rs::{Element, IonResult};

const DOC: &str = r#"
$ion_symbol_table::{
  imports: [{ name: "com.example.not_in_catalog", version: 1, max_id: 3 }],
  symbols: ["local_a"]
}
{ $13: "hello" }
"#;

fn main() -> IonResult<()> {
    let elements = Element::read_all(DOC.as_bytes())?; // <- errors here
    println!("{elements:?}");
    Ok(())
}
```

### Actual behavior

Error message:

```text
symbol table import failed, could not find table with name='"com.example.not_in_catalog"' and version=1
```

Full `Debug` output:

```text
Decoding(DecodingError { description: "symbol table import failed, could not find table with name='\"com.example.not_in_catalog\"' and version=1", position: None })
```

### Expected behavior

The document parses successfully: the import contributes SIDs 10–12 with unknown
text (dummy table padding), the local symbol `"local_a"` becomes SID 13, and the value reads
as `{local_a: "hello"}`.

<details>
<summary>Equivalent binary Ion 1.0 repro (same error)</summary>

```rust
/// Hand-assembled binary equivalent of the text document above.
fn binary_with_import() -> Vec<u8> {
    let mut b = vec![0xE0, 0x01, 0x00, 0xEA]; // IVM 1.0
    b.extend([0xEE, 0xB6]); // annotation wrapper, len 54
    b.extend([0x81, 0x83]); // annot_length=1, annotation SID 3 ($ion_symbol_table)
    b.extend([0xDE, 0xB2]); // struct, len 50
    b.extend([0x86]); //   field $6 (imports)
    b.extend([0xBE, 0xA5]); //   list, len 37
    b.extend([0xDE, 0xA3]); //     struct, len 35
    b.extend([0x84, 0x8E, 0x9A]); //       $4 (name): string, len 26
    b.extend(b"com.example.not_in_catalog");
    b.extend([0x85, 0x21, 0x01]); //       $5 (version): int 1
    b.extend([0x88, 0x21, 0x03]); //       $8 (max_id): int 3
    b.extend([0x87]); //   field $7 (symbols)
    b.extend([0xB8, 0x87]); //   list len 8, string len 7
    b.extend(b"local_a");
    // { $13: "hello" }
    b.extend([0xD7, 0x8D, 0x85]); // struct len 7, field $13, string len 5
    b.extend(b"hello");
    b
}
```

`Element::read_all(&binary_with_import())` produces the identical `Decoding` error.

</details>

## Spec citation

From the [Ion 1.0 symbols spec](https://amazon-ion.github.io/ion-docs/docs/symbols.html),
*Local Symbol Tables → Imports*:

> - Select a shared symbol table instance as follows:
>   - Query the catalog to retrieve the specified table by `name` and `version`.
>   - If an exact match is not found:
>     - If `max_id` is undefined, implementations MUST raise an error and halt processing.
>     - Otherwise query the catalog to retrieve the table with the given `name` and the greatest version available.
>   - If no table has been selected, substitute a dummy table containing `max_id` undefined symbols.

Note: the spec's *Data Model* section does permit an implementation to error when encountering a symbol with unknown text and a valid SID (other than `$0`) but that is not what happens here. The reader errors during import processing, before any value is read, and the only symbol the document actually uses (`$13`) has known local text.

## Behavior of other Ion implementations

ion-java 1.11.10 (default `IonSystem`/`IonLoader`, no catalog entries) reads the equivalent document successfully, substituting a dummy table for the unresolvable import:

```text
{local_a: "hello"}   // field name text = "local_a", SID 13
```

A variant that uses an in-gap SID (`{ $10: "world" }`) also parses: the symbol is preserved with unknown text and round-trips as `$10`. ion-rs diverges from ion-java here.

## Root cause pointer

`SystemReader::process_imports` (`src/lazy/system_reader.rs`) returns the decoding error unconditionally on a catalog miss, before `max_id` is ever read:

```rust
let shared_table = match catalog.get_table_with_version(name.as_ref(), version) {
    Some(table) => table,
    None => return IonResult::decoding_error(
        format!("symbol table import failed, could not find table with name='{name}' and version={version}")),
};
let max_id = match import.get("max_id")? { ... };   // never reached on miss
```

The unknown-text padding logic (resizing with `Symbol::unknown_text()` when `max_id > symbols.len()`) exists a few lines below, but is only reachable when the catalog lookup succeeds.

Looking at the history, this appears to have been intentional: PR #578 ("adds changes for shared symbol table support in reader", Jul 2023) includes a commit "adds changes to return error for non existing table", and a follow-up commit added the affected ion-tests conformance vectors to the skip list ("added back the related tests for skip-list"). It leaves conforming Ion 1.0 documents unreadable, and it is presumably why vectors such as `good/localSymbolTableImportZeroMaxId.ion` and `good/testfile35.ion` (which imports with `max_id: 2` and uses the imported symbols) remain in `ELEMENT_GLOBAL_SKIP_LIST` ("Requires importing shared symbol tables").

Relatedly, the spec's fallback to "the greatest version available" on an exact-version miss also appears unimplemented: only `get_table_with_version` is consulted, with no best-available fallback.

**Proposed behavior:** on a catalog miss, read `max_id` first; if it is defined, fall back to the greatest available version of the named table or, failing that, substitute a dummy table of `max_id` unknown-text symbols; error only when `max_id` is undefined. This would allow revisiting the skipped vectors above (round-trip skip entries may need separate writer-side
work) and would require updating the in-repo test `non_existent_shared_symbol_table_imports`. The missing greatest-version fallback could also be split into a separate issue if preferred. 

This concerns Ion 1.0 stream processing, which is unchanged by Ion 1.1's module system.

## Impact

A consumer reading Ion documents that import shared symbol tables it does not have cannot
read those documents at all — the exact scenario the spec's dummy-table substitution exists
to handle (the spec's *Data Model* section notes unknown-text symbols "can happen as a result
of not having access to a shared symbol table being imported"). Because the failure occurs on
the stable, default-features `Element` API and the catalog workaround requires an
experimental feature flag, affected users of the stable API currently have no recourse.

hsbakshi:
Fixes #1036.

## What

Per the [Ion 1.0 symbols spec](https://amazon-ion.github.io/ion-docs/docs/symbols.html) (*Local Symbol Tables → Imports*), when a shared symbol table import cannot be resolved to an exact `(name, version)` match in the catalog:

- If `max_id` is undefined (missing, null, not an int, or negative), this remains a decoding error — unchanged.
- Otherwise, the reader falls back to the greatest available version of the named table (`Catalog::get_table`), and if the catalog has no table with that name at all, substitutes a dummy table. In both cases the imported subsequence is sized to exactly `max_id`, truncating or padding with placeholder symbols.

This revision implements the notes from @popematt's comment on #1036:

1. **Limit on placeholder materialization**: `MAX_IMPORT_PLACEHOLDER_SYMBOLS` (1,000,000), accounted cumulatively across all imports in a directive and checked before any allocation (`checked_add` + `try_reserve` as backstops). Worst case ≈ tens of MB. The two ion-tests vectors with pathological `max_id` values (~268M and ~2.1B) now fail with a clear cap error instead of exhausting memory; a non-materialized (lazy) substitute-table representation would remove the cap entirely and un-skip them — happy to file that as a follow-up issue (it dovetails with the `Vec<Arc<SymbolTable>>` restructure you sketched, which I assessed but deferred: the run-length variant is blocked by `symbol_for(&self) -> Option<&Symbol>` returning a reference).
2. **`SymbolText::UnknownImport` + refuse to write**: placeholder symbols carry a distinct variant (internal enum — no public API shape change; the baseline gains three hand-written comparison impls semantically identical to the previous derives). Equality/ordering/hashing remain text-based, so `Eq`/`IonDataEq` contracts are preserved (placeholder ≡ `$0`, consistent with the existing unknown-text handling; spec-strict substitute-symbol equivalence by import location is intentionally not implemented yet and is noted in the docs). Writing refuses at the typed transcription layer — value, annotation, and field-name positions, eager and lazy — with an error that tells the user the catalog fix.
3. Import-location preservation for round-trip is left for a later issue/PR, per your comment; the placeholder flag carries a TODO pointing at the table-side representation so that work knows where to absorb it.

## Behavior changes (explicit disclosure)

- **error→ok**: documents with unresolvable imports that declare `max_id` now parse (previously an unconditional decoding error). This matches ion-java, which reads the equivalent documents successfully.
- **ok→error (one edge)**: a *resolved* import whose `max_id` overstates the table's size previously padded the gap with plain unknown-text symbols, which wrote out as `$0`. Those gap SIDs are substitute symbols too (a later version of the table could define their text), so they now use the placeholder variant and refuse to write. Pinned by a dedicated test. If you'd rather keep `$0`-writability for the resolved-import case, that's a two-line change — flagging it for an explicit decision.
- **Write-refusal boundary**: on the default-features API, typed encoding (`Element::encode_as`/`encode_to`) refuses placeholders everywhere. `Display` renders them as `$0` (it produces Ion text and has no error channel); `Debug` distinguishes them (`$0 (unresolved import)`). The raw-level conversions (`From<SymbolRef> for RawSymbolRef` etc.) are infallible and lower placeholders to `$0`; they are documented as an intentional lossy escape hatch, are pinned by test, and are only consumed by writer APIs behind `experimental-reader-writer`.

## Conformance vectors

Un-skipped and passing: `good/item1.10n`, `good/localSymbolTableImportZeroMaxId.ion`, `good/testfile35.ion`, `good/subfieldVarInt.ion` (global list), `good/subfieldVarUInt15bit.ion`, `good/subfieldVarUInt16bit.ion` — including round-trip permutations where applicable. Still skipped, with updated comments: `subfieldVarUInt.ion` / `subfieldVarUInt32bit.ion` (placeholder cap; see follow-up above), `item1.10n` round-trip (writes a gap-SID placeholder, which now refuses), `subfieldVarInt.ion` round-trip (pre-existing unrelated decimal-buffer writer error).

## Tests

Spec-branch coverage (exact match / greatest-version fallback / dummy substitution / undefined-`max_id` error / invalid-`max_id`-as-undefined / truncate & pad / `max_id: 0`), LST-append after substitution, IVM reset, gap SIDs as field names and annotations, refusal in nested containers and lazy transcription, the resolved-import overstated-`max_id` edge, an Ion 1.1 stream carrying a 1.0-style LST (same path), overflow/`try_reserve` regression tests, and size-regression pins for `SymbolRef`/`ValueRef` (`ValueRef` is unchanged at 96 bytes; `SymbolRef` grows 16→24, absorbed by larger variants).

`cargo test --workspace --all-features`, clippy with `-D warnings`, fmt, docs, and the public-API baseline checks are all green.

## Context

I'm not an Ion expert — I validated the intended behavior against the spec text and empirically against ion-java 1.11.10. Happy to rework any of the design points above, and to file the follow-up issues (lazy substitute tables / configurable cap / import-location round-trip) if this direction looks right.

### DIFF
--- a/src/lazy/system_reader.rs
+++ b/src/lazy/system_reader.rs
@@ -23,6 +23,23 @@ use crate::{
 use std::ops::Deref;
 use std::sync::Arc;
 
+/// The maximum number of placeholder (unknown-text) symbols that the reader is willing to
+/// materialize while processing a single symbol table directive's `imports` list.
+///
+/// When a shared symbol table import cannot be resolved, the Ion 1.0 spec requires the reader
+/// to reserve exactly `max_id` symbol IDs for it, padding with unknown-text placeholders as
+/// needed. Because `max_id` is attacker-controlled and each placeholder occupies memory, a
+/// tiny stream could otherwise instruct the reader to allocate an enormous symbol table (for
+/// example, a `max_id` of 2^31 would materialize billions of placeholder symbols). Capping
+/// the number of placeholders bounds that amplification: streams whose unresolvable imports
+/// require more placeholders than this limit produce a decoding error instead.
+///
+/// One million placeholder symbols occupy tens of megabytes--far more than any known
+/// legitimate use of substituted imports requires, but small enough to keep a malicious
+/// stream from exhausting memory. Symbols that are actually present in a matching catalog
+/// entry do not count against this limit.
+pub(crate) const MAX_IMPORT_PLACEHOLDER_SYMBOLS: usize = 1_000_000;
+
 /// A binary reader that only reads each value that it visits upon request (that is: lazily).
 ///
 /// Unlike [`crate::lazy::reader::Reader`], which only exposes values that are part
@@ -570,6 +587,9 @@ impl<Encoding: Decoder, Input: IonInput> SystemReader<Encoding, Input> {
         catalog: &dyn Catalog,
         imports: LazyValue<'_, Encoding>,
     ) -> IonResult<()> {
+        // The number of placeholder symbols materialized for unresolvable imports so far
+        // while processing this symbol table directive's `imports` list.
+        let mut total_placeholders: usize = 0;
         match imports.read()? {
             // Any symbol other than `$ion_symbol_table` is ignored.
             ValueRef::Symbol(symbol_ref) if symbol_ref == "$ion_symbol_table" => {
@@ -597,37 +617,95 @@ impl<Encoding: Decoder, Input: IonInput> SystemReader<Encoding, Input> {
                         _ => Ok(1),
                     }?;
 
-                    let shared_table = match catalog.get_table_with_version(name.as_ref(), version) {
-                        Some(table) => table,
-                        None => return IonResult::decoding_error(
-                            format!("symbol table import failed, could not find table with name='{name}' and version={version}")
-                        ),
-                    };
-
-                    let max_id = match import.get("max_id")? {
+                    // If the max_id is unspecified, null, negative, or an invalid data type,
+                    // we treat it as undefined.
+                    // See: https://amazon-ion.github.io/ion-docs/docs/symbols.html
+                    let max_id: Option<usize> = match import.get("max_id")? {
                         Some(ValueRef::Int(i)) if i >= Int::ZERO => {
-                            usize::try_from(i).map_err(|_| {
+                            Some(usize::try_from(i).map_err(|_| {
                                 IonError::decoding_error(
                                     "found a `max_id` beyond the range of usize",
                                 )
-                            })?
+                            })?)
                         }
-                        // If the max_id is unspecified, negative, or an invalid data type, we'll import all of the symbols from the requested table.
-                        _ => shared_table.symbols().len(),
+                        _ => None,
                     };
 
-                    let num_symbols_to_import = shared_table.symbols().len().min(max_id);
+                    // Look for an exact match in the catalog. If the requested version is not
+                    // available and `max_id` is defined, fall back to the highest available
+                    // version of the named table; if the catalog has no table with that name,
+                    // substitute a dummy table by padding with `max_id` unknown-text symbols.
+                    // If there is no exact match and `max_id` is undefined, the reader cannot
+                    // know how many symbol IDs the import occupies; this is an error.
+                    let shared_table = match catalog.get_table_with_version(name.as_ref(), version)
+                    {
+                        Some(table) => Some(table),
+                        None if max_id.is_none() => return IonResult::decoding_error(
+                            format!("symbol table import failed, could not find table with name='{name}' and version={version} and no valid max_id was specified")
+                        ),
+                        None => catalog.get_table(name.as_ref()),
+                    };
+
+                    let table_symbols: &[Symbol] =
+                        shared_table.map(|table| table.symbols()).unwrap_or(&[]);
+
+                    // If `max_id` is undefined, import all of the found table's symbols.
+                    // (An exact match is guaranteed to have been found in that case.)
+                    let max_id = max_id.unwrap_or(table_symbols.len());
+
+                    // Import up to `max_id` symbols from the table...
+                    let num_symbols_to_import = table_symbols.len().min(max_id);
 
                     pending_lst
                         .imported_symbols
-                        .extend_from_slice(&shared_table.symbols()[..num_symbols_to_import]);
+                        .extend_from_slice(&table_symbols[..num_symbols_to_import]);
 
-                    if max_id > shared_table.symbols().len() {
+                    // ...and if the table did not have enough symbols (or was not found at all),
+                    // pad the remaining symbol IDs with unknown text.
+                    if max_id > num_symbols_to_import {
                         let num_pending_symbols = pending_lst.imported_symbols().len();
-                        let num_placeholders = max_id - shared_table.symbols().len();
+                        let num_placeholders = max_id - num_symbols_to_import;
+                        // Placeholder symbols are materialized eagerly, so `max_id` values
+                        // (which are attacker-controlled) must be capped to keep a small
+                        // stream from demanding an enormous allocation. See the doc comment
+                        // on `MAX_IMPORT_PLACEHOLDER_SYMBOLS` for details.
+                        total_placeholders = total_placeholders
+                            .checked_add(num_placeholders)
+                            .filter(|&total| total <= MAX_IMPORT_PLACEHOLDER_SYMBOLS)
+                            .ok_or_else(|| {
+                                IonError::decoding_error(format!(
+                                    "symbol table import (name='{name}', max_id={max_id}) requires \
+                                     padding with more placeholder symbols than this implementation \
+                                     is willing to materialize (limit: {MAX_IMPORT_PLACEHOLDER_SYMBOLS})"
+                                ))
+                            })?;
+                        // In practice this addition cannot overflow: `num_placeholders` was
+                        // just capped at `MAX_IMPORT_PLACEHOLDER_SYMBOLS` (above), and
+                        // `num_pending_symbols` is the length of an existing `Vec<Symbol>`,
+                        // which is far smaller than `usize::MAX - MAX_IMPORT_PLACEHOLDER_SYMBOLS`.
+                        // The checked arithmetic is retained purely as defense in depth.
+                        let num_symbols_after_import = num_pending_symbols
+                            .checked_add(num_placeholders)
+                            .ok_or_else(|| {
+                                IonError::decoding_error(
+                                    "internal error: combined symbol table size overflowed usize",
+                                )
+                            })?;
+                        // The cap above bounds the allocation to a modest size; reserve the
+                        // space fallibly anyway so that allocation failure (e.g. under severe
+                        // memory pressure) surfaces as a decoding error instead of aborting
+                        // the process.
+                        pending_lst
+                            .imported_symbols
+                            .try_reserve(num_placeholders)
+                            .map_err(|_| {
+                                IonError::decoding_error(format!(
+                                    "could not allocate space for a symbol table import with max_id {max_id}"
+                                ))
+                            })?;
                         pending_lst.imported_symbols.resize(
-                            num_pending_symbols + num_placeholders,
-                            Symbol::unknown_text(),
+                            num_symbols_after_import,
+                            Symbol::unknown_import_placeholder(),
                         );
                     }
                 }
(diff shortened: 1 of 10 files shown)


src/lazy/system_reader.rs
```diff
@@ -597,37 +617,95 @@ impl<Encoding: Decoder, Input: IonInput> SystemReader<Encoding, Input> {
...
+                            .map_err(|_| {
+                                IonError::decoding_error(format!(
+                                    "could not allocate space for a symbol table import with max_id {max_id}"
+                                ))
+                            })?;
                         pending_lst.imported_symbols.resize(
-                            num_pending_symbols + num_placeholders,
-                            Symbol::unknown_text(),
+                            num_symbols_after_import,
+                            Symbol::unknown_import_placeholder(),
```
popematt, line 708: These placeholder symbols get interned into the text index, and I think that corrupts symbol resolution. They're drained into the active table in `expanded/mod.rs` (`imported_symbols.drain(..)` into `symbol_table.add_symbol`), and `add_symbol` (`symbol_table.rs:171`) inserts every symbol into `ids_by_text`. `Borrow<str> for Symbol` maps any unknown-text symbol to `""` (`types/symbol.rs:244`), so this writes a `"" -> <placeholder SID>` entry.

That means `sid_for("")` now resolves to a placeholder SID, and the consequence is that a real empty-text symbol `''` will get remapped to a placeholder SID (because imported symbols come before local symbols and when writing Ion data, implementations MUST select the lowest known ID).
The placeholder SID has no defined text (or different text if another reader has access to the correct shared symbol table), and the result is silent data corruption, not just placeholder-vs-placeholder collapse.

src/lazy/system_reader.rs
```diff
@@ -23,6 +23,23 @@ use crate::{
...
+/// tiny stream could otherwise instruct the reader to allocate an enormous symbol table (for
+/// example, a `max_id` of 2^31 would materialize billions of placeholder symbols). Capping
+/// the number of placeholders bounds that amplification: streams whose unresolvable imports
+/// require more placeholders than this limit produce a decoding error instead.
+///
+/// One million placeholder symbols occupy tens of megabytes--far more than any known
+/// legitimate use of substituted imports requires, but small enough to keep a malicious
+/// stream from exhausting memory. Symbols that are actually present in a matching catalog
+/// entry do not count against this limit.
+pub(crate) const MAX_IMPORT_PLACEHOLDER_SYMBOLS: usize = 1_000_000;
```
popematt, line 41: The running total is cumulative across the imports in a single directive, so the naive case does get caught, but it's still possible to create a DoS vector using repetition.

A minimal malicious directive can be constructed in 24 bytes of Ion binary.
```
$ion_symbol_table::{ imports:[{ name:"a", version:1, max_id:999_999 }] }
```

For this import, there will be ~1M placeholder writes when `process_imports` resizes `imported_symbols` (~32 MB) and ~1M `Symbol` clones when `apply_pending_context_changes` drains that into `symbols_by_id` (~32 MB). The result is that for every 24 bytes of input, there are many megabytes of memory writes. A 10 kB payload could contain >400 such directives, which would result in many gigabytes of writes.

Resident memory stays near 32 MB the whole time, because `clear`, `truncate`, and `drain` retain capacity and the buffers are reused. The attack doesn't show up as memory growth; it shows up as a reader thread that never returns. Lowering the cap helps proportionally (a cap of 1,000 turns each directive into ~32 KB of work), but sustained or parallel requests still reach gigabyte-scale traffic, so it reduces the severity rather than removing the vector.

src/types/symbol.rs
```diff
@@ -29,7 +40,7 @@ impl SymbolText {
             SymbolText::Shared(s) => s.as_ref(),
             SymbolText::Owned(s) => s.as_str(),
             SymbolText::Static(s) => s,
-            SymbolText::Unknown => return None,
+            SymbolText::Unknown | SymbolText::UnknownImport => return None,
```
popematt, line 43: This causes an unknown imported symbol to compare equal to `$0` (and to imported unknown symbols from other import locations) through `impl PartialEq for Symbol`, and therefore through `IonEq` and `IonDataOrd` as well. Per the Ion data model, imported unknown-text symbols are not equivalent to `$0`, and are equivalent to each other only when they share an import location, so this violates the equivalence semantics the spec defines and that `IonData` is meant to honor. This is one of the `Eq`/`IonDataEq` risks I raised in #1036.

src/raw_symbol_ref.rs
```diff
@@ -115,6 +115,18 @@ impl AsRawSymbolRef for &str {
     }
 }
 
+// This conversion is infallible and therefore intentionally lossy: a `Symbol` that is a
```
popematt, line 118: Why not update `RawSymbolRef` so that it's not lossy?

src/symbol_ref.rs
```diff
@@ -7,14 +7,55 @@ use std::hash::{Hash, Hasher};
...
+    // everywhere except during transcription (the `WriteAsIon` implementations used when
+    // writing an `Element`, `Value`, `LazyValue`, `SymbolRef`, or `Symbol`), which refuses
+    // to encode it. Conversions to `RawSymbolRef` (used by the value writer APIs for symbol
+    // values, struct field names, and annotations) are infallible and intentionally lossy:
+    // they discard this flag and map the placeholder to `$0` (symbol ID 0).
+    //
+    // TODO: absorb into the symbol table representation (tracked SID ranges of unresolved
+    //       imports) when symbol tables move to a slab/Vec<Arc<SymbolTable>> design; that
+    //       also enables import-location preservation for round-trip.
+    is_unknown_import_placeholder: bool,
```
popematt, line 25: `SymbolRef` holds an `Option<&'a str>` today, and I think the right long-term solution is probably replacing the `Option` with something like this:
```rust
enum SymbolRefKind {
    Text(&'a str),
    Unknown,
    UnknownImported( /* table and offset */),
}
```

popematt (changes requested):
When I commented on #1036, I thought that this approach might be possible, but this ended up hitting all of the risks that I mentioned, some of them in subtle ways that I didn't even think of at the time. That's at least partly on me for underestimating how complex this would be.

I think we need to break it down into smaller steps. Here's what I think needs to be done, in an order that can build on each other without introducing an intermediate state where the library would be un-releaseable.

  1. Fix the `$0` -> `''` interning
  2. Add the internal import-location representation, no behavior change.
  3. Spec equivalence, unit-tested against the equivalence matrix
  4. Sparse symbol-table storage
  5. Writer refuses located symbols (interim)
  6. Reader produces located symbols
  8. Writer round-trips by reproducing the import

I'm willing to collaborate on a more detailed implementation plan, if you like.

Issue #476, opened 2023-03-08
Shrink stack sizes for common types


zslayton:
Some types have large stack sizes that are dominated by fields that are rarely used.

### `UInteger`
Use of the `BigUInt` variant is rare; it could be boxed to cut the enum size by 16 bytes.
```
print-type-size type: `types::integer::UInteger`: 24 bytes, alignment: 8 bytes
print-type-size     variant `BigUInt`: 24 bytes
print-type-size         field `.0`: 24 bytes
print-type-size     variant `U64`: 8 bytes
print-type-size         field `.0`: 8 bytes
```

### `Decimal`
`Decimal` is dominated by the size of `UInteger` above:
```
print-type-size type: `types::decimal::Decimal`: 40 bytes, alignment: 8 bytes
print-type-size     field `.coefficient`: 32 bytes
print-type-size     field `.exponent`: 8 bytes
```
so shrinking `UInteger` would shrink `Decimal`.

### `Timestamp`
`Timestamp` is dominated by the `Decimal` representation of `fractional_seconds`:
```
int-type-size type: `types::timestamp::Timestamp`: 64 bytes, alignment: 8 bytes
print-type-size     field `.fractional_seconds`: 40 bytes
print-type-size     field `.offset`: 8 bytes
print-type-size     field `.date_time`: 12 bytes
print-type-size     field `.precision`: 1 bytes
print-type-size     end padding: 3 bytes
```
even though `fractional_seconds` will only be used when the timestamp precision is >nanoseconds.

### `Value`
`Value` is dominated by `Timestamp` and `Decimal`:
```
print-type-size type: `value::owned::Value`: 64 bytes, alignment: 8 bytes
print-type-size     variant `Timestamp`: 64 bytes
print-type-size         field `.0`: 64 bytes
print-type-size     variant `Decimal`: 40 bytes
print-type-size         field `.0`: 40 bytes
print-type-size     variant `Integer`: 32 bytes
print-type-size         field `.0`: 32 bytes
print-type-size     variant `Symbol`: 32 bytes
print-type-size         field `.0`: 32 bytes
print-type-size     variant `String`: 24 bytes
print-type-size         field `.0`: 24 bytes
print-type-size     variant `Blob`: 24 bytes
print-type-size         field `.0`: 24 bytes
print-type-size     variant `Clob`: 24 bytes
print-type-size         field `.0`: 24 bytes
print-type-size     variant `Float`: 8 bytes
[...]

Issue #1035, opened 2026-08-11
`experimental-chrono` conversions should be fallible


popematt:
The `From<DateTime<FixedOffset>> for Timestamp` impl (behind `experimental-chrono`) accepts chrono values outside Ion's valid range without returning an error. Chrono permits year 0 and years above 9999; Ion does not. The conversion silently produces a `Timestamp` with masked/clamped field values rather than signaling the incompatibility. This should be a `TryFrom` that returns `IonError` for out-of-range inputs.

```rust
#[cfg(feature = "experimental-chrono")]
#[test]
fn from_chrono_rejects_out_of_range_year() {
    use chrono::{FixedOffset, NaiveDate, TimeZone};
    use ion_rs::Timestamp;
    use std::convert::TryFrom;
    let offset = FixedOffset::east_opt(0).unwrap();
    let dt = offset
        .from_local_datetime(
            &NaiveDate::from_ymd_opt(20000, 1, 1).unwrap().and_hms_opt(0, 0, 0).unwrap()
        )
        .unwrap();
    // Currently silently produces a corrupt Timestamp; should return Err
    let result = Timestamp::try_from(dt);
    assert!(result.is_err());
}
```

popematt:
The same thing also applies to `NaiveDateTime`. Chrono's `NaiveDateTime` supports year 0 and negative years. The current `From` impl panics (via `.expect()`) when the year is outside Ion's 1-9999 range. It should be a fallible `TryFrom` that returns an error instead of panicking.

```rust
#[cfg(feature = "experimental-chrono")]
#[test]
fn from_naive_datetime_rejects_year_zero() {
    use chrono::NaiveDate;
    use ion_rs::Timestamp;
    use std::convert::TryFrom;
    let dt = NaiveDate::from_ymd_opt(0, 1, 1).unwrap().and_hms_opt(0, 0, 0).unwrap();
    // Currently panics; should return Err
    let result = Timestamp::try_from(dt);
    assert!(result.is_err());
}
```

popematt:
The solution for this and for #1033 should probably be to replace the `experimental-chrono` feature with a `jiff-compat` feature that has fallible conversions. (See also #838.)

Issue #1033, opened 2026-08-11
Binary encoder can write timestamp with year 0 (invalid Ion)


popematt:
When a timestamp with a small year and positive offset is converted to UTC via `to_utc()`, the result can have year 0 (e.g., `0001-01-01T00:30+01:00` → `0000-12-31T23:30Z`). The binary encoder uses `to_utc()` to obtain UTC fields for serialization and writes this year 0 to the output stream. Year 0 is invalid per the Ion spec, and the crate's own reader rejects it on decode, producing a write→read round-trip failure.

```rust
#[test]
fn binary_encoder_writes_year_zero() {
    use ion_rs::*;
    let ts = Timestamp::with_year(1)
        .with_month(1)
        .with_day(1)
        .with_hour_and_minute(0, 30)
        .with_offset(60) // +01:00
        .build()
        .unwrap();
    let elem: Element = ts.into();
    let binary = elem.encode_as(v1_0::Binary).unwrap();
    // This should succeed but currently fails:
    let roundtrip = Element::read_one(&binary);
    assert!(roundtrip.is_ok(), "round-trip failed: {:?}", roundtrip.err());
}
```

popematt:
https://github.com/amazon-ion/ion-rust/pull/1041#discussion_r3960538969 also points out that an illegal year can also be written via inappropriate wrapping of the year.

PR #1041, opened 2026-08-18
perf(timestamp): replace internals with 16-byte packed representation

popematt:
### Issue #, if available:

Related to #476 

### Description of changes:

I am working on implementing my own Ion reader implementation, and discovered that reading a `Timestamp` was a significant bottleneck in my mixed-data workloads. Although the read/write performance of `ion-rust` doesn't change much with this PR, this refactoring of `Timestamp`, when combined with my own reader implementation resulted in a ~15x improvement in read speed relative to the `v1.0.1` release.

`Timestamp` previously used chrono's `NaiveDateTime`, an `Option<FixedOffset>`, a `TimestampPrecision` enum, and an
`Option<Mantissa>` for fractional seconds. This totaled 80 bytes per instance and required heap allocation for the `Decimal`-backed mantissa. This PR replaces all of that with two `u64` fields:

- `packed_fields` -- date-time components in the high bits (year, month, day, hour, minute, second) for natural chronological ordering via integer comparison; metadata (precision, subsecond digits, offset) in the low bits
- `attoseconds` -- fractional seconds normalized to 10^-18 scale

This PR also eliminates the unnecessary `Decimal` allocation in the `TimestampBuilder::build()` method.

#### Design Choices

- **`Ord` without division.** Cross-offset comparison normalizes via linear-minute arithmetic and in-place bit patching. Division is only needed when the offset requires the day-rollover path (which falls back to `add_offset_to_utc`).
- **`PartialEq` short-circuits.** Seconds and attoseconds are unaffected by offset normalization, so mismatches there reject early without entering the comparison machinery.
- **`IonEq` / `IonDataOrd` / `IonDataHash` are single-instruction operations** -- direct `u64` comparison or hash of the packed fields.
- **`Copy` deliberately not derived** -- leaves room for a future heap fallback if >18-digit fractional precision is ever needed, without a semver break.
- **chrono dependency moved behind `experimental-chrono` feature gate.** The default build has no chrono dependency any more.

#### Performance (Apple M-series / AWS Graviton, 1000-op batches)

| Operation                        | Improvement    |
| -------------------------------- | -------------- |
| Construction (TimestampBuilder)  | 2-4x faster    |
| Clone                            | 6-10x faster   |
| Equality / Ordering              | 1.3-6x faster  |
| IonEq / IonDataOrd / IonDataHash | 2-10x faster |
| Formatting (Display)             | 10-21% faster  |
| Element read/write               | 1-6% faster    |

#### Known limitations

- Fractional precision is capped at 18 digits (attosecond). The `timestampsLargeFractionalPrecision` and `T6_large` conformance tests are skipped.
- `cmp` for the specific case of same-instant-different-offset shows a ~30% regression on Graviton relative to the old chrono-backed implementation (which stored epoch seconds internally and could compare with a single i64 subtract). This case is still single-digit nanoseconds per operation, so unlikely to be an issue in practice, and well worth the improvements to construction and formatting.

---

### DIFF
--- a/src/types/timestamp.rs
+++ b/src/types/timestamp.rs
@@ -1,140 +1,381 @@
-use crate::decimal::Sign;
 use crate::ion_data::{IonDataHash, IonDataOrd, IonEq};
-use crate::result::{IonError, IonFailure, IonResult};
-use crate::types::{CountDecimalDigits, Decimal};
-use chrono::{
-    DateTime, Datelike, FixedOffset, LocalResult, NaiveDate, NaiveDateTime, TimeZone, Timelike,
-};
+use crate::result::{IonFailure, IonResult};
+use crate::types::Decimal;
+#[cfg(feature = "experimental-chrono")]
+use crate::IonError;
+#[cfg(feature = "experimental-chrono")]
+use chrono::{DateTime, Datelike, FixedOffset, NaiveDate, NaiveDateTime, TimeZone, Timelike};
 use std::cmp::Ordering;
 use std::fmt::{Debug, Display, Formatter};
 use std::hash::{Hash, Hasher};
 use std::marker::PhantomData;
 
 /// Indicates the most precise time unit that has been specified in the accompanying [Timestamp].
+// IMPL NOTE: Discriminant values are cast to u64 and stored in the packed bit layout.
+// They must remain stable. They are not exposed in the public API.
 #[derive(Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Ord, Default, Hash)]
 pub enum TimestampPrecision {
     /// Year-level precision (e.g. `2020T`)
     #[default]
-    Year,
+    Year = 0,
     /// Month-level precision (e.g. `2020-08T`)
-    Month,
+    Month = 1,
     /// Day-level precision (e.g. `2020-08-01T`)
-    Day,
+    Day = 2,
     /// Minute-level precision (e.g. `2020-08-01T12:34Z`)
-    HourAndMinute,
+    HourAndMinute = 3,
     /// Second-level precision or greater. (e.g. `2020-08-01T12:34:56Z` or `2020-08-01T12:34:56.123456789Z`)
-    Second,
+    Second = 4,
 }
 
-// [Default] cannot be derived for enum types. Providing a manual implementation of this type
-// allows us to derive Default for [Timestamp].
-
-/// Stores the precision of a Timestamp's fractional seconds, if present. This type is not
-/// self-contained; if the Timestamp has a precision that is less than or equal to nanoseconds
-/// (i.e. fewer than 10 digits), the fractional seconds value will be stored in the Timestamp's
-/// NaiveDateTime component and the Mantissa will indicate the number of digits from that value
-/// that should be used. If the precision is 10 or more digits, the Mantissa will store the value
-/// itself as a Decimal with the correct precision.
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub enum Mantissa {
-    /// The number of digits of precision in the Timestamp's fractional seconds. For example, a
-    /// value of `3` would indicate millisecond precision. A value of `6` would indicate
-    /// microsecond precision. All precisions less than or equal to nanoseconds should use
-    /// this representation when possible.
-    Digits(u32),
-    /// Specifies the fractional seconds precisely as a `Decimal` in the range `>= 0` and `< 1`.
-    /// The Decimal will have the correct precision; the complete value can and should be used.
-    /// This representation should only be used for precisions greater than nanoseconds as it can
-    /// require allocations.
-    Arbitrary(Decimal),
-}
-
-impl Mantissa {
-    fn decimals_equal(d1: &Decimal, d2: &Decimal) -> bool {
-        // See the [EmptyMantissa] trait for details about `is_empty()`
-        (d1.is_empty() && d2.is_empty())
-            // Exact equality test
-            || d1.eq(d2)
-            // Coefficient zeros' signs don't have to match for fractional seconds.
-            || (d1.coefficient().is_zero() && d2.coefficient().is_zero() && d1.exponent == d2.exponent)
-    }
-
-    fn decimals_compare(d1: &Decimal, d2: &Decimal) -> Ordering {
-        // See the [EmptyMantissa] trait for details about `is_empty()`
-        if d1.is_empty() && d2.is_empty() {
-            Ordering::Equal
-        } else if d1.coefficient().is_zero() && d2.coefficient().is_zero() {
-            // Coefficient zeros' signs don't have to be compared for fractional seconds.
-            d1.exponent.cmp(&d2.exponent)
-        } else {
-            // Exact comparison test
-            d1.cmp(d2)
-        }
-    }
-}
-
-trait EmptyMantissa {
-    /// Returns true if the Mantissa's value is equivalent to not having specified a
-    /// sub-second precision at all. For example, `Mantissa::Digits(0)` or
-    /// `Mantissa::Arbitrary(Decimal::new(0, 0))`.
-    fn is_empty(&self) -> bool;
-}
-
-impl EmptyMantissa for Decimal {
-    fn is_empty(&self) -> bool {
-        self.coefficient().is_zero() && self.exponent == 0
-    }
-}
-
-impl EmptyMantissa for Mantissa {
-    fn is_empty(&self) -> bool {
-        match self {
-            // Look at zero digits of the DateTime's nanoseconds
-            Mantissa::Digits(0) => true,
-            // Or a Decimal with a coefficient of zero (any sign) and an exponent of zero.
-            Mantissa::Arbitrary(d) => d.is_empty(),
-            _ => false,
-        }
-    }
-}
-
-/// Returns the first `num_digits` digits of the specified `value`.
-// This is used in Timestamp's implementation of [PartialEq].
-fn first_n_digits_of(num_digits: u32, value: u32) -> u32 {
-    let total_digits = value.count_decimal_digits();
-    if total_digits <= num_digits {
-        return value;
-    }
-    // Truncate the trailing digits
-    value / 10u32.pow(total_digits - num_digits)
-}
-
-/// Constructs a [FixedOffset] at the specified offset seconds from UTC. If the specified offset
+/// Constructs a [`FixedOffset`] at the specified offset seconds from UTC. If the specified offset
 /// is out of bounds, this method will panic.
+// Only the tests below construct a `FixedOffset` this way; production conversions go through
+// `try_to_datetime_fixed_offset`, which reports an out-of-range offset as an `IonError`.
+#[cfg(all(test, feature = "experimental-chrono"))]
 fn offset_east(seconds_east: i32) -> FixedOffset {
-    FixedOffset::east_opt(seconds_east)
-        // This error case is expected to be handled before this method is called.
-        .expect("seconds_east was outside the supported range")
+    FixedOffset::east_opt(seconds_east).expect("seconds_east was outside the supported range")
 }
 
-/// Constructs a [`DateTime<FixedOffset>`] at the specified offset using the fields of
-/// [`NaiveDateTime`] representing the desired UTC datetime.
-fn datetime_at_offset(utc_datetime: &NaiveDateTime, seconds_east: i32) -> DateTime<FixedOffset> {
-    offset_east(seconds_east).from_utc_datetime(utc_datetime)
+// ─── Packed bit layout ────────────────────────────────────────────────
+//
+// All date-time fields are stored as LOCAL time in a single u64.
+// A second u64 holds the attoseconds payload.
+//
+// Date-time fields are in the HIGH bits (most-significant first) so that
+// a numeric comparison of the packed value yields chronological order.
+// Metadata (offset, precision, subsecond_digits) is in the LOW bits.
+//
+// Bit layout of `packed` (MSB = bit 63):
+//
+//   [63:50] year                (14 bits, 0-9999)
+//   [49:46] month               (4 bits, 1-12)
+//   [45:41] day                 (5 bits, 1-31)
+//   [40:36] hour                (5 bits, 0-23)
+//   [35:30] minute              (6 bits, 0-59)
+//   [29:24] second              (6 bits, 0-59)
+//   [23:21] precision           (3 bits, 0-4 maps to TimestampPrecision)
+//   [20:16] subsecond_precision (5 bits, 0 = none, 1-18 = digit count)
+//   [15:4]  offset              (12 bits, biased unsigned: stored = minutes + 1440;
+//                                valid range 1..=2879; 0 = unknown)
+//   [3:0]   spare               (4 bits)
+//
+// `attoseconds`: fractional seconds normalized to 10^-18 scale.
+// `subsecond_precision` records display precision (how many digits to render).
+//
+// Invariants:
+//
+// 1. Fields below the declared precision hold their default values:
+//    - Year:          month=1, day=1, hour=0, minute=0, second=0
+//    - Month:         day=1, hour=0, minute=0, second=0
+//    - Day:           hour=0, minute=0, second=0
+//    - HourAndMinute: second=0
+//    - Second:        (no fields required to be zero)
+//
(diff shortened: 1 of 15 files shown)


popematt:
Also closes #411 as no longer relevant.

popematt:
This will also mostly solve #838.

src/types/timestamp.rs
```diff
@@ -1,140 +1,381 @@
...
-    /// require allocations.
-    Arbitrary(Decimal),
+/// Constructs a [`FixedOffset`] at the specified offset seconds from UTC. If the specified offset
+/// is out of bounds, this method will panic.
+// Only the tests below construct a `FixedOffset` this way; production conversions go through
+// `try_to_datetime_fixed_offset`, which reports an out-of-range offset as an `IonError`.
+#[cfg(all(test, feature = "experimental-chrono"))]
+fn offset_east(seconds_east: i32) -> FixedOffset {
+    FixedOffset::east_opt(seconds_east).expect("seconds_east was outside the supported range")
 }
```
almann: What tests are dependent on this, should this be in the test module itself with a gate on it?

src/types/timestamp.rs
```diff
@@ -1,140 +1,381 @@
...
-use crate::types::{CountDecimalDigits, Decimal};
-use chrono::{
-    DateTime, Datelike, FixedOffset, LocalResult, NaiveDate, NaiveDateTime, TimeZone, Timelike,
-};
+use crate::result::{IonFailure, IonResult};
+use crate::types::Decimal;
+#[cfg(feature = "experimental-chrono")]
+use crate::IonError;
+#[cfg(feature = "experimental-chrono")]
+use chrono::{DateTime, Datelike, FixedOffset, NaiveDate, NaiveDateTime, TimeZone, Timelike};
```
almann: Is it possible to constraint all `experimental-chrono` dependencies to a submodule of `types::timestamp`?

src/types/timestamp.rs
```diff
@@ -1,140 +1,381 @@
...
+//   [49:46] month               (4 bits, 1-12)
+//   [45:41] day                 (5 bits, 1-31)
+//   [40:36] hour                (5 bits, 0-23)
+//   [35:30] minute              (6 bits, 0-59)
+//   [29:24] second              (6 bits, 0-59)
+//   [23:21] precision           (3 bits, 0-4 maps to TimestampPrecision)
+//   [20:16] subsecond_precision (5 bits, 0 = none, 1-18 = digit count)
+//   [15:4]  offset              (12 bits, biased unsigned: stored = minutes + 1440;
+//                                valid range 1..=2879; 0 = unknown)
+//   [3:0]   spare               (4 bits)
```
almann: *Minor*: I would just call this `unused` or `reserved`.

src/types/timestamp.rs
```diff
@@ -1,140 +1,381 @@
...
-            d1.cmp(d2)
-        }
+// ─── Packed bit layout ────────────────────────────────────────────────
+//
+// All date-time fields are stored as LOCAL time in a single u64.
+// A second u64 holds the attoseconds payload.
+//
+// Date-time fields are in the HIGH bits (most-significant first) so that
+// a numeric comparison of the packed value yields chronological order.
+// Metadata (offset, precision, subsecond_digits) is in the LOW bits.
```
almann: I might lead with something like the following because it was not clear on first read until I figured things out later:

```rust
// `Timestamp` is two 64-bit words. The time *value* needs 100 bits; 40 for the
// date-time fields (year..second) and 60 for the fraction (10^18 - 1 < 2^60)
// which is more than one word holds. It is intentional that the
// date-time component and the fraction each get their own word: the date-time
// occupies the top of `packed_fields`, and the fraction gets `attoseconds` to itself.
//
// The 24 bits left over in `packed_fields` hold the metadata, which is why precision,
// subsecond_precision, and offset sit *between* `second` and the fraction in
// significance order. Two consequences for future changes:
//
// - year..second must stay in the top bits, in descending significance, with no
//   metadata interleaved: `Ord` and `PartialEq` compare `packed & DATETIME_MASK` as a
//   single integer.
// - `ion_cmp` compares `packed_fields` and then `attoseconds` as raw integers, so this
//   layout *is* the `IonDataOrd` total order — date-time, precision,
//   subsecond_precision, offset, fraction. Permuting the metadata fields changes how
//   `IonData`-wrapped timestamps sort.
```
Also this should be closer to the timestamp struct declaration itself.

src/types/timestamp.rs
```diff
@@ -1,140 +1,381 @@
...
+const SUBSECOND_PRECISION_MASK: u64 = (1 << SUBSECOND_PRECISION_BITS) - 1;
+
+/// Mask covering only the date-time fields (year through second).
+/// Comparing `packed & DATETIME_MASK` gives chronological order.
+const DATETIME_MASK: u64 = (YEAR_MASK << YEAR_SHIFT)
+    | (MONTH_MASK << MONTH_SHIFT)
+    | (DAY_MASK << DAY_SHIFT)
+    | (HOUR_MASK << HOUR_SHIFT)
+    | (MINUTE_MASK << MINUTE_SHIFT)
+    | (SECOND_MASK << SECOND_SHIFT);
```
almann, line 79: Could these static asserts be useful to avoid compile time invariance breaks for changes?
```rs
const _: () = assert!(YEAR_SHIFT + YEAR_BITS == 64, "fields must fill the word");
// The date-time region must be contiguous and top-aligned; `Ord` compares it as an integer.
const _: () = assert!(DATETIME_MASK.leading_zeros() == 0);
const _: () = assert!(DATETIME_MASK.trailing_zeros() == SECOND_SHIFT as u32);
const _: () = assert!(DATETIME_MASK.count_ones() == 64 - SECOND_SHIFT as u32);
const _: () = assert!(
    10u128.pow(MAX_FRAC_DIGITS as u32) <= u64::MAX as u128,
    "attoseconds scale must fit in the u64 payload"
);
```

src/types/timestamp.rs
```diff
@@ -1,140 +1,381 @@
...
+
+/// Bias added to offset-in-minutes before storage. With this bias the valid
+/// range (-1439..=+1439) maps to 1..=2879, leaving 0 free as the "unknown" sentinel.
+const OFFSET_BIAS: i16 = 1440;
+const OFFSET_UNKNOWN_SENTINEL: u16 = 0;
```
almann: This was a bit hard for me to read, maybe:

```rust
/// Offset-in-minutes is stored biased so the field's zero value can serve as the
/// "unknown offset" sentinel:
///
///     stored  = minutes + OFFSET_BIAS   // -1439..=1439  ->  1..=2879
///     minutes = stored  - OFFSET_BIAS
///
/// Note that `+00:00` stores as 1440, distinct from _unknown_.
const OFFSET_BIAS: i16 = 1440;

/// Stored `offset` value meaning "no offset specified" (Ion's `-00:00`).
const OFFSET_UNKNOWN_SENTINEL: u16 = 0;
```

src/types/timestamp.rs
```diff
@@ -1,140 +1,381 @@
...
+use crate::result::{IonFailure, IonResult};
+use crate::types::Decimal;
+#[cfg(feature = "experimental-chrono")]
+use crate::IonError;
+#[cfg(feature = "experimental-chrono")]
+use chrono::{DateTime, Datelike, FixedOffset, NaiveDate, NaiveDateTime, TimeZone, Timelike};
 use std::cmp::Ordering;
 use std::fmt::{Debug, Display, Formatter};
 use std::hash::{Hash, Hasher};
 use std::marker::PhantomData;
```
almann, line 8: Consider the following failing tests:

```rs
    // -----------------------------------------------------------------------
    // Validation gaps.
    //
    // `build` calls `validate_field_ranges` before narrowing `u32` fields to
    // `u8`/`u16`; `build_utc_fields_at_offset` does not, and the validation
    // inside `from_utc_fields`/`from_fields` only ever sees the already-wrapped
    // value. The two public chrono conversions also disagree about years
    // outside Ion's 1..=9999 range: one panics, the other silently masks.
    //
    // Every test below asserts the behavior the code intends.
    // -----------------------------------------------------------------------

    /// `build`'s own comment cites `month: 268` becoming `12` as the reason it
    /// validates before narrowing. Its sibling skips that guard.
    #[rstest]
    #[case::year(67_557, 1, 1, 0, 0)]
    #[case::month(2021, 268, 1, 0, 0)]
    #[case::day(2021, 1, 261, 0, 0)]
    #[case::hour(2021, 1, 1, 261, 0)]
    #[case::minute(2021, 1, 1, 0, 286)]
    fn build_utc_fields_at_offset_rejects_out_of_range_fields(
        #[case] year: u32,
        #[case] month: u32,
        #[case] day: u32,
        #[case] hour: u32,
        #[case] minute: u32,
    ) {
        let built = TimestampBuilder::with_ymd(year, month, day)
            .with_hour_and_minute(hour, minute)
            .build_utc_fields_at_offset(0);
        assert!(
            built.is_err(),
            "expected {year}-{month}-{day}T{hour}:{minute} to be rejected, got `{}`",
            built.unwrap()
        );
    }

    /// The same defect reached through the public API. A binary 1.0 timestamp
    /// with a known offset at HourAndMinute precision routes to
    /// `build_utc_fields_at_offset`, so an out-of-range VarUInt month decodes
    /// as a plausible wrong month instead of raising a decoding error.
    #[test]
    fn binary_1_0_rejects_timestamp_with_out_of_range_month() {
        // 0x6L       timestamp, L-byte body
        // 0xBC       VarInt  offset = +60 minutes (a *known* offset)
        // 0x0F 0xE5  VarUInt year   = 2021
        // ...        VarUInt month
        // 0x81       VarUInt day    = 1
        // 0x80 0x80  VarUInt hour   = 0, minute = 0
        const IVM: [u8; 4] = [0xE0, 0x01, 0x00, 0xEA];

        // Control: month = 12 as a one-byte VarUInt reads as December.
        let mut control = IVM.to_vec();
        control.extend_from_slice(&[0x67, 0xBC, 0x0F, 0xE5, 0x8C, 0x81, 0x80, 0x80]);
        let control = Element::read_one(&control).expect("control timestamp must decode");
        assert_eq!(
            control.as_timestamp().expect("control is a timestamp").month(),
            12
        );

        // month = 268 as a two-byte VarUInt, and `268 as u8` is also 12.
        let mut wrapped = IVM.to_vec();
        wrapped.extend_from_slice(&[0x68, 0xBC, 0x0F, 0xE5, 0x02, 0x8C, 0x81, 0x80, 0x80]);
        let decoded = Element::read_one(&wrapped);
        assert!(
            decoded.is_err(),
            "month 268 should be a decoding error, decoded as `{}`",
            decoded.unwrap()
        );
    }

    /// Invariant 4 (`attoseconds < 10^18`) lives only in
    /// `validate_field_ranges`. `with_milliseconds(1000).build()` is already
    /// covered as an error above; the unvalidated path accepts it and then
    /// renders more digits than the declared precision.
    #[test]
    fn over_range_subseconds_do_not_outrun_declared_precision() {
        let built = TimestampBuilder::with_ymd(2024, 1, 1)
            .with_hms(0, 0, 0)
            .with_milliseconds(1023)
            .build_utc_fields_at_offset(0);
        let Ok(timestamp) = built else {
            return; // Rejected, as `build` would have. Nothing further to check.
        };
        assert_eq!(timestamp.subsecond_digit_count(), 3);
        let rendered = timestamp.to_string();
        let fraction: String = rendered
            .split('.')
            .nth(1)
            .unwrap_or_default()
            .chars()
            .take_while(|c| c.is_ascii_digit())
            .collect();
        assert_eq!(
            fraction.len(),
            3,
            "declared 3 subsecond digits but rendered `{rendered}`"
        );
    }

    /// chrono's `NaiveDateTime` spans roughly ±262143, while `from_fields`
    /// rejects year 0 and years above 9999, so this `expect` is reachable.
    #[cfg(feature = "experimental-chrono")]
    #[test]
    fn from_naive_datetime_does_not_panic_outside_ion_year_range() {
        let year_zero = NaiveDate::from_ymd_opt(0, 1, 1)
            .unwrap()
            .and_hms_opt(0, 0, 0)
            .unwrap();
        let previous_hook = std::panic::take_hook();
        std::panic::set_hook(Box::new(|_| {}));
        let outcome = std::panic::catch_unwind(|| Timestamp::from(year_zero));
        std::panic::set_hook(previous_hook);
        assert!(
            outcome.is_ok(),
            "`From<NaiveDateTime>` panicked on year 0 instead of reporting an error"
        );
    }

    /// The sibling conversion bypasses validation entirely and calls
    /// `pack_masked`, so an out-of-range year is masked to 14 bits rather than
    /// panicking. Whichever policy is right, the two should agree.
    #[cfg(feature = "experimental-chrono")]
    #[test]
    fn from_fixed_offset_datetime_does_not_silently_wrap_year() {
        use chrono::TimeZone;
        let far_future = FixedOffset::east_opt(0)
            .unwrap()
            .with_ymd_and_hms(20_000, 1, 1, 0, 0, 0)
            .unwrap();
        let timestamp = Timestamp::from(far_future);
        // 20000 & 0x3FFF == 3616
        assert_ne!(
            timestamp.year(),
            3616,
            "year 20000 was masked to {} rather than rejected or clamped",
            timestamp.year()
        );
    }
```

almann (changes requested):
Some  cases we should either add or explicitly defer (I think some are covered by the issues that are out of scope).

src/types/timestamp.rs
popematt: Yes. I wanted to do that initially, but figured it was already a big enough diff, and that I could do it as a followup. I don't mind adding it now.

src/types/timestamp.rs
popematt: Moved to `chrono_interop` module.

src/types/timestamp.rs
popematt: First 3 are fixed. The last two now silently accept the invalid year, to match the existing behavior. I've update #1033 and #1035 as a result of this.

src/types/timestamp.rs
```diff
@@ -1213,17 +1134,53 @@ impl TimestampBuilder<HasSeconds> {
...
+        } else {
+            let digits = fractional_seconds.exponent.unsigned_abs();
+            let coefficient = fractional_seconds
+                .coefficient()
+                .magnitude()
+                .as_u128()
+                .unwrap_or(0) as u64;
+            // Convert coefficient at `digits` scale to attoseconds (10^-18).
+            // attoseconds = coefficient * 10^(18 - digits)
+            self.fractional_digits = digits.min(MAX_FRAC_DIGITS as u64) as u8;
```
almann: This technically truncates precision, the 19 digit case of fractional zero would turn into 18, if we want to be consistent, we should error out like we do in other `>` precision we support cases.

src/types/timestamp.rs
```diff
@@ -1213,17 +1134,53 @@ impl TimestampBuilder<HasSeconds> {
         nanoseconds: u32,
         precision_digits: u32,
     ) -> TimestampBuilder<HasFractionalSeconds> {
-        self.fractional_seconds = Some(Mantissa::Digits(precision_digits));
-        self.nanoseconds = Some(nanoseconds);
+        self.attoseconds = (nanoseconds as u64).saturating_mul(1_000_000_000);
+        self.fractional_digits = precision_digits as u8;
```
almann: This overflows silently, is that intentional?  Should this be a documented precondition?

src/types/timestamp.rs
```diff
@@ -1213,17 +1134,53 @@ impl TimestampBuilder<HasSeconds> {
         nanoseconds: u32,
         precision_digits: u32,
     ) -> TimestampBuilder<HasFractionalSeconds> {
-        self.fractional_seconds = Some(Mantissa::Digits(precision_digits));
-        self.nanoseconds = Some(nanoseconds);
+        self.attoseconds = (nanoseconds as u64).saturating_mul(1_000_000_000);
```
almann: I think assumes that the nanoseconds are structured correctly, as in the precision of the value matches the precision digits, should that assumption be documented or is this an error?

almann (changes requested):
Can you check on these issues, I think they are correctness bugs or at a minimum the preconditions should be documented.

src/types/timestamp.rs
popematt: The intention of this method (and the previous behavior) is the truncate to the number of precision digits (although the previous implementation did when you read the field rather than when the instance was constructed). I've updated it accordingly.

almann (approved):
:shipit:

PR #1044, opened 2026-09-03
perf: add a 16-byte, align-8 overflowing integer representation 

popematt:
**Issue #, if available:**

Supports #476

**Description of changes:**

This adds `OverflowingInt`, a compact sign-magnitude integer that will back the crate's numeric data-model types. It is a 16-byte tagged union where magnitudes up to 126 bits are stored inline in two `u64` words, and anything larger heap-allocates a `Box<BigUint>`.  By using `u64` words, we get an alignment of 8 bytes (on aarch64), which is smaller than then 16 byte alignment of `IntData` and `UIntData`. By using a `union`, we can steal a bit from the payload to use as the union discriminant so that we don't need to add another byte (plus additional padding) for an enum discriminant. In followup PRs, this new primitive will be used for `Decimal`, `UInt`, and `Int`, and it could also be used to restore arbitrary precision for `Timestamp` after #1041 lands.

The type owns three invariants, all enforced by construction rather than by callers:

  - Tag agreement: bit 0 of the word at offset 0 identifies the active variant. The tag is read through a dedicated `u64` union arm, never by reading the pointer bytes as an integer.
  - Canonicality: a magnitude is heap-backed if and only if it does not fit inline. Every constructor demotes a value that fits. This is what makes comparison, hashing, and the zero test correct without ever promoting a value.
  - Freed exactly once: `Drop` lives on this type alone, so the future wrappers carry no `unsafe` and rely on drop glue.

The type is immutable—the only `&mut self` method is `Drop`, and no accessor lends out a `&mut` into the payload. Comparison and hashing are hand-written, allocation-free, and follow decArith `compare-total` (so `-0 < +0`).

I've added a WASM test target to make sure that it works on 32-bit systems. I've also added miri coverage for 64-bit LE, 64-bit BE, and 32-bit LE architectures to ensure that it is sound.

### DIFF
--- a/.github/workflows/miri.yml
+++ b/.github/workflows/miri.yml
@@ -49,3 +49,75 @@ jobs:
       - name: Test with Miri (windows)
         if: matrix.os == 'windows-latest'
         run: $env:MIRIFLAGS="-Zmiri-disable-isolation -Zmiri-tree-borrows" ; cargo miri test serde --features "${{ matrix.features }}"
+
+  unsafe-representations:
+    name: Miri over the OverflowingInt representation
+    # A single runner: Miri interprets the abstract machine, so these tests —
+    # which have no syscalls, FFI, clock, or `cfg`-conditional code — behave
+    # identically across host OSes. The dimensions the layout actually depends on
+    # are pointer width (covered by the wasm32 `cargo check` in `rust.yml`) and
+    # endianness, exercised below by a big-endian target rather than by a second
+    # OS (all GitHub runners are little-endian).
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != 'amazon-ion/ion-rust'
+    permissions:
+      checks: write
+    env:
+      # Module-path test selection, defined once. Each dependent part appends its
+      # own module (space-separated; the value word-splits into one filter each),
+      # e.g. `overflowing_int coefficient integer annotations`.
+      MIRI_TEST_SELECTION: overflowing_int
+
+    steps:
+      - name: Git Checkout
+        uses: actions/checkout@v5
+        with:
+          submodules: recursive
+      - name: Rust Toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
+          components: miri
+          rustflags: "" # Default: "-D warnings', we do not want that for now.
+          override: true
+      - name: Setup Miri
+        run: cargo miri setup
+      # The `serde` job above filters by test name and will not select these
+      # tests. This job selects by MODULE PATH via `MIRI_TEST_SELECTION` (above),
+      # which fails closed — a new test in a selected module runs with no per-name
+      # registration. Filters go after `--` (Cargo rejects multiple bare filters).
+      #
+      # `-Zmiri-strict-provenance` is in the flag set from the start: the union
+      # reads its tag through a dedicated arm so it never materializes the boxed
+      # pointer as an integer, and this flag is what enforces that.
+      - name: Test with Miri (native x86-64, 64-bit, little-endian)
+        run: MIRIFLAGS="-Zmiri-disable-isolation -Zmiri-tree-borrows -Zmiri-strict-provenance" cargo miri test -- $MIRI_TEST_SELECTION
+      # 32-bit pointer width: `Box` is 4 bytes, so `HeapValue`'s layout and
+      # trailing padding differ from the 64-bit case, and `u64` aligns to 4 under
+      # this ABI. Exercises the unsafe reads against a layout the 64-bit run
+      # cannot produce. (The wasm32 `cargo check` covers the 32-bit assertions at
+      # compile time; this runs the unsafe under a 32-bit interpreter.)
+      - name: Test with Miri (i686, 32-bit, little-endian)
+        run: MIRIFLAGS="-Zmiri-disable-isolation -Zmiri-tree-borrows -Zmiri-strict-provenance" cargo miri test --target i686-unknown-linux-gnu -- $MIRI_TEST_SELECTION
+      # Big-endian target: the union reconstructs its magnitude from `u64` word
+      # values rather than raw bytes, so the layout must be endian-independent.
+      # Miri interprets the foreign target directly (no s390x toolchain needed),
+      # which is a stronger check than any little-endian second OS.
+      - name: Test with Miri (s390x, 64-bit, big-endian)
+        run: MIRIFLAGS="-Zmiri-disable-isolation -Zmiri-tree-borrows -Zmiri-strict-provenance" cargo miri test --target s390x-unknown-linux-gnu -- $MIRI_TEST_SELECTION
+
+  confirm-miri:
+    # A single "join" over both Miri jobs so branch protection can require one
+    # stable check name and neither the serde matrix nor the unsafe-representation
+    # coverage can regress without blocking merge.
+    needs: [build, unsafe-representations]
+    name: Miri Confirmation
+    runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


src/types/overflowing_int.rs
```diff
@@ -0,0 +1,1257 @@
...
+    /// The tag word alone. Every tag read goes through this arm, never through
+    /// `words` by value — reading `words` while the heap variant is active would
+    /// materialize the `Box` as an integer.
+    tag: u64,
+    /// Inline variant (tag bit == 1): `words[0]` holds the tag, sign, and the low
+    /// 62 magnitude bits; `words[1]` holds the high 64 magnitude bits.
+    words: [u64; 2],
+    /// Heap variant (tag bit == 0).
+    heap: ManuallyDrop<HeapValue>,
+}
```
almann: You should statically assert the size of this type.

src/types/overflowing_int.rs
```diff
@@ -0,0 +1,1257 @@
...
+/// always a fully-initialized `u64` this type controls, independent of pointer
+/// representation, width, and endianness.
+#[repr(C)]
+struct HeapValue {
+    /// Bit 0: always 0 (the heap tag). Bit 63: sign (1 = negative).
+    meta: u64,
+    /// The magnitude. Always `>= 2^126` by canonicality, and therefore never
+    /// zero.
+    value: Box<BigUint>,
+}
```
almann, line 129: Now that we have effectively a tag bit and sign bit, can we/should we leverage the other 62 bits for something--seems like we have at least some words that can fit inline or even 62-bits of _length_.  The downside being that we would have to have a custom encoding for the heap value, but something we should at least consider or dismiss.

src/types/overflowing_int.rs
```diff
@@ -0,0 +1,1257 @@
...
+    // Inline: tag = 1 (bit 0), sign = negative (bit 1 = 1), magnitude = 0.
+    pub(crate) const NEGATIVE_ZERO: Self = Self { words: [0b11, 0] };
+
+    // ===== Construction =====
+
+    /// Builds the inline variant. The caller guarantees `magnitude < 2^126`.
+    #[inline]
+    fn from_inline(sign: Sign, magnitude: u128) -> Self {
+        debug_assert!(magnitude < INLINE_LIMIT, "magnitude does not fit inline");
+        let low_mask = (1u128 << INLINE_LOW_MAGNITUDE_BITS) - 1;
```
almann: This is a constant, right? I would just define it as such for readability despite the compiler likely folding it.

src/types/overflowing_int.rs
```diff
@@ -0,0 +1,1257 @@
...
+        let low = (magnitude & low_mask) as u64;
+        // bits 2..64 of `words[0]` hold the low magnitude bits; bit 1 the sign,
+        // bit 0 the tag. `words[1]` holds the high magnitude bits.
+        let w0 = (low << 2) | (sign_bit(sign) << 1) | 1;
+        let w1 = (magnitude >> INLINE_LOW_MAGNITUDE_BITS) as u64;
```
almann: Something I am wondering is if this becomes simpler if the high bits are the sign and tag. so you would avoid a single shift on the low word and make it consistent to the heap overflow meta-data.

For clarity I would make both cases have bit 63 be the tag, and bit 62 be the sign.  The order is aesthetic, but the consistency is more than aesthetic as the logic is reusable across both cases.

.github/workflows/miri.yml
```diff
@@ -49,3 +49,75 @@ jobs:
...
+      # The `serde` job above filters by test name and will not select these
+      # tests. This job selects by MODULE PATH via `MIRI_TEST_SELECTION` (above),
+      # which fails closed — a new test in a selected module runs with no per-name
+      # registration. Filters go after `--` (Cargo rejects multiple bare filters).
+      #
+      # `-Zmiri-strict-provenance` is in the flag set from the start: the union
+      # reads its tag through a dedicated arm so it never materializes the boxed
+      # pointer as an integer, and this flag is what enforces that.
+      - name: Test with Miri (native x86-64, 64-bit, little-endian)
+        run: MIRIFLAGS="-Zmiri-disable-isolation -Zmiri-tree-borrows -Zmiri-strict-provenance" cargo miri test -- $MIRI_TEST_SELECTION
```
almann: I think these will fail open if the test renames--it might be worth adding a script to cover this and fail if the test selection is empty or incomplete.

Should we be just running everything in miri?

src/types/overflowing_int.rs
```diff
@@ -0,0 +1,1257 @@
+//! A compact, immutable sign-magnitude integer used to back the crate's numeric
+//! types.
+//!
+//! [`OverflowingInt`] is a 16-byte, align-8 tagged union. It stores a
+//! sign-magnitude integer inline when the magnitude fits in 126 bits and
+//! heap-allocates a [`BigUint`] beyond that. Align 8 is the point of the type:
+//! an inline `i128` (or a `Box<BigInt>` paired with a sign field) forces align
+//! 16, which pads every enclosing enum.
```
almann: "Align at most 8" (as per your static check) since it is possible to align your type at 4 on some platforms (e.g., i686 which you have miri tests for).

.github/workflows/rust.yml
```diff
@@ -143,15 +143,39 @@ jobs:
           LC_ALL=C sort -u /tmp/public-api-bigdecimal.txt > /tmp/public-api-bigdecimal-sorted.txt
           comm -13 /tmp/public-api-default-sorted.txt /tmp/public-api-bigdecimal-sorted.txt > /tmp/public-api-bigdecimal-delta.txt
           diff -u api/bigdecimal.txt /tmp/public-api-bigdecimal-delta.txt
+  wasm:
+    # `cargo check` on a 32-bit pointer width. This is what actually exercises
+    # the pointer-width-dependent layout claims that a `size_of` assertion on a
+    # 64-bit host cannot: on wasm32 the heap payload is smaller, yet the layout
+    # must still hold. wasm32 is the realistic 32-bit target since ion-rs enables
+    # chrono's `wasmbind`.
```
almann: What layout difference is this intended to validate? On wasm32, `Box<BigUint>` is four bytes, but `u64` remains eight-byte aligned, so HeapValue is still 16 bytes with four bytes of trailing padding.
The i686 Miri run is what exercises a 12-byte HeapValue.

More generally, the union’s 16-byte layout relies on HeapValue being no larger than 16 bytes and no more aligned than eight. `[u64; 2]` fixes the union’s minimum size at 16, while the const
assertions reject targets where those assumptions do not hold.

If the goal is just running the tests on wasm target, then that is just good and I wouldn't comment on this.

almann (changes requested):
Please see my comment about consistent layout--everything else seems reasonable.

.github/workflows/miri.yml
popematt: > Should we be just running everything in miri?

I like this, and I will add that as a follow up task. I may need to a `#[cfg_attr(miri, ignore)]` for some tests that use things that are unsupported by Miri, and if there are test failures, I would try to fix them at the same time.

src/types/overflowing_int.rs
popematt: The heap path is supposed to be cold/uncommon, so I'd rather prioritize simplicity over trying to maximize every bit on that path.

.github/workflows/rust.yml
popematt: The goal is to test on wasm generally (to make sure that everything is sound on 32 bit). It's not necessarily a big deal for this one, but I'm planning on making a few more unsafe changes where it might be more relevant.

src/types/overflowing_int.rs
popematt: I've updated the layout.

almann (approved):
:shipit:

PR #1048, opened 2026-09-09
perf(struct): store fields in a boxed slice with a lazy index

popematt:
**Issue #, if available:**

Supports #476

**Description of changes:**

This replaces `Struct`'s `Vec<(Symbol, Element)>` and eager `HashMap<Symbol, SmallVec<[usize; 1]>>` with a single boxed slice plus lazily-built indexes. `Struct` goes from 48 to 32 bytes on 64-bit targets (asserted; 16 on 32-bit), `get()` stays sub-linear for large structs, and constructing a struct no longer allocates a map or a per-field `SmallVec`.

`Fields` is now a `slots: Box<[(Symbol, Element)]>` and an `aux: OnceLock<Box<FieldsAux>>`. `FieldsAux` holds two independently lazy indexes:

- a `hashbrown::HashTable<HashIndex>` that `get()` builds and probes once a struct is large enough to be worth it, and
- a `by_field: Box<[u32]>` that orders the slots into the `IonDataOrd` order for `PartialEq`, `IonEq`, `IonDataOrd`, and `IonDataHash`.

A struct that is only iterated builds neither index. Using a boxed slice rather than a `Vec` is deliberate: a struct is immutable once built, so the capacity word is dead weight, and dropping it is part of what gets `Struct` down to 32 bytes.

The PR is three commits: the storage rewrite, a `struct_ops` benchmark, and the build-time threshold knob described below.

#### Tunable `LINEAR_SCAN_THRESHOLD`

`get()` scans linearly at or below `LINEAR_SCAN_THRESHOLD` and builds and probes the hash index above it. The threshold is a compile-time `const` that can be overridden at build time with `ION_RS_STRUCT_LINEAR_SCAN_THRESHOLD=<n> cargo build`. It is parsed in a `const fn`, so a malformed value is a compile error rather than a silent fallback, and the `build.rs` entry makes cargo rebuild when it changes.

The default is 16. I chose it on aarch64 against a hit-dominated workload, where the linear-vs-hash crossover measured around 10 fields at 2 lookups per field, ~19 at 1x, and ~28 at 0.5x. The asymmetry is what pushed the default low: setting the threshold too high pays linear's O(n²) tail as structs grow, while setting it too low pays only a bounded index build of roughly 140 ns plus ~14 ns per field. Erring low is the cheaper mistake. The `struct_threshold_model` bench in this PR is the instrument those numbers came from, and it can re-derive the crossover if the hardware or the `Symbol`/`Element` layout changes.

I'm happy to drop the environment override if we would rather not carry a tuning knob in the public build. Users aren't expected to need to set this (the default value should be good), but it provides an escape hatch for users who want to optimize for other access patterns.

#### Performance

Measured on Graviton2 against the current `main`:

- `PartialEq` is 90% to 99% faster, `ion_cmp` 57% to 83% faster, and `ion_data_hash` 11% to 38% faster across 1 to 100 fields. The old implementations rebuilt a sorted `Vec` (or did a per-field map lookup) on every call; the sorted order is now built once and cached in the `OnceLock`.
- `get()` is faster across the board except for a key sitting at the back of a struct within the linear band, whose cost is bounded by the threshold-sized scan and recovers as soon as the hash index engages.
- Construction does fewer allocations, since there is no map and no per-duplicate `SmallVec`, resulting in small but meaningful deserialization improvements on some real-world data that I used. (Don't have an exact figure since it's somewhat entangled with the rest of my work to reduce the size and align of `Element`.)

There is one regression introduced. Calling `get_all()` is now a scan over the slice rather than a map lookup. Duplicate field names are uncommon and the scan is over one contiguous allocation, so I think this should be acceptable. A Graviton L1 data cache is 64 kB, so the entire `slots` vec will fit in the L1 cache for struct with even several hundred fields. (You only need ~8 kB of cache to fit 100 fields.) If it turns out to be a problem, we can leverage the `by_fields` index to perform a binary search of field names for large structs.

#### Testing

- The full test suite, the `ion-tests` corpus, and `tests/ion_data_consistency.rs` (the `good/equivs` `Eq`/`Ord`/`Hash` gate) pass.
- `--features experimental-ion-hash` passes; the struct digest is unchanged, since it still hashes the per-field _digests_ in their own sorted order.
- `cargo check --target wasm32-unknown-unknown` passes, which is what checks the pointer-width byte-size assertions.
- New unit tests cover both lookup regimes: duplicate, absent, and unknown-text (`$0`) lookups; the `$0`-versus-empty-text case, whose keys hash alike but compare unequal; multiset equality across the equivalence classes crossed with both regimes; the threshold boundary; `Send`/`Sync`; and a concurrent first `get()`.

----

### DIFF
--- a/src/types/struct.rs
+++ b/src/types/struct.rs
@@ -4,80 +4,186 @@ use crate::ion_data::{IonDataHash, IonDataOrd, IonEq};
 use crate::symbol_ref::AsSymbolRef;
 use crate::text::text_formatter::FmtValueFormatter;
 use crate::Symbol;
-use smallvec::SmallVec;
+use hashbrown::HashTable;
+use rustc_hash::FxHasher;
 use std::cmp::Ordering;
-use std::collections::{HashMap, VecDeque};
+use std::collections::VecDeque;
 use std::fmt::{Display, Formatter};
-use std::hash::Hasher;
+use std::hash::{Hash, Hasher};
+use std::sync::OnceLock;
 
-// A convenient type alias for a vector capable of storing a single `usize` inline
-// without heap allocation. This type should not be used in public interfaces directly.
-type IndexVec = SmallVec<[usize; 1]>;
+/// Field counts at or below this length are served by a linear scan over `slots`; above it, `get()`
+/// lazily builds and probes a hash index. The crossover is a benchmark-derived tuning default
+/// (measured on aarch64 / Graviton2, with a workload of roughly one lookup per field), not a
+/// contract. Revisit if `Element`/`Symbol` size or real `get()` access patterns change.
+const LINEAR_SCAN_THRESHOLD: usize = 48;
 
-// This collection is broken out into its own type to allow instances of it to be shared with Arc/Rc.
+/// One entry in the lazily built hash lookup index. Stores the cached field-name hash (so probes
+/// don't rehash the stored name) and the index of the corresponding pair in [`Fields::slots`].
+/// Only fields with **known** text are indexed; unknown-text (`$0`) lookups scan `slots` directly.
+#[derive(Debug, Clone, Copy)]
+struct HashIndex {
+    hash: u64,
+    slot_index: u32,
+}
+
+/// Lazily built auxiliary indexes over [`Fields::slots`]. Boxed behind a single `OnceLock` so an
+/// iterate-only struct never allocates either index, and so `Fields` itself stays pointer-small.
+#[derive(Debug, Clone, Default)]
+struct FieldsAux {
+    /// Hash lookup index over known-text fields, built on the first large-struct `get()`.
+    table: OnceLock<Box<HashTable<HashIndex>>>,
+    /// Slot indices sorted into canonical Ion field order, built on the first
+    /// equality / ordering / hashing operation. Covers **all** slots, including unknown-text ones.
+    by_field: OnceLock<Box<[u32]>>,
+}
+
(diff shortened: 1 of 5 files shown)


src/types/struct.rs
```diff
@@ -4,80 +4,222 @@ use crate::ion_data::{IonDataHash, IonDataOrd, IonEq};
...
+                })
+                .map(|entry| &self.slots[entry.slot_index as usize].1)
         }
     }
 
-    /// Gets the last value in the Struct that is associated with the specified field name.
+    /// Looks up the unknown-text (`$0`) field name. These share a hash with the empty-text field
+    /// name but compare unequal, so they are never indexed; a direct scan for the first text-less
+    /// field is both correct and cheap.
+    fn get_unknown_text(&self) -> Option<&Element> {
```
piyushk-amzn: get($0) regresses from O(1) to O(n). At least put a note about it in here and release notes.

piyushk-amzn (approved):
Looks good overall. One minor comment about the $0 look up cost increasing. Worth documenting.

src/types/struct.rs
popematt: Actually, this made me realize that (a) we are storing more than what is necessary in the hash index, and (b) we can actually just hash the $0 case. I'm fixing it.

almann (approved):
Approved based on @piyushk-amzn feedback.

PR #1049, opened 2026-09-14
perf(binary): stack-encode int/decimal magnitudes

popematt:
**Issue #, if available:**

N/A

**Description of changes:**

`UIntData::to_be_bytes` heap-allocates a `Vec` per `Int` value and per decimal coefficient. Encode `u128`-fitting magnitudes on a stack `[u8; 16]` instead; `BigUint` keeps the `Vec` on a `cold_path!`. Trimming stays byte-identical (zero -> `0x00`).

- consolidate the trimming rule in `UIntData::small_to_be_bytes`; `to_be_bytes`/`to_le_bytes` delegate, and `DecodedInt::write` shares its sign-magnitude tail across the hot and cold paths
- `encode_decimal` writes the coefficient from `Decimal` fields via `DecodedInt::write(&Int)`, dropping two per-decimal clones
- reuse `as_small_value()` and `#[inline]` it so the hot path inlines in non-LTO downstream builds
- add byte-exact tests for both encoders (widths 1-17, `u128` seam, sign padding, 13->14 header boundary); prior tests only compared decoded elements

This improves the cycle count for encoding integer values and integer fields (i.e. in timestamps, decimals) by as much as ~30% in some informal benchmarks I ran.


almann (approved):
:shipit:

PR #1053, opened 2026-09-17
perf(element): pack Annotations into a one-word tagged pointer

popematt:
**Issue #, if available:**

#476 

**Description of changes:**

Replace the `Vec<Symbol>` backing with a one-word tagged pointer, so `Annotations` occupies one word in every `Element` instead of three. The tagged pointer and all of its `unsafe` live in a private `AnnotationsRepr`; the public `Annotations` derives its traits and delegates, so its public API is unchanged.

The empty case — overwhelmingly the common one — is a null pointer that touches no heap; the 1- and 2-annotation arms carry a single indirection. The tag in the low two bits selects the four storage shapes (empty, `Box<Symbol>`, `Box<[Symbol; 2]>`, `Box<Box<[Symbol]>>`).

Why `Box<Box<[Symbol]>>`? The outer box is a thin pointer to the inner fat pointer. We might be able to avoid the double indirection by storing a pointer to a `ManyAnnotations(usize, [Symbol])`, but >2 annotations is thought to be pretty uncommon, so the cost of the second indirection is minimal compared to the extra complexity that would be required to manage a custom DST.

Construction and teardown route through two symmetric primitives, `from_boxed_slice` and `into_boxed_slice`, that reuse each arm's allocation rather than rebuilding it: constructing from an exact-sized `Vec` reuses its buffer as the compact box on the 1/2 arms, and the consuming iterator is allocation-free on every arm. `Clone`, `Drop`, and `IntoIterator` all route through these. `PartialEq` and `ion_cmp` short-circuit on pointer identity (empty, or a value compared to itself). Per-arm allocation counts are recorded by the new `annotations_allocations` bench.


src/element/annotations.rs
```diff
@@ -172,14 +179,331 @@ where
...
+            2 => unsafe {
+                Self::from_tag_and_ptr(Self::TAG_TWO, Box::into_raw(boxed).cast::<[Symbol; 2]>())
+            },
+            // Marked cold: this is the only arm that reaches the allocator, for the outer thinning
+            // cell, and the compact arms are the common shapes.
+            3.. => cold_path! {{
+                // SAFETY: `TAG_MANY`'s documented pointee is `Box<[Symbol]>`, exactly what is boxed
+                // here, and it is not `TAG_EMPTY`.
+                unsafe { Self::from_tag_and_ptr(Self::TAG_MANY, Box::into_raw(Box::new(boxed))) }
+            }},
```
popematt: 🗺️ Cold path because it allocates. The corresponding arm in `as_slice` and `into_boxed_slice` have no allocation, so I didn't bother `cold_path!`-ing them.

piyushk-amzn (commented):
Nice optimization Matt to reduce overall memory size for large ion payloads.

One overall comment is to put an integ test (tests/annotations_allocations.rs), to validate zero heap allocations on into_iter call, so it is not just a benchmark test and is caught during cargo test. You can just have one test to validate, so you don't run into multiple tests running in parallel, modifying the global state.

almann (approved):
Approved as per @piyushk-amzn 's review

Issue #1038, opened 2026-08-15
Reduce per-document fixed costs for small binary Ion 1.0 documents (~half of Element::read_one time on a 228-byte document)


hsbakshi:
### Summary

When small binary Ion 1.0 documents are read one at a time (a fresh reader per document), most of the cost is per-document setup rather than value decoding. We measured this on a ~228-byte binary document: a 10-field struct with a local symbol table.

Reading that document with `Element::read_one` (the stable, default-feature entry point, and the first example in the `lib.rs` docs) costs ~11.2 µs total. The large majority is fixed cost: constructing the reader, then processing the IVM and the LST. Those fixed phases cost **~6.2 µs** measured on the binary lazy path; the`AnyEncoding` path that `Element` routes through costs more.

Phase split on the `v1_0::Binary` lazy path (same document, release build):

| Phase | ns/op | share |
|---|---:|---:|
| Reader construction | 875 | 12% |
| IVM + LST processing | 5,324 | **73%** |
| User value reads (10 fields) | 1,123 | 15% |

The allocation picture matches: of 52 heap allocations per document, the two fixed phases own 50; the lazy value-reading path itself is already efficient (~2 allocations). Any workload that reads one document per
record, message, or database item (a stream of independent small documents) pays this fixed cost on every read.

### Where the cost comes from and what we propose

**1. Per-document system-symbol-table rebuild** (largest contributor). 

Ion defines 9 built-in "system symbols", fixed strings like `name` and `version`. Today, building or resetting a symbol table heap-allocates all 9 of them with `Arc::from` and re-inserts them into the reverse text->SID index once per document, even though the strings never change. An in-code `TODO` at `symbol_table.rs` flag this ("we don't need to allocate these each time"). The codebase already has `Symbol::static_text`, so the fix can point at the `'static` strings directly, with no new machinery.

There is a second saving on the same path. When an IVM re-declares the version the stream is already using (the common case in concatenated Ion 1.0 documents), the reset can reuse the existing prefix-only truncation (already used on the non-version-switch LST path) instead of a full rebuild. The spec-mandated symbol-table reset still occurs; only the
reallocation is avoided. The regression test added in #1022 (`ivm_resets_symbol_table_in_concatenated_documents`) is the behavioral gate here; it passes unmodified with this change.

**2. Question for maintainers: macro-table reset on Ion 1.0 IVMs.** 

[...]

Issue #1050, opened 2026-09-16
Run all `--lib` tests in miri


popematt:
> Should we be just running everything in miri?

I like this, and I will add that as a follow up task. I may need to a `#[cfg_attr(miri, ignore)]` for some tests that use things that are unsupported by Miri, and if there are test failures, I would try to fix them at the same time.

_Originally posted by @popematt in https://github.com/amazon-ion/ion-rust/pull/1044#discussion_r4030339521_

...if it's a reasonable length of time to do so on a GitHub Actions runner.

PR #1054, opened 2026-09-18
Make `Writer` instances reusable

popematt:
**Issue #, if available:**

Related to #1038, but this affects the writer side.

**Description of changes:**

Adds an opt-in way to reuse a managed `Writer` across documents instead of constructing a fresh one each time. Building a writer per document allocates a fresh bump arena plus a symbol table and two sub-writers; a service that encodes many small documents (one per request/message) pays that setup on every value. This lets a caller park an idle, sink-less writer and rebind it to a new sink, amortizing the setup while keeping output byte-for-byte identical to a freshly built writer.

The feature is gated behind the `experimental-reader-writer` feature and does not change the crate's default public API.

The API is three methods on the managed `Writer`, available only for encodings whose raw writer can be reused (Ion 1.0 binary and text):

- `Writer::<E, ()>::idle(config)` — build a sink-less writer that can be parked and reused.
- `idle.attach(sink)` — bind a sink and get back an active writer, ready to encode.
- `writer.detach() -> (Writer<E, ()>, Output)` — reclaim the sink-less writer and the sink so the writer can be parked again.
- `Writer::<E, ()>::set_config(config)` — re-configure a parked writer in place (e.g. hand a compact-configured writer to a caller that wants pretty text) without discarding its warm scratch state.

A sealed marker trait, `Reusable`, gates the API at compile time: only Ion 1.0 encodings implement it, so `Writer<BinaryEncoding_1_0, _>` and `Writer<TextEncoding_1_0, _>` can be reused and the Ion 1.1 encodings cannot. Attempting to reuse a 1.1 writer is a compile error rather than a runtime fallback. Ion 1.1 is excluded deliberately: its writers own a macro table that reuse would need to reset, which is not yet implemented.

#### Behavior and guarantees

- **Byte-identical reuse.** A reused writer produces the same bytes as a fresh writer per document, for both binary and text 1.0 (covered by tests).
- **`detach` is infallible and discards any un-flushed document.** It performs no I/O, so it is safe to call from a `Drop` impl where an error could not be reported. Callers commit a document by calling `flush` (or `close`) first; `flush` is a top-level-only operation, so only complete top-level values ever reach the sink.
- **Bounded retained memory.** A parked writer's scratch arena, output buffers, and symbol table are shrunk back toward a small size when they have grown past an internal cap, so a long-lived pooled writer does not hold memory sized to the largest document it ever encoded. Each idle binary writer can retain up to ~1 MB of memory, and each idle text writer can retain up to ~512 kB of memory.

#### Move soundness (`AliasableBump`)

The raw binary 1.0 writer keeps a raw pointer to a top-level encoding buffer that lives inside its bump arena and captures a reference back to the allocator. Reuse moves the writer, which would relocate an inline `Bump` and dangle that captured reference on the next growing write. `Box<Bump>` does not fix this — its uniqueness assertion invalidates the captured reference on move/reborrow under Stacked/Tree Borrows. The allocator is now held in `AliasableBump`, an `Rc<Bump>` wrapper whose shared deref makes no uniqueness assertion, so the captured reference stays valid across moves. This is verified with Miri under both Stacked Borrows and `-Zmiri-tree-borrows`.

#### Testing

- Byte-identity and read-back round-trip tests for binary and text 1.0, including a warm-capacity-retention test that fails if a regression reallocates on every `detach`.
- Arena/buffer/symbol-table shrink tests, and an unflushed-discard test.
- `compile_fail` doctests confirming the Ion 1.1 encodings do not resolve the reuse API.
- Miri (Stacked and Tree Borrows) over the reuse and writer-move tests.
- `cargo public-api` confirms the default public API baseline is unchanged.

#### Performance

Informal testing on Graviton 2 with a variety of document sizes showed that reusing a binary writer can decrease the overall time to write an Ion document by amortizing the setup cost.

| Encoded size | fresh writer | reused writer | Δ |
|---|---|---|---|
| 6 B (single int) | 1.430 µs | 0.932 µs | −35% |
| 164 B | 2.883 µs | 1.882 µs | −35% |
| 1.03 KB | 6.554 µs | 5.269 µs | −20% |
| 9.96 KB | 43.46 µs | 42.19 µs | −2.9% |
| 99 KB | 444.6 µs | 444.8 µs | ~0 |

#### Example

`examples/reusable_writer.rs` demonstrates a generic thread-local reusable writer with an RAII guard: `acquire` hands out a guard that derefs to the writer, `finish` flushes and returns the writer to the thread's slot, and dropping the guard without `finish` discards the in-progress document and returns the writer anyway.


.github/workflows/miri.yml
```diff
@@ -66,10 +66,17 @@ jobs:
...
+      # folded scalar joins the lines with spaces and the value word-splits into one filter each.
+      # (No `#` comments inside the block -- they would become literal filter words. Document each
+      # module here instead.)
+      #   overflowing_int      -- 16-byte overflowing integer representation
+      #   move_soundness_tests -- raw binary writer stays sound when moved (AliasableBump arena)
+      # (The reusable-writer `reuse_tests` verify behavior, not unsafe soundness, so they run under
+      # normal `cargo test` only -- their large allocations are prohibitively slow under Miri.)
+      MIRI_TEST_SELECTION: >-
+        overflowing_int
+        move_soundness_tests
```
almann, line 79: Can we push these selections into a script or something?  Should our wrapper script just specify the tests that make sense to run this way for local builds/testing?

What is the implication if the tests ever get renamed and this doesn't change--does our script actually catch it?

src/lazy/encoder/binary/mod.rs
```diff
@@ -1,2 +1,126 @@
...
+    /// use-after-free).
+    pub(crate) unsafe fn reset(&mut self) {
+        // In normal operation this handle holds the only `Rc` (it is never cloned), so `get_mut`
+        // succeeds and we reset in place. If the arena were ever shared, we cannot obtain unique
+        // access, so fall back to a fresh arena rather than panicking -- the same defensive
+        // strategy as `EncodingContext::make_allocator_mut`, and it keeps `reset` panic-free (no
+        // `unwrap`/`expect` on a hot path).
+        match Rc::get_mut(&mut self.0) {
+            Some(bump) => bump.reset(),
+            None => self.0 = Rc::new(BumpAllocator::new()),
```
almann: This is very puzzling to me--if `None` ever happens, we have fundamentally broken something, right?  It is unclear why this fall-back is actually correct (in fact why is it not a `panic!`?).

I think this fallback actually obscures what would be a really bad state to be in, I would think `unreachable!` is actually the right answer, but maybe I am missing something.

almann (changes requested):
Looks generally good--but the "make a new allocator on invariant break" seems problematic.

.github/workflows/miri.yml
popematt: Tracked as #1050

src/lazy/encoder/binary/mod.rs
popematt: I think you're right, but I'm going to take a closer look just to be certain.

almann (approved):
:shipit:

PR #1055, opened 2026-09-21
feat(timestamp): add `TimestampBuilder::localize_to_offset`

popematt:
**Issue #, if available:**

None

**Description of changes:**

* Provides a public way to construct a `Timestamp` from fields given in UTC plus an offset—the natural shape of an Ion binary timestamp. This handles shifting the fields into local time on `build()` and replaces the `pub(crate)` `build_utc_fields_at_offset`.
* Adds a `SupportsOffset` marker trait so that it's clean if we decide to make `TimestampBuilder` public.

This is a useful/important change for users that might want to handle their own parsing of `Timestamp`s (e.g. domain types with built-in deserializers, alternative Ion reader implementations).


src/types/timestamp.rs
```diff
@@ -1093,18 +1115,52 @@ impl TimestampBuilder<HasDay> {
...
+/// finer. It bounds the single blanket impl below that provides [`TimestampBuilder::with_offset`]
+/// and [`TimestampBuilder::localize_to_offset`], so those two setters are defined once rather than
+/// duplicated across each state's `impl` block.
+///
+/// The trait is declared `pub` (required: it bounds `pub` methods) but lives in the private `types`
+/// module and is not re-exported, so it is effectively sealed — no type outside the crate can name
+/// it, let alone implement it, and the set of offset-bearing states cannot be extended downstream.
+/// (If the builder is ever exported as nameable public API, add a private supertrait to keep that
+/// sealing explicit once the module no longer hides it.)
+pub trait SupportsOffset {}
```
piyushk-amzn: Consider sealing this trait rather than relying on module being private.

piyushk-amzn (approved):
Looks good. One minor comment.

src/types/timestamp.rs
popematt: It is effectively as-good-as-sealed right now, and we have a public API check in CI, so it won't get exposed by accident.

